PR #1, opened 2016-01-31
UX-1# rest-api module... wip

vivek:
- Most model classes, still wip
- BlueOceanUI.doResT() handles /rest and tries to delegates rest of the path to
  ApiHead. It does not work though, needs more investigations
- @Operation annotation intercepts incoming HTTP requests and does the following:
   enforces handling of specific HTTP method, captures exceptions to return JSON error
   message
- WebException class to abstract HTTP specific exception.
- ProfileService defined and Standalone implementation as example. More work needed.

### DIFF
--- a/rest-api/src/main/java/io/jenkins/blueocean/rest/model/hal/Links.java
+++ b/rest-api/src/main/java/io/jenkins/blueocean/rest/model/hal/Links.java
@@ -1,0 +1,64 @@
+package io.jenkins.blueocean.rest.model.hal;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.jenkins.blueocean.rest.guice.JsonConverter;
+
+/**
+ * Implementation of
+ * <a href="https://tools.ietf.org/html/draft-kelly-json-hal-07">JSON Hipertext application language (HAL)</a>.
+ *
+ * <ul>
+ *     <li>Only -links element is implemented</li>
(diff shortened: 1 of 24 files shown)


vivek:
@michaelmccaskill  please take a look.

michaelmccaskill:
@vivek  I think you meant someone else named Michael. I'm not familiar with this project.

michaelneale:
@vivek thanks. hahha, yes wrong michael, taking a look.

michaelneale:
LGTM - however, perhaps this is broader than just #UX-1. 

The Pipelines/run stuff maybe should be its own module, and User/Org another. Perhaps put this against another ticket as this is infrastructural for REST/Json style interfaces.

rest-api/src/main/java/io/jenkins/blueocean/rest/model/hal/Links.java
```diff
@@ -0,0 +1,64 @@
...
+ **/
+public abstract class Links {
+    public final Link self;
+
+    public Links(@JsonProperty("self") Link self) {
+        this.self = self;
+    }
+
+    public static void main(String[] args) {
+        System.out.println(JsonConverter.fromJava(new Links(new Link("/users/23")) {
```
michaelneale: need to get these samples out of main into a test cases?

PR #202, opened 2016-05-26
Fix jenkinsfile

michaelneale:
A tidy up of Jenkinsfile, that specifies the "docker" label is required. 
No more hipchat from here (as OSS)

ping @rtyler 

@reviewbybees


michaelneale:
@tfennelly some old CI server feels the need to update github even thought he job was disabled .. I might rework this next week anyway. Add some stages and stuff!

michaelneale:
will open this from another angle

PR #207, opened 2016-05-27
[UX-181] Introduce api to see queued and running items

imeredith:
This is a WIP.

### DIFF
--- a/blueocean-plugin/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
+++ b/blueocean-plugin/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
@@ -56,6 +56,11 @@ public class AbstractRunImpl<T extends Run> extends BlueRun {
     }
 
     @Override
+    public String getQueueId() {
+        return Long.toString(run.getQueueId());
+    }
+
+    @Override
     public String getPipeline() {
         return run.getParent().getName();
     }
@@ -155,7 +160,7 @@ public class AbstractRunImpl<T extends Run> extends BlueRun {
         return null; // default
     }
 
-    protected static BlueRun getBlueRun(Run r){
+    public static BlueRun getBlueRun(Run r){
         //TODO: We need to take care several other job types
         if (r instanceof FreeStyleBuild) {
             return new FreeStyleRunImpl((FreeStyleBuild)r);
--- a/blueocean-plugin/src/main/java/io/jenkins/blueocean/service/embedded/rest/queue/BlueTaskQueueImpl.java
+++ b/blueocean-plugin/src/main/java/io/jenkins/blueocean/service/embedded/rest/queue/BlueTaskQueueImpl.java
@@ -1,0 +1,52 @@
+package io.jenkins.blueocean.service.embedded.rest.queue;
+
+import com.google.common.base.Function;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import hudson.model.Queue;
+import io.jenkins.blueocean.rest.model.BlueRun;
+import io.jenkins.blueocean.rest.model.queue.BlueTask;
+import jenkins.model.Jenkins;
+
+import javax.annotation.Nullable;
+
+/**
+ * @author Ivan Meredith
+ */
+public class BlueTaskQueueImpl extends BlueTask {
+    private Queue.Item item;
+
+    public BlueTaskQueueImpl(Queue.Item item) {
+        this.item = item;
+    }
+
+    @Override
+    public String getId() {
+        return Long.toString(item.getId());
+    }
+
+    @Override
+    public TaskType getType() {
+        return TaskType.QUEUE_ITEM;
+    }
+
+    @Override
+    public String getName() {
+        return item.getDisplayName();
+    }
+
+    @Override
+    public BlueRun getRun() {
+        return null;
+    }
+
+    public static Iterable<BlueTaskQueueImpl> getQueuedTasks() {
+        ImmutableList<Queue.Item> items = ImmutableList.copyOf(Jenkins.getInstance().getQueue().getItems());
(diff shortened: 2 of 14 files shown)


i386:
@imeredith don't forget the `README.md` update. Great work on getting this working though.

blueocean-plugin/src/main/java/io/jenkins/blueocean/service/embedded/rest/queue/BlueTaskQueueImpl.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public BlueRun getRun() {
+        return null;
+    }
+
+    public static Iterable<BlueTaskQueueImpl> getQueuedTasks() {
+        ImmutableList<Queue.Item> items = ImmutableList.copyOf(Jenkins.getInstance().getQueue().getItems());
+        return Iterables.transform(items, new Function<Queue.Item, BlueTaskQueueImpl>() {
```
tfennelly: So is this list effectively being cloned twice every time?

Why not just a simple `for` loop to create the `Collection` of `BlueTaskQueueImpl` instances? Aside from the fact that I had to read it twice to get what was happening, it seems like we're creating extra collections than are needed.

tfennelly: E.g.

```java
public static Iterable<BlueTaskQueueImpl> getQueuedTasks() {
    List<BlueTaskQueueImpl> blueItems = new ArrayList<BlueTaskQueueImpl>();
    for (Queue.Item item : Jenkins.getInstance().getQueue().getItems()) {
        blueItems.add(new BlueTaskQueueImpl(item));
    }
    return blueItems.iterable();
}
```

blueocean-plugin/src/main/java/io/jenkins/blueocean/service/embedded/rest/queue/BlueTaskRunImpl.java
```diff
@@ -0,0 +1,54 @@
...
+public class BlueTaskRunImpl extends BlueTask {
+    private Run run;
+
+    public BlueTaskRunImpl(Run run) {
+        this.run = run;
+    }
+
+    public String getId() {
+        return Long.toString(run.getQueueId());
+    }
```
tfennelly: I would thing we need a `getQueueId` and a `getId`.

blueocean-plugin/src/test/java/io/jenkins/blueocean/service/embedded/TasksTest.java
```diff
@@ -0,0 +1,81 @@
...
+            } else {
+                id =  p.getQueueItem().getId();
+            }
+
+            Map r = request().get("/tasks/" + id).build(Map.class);
+            Assert.assertEquals(r.get("id"), Long.toString(id));
+        }
+
+    }
+}
```
tfennelly: A few comments on the tests would be good I think i.e. what the tests are testing for. They look like they are nearly doing the same thing but just some subtle differences that are not obvious for those not familiar with the code.

tfennelly:
I think, in general, some Javadoc would be good on the new code.

I'm not sure if it's covered in this, but if I get a list of the runs that are currently executing, can I get the queueId of the original `Queue.Item` task that that run was created from i.e. correlate between them?

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/queue/BlueTask.java
```diff
@@ -0,0 +1,34 @@
...
+    @Exported(name = TYPE)
+    public abstract TaskType getType();
+
+    @Exported(name = NAME)
+    public abstract String getName();
+
+    @Exported(name = RUN, inline = true)
+    public abstract BlueRun getRun();
+
+    public enum TaskType {
```
vivek: Whats the purpose of task type is? Enum type makes it specific to only these two types. For example, if my plugin wants to add it's own Task, how would the getType work? I think we should remove getType(), just the subclasses of BlueTask is what the type of task is.

blueocean-plugin/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
```diff
@@ -155,7 +160,7 @@ public BluePipelineNodeContainer getNodes() {
         return null; // default
     }
 
-    protected static BlueRun getBlueRun(Run r){
+    public static BlueRun getBlueRun(Run r){
```
vivek: Why this change in scope?

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/queue/BlueTask.java
```diff
@@ -0,0 +1,34 @@
...
+    public abstract String getId();
+
+    @Exported(name = TYPE)
+    public abstract TaskType getType();
+
+    @Exported(name = NAME)
+    public abstract String getName();
+
+    @Exported(name = RUN, inline = true)
+    public abstract BlueRun getRun();
```
vivek: Not sure I understand, why is a task return run? Each of the exported methods should have comments to tell what its about, will be helpful to understand.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/queue/BlueTask.java
```diff
@@ -0,0 +1,34 @@
...
+    public static final String RUN = "run";
+
+    /**
+     * @returns Id of queue item. Needs to match up with {@link BlueRun#getQueueId()}.
+     */
+    @Exported(name = ID)
+    public abstract String getId();
+
+    @Exported(name = TYPE)
+    public abstract TaskType getType();
```
vivek: Not sure if we need it.

PR #213, opened 2016-05-27
UX-495# Module refactoring

vivek:
Related to issue # https://issues.jenkins-ci.org/browse/HOSTING-98

Summary of this pull request: 
- Needed by HOSTING-98 ticket
- Rename blueocean-plugin artifactId and module name to blueocean-rest-impl
- Create new aggreagator module blueocean-plugin with artifactId blueocean
- README changes
- Added LICENSE to each module

@reviewbybees


i386:
License changes LGTM

vivek:
@i386 @michaelneale BTW, I have tested running mvn hpi:run inside blueocean-plugin and some sanity testing, blueocean loads up fine. Will be good to have another eye verifies README changes and some manual tests.

michaelneale:
@vivek I think this is ok - can you verify it works when running as hpi plugins stand alone? (not just hpi)?

vivek:
Isn't that what dog food server does? Anyways am back on Tuesday, will do
it then. If you get time on your Monday feel free to do it, merge it?
On Sun, May 29, 2016 at 4:29 PM Michael Neale <notifications@github.com>
wrote:

> @vivek <https://github.com/vivek> I think this is ok - can you verify it
> works when running as hpi plugins stand alone? (not just hpi)?
>
> —
> You are receiving this because you were mentioned.
>
>
> Reply to this email directly, view it on GitHub
> <https://github.com/jenkinsci/blueocean-plugin/pull/213#issuecomment-222389179>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AACU-2EAjijA21fksMpi4zBvoBI03DTwks5qGiFHgaJpZM4Io4gZ>
> .
>
-- 
Sent from my iPhone

michaelneale:
@vivek :bee: - if you can verify it works in non "mvn hpi:run" mode that would be appreciated.

tfennelly:
🐝 

I would also rename the `blueocean-plugin` dir to be just `blueocean` (matching it's `artifactId`), but we can do that in another PR if others agree.

tfennelly:
BTW, I think there was another PR similar to this with the only difference being that it had a slightly different name for the `blueocean-rest-impl` and it renamed `blueocean-plugin` to `blueocean`. Where is that, or am I imagining things? If should get closed/deleted.

michaelneale:
@tfennelly I think that PR was closed. I think we can do with this one 🐝

tfennelly:
@michaelneale sure this is perfect ... I just wanted to get rid of the other if it was still around.

I already gave a bee, but ... 🐝

vivek:
@michaelneale I tested with tomcat and jenkins embedded jetty/winstone, copying over all blue*.hpi and their dependencies (sse-event, metric etc.) works fine.

PR #214, opened 2016-05-27
docker image so one can test blue ocean

daniel-beck:
Restoring the PR that was lost when we removed the fork and instead transferred the repository from cloudbees to jenkinsci.

Previous discussion:

> screen shot

FYI @ndeloof


michaelneale:
I like this, I think @i386 has wanted to have this for some time (this could also power any "dogfood server"). 

One comment - as we may need a Dockerfile in the root to represent the build environment. At the moment we use an "off the shelf" image - cloudbees/java-build-tools, but say we wanted to customise that, it would make sense to have Dockerfile side by side with Jenkinsfile for that purpose (no technical reason, just seems like a reasonable convention). 

Would this make sense in a "dist" or "docker" or some other top level dir? or perhaps "Dockerfile.distribution" ??

ndeloof:
@michaelneale I use to have 2 Dockerfiles (Dockerfile.build + Dockerfile) and script to chain them together, see https://github.com/cfpio/callForPapers/ for sample

Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+	curl -LO http://updates.jenkins-ci.org/latest/jquery-detached.hpi; \
+	curl -LO http://updates.jenkins-ci.org/latest/ace-editor.hpi; \
+	curl -LO http://updates.jenkins-ci.org/latest/script-security.hpi; \
+	curl -LO http://updates.jenkins-ci.org/latest/cloudbees-folder.hpi; \
+	curl -LO http://updates.jenkins-ci.org/latest/script-security.hpi; \
+	curl -LO http://updates.jenkins-ci.org/latest/junit.hpi; \
+	curl -LO http://updates.jenkins-ci.org/latest/plain-credentials.hpi; 
+
+RUN cd /usr/share/jenkins/ref/plugins; \
+	curl -LO http://archives.jenkins-ci.org/plugins/credentials/2.0.4/credentials.hpi;
```
dragoonis: this will create an additional docker layer, do you want to instead only have a single `RUN` section?

ndeloof: this is a temporary solution, waiting for 2.0.7 to be available on all mirrors

dragoonis: @ndeloof okay: 2.0.7 of what? (I'm still learning)

ndeloof: credentials 2.0.7 (i.e latest) - 2.0.5 introduced a major regression, so a fix release occurred yesterday. But mirror sync take some time.

batmat:
:+1:

docker-demo/set_welcome_message.groovy
```diff
@@ -0,0 +1,8 @@
+jenkins.model.Jenkins.instance.setSystemMessage(
+"""
+<p>Welcome to Jenkins Blue Ocean Demo. Click on the image below: </p>
+<p><a href="/blue"><img src="/userContent/blueocean-demo.png" /></a></p>
+
+<p>log in with <em>admin/admin</em></p>
```
batmat: @ndeloof as per the comment posted in #215, could you maybe add \<strong> around that line, so that users have bigger chances to read it and know the admin user credentials? Thanks

ndeloof: d4539f2ed427147651c81eba49ddd7d24f7c8ae2

michaelneale:
@ndeloof would be nice to use this as the ci "dogfood" image, but not sure how suitable it would be (given requires admin password, vs using github oauth ultimately). Perhaps find @imeredith on irc and hash out what could be done to make this the "one image"?

ndeloof:
@michaelneale @imeredith as auth is setup by init groovy script, can: 

- be tweaked by env parameters, so we can switch to oauth for ci dogfood
- just define a derived image with a distinct / complementary init script
- bind mount an init script at runtime for dogfood specific needs

lotharschulz:
You need to access the url: *http://[dockerIP]:8080* in order to test the [image](https://hub.docker.com/r/jenkinsci/blueocean/) on operating systems that require [docker machine](https://docs.docker.com/machine/) like OSX.
*[localhost:8080](http://localhost:8080)* would not work in those cases. Please adapt the [image description](https://hub.docker.com/r/jenkinsci/blueocean/).

ndeloof:
@lotharschulz docker4desktop will be released in few days at dockercon, so this won't be relevant anymore.

nicosuave:
@daniel-beck some links in the blueocean interface (such as New Project) are bound to `/jenkins` urls, so they 404. I'm not sure how to best set the `JERKINS_PREFIX=/jenkins` to resolve this.

batmat:
@nicosuave known issue indeed https://issues.jenkins-ci.org/browse/JENKINS-35202

i386:
The context path issue was fixed in #218 – just waiting on someone to verify it.

imeredith:
@ndeloof The blue* plugins need to be pinned as they are not in the UC. I was doing this by adding .override to the rile names with the old plugins.sh file. This doesn't work with the install-plugins script unforunatly.

Also it would be good to only set the security realm if none is already set.

ndeloof:
@imeredith I'm not aware of this `.override` (do you maybe mean `.pinned` ?). https://github.com/jenkinsci/docker/blob/master/install-plugins.sh could easily support this if needed. I'm a bit curious, as jenkins 2.x does not bundle any plugin anymore, do we still have some pinned plugin logic there ? 

Then `enable_security` script has been designed to offer a demo environment, maybe it would make sense to separate and have this Dockerfile to just distribute jenkins 2.x with blueocean plugins, then a derived -demo image to include those files ?

imeredith:
@ndeloof  https://github.com/jenkinsci/docker/blob/master/jenkins.sh#L16 override is used here.

I forked repo because I wasn't sure about copying in plugins like I did.

ndeloof:
@imeredith so IIUC issue is to provide end-user a way to get latest plugins while reusing an existing jenkins_home, as there's no way to update blueocean-* plugins from update center until they get published, right ? 
https://github.com/jenkinsci/blueocean-plugin/pull/214/commits/b0e7a7407d611f0ae5ed49aab3b8ba0fa0fe3ed2 do fix this

i386:
@ndeloof will this be deployed to dockerhub under Jenkins?

imeredith:
@ndeloof cheers thats exactly what we needed.

@i386 @ndeloof As for deployment I have created a blueocean org on dockerhub and will publish from the dogfood server. If it gets deployed else where thats up to other

i386:
@ndeloof great! can you please add @michaelneale, @imeredith, @kohsuke and @vivek admin access to the new org please?

ndeloof:
@imeredith why not publish under jenkinsci dockerhub account ?

i386:
@ndeloof @imeredith actually on second thoughts we probably want it in the jenkins docker hub account as it is better exposure.

rtyler:
@i386 this could be brought into the "trusted" CI environment we have for the project which builds and publishes a number of other Docker containers. That would need be expressed in the `Jenkinsfile` though

ndeloof:
@rtyler I guess this is not http://jenkins.ci.cloudbees.com/job/docker you have in mind ?

rtyler:
@ndeloof Correct. There is a Jenkins cluster off the public internet that we operate which has a number of keys to publish/sign various things. The docker containers for [jenkins-infra](https://github.com/jenkins-infra) are all published from this cluster, for example.

ndeloof:
@rtyler ok. Could you please then migrate the weekly docker image build there ?

rtyler:
@ndeloof Will you kick an INFRA ticket my way for that?

michaelneale:
We don't want the "dogfood" server pushing the jenkinsci account, so as long as that constraint is met...

michaelneale:
@imeredith @i386 can we merge this one? Would be handy to have it in here, or are we waiting on @ndeloof ?

ndeloof:
nothing more to do on my side afaik

batmat:
:+1: IMO even if we ever want to improve that again, there's already some value for users merging as the Docker image will immediately show blue ocean and an image with a link to it on the front screen.

rtyler:
merge it already

smazurov:
[Blueocean](https://hub.docker.com/r/jenkinsci/blueocean/) is already on dockerhub under jenkinsci user.

PR #217, opened 2016-05-30
Fixing afterClose navigation on Pipeline detail view (modal window)

mkramb:
When opening the pipeline details view (modal window) and then navigating between tabs and finally dismissing the window, the afterClose navigation url is not correct, and is showing an empty page.

First part of the fix is at routing, so that we can define fallback redirect (pipelines/activity) and then we have to be aware that opener url (previous) is passed from context when component is initialised and should not be set from render method, which is of course called every time we switch between tabs.

@reviewbybees

### DIFF
--- a/blueocean-dashboard/src/main/js/components/RunDetails.jsx
+++ b/blueocean-dashboard/src/main/js/components/RunDetails.jsx
@@ -44,9 +44,12 @@ class RunDetails extends Component {
                 },
                 config = {},
             } = this.context;
+
             config.pipeline = pipeline;
+
             this.props.fetchRunsIfNeeded(config);
             this.props.setPipeline(config);
+            this.opener = this.props.previous;
         }
     }
     navigateToChanges() {
@@ -60,20 +63,16 @@ class RunDetails extends Component {
             || this.props.isMultiBranch === null) {
             return null;
         }
+
         const {
-            context: {
-                router,
-                location,
-                params: {
-                    branch,
-                    runId,
-                    pipeline: name,
-                },
+            router,
+            location,
+            params: {
+                branch,
+                runId,
+                pipeline: name,
             },
-            props: {
-                previous,
-            },
-        } = this;
+        } = this.context;
 
         const baseUrl = cleanBaseUrl(this.context.location.pathname);
 
@@ -83,12 +82,11 @@ class RunDetails extends Component {
         result.name = name;
 
         const afterClose = () => {
+            const fallback = `/pipelines/${name}/`;
+
+            location.pathname = this.opener || fallback;
             location.hash = `#${branch}-${runId}`;
-            if (previous) {
-                location.pathname = previous;
-            } else {
-                location.pathname = `/pipelines/${name}/activity/`;
-            }
+
             router.push(location);
         };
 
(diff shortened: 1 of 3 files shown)


michaelneale:
Very cool! could you open a ticket on https://issues.jenkins-ci.org - with the "blueocean-plugin" component just to track this?

i386:
@mkramb nice to hear from you again 👍

michaelneale:
ping @sophistifunk - any feedback here?

sophistifunk:
I'll check this out and poke it with a stick later this evening.

blueocean-dashboard/src/main/js/components/RunDetails.jsx
```diff
@@ -44,9 +44,12 @@ class RunDetails extends Component {
                 },
                 config = {},
             } = this.context;
+
             config.pipeline = pipeline;
+
             this.props.fetchRunsIfNeeded(config);
             this.props.setPipeline(config);
+            this.opener = this.props.previous;
```
scherler: Why do you set this.opener and not just use this.props.previous later on?

scherler:
In general the only "change" is the redirect route, the other code is the same just a bit different style, correct? I would like to see the PR only with the Redirect in the Routes since the other is IMO noise.

mkramb:
@scherler 

Previously this.prop.previous was always set from render() function, which is wrong, because it was pointing to previous router change (previous tab) and not to window opener component. That is also the reason this is done in componentWillMount - which is called only once, point being after tab selection is changed, render() is called again (with different previous information). 

So there are three changes:

 - going to pipelines > activity and opening the details (modal window) and if we then also change tab selection from initial "pipelines" to something else, after that dismissing of modal window is broken, url which is used is from the previous tab and not from modal "opener"

 - better definitions of Routes in "PipelineRoutes", from modal window we can just redirect to "PipelinePage" route page and have definition which is the actual default Route in router definition, and not having that hardcoded in "RunDetails" component

 - and lastly I'm removing previous defined hash in location object from "Runs" component, so that we can set that on afterClose

The change/fix is that now you can dismiss a window if you before that switch between tabs (Pipeline, Changes, Tests, Artifacts). Please let me know if this makes it more sense.

mkramb:
@michaelneale @scherler

I opened a ticket with bug information https://issues.jenkins-ci.org/browse/JENKINS-35221,
and also added screenshots so it's easier to see/understand the bug.

scherler:
@mkramb thanks for the explanation, I only looked at the diff so did not see the componentWillMount. Further thanks for explaining how to reproduce the bug and filling the issue.

All good, only one small detail, IMO we do not need to "blank" the hash in Runs since we will always set it in the detail page however reset makes sure you do not pass an already existing to the detailView. ...but if you do so, we need to do it in PullRequest and Branches as well out of consistency.

:bee:

mkramb:
@scherler

I completely agree regarding removing location.hash in Runs component - not needed, reverted that file. Let me know if that works for you, thanks for the review.

sophistifunk:
Looks good to me. @scherler is *supposed* to be on holiday, he doesn't seem too good at it tho ;-)

sophistifunk:
@mkramb are you happy for me to merge this?

mkramb:
@sophistifunk 
Yes I am. Thanks.

i386:
@mkramb thanks for the changes 🎩 👍

PR #218, opened 2016-05-31
bug/JENKINS-35202

sophistifunk:
Related to issue #JENKINS-35202 

Summary of this pull request: 

@reviewbybees 

 * Update Pipelines to use the rootUrl when linking to 'new job' in legacy UI
 * Update test

### DIFF
--- a/blueocean-dashboard/src/main/js/components/Pipelines.jsx
+++ b/blueocean-dashboard/src/main/js/components/Pipelines.jsx
@@ -11,7 +11,7 @@ const { array } = PropTypes;
 export default class Pipelines extends Component {
 
     render() {
-        const { pipelines } = this.context;
+        const { pipelines, config } = this.context;
 
         // Early out
         if (!pipelines) {
@@ -30,16 +30,15 @@ export default class Pipelines extends Component {
             { label: '', className: 'favorite' },
         ];
 
+        const baseUrl = config.getRootURL();
+        const newJobUrl = `${baseUrl}view/All/newJob`;
+
         return (
             <Page>
                 <PageHeader>
                     <Title>
                         <h1>Dashboard</h1>
-                        <a
-                          target="_blank"
-                          className="btn-inverse"
-                          href="/jenkins/view/All/newJob"
-                        >
+                        <a target="_blank" className="btn-inverse" href={newJobUrl}>
                             New Pipeline
                         </a>
                     </Title>
@@ -65,4 +64,5 @@ export default class Pipelines extends Component {
 
 Pipelines.contextTypes = {
     pipelines: array,
+    config: PropTypes.object,
 };
--- a/blueocean-dashboard/src/test/js/pipelines-spec.js
+++ b/blueocean-dashboard/src/test/js/pipelines-spec.js
@@ -13,10 +13,17 @@ const
 describe("pipelines", () => {
   let tree;
 
+  const config = {
+    getRootURL: () => "/"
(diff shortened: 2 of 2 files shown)


sophistifunk:
This *should* work, but I don't have the correct set-up to actually test it in-container with a "growed up" Jenkins install that doesn't live on /jenkins/

i386:
🐝 LGTM

naphthalene:
This does not resolve the issue for me, when I do `mvn hpi:run` I still get a 404 when hitting `http://localhost:8080/jenkins/blue`
Am I missing something?

dragoonis:
@sophistifunk this is a nice simple PR for me to learn about doing this in other areas of the codebase. I've been wondering about a /jenkins/blue/view/All/newJob page, instead of just /jenkins/view/All/newJob .. is this in the works? what's the deal with this section? It's something I could try to make a start on and get my feet wet.

i386:
@dragoonis yeah there is a Pipeine creation design in the works - we need to get the design up to scratch before we put it out for feedback. For context, we are not focusing on solving the creation for other job types in 1.0 of Blue Ocean - we are really keen on making the best experience for a pipeline :)

If you are looking to get your feet wet we currently don't have any pagination on the dashboard or pipeline summary screens. We want to put a simple "show more" button on them so a user can have more pipelines displayed. Let me know if that sounds good to you :)

(And I need to get myself into gear and publish a roadmap!)

sophistifunk:
@naphthalene I don't think your issue is related to this bug, which was about correcting a URL that linked from Blue Ocean back into the existing Jenkins UI in order to create a new job. My naive guess based on your stack trace is that for some reason the actual Blue Ocean HPI isn't being loaded into the Jenkins instance, but I'm woefully uneducated when it comes to the internals of Stapler URL processing. You'll definitely have more luck with troubleshooting this on the mailing list than I can give you, I'm sorry to say.

i386:
@dragoonis I copied these out of our backlog (we are in the process of migrating it to the public issue tracker) https://issues.jenkins-ci.org/browse/JENKINS-35235 https://issues.jenkins-ci.org/browse/JENKINS-35236

sophistifunk:
@dragoonis I'm happy to answer any questions you have about this code; I'm not sure we have any other instances of this specific problem, as most of our links are generated by react-router's `<Link>` component, as they're between different Blue Ocean routes, rather than a link back to the existing UI which was the case here.

That being said, I didn't know we had this particular problem, so the fact that I think it's the only occurrence should definitely be taken with a grain of salt :)

dragoonis:
@i386 by default you show 5, then a Show More button, would you like the ajax call to pull in another 5 every time, or display 25 pipelines when you press Show More?

i386:
@dragoonis showing 25 at a time sounds like a good start then we can start tweaking that number. Probably want it to be configurable. 

Excited that you want to contribute btw! I'll publish some more issues this week so there's more interesting things to work on :)

i386:
by configurable I mean for the API wherever that component get gets placed

dragoonis:
@i386 okay, from what I understand is:

The initial lookup should be configurable, and default to 25 (```config.someKey || 25```), and if there's any more records we display Show More button.

Should more than 25 be pulled in the initial lookup, but just not displayed? or should we limit the initial load at 25 and then AJAX for another 25?

i386:
@dragoonis whatever you think is easiest todo for the moment :)

PR #219, opened 2016-05-31
The beginning of the blue ocean installer

dragoonis:
@reviewbybees

### DIFF
--- a/bin/setup.sh
+++ b/bin/setup.sh
@@ -8,6 +8,4 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
 # todo - check for version of java installed
 
-mvn clean install \
-&& cd ${DIR}/blueocean-plugin \
-&& mvn hpi:run
+mvn clean install "$@"
--- a/bin/start.sh
+++ b/bin/start.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# todo - check if maven clean install has been run prior
+
+cd ${DIR}/blueocean-plugin \
+&& mvn hpi:run
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# todo - check for mvn being installed
+
+# todo - check for version of maven installed
+
+# todo - check for version of java installed
+
+mvn clean install \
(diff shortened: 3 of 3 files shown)


vivek:
@dragoonis it's not really an installer, more of a convenience script to build and run blueocean plugin locally? Nothing wrong with it just that there are going to be different variations of how folks may like to build and/or run. Looks confusing to me in terms of what its trying to do and the usage it serves. 🐛

michaelneale:
this may be useful as currently people have to follow the readme in detail to know what to try. Perhaps renaming it to make intent clear...

i386:
perhaps move to {{bin/start.sh}}

vivek:
Calling bin/start.sh is better. Although 'mvn clean build' should be optional/conditional as some may like to just run it (mvn hpi:run) if it's previously built.

michaelneale:
there is also "-DcleanNode" which can sometimes be run to give a fresh environment. Perhaps encoding that into bin/start.sh may make sense... (maybe "dev.sh" is better than start.sh - as this is development mode)

dragoonis:
@vivek I agree with your points, this is a convenience script;

@michaelneale is correct, IMO, because we have to keep on referring people to the readme and we have no clue what parts they have or haven't ran from the readme. Having a script for newcomes to just execute and have a setup at /jenkins/blue/, this is the goal here.

I guess a good question could be, what is the **default** way to spin this up? `mvn clean install` with no flags, is the README suggestion. If anyone wants to deviate from that then they shouldn't use `setup.sh`

**Setup**
This script's usefulness is to check the local system for necessary deps such as having java, maven and other bits are prerequisites. This will educate the blueocean onboarder that they don't have the necessary bits in place before even being able to perform `bin/start.sh`

How about `bin/setup.sh` and `bin/start.sh` ?

dragoonis:
I've pushed an update to this PR.

To answer @vivek and @michaelneale's point about passing custom flags to the maven install command, I've built it with ```mvn clean install "$@"``` this way folks can optionally pass extra bits to it if they want to.

vivek:
@dragoonis Looks good. Maybe you can add some comments at the top of setup.sh and start.sh scripts to say what they do?

michaelneale:
I like it - if @vivek thinks this is ok, lets get this in. Maybe put a note in the README for people to try this (for those that want a quick start).

dragoonis:
Okay, I will add descriptions to the 2 shell scripts and I will update the project readme to reflect these scripts too.

michaelneale:
ok @dragoonis, sounds good.

tfennelly:
Looks good to me. I wouldn't bother with passing parameters to the build script. If people want more than just the ability to build it and run it (without extra params), then they should start using normal `mvn` directly.

PR #221, opened 2016-06-01
Paginating the Pipelines View

dragoonis:
This was a feature request by @i386 

I've done some initial parts on showing the `Show More` button when the number of pipelines reaches a certain amount. There's just a bit of complexity here that I've discovered and wanted to share.

- the crud subscriber code that I've stumbled across will pull in a set of data and doesn't know anything about the paginated state.
- the UI would have to then persist the current offset/limit that it's on somewhere. Also a few other parts of the code, would have to be updated to respect this state paginated state.
- I foresee a bunch of code needing written to allow new pipelines to make their way to Blue Ocean subscriber jobs and update the UI with the pagination logic in mind, always.

**Simplified Approach**
I'm proposing a simplification, so rather than having Pagination at all with a `Show More` button, we just make a `Show All` button.

This would allow me to not even care about pagination offset/limit states and continue to always pull in `all pipelines` via the `/rest/` call and just toggle the UI upon initial page load.

Keen to get some feedback on this change in direction, and any additional input that should be considered.

### DIFF
--- a/blueocean-dashboard/src/main/js/components/Pipelines.jsx
+++ b/blueocean-dashboard/src/main/js/components/Pipelines.jsx
@@ -30,8 +30,6 @@ export default class Pipelines extends Component {
             { label: '', className: 'favorite' },
         ];
 
-        console.log(config);
-
         const isShowMoreButtonVisible = pipelines.length > 5;
 
         const baseUrl = config.getRootURL();
@@ -61,7 +59,6 @@ export default class Pipelines extends Component {
                             }
 
                             {isShowMoreButtonVisible && <tr><td colspan="5"><button className="">Show More</button></td></tr>}
-
 
                         </Table>
                     </article>
--- a/blueocean-dashboard/src/main/js/components/Pipelines.jsx
+++ b/blueocean-dashboard/src/main/js/components/Pipelines.jsx
@@ -10,8 +10,22 @@ const { array } = PropTypes;
 
 export default class Pipelines extends Component {
 
+    constructor(props) {
+        super(props);
+        this.state = {
+            showAllPipelines: false,
+        };
+        this.onShowAllClick = this.onShowAllClick.bind(this);
+    }
+
+    onShowAllClick() {
+        this.state.showAllPipelines = true;
+        this.forceUpdate();
+    }
+
     render() {
         const { pipelines, config } = this.context;
+        const numInitialPiplinesToDisplay = 5;
 
         // Early out
         if (!pipelines) {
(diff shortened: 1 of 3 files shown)


dragoonis:
In commit 2b9545d I have pushed the functional code that only displays the first 5 rows, with a Show All button. When you press Show All, the button disappears and all of the piplines are rendered.

**Things needing looked at**
- I've hard coded the initial number of piplines to show, as a `const` instead of some `configurable value`. I have added this to `config.js` but given that this is my first React app I'm unsure how to pass it through. I know this will be changed from `5` to something like `20` rows to display but I kept it small for testing purposes so that we don't need to make 20 pipelines to test it out. I will make it the right value once we're happy with the implementation. https://github.com/dragoonis/blueocean-plugin/blob/2b9545d149354fcb6cb5e9da95698849063e899b/blueocean-dashboard/src/main/js/components/Pipelines.jsx#L28

- I've re-rendered the pipelines list by using `this.forceUpdate()`, which was documented as a way to re-trigger the `render()` function. However there are notices on the documentation saying that this isn't always recommended and can cause perf issues. I'm wondering if there's a more recommended way to re-calling the render function without this `forceUpdate()` technique?

michaelneale:
Hey @dragoonis did @cliffmeyers get in touch with you? He was going to spend some time helping you on this.

dragoonis:
Not yet Michael. I've not been on IRC much and won't be over the next few days.

I'm going to be around on GitHub most of the days though.

cliffmeyers:
@dragoonis apologies. I may have missed the message where Michael asked me to look at your PR. I'll review this tomorrow when I am back at the office and provide some feedback!

dragoonis:
No problem Cliff. I'm sure you're working on more important things than pagination.

The code works, but as mentioned needs a bit more love in some areas.

blueocean-dashboard/src/main/js/components/Pipelines.jsx
```diff
@@ -10,8 +10,22 @@ const { array } = PropTypes;
...
+            showAllPipelines: false,
+        };
+        this.onShowAllClick = this.onShowAllClick.bind(this);
+    }
+
+    onShowAllClick() {
+        this.state.showAllPipelines = true;
+        this.forceUpdate();
+    }
+
```
cliffmeyers: I know you're new to React: the best way to do this is to use the setState() method which triggers an "update" (which triggers `render()` again:

e.g. 
```
this.setState({
    showAllPipelines: true
});
```

cliffmeyers:
@dragoonis I discussed the requirements for this a bit with @i386 and there's some things that we need to consider that might change the code design in a major way:

* We need to support limiting up to X records, then click "Show More" to display 2X, then 3X, and so on.
* We need to preserve that number of records displayed if the user navigates to another page and then clicks back
* We also need to preserve the scroll state in that scenario.

As long as our content is growing large enough to scroll the entire page, then the browser should give us the third item for free.

If the user has clicked "Show More" twice (e.g. displaying 3 pages of data), we need to preserve that state in some way. I would recommend updating the URL to reflect the number of pages being displayed. We could add a query param called `page` which if unspecified would default to a value of 1 As the user clicks show more, we can replace the current history item with the updated URL. If the user navigates away and presses back, they'll get pipelines?page=3 which should load the correct data, and the browser will hopefully retain the scroll position.

I haven't written any code to validate this but it's the most reasonable approach I have just thinking through it in my head. You can refer to `PipelinesRoutes` and the code for the pipeline details view to see how integration with "params" is done.

tfennelly:
I think we should avoid a "Show All" button (as stated in the description). That option didn't work well in the Build History widget in "classic" Jenkins as jobs can have 1,000s of runs.

michaelneale:
Yes, good point. People have to opt in to limiting history, so show all would be a bad idea ("show more"?)

cliffmeyers:
@dragoonis I know you said on IRC you'd be offline for a while but I just thought I'd follow up with you now. Are you still interested in enhancing this PR to make it align with the UX guidelines?

i386:
@dragoonis @cliffmeyers ill close this one out until theres a decision to resume it again :)

PR #222, opened 2016-06-01
UX-504# Log response has missing X-Text-Size header

vivek:
Related to issue # . 

Summary of this pull request: 
.
.
.

@reviewbybees 

This header was added but after call to ServletResponse.getWriter(). Servlet
spec doesn't talk about it explicitly but certain servlet container implementation
commits the response such that headers added after call to ServletResponse.getWriter()
are ignored. Fix was to move this code before call to ServletResponse.getWriter().


michaelneale:
🐝

PR #223, opened 2016-06-01
[WIP] added Docker build file

robtec:
Related to issue # . 

Summary of this pull request: 

**Motivation**

- Useful if I want to get Blue Ocean up and running without having java, maven etc installed

- Common image to test features/PRs

**Building the image**

`docker build -f Docker.build -t jenkinsci/blueocean-plugin .`

**Run the Container (interactive)**

`docker run -it -v ~/.m2:/root/.m2 jenkinsci/blueocean-plugin mvn clean install`

**TODO**

Expose `8080` and create entry point file which builds (`mvn clean install`) and runs (`mvn hpi:run`)

Initial thoughts guys? 😄 

@reviewbybees

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,0 +1,10 @@
+FROM maven:3
+
+# install nodejs and gulp
+RUN curl -sL https://deb.nodesource.com/setup_4.x | bash - && \
+    apt-get install -y nodejs && \
+    npm install gulp -g
+
+ADD . /etc/blue-ocean
+
+WORKDIR /etc/blue-ocean


michaelneale:
Thanks @robtec - is this similar to: https://github.com/jenkinsci/blueocean-plugin/pull/214 ? 

This does look handy for those that want to run it in dev mode quickly with only docker. Perhaps putting this in a /bin directory might make sense? (I can't remember if the Dockerfile can refer to a context one level above its directory though).

michaelneale:
@robtec I like this. Could you put a description of it in the file itself, just like you did in the PR? Then when people discover it it will be explained. Perhaps a section in the main README.md too. This could indeed be handy for PRs etc.. Is gulp and nodejs required? Mvn should pull it down (last I tried with that image, I needed to setup a user for git so the tests would run, did you not see that break?)

ndeloof:
suggestion : `COPY` package.json and pom.xml before you `COPY .` command, so you can resolve dependencies and won't have to re-run dependency resolution any time you change a file on project and rebuild container.

batmat:
+1, in the same vein of what @ndeloof says, you possibly want to "warm up" the maven local cache (aka the _repository_) so that you don't have to download all the maven dependencies each time you run mvn clean install from the CLI. It could accelerate the build with an order of magnitude.

See https://github.com/batmat/jenkins-pr-tester/blob/master/Dockerfile#L13L19 for an example.

michaelneale:
@robtec you able to address the above by @batmat and @ndeloof ?

ydubreuil:
Hi, I missed your PR before starting to work on #268. I think it supersedes Dockerfile in this PR and supports keeping the local maven repository around for following builds. Would you mind taking a look?

robtec:
@ydubreuil looks cool - if the guys are happy with #268 then I can close this PR

michaelneale:
@robtec cool yes. Feel free to close.

blueoceandeploy:

PR #228, opened 2016-06-03
UX-533# Don't recurse in to sub folders

vivek:
Related to issue # . 

Summary of this pull request: 
.
.
.

@reviewbybees


michaelneale:
🐝  looks good. 

This passes, but I think @cliffmeyers  was running a test with his account that marked it as failure. @cliffmeyers if testing a repo - use a different repo otherwise it marks pull requests that are in progress.

PR #232, opened 2016-06-06
[UX-431] Externalize some packages

tfennelly:
Related to issue [UX-431](https://cloudbees.atlassian.net/browse/UX-431). 

Externalizes some of the NPM packages used in the web and dashboard  plugins.

I had to make a bugfix to js-builder and so also included a possible fix for [JENKINS-35180](https://issues.jenkins-ci.org/browse/JENKINS-35180)

@reviewbybees

### DIFF
--- a/js-extensions/package.json
+++ b/js-extensions/package.json
@@ -21,7 +21,7 @@
     "react-dom": "^0.14.7 || ^15.0.0"
   },
   "devDependencies": {
-    "@jenkins-cd/js-builder": "0.0.33-beta7",
+    "@jenkins-cd/js-builder": "0.0.33",
     "@jenkins-cd/js-test": "1.1.1",
     "gulp": "^3.9.1",
     "react-tools": "^0.13.3"
--- a/js-extensions/package.json
+++ b/js-extensions/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jenkins-cd/js-extensions",
-  "version": "0.0.14",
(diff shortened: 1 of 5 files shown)


js-extensions/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@jenkins-cd/js-extensions",
-  "version": "0.0.14",
+  "version": "0.0.15-beta4",
```
tfennelly: Yes ... I need to do a proper (non `--tag` release) before merging this PR.

ianceicys: I've taken a look at this PR on my windows machine and I can successfully build the dashboard now but there are several test failures. 

Video(4:30 min) with my repo steps: [https://www.youtube.com/watch?v=KFe-ubQ-VdA?version=3&vq=hd1080]

Build Summary: https://gist.github.com/ianceicys/eafeba18ae28435cdaaf76dc4c7bb385

nohup test failure - https://gist.github.com/anonymous/4759fee0a0218fe9cb5aa4324a2111d1

sh test failure - https://gist.github.com/anonymous/c554be4948c2bd949de1222dc437c005

michaelneale: A lot of the tests use "sh" steps in pipeline, which I don't think will work on windows. They could be changed to simply echo output - and then work across platforms (and would serve the same purpose, as long as the assertions are still there).

michaelneale:
🐝  I think this looks good.

PR #233, opened 2016-06-06
feature/ux 529 - filter out folders from the Pipelines view

cliffmeyers:
Related to issue # UX-529. 

Summary of this pull request: 
* Add some logic to filter out the folders

I wanted to experiment with a new concept in this PR which is using Immutable.js Record subclasses. I think it would be an improvement if our data was typed, mainly for use of debugging but also giving us the ability to bake some relevant methods directly into the data itself.

@reviewbybees

### DIFF
--- a/blueocean-dashboard/src/main/js/components/Pipelines.jsx
+++ b/blueocean-dashboard/src/main/js/components/Pipelines.jsx
@@ -1,6 +1,6 @@
 import React, { Component, PropTypes } from 'react';
 import PipelineRowItem from './PipelineRowItem';
-import { PipelineRecord } from './records';
+import { FolderRecord, PipelineRecord } from './records';
 
 import { Page, PageHeader, Table, Title } from '@jenkins-cd/design-language';
 import { ExtensionPoint } from '@jenkins-cd/js-extensions';
@@ -18,7 +18,12 @@ export default class Pipelines extends Component {
         }
 
         const pipelineRecords = pipelines
-            .map(data => new PipelineRecord(data))
+            .map(data =>
+                'numberOfFolders' in data ?
+                    new FolderRecord(data) :
+                    new PipelineRecord(data)
+            )
+            .filter(data => !(data instanceof FolderRecord))
             .sort(pipeline => !!pipeline.branchNames);
 
         const headers = [
--- a/blueocean-dashboard/src/main/js/components/Pipelines.jsx
+++ b/blueocean-dashboard/src/main/js/components/Pipelines.jsx
@@ -1,6 +1,6 @@
 import React, { Component, PropTypes } from 'react';
 import PipelineRowItem from './PipelineRowItem';
-import { FolderRecord, PipelineRecord } from './records';
+import { PipelineRecord } from './records';
 
 import { Page, PageHeader, Table, Title } from '@jenkins-cd/design-language';
 import { ExtensionPoint } from '@jenkins-cd/js-extensions';
@@ -18,12 +18,8 @@ export default class Pipelines extends Component {
         }
 
         const pipelineRecords = pipelines
-            .map(data =>
-                'numberOfFolders' in data ?
-                    new FolderRecord(data) :
-                    new PipelineRecord(data)
-            )
-            .filter(data => !(data instanceof FolderRecord))
+            .map(data => new PipelineRecord(data))
+            .filter(data => !data.isFolder())
             .sort(pipeline => !!pipeline.branchNames);
 
         const headers = [
--- a/blueocean-dashboard/src/main/js/components/records.jsx
+++ b/blueocean-dashboard/src/main/js/components/records.jsx
@@ -65,6 +65,15 @@ export const RunsRecord = Record({
     pullRequest: PullRequestRecord,
 });
 
+export class FolderRecord extends Record({
+    displayName: '',
+    fullName: '',
+    name: '',
+    numberOfFolders: 0,
+    numberOfPipelines: 0,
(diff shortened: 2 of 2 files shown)


blueocean-dashboard/src/main/js/components/records.jsx
```diff
@@ -3,19 +3,23 @@ import Immutable from 'immutable';
 since we would return a function, */
```
cliffmeyers: There was a note here made about not being able to extend Record - I'm curious if anyone knows what the problem was?

cliffmeyers:
I have a larger question which involves where we are creating most of our Immutable records. My gut intuition is that they'd be created in the stores where the bulk of data management would be done. Right now they're mostly being creating in our UI components. Is this something we should we be looking at?

michaelneale:
@cliffmeyers  is this ready for review?

cliffmeyers:
@michaelneale yeah, it's fair to say it is. I have some open questions but I guess those can be addressed in a proper review.

blueocean-dashboard/src/main/js/components/records.jsx
michaelneale: as in - having a PipelineItem parent type that pipeline record and folder record descend from?

blueocean-dashboard/src/main/js/components/records.jsx
```diff
@@ -65,6 +69,15 @@ export const RunsRecord = Record({
     pullRequest: PullRequestRecord,
 });
 
+export class FolderRecord extends Record({
```
michaelneale: Still not sure if a separate folder record is the right way. I can see what you have done, and it makes sense, but will we have records of all types at some point and then the listing is always heterogenous? (not sure why, but heterogenous lists seem... odd... although I know this is what the json returns). I suppose it is better than a generic object with half the things null. 

My kingdom for sum types ;)

michaelneale:
@sophistifunk  this looks good to me - any objections?

blueocean-dashboard/src/main/js/components/Pipelines.jsx
```diff
@@ -18,7 +18,12 @@ export default class Pipelines extends Component {
         }
 
         const pipelineRecords = pipelines
-            .map(data => new PipelineRecord(data))
+            .map(data =>
+                'numberOfFolders' in data ?
+                    new FolderRecord(data) :
+                    new PipelineRecord(data)
+            )
+            .filter(data => !(data instanceof FolderRecord))
```
sophistifunk: Why not just `.filter(data => numberOfFolders in data)` instead of all this shenanigans creating extra stuff that's thrown away?

blueocean-dashboard/src/main/js/components/records.jsx
```diff
@@ -3,19 +3,23 @@ import Immutable from 'immutable';
 since we would return a function, */
 /* eslint new-cap: [0] */
 const { Record } = Immutable;
-export const PipelineRecord = Record({
+export class PipelineRecord extends Record({
```
sophistifunk: This makes no sense to me, why would you extend a reference which is then thrown away?

sophistifunk:
I like types. For types we should use Flow+ES6 classes, or TS. 

> [rant deleted]

blueocean-dashboard/src/main/js/components/records.jsx
sophistifunk: Is the return value of Record(....) even a proper constructor?

blueocean-dashboard/src/main/js/components/Pipelines.jsx
michaelneale: yeah - I think that may be better and it maps to the data closer. 

The day may come where we need a folder type, but I don't think like this.

blueocean-dashboard/src/main/js/components/records.jsx
cliffmeyers: I guess that's where I was at. It's not the end of the world to add the props for num of folders and pipelines but it makes it less clear to what is what. What if the Record for Pipeline had an `isFolder()` method? That would at least a bit more clear.

blueocean-dashboard/src/main/js/components/records.jsx
cliffmeyers: For reference, it is, see the end of these docs: https://facebook.github.io/immutable-js/docs/#/Record

blueocean-dashboard/src/main/js/components/Pipelines.jsx
cliffmeyers: Yep, all good, just trying new things to provoke discussion, which it did :)

blueocean-dashboard/src/main/js/components/Pipelines.jsx
```diff
@@ -18,12 +18,8 @@ export default class Pipelines extends Component {
...
 
         const pipelineRecords = pipelines
-            .map(data =>
-                'numberOfFolders' in data ?
-                    new FolderRecord(data) :
-                    new PipelineRecord(data)
-            )
-            .filter(data => !(data instanceof FolderRecord))
+            .map(data => new PipelineRecord(data))
+            .filter(data => !data.isFolder())
```
cliffmeyers: @michaelneale @sophistifunk here's a different take on it.

sophistifunk: Much nicer 👍

michaelneale:
🐝

PR #234, opened 2016-06-07
[Ux-449] Jacoco test coverage

imeredith:
Adds Jacoco test reports to blueocean-rest-impl.

We have 79% coverage atm.


michaelneale:
🐝  nice work. That's pretty good coverage.

i386:
Nice 🎉

tfennelly:
🐝  🍻  awesome ... 79% is really good !!!

I wonder what it is for the JS code. It's arguably more important there due to the lack of typing.

PR #235, opened 2016-06-07
UX-553# In case accept header is */* or text/*, defaults to text/plain.

vivek:
Related to issue # . 

Summary of this pull request: 
@michaelneale previous log fix caused this regression where default behavior changed to serve text/html logs. To prevent future regression a test to check text/plain being default type is added.

@reviewbybees


michaelneale:
🐝  thanks for jumping on this. Looks great.

PR #236, opened 2016-06-07
UX-550# Pipeline API: /pipelines

vivek:
Related to issue # . 

Summary of this pull request: 
.
.
.

@reviewbybees

### DIFF
--- a/blueocean-rest/README.md
+++ b/blueocean-rest/README.md
@@ -92,9 +92,28 @@ $$
         }
     }
 
-## Get Pipelines
+## Get Pipelines for an organization
+
+Pipelines are sorted by pipeline name alphabetically
 
     curl -v -X GET  http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/
+    
+    [ 
+      {
+      "organization" : "jenkins",
+      "name" : "pipeline1",
+      "displayName": "pipeline1",
+      "fullName" : "pipeline1",      
+      "weatherScore": 100,
+      "estimatedDurationInMillis": 280,
+      } 
+    ]
+
+## Get Pipelines
+
+Pipelines are sorted by pipeline name alphabetically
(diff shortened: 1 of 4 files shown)


imeredith:
@vivek @i386 Why are we doing this? In a recent discussion about queued items it was decided not to global scope them... so why pipelines?

blueocean-rest/README.md
```diff
@@ -107,6 +109,23 @@ $$
       } 
     ]
 
+## Get Pipelines
+
+Pipelines are sorted by pipeline name alphabetically
+
+    curl -v -X GET  http://localhost:8080/pipelines/
```
imeredith: should be /rest/pipelines/ ?

michaelneale:
@imeredith the feature request was specifically for pipelines across organisations. It isn't the main pipeline api that is currently in use.

imeredith:
@michaelneale i realise that, i just wonder why we want one and not the other.

blueocean-rest/README.md
vivek: yes, good catch, will update it.

vivek:
@imeredith see https://cloudbees.atlassian.net/browse/UX-310, thats driving this API, that is list of pipelines across organizations.

michaelneale:
@imeredith I believe we want both

imeredith:
@vivek I know, it just doesn't make sense to me as we seemed to be scoping everything to org, including queued items.

imeredith:
@michaelneale probably, it just struck me as odd after last weeks discussions

imeredith:
🐝 from me, as long as @i386 is happy with the global scope, which he should be since i assumed he was one that specc'ed it :)

i386:
@imeredith yes, happy with this. @kohsuke has known this is coming for a while :)

PR #239, opened 2016-06-08
[UX-163] Building on Windows

tfennelly:
Related to issue [UX-163](https://cloudbees.atlassian.net/browse/UX-163) . 

@jenkinsci/code-reviewers  @reviewbybees 

__TODO__:

- [x] Remove JDL symlink and copy assets for Windows
- [x] Fix frontend build in parent pom (see https://github.com/jenkinsci/plugin-pom/pull/26 ).
- [x] Update parent pom dependency after https://github.com/jenkinsci/plugin-pom/pull/26 merge

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ npm-debug.log
 **/.settings/
 **/.project
 **/.classpath
+blueocean-web/src/main/webapp/
--- a/blueocean-web/src/main/js/blueocean.js
+++ b/blueocean-web/src/main/js/blueocean.js
@@ -1,3 +1,6 @@
+// Apply polyfills before we do anything.
+require('./polyfills');
+
 // Initialise the Blue Ocean app.
 const init = require('./init.jsx');
 
--- a/blueocean-web/src/main/js/blueocean.js
+++ b/blueocean-web/src/main/js/blueocean.js
@@ -1,6 +1,3 @@
-// Apply polyfills before we do anything.
-require('./polyfills');
-
 // Initialise the Blue Ocean app.
 const init = require('./init.jsx');
 
--- a/README.md
+++ b/README.md
@@ -45,6 +45,12 @@ Builds all maven modules (run this the first time you check things out, at least
 $ mvn clean install
 ```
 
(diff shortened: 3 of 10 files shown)


tfennelly:
The bablejs polyfill fixed the ES6 errors on IE, but now we're hitting a `react-redux` issue.

After some quick hacking I can see that the `children` object on the following code is failing on a `ReactElement.isValidElement(children)` check because `object.$$typeof === REACT_ELEMENT_TYPE` is `false`. Showing up as a `Symbol`, whatever that means.

````javascript
Provider.prototype.render = function render() {
    var children = this.props.children;

    return _react.Children.only(children);
  };
```

More digging tomorrow.

michaelneale:
@cliffmeyers is there a master configured with write acess to the repo? as something is marking this build as failed but it isn't CI

blueocean-web/src/main/js/blueocean.js
```diff
@@ -1,3 +1,6 @@
+// Apply polyfills before we do anything.
+require('./polyfills');
```
michaelneale: Is this just for building on windows, or is this related to IE?

.gitignore
```diff
@@ -13,3 +13,4 @@ npm-debug.log
 **/.settings/
 **/.project
 **/.classpath
+blueocean-web/src/main/webapp/
```
michaelneale: how come we need this?

michaelneale:
Looking good - I assume still WIP? 

I had a few inline questions. Mostly - is this just for building on windows or is the polyfill for IE support?

blueocean-web/src/main/js/blueocean.js
tfennelly: That's needed to run on IE ... the earlier changes were for building (it should build fine now with `-DskipTests`.

.gitignore
tfennelly: We just want the folder there for holding the assets, which get put into it at build time. We don't want any of the contents checked in.

tfennelly:
@michaelneale Changing it from a Wip.  Getting this running on IE is not going to be a trivial task, so pushing that part to a separate task for now.

michaelneale:
🐝

PR #240, opened 2016-06-08
feature/ux 310 - All routable URLs must include the organization in the path

cliffmeyers:
Related to issue # UX-310. 

Summary of this pull request: 
* Rework route structure to hang off parent `/organizations/:org` (e.g. org="jenkins") with the exception of `/pipelines` which shows all pipelines
* Update UI to display org name in header and pipelines table, depending on context

While I was working through the code, I had planned to use two higher-level components to wrap OrganisationPipelines and provide it with different functions via props to fetch data either from the "all pipelines" endpoint or the "organization's pipelines" endpoint. There ended up being a lot of duplicate code, and it was easier and much less plumbing just to pass an optional organization name to action method and have it select the right endpoint.

_Tomorrow I plan to eliminate the two higher level components and just put the logic directly into `OrganisationPipelines` - but feel free to comment on anything else in the mean time._

@reviewbybees


blueocean-dashboard/src/main/js/AllPipelines.jsx
```diff
@@ -0,0 +1,14 @@
...
+
+    render() {
+        return (
+            <OrganisationPipelines { ... this.props } />
+        );
+    }
+
+}
+
+export default AllPipelines;
```
cliffmeyers: Tried an experiment with using higher-order components to wrap our old `OrganisationPipelines` and pass in organization (see: `OrganizationPipelines`). As I worked through it, these components became simpler and simpler, to the point where they may not even be necessary. (This one clearly isn't, so should probably nuke it and use the original directly).

blueocean-dashboard/src/main/js/Dashboard.jsx
```diff
@@ -0,0 +1,35 @@
...
+    children: node, // From react-router
+    location: object, // From react-router
+};
+
+Dashboard.childContextTypes = {
+    params: object, // From react-router
+    location: object, // From react-router
+};
+
+export default Dashboard;
```
cliffmeyers: I refactored some of the truly global things like params and location out of OrganizationPipelines because as we add more routes, odds are the may not descend from OrgPipelines but we still want these items to be available in all contexts.

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
```diff
@@ -0,0 +1,20 @@
...
+        );
+    }
+
+}
+
+OrganizationPipelines.contextTypes = {
+    params: PropTypes.object,
+};
+
+export default OrganizationPipelines;
```
cliffmeyers: Another higher order component, similar to `AllPipelines` - this one passes in the organization to provide that context. Honestly we could probably just do a little extra logic in OrganisationPipelines and discard them both, but I figured I'd leave them in for now to get some feedback.

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -136,12 +140,16 @@ export const actions = {
...
-            const url = `${config.getAppURLBase()}` +
-                '/rest/organizations/jenkins/pipelines/';
+            const baseUrl = config.getAppURLBase();
+            const url = !organizationName ?
+                `${baseUrl}/rest/pipelines/` :
+                `${baseUrl}/rest/organizations/${organizationName}/pipelines/`;
+
             if (!pipelines) {
                 return dispatch(actions.generateData(
                     url,
```
cliffmeyers: The change here is that `organizationName` is optional but, if supplied, will fetch from a different endpoint.

blueocean-dashboard/src/main/js/OrganisationPipelines.jsx
```diff
@@ -30,14 +30,12 @@ class OrganisationPipelines extends Component {
...
-            params,
-            location,
         };
     }
 
     componentWillMount() {
         if (this.context.config) {
-            this.props.fetchPipelinesIfNeeded(this.context.config);
+            this.props.fetchPipelinesIfNeeded(this.context.config, this.props.organization);
             const _this = this;
```
cliffmeyers: It probably makes a lot more sense just to check the route params here and pass them in. Actually the more I explain the old code, the less I like it.

blueocean-dashboard/src/main/js/AllPipelines.jsx
michaelneale: Should this be `OrganizationPipelines` with z not s? 

So in the end you moved back away from these higher level components?

blueocean-dashboard/src/main/js/AllPipelines.jsx
```diff
@@ -0,0 +1,14 @@
+import React, { Component } from 'react';
+import OrganisationPipelines from './OrganisationPipelines';
+
+class AllPipelines extends Component {
+
+    render() {
+        return (
+            <OrganisationPipelines { ... this.props } />
```
michaelneale: typo - using queens english ;)

michaelneale:
Looking good. As I mentioned on the ticket, the org name prefix for pipelines "looks odd" but otherwise, it seems to work well. 

Some comments above on the code (maybe just spelling consistency). Nice.

blueocean-dashboard/src/main/js/AllPipelines.jsx
cliffmeyers: Yeah, originally I was going to use these higher level components to set the `fetchPipelines` and `fetchPipelinesIfNeeded` properties of `OrganisationPipelines`, with different action funcs that would invoke a different REST URL. It requires quite a bit of plumbing, and I realized it made sense just to update the existing action functions with a `organizationName` property that would cause one URL or the other to be called. It's clear now the higher-level components are doing so little that I can just drive the logic back into `OrganisationPipelines` because it's so simple.

blueocean-dashboard/src/main/js/AllPipelines.jsx
cliffmeyers: I noticed the REST API is spelled like "organization" so I wanted to be consistent with that. Will do some renaming before this PR is done to get it all consistent.

michaelneale:
as this touches routes would like @sophistifunk to give it the nod (is this ready to review @cliffmeyers ?)

cliffmeyers:
@michaelneale it might help if I removed the higher-level components first. It's actually a lot of noise in what turned out to be a pretty straightforward PR. I probably won't get to that until tomorrow AM ET unfortunately.

michaelneale:
@cliffmeyers no problems - always nice when things "turn out" straightforward. 
Looking good so far!

i386:
@michaelneale let me handle the oddness of the prefix :)

michaelneale:
ok, ignore oddness about org prefix, but I like this. Will await more tidying up.

cliffmeyers:
Okay, this is a lot simpler now that the higher-order components are gone. Ready for review now.

cliffmeyers:
When this is merged, we also want to merge cloudbees/blueocean-acceptance-test#1

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
```diff
@@ -20,6 +20,8 @@ export default class PipelinePage extends Component {
             return null; // Loading...
         }
 
+        const baseUrl = removeLastUrlSegment(this.context.location.pathname);
```
sophistifunk: Is there a reason we're constructing a URL by chopping up what's in the browser's state, rather than the app's base URL? It makes it harder to figure out the URL by reading the code, and will almost certainly break as soon as the URLs get deeper because somebody has changed routes further down the component/route tree.

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -120,11 +120,15 @@ export const actions = {
...
      * @param config Application configuration.
+     * @param organizationName (optional)
      */
-    fetchPipelines(config) {
+    fetchPipelines(config, organizationName) {
         return (dispatch) => {
-            const url = `${config.getAppURLBase()}` +
-                '/rest/organizations/jenkins/pipelines/';
+            const baseUrl = config.getAppURLBase();
+            const url = !organizationName ?
```
sophistifunk: Could we ditch the `!` and just swap the following two lines? Then it reads as "if we have organisations, use them in the url like so, otherwise do this"

blueocean-dashboard/src/main/js/redux/actions.js
sophistifunk: Not related to this PR I know, but this, like the multibranch thing, is gonna come back to bite us one day. 

Related: everybody's sick of my pessimism.

sophistifunk:
Just a couple of small suggestions, but on the whole LGTM

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
cliffmeyers: You're right, in this case we can just use `/organizations/${organization}/${pipelineName}/ as the baseUrl. I wonder if instead of hardcoding the URL paths in various views if there would be any benefit to a utility that would create the URLs for us? From a maintenance standpoint, it might be easier if they were in one place.

blueocean-dashboard/src/main/js/redux/actions.js
cliffmeyers: Fixed in commit 43c258c

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
cliffmeyers: Fixed in commit 34bd840accff5357c2dd8f224f32dd23231cfc27

PR #241, opened 2016-06-09
[UX-181] Queue api for pipelines

imeredith:


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/MultiBranchPipelineImpl.java
```diff
@@ -4,11 +4,7 @@
 import hudson.model.Result;
 import hudson.model.Run;
 import io.jenkins.blueocean.commons.ServiceException;
-import io.jenkins.blueocean.rest.model.BlueMultiBranchPipeline;
-import io.jenkins.blueocean.rest.model.BluePipeline;
-import io.jenkins.blueocean.rest.model.BluePipelineContainer;
-import io.jenkins.blueocean.rest.model.BlueRun;
-import io.jenkins.blueocean.rest.model.BlueRunContainer;
+import io.jenkins.blueocean.rest.model.*;
```
michaelneale: I think explicit imports preferred (there is no intelliJ config checked in to enforce though).

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/PipelineApiTest.java
```diff
@@ -4,15 +4,10 @@
...
 import hudson.FilePath;
 import hudson.Launcher;
-import hudson.model.AbstractBuild;
-import hudson.model.BuildListener;
-import hudson.model.FreeStyleBuild;
-import hudson.model.FreeStyleProject;
-import hudson.model.Project;
-import hudson.model.Result;
-import hudson.model.Run;
+import hudson.model.*;
```
michaelneale: explicit preferred

michaelneale:
nice! - my only comments are explicit imports. 

@vivek looks good to you?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/MultiBranchPipelineImpl.java
vivek: Yes, explicit imports please:)

vivek:
@imeredith Looks good! Take care of explicit import and it's good to go.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/MultiBranchPipelineImpl.java
imeredith: This was my intellij

vivek: @imeredith There is setting to prevent wildcard imports, I set it to high value 9999, its hidden inside Settings->CodeStyle->Java-> 'class count to using import with *'

PR #244, opened 2016-06-13
Fixed docker build.

bmustiata:
Currently the docker build is not working, since the impl rest plugin is not included, and the blueocean plugin seems to have a funny name in the maven build.

Also I changed the `COPY` commands to `ADD`, since `COPY` is deprecated.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,7 +1,8 @@
 *
-!blueocean-commons/target/blueocean-commons.hpi 
-!blueocean-dashboard/target/blueocean-dashboard.hpi 
-!blueocean-plugin/target/blueocean-plugin.hpi 
-!blueocean-rest/target/blueocean-rest.hpi 
-!blueocean-web/target/blueocean-web.hpi 
+!blueocean-commons/target/blueocean-commons.hpi
+!blueocean-dashboard/target/blueocean-dashboard.hpi
+!blueocean-plugin/target/blueocean.hpi
+!blueocean-rest/target/blueocean-rest.hpi
+!blueocean-rest-impl/target/blueocean-rest-impl.hpi
+!blueocean-web/target/blueocean-web.hpi
 !docker-demo/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
 FROM jenkinsci/jenkins:latest
 
-ADD blueocean-commons/target/blueocean-commons.hpi /usr/share/jenkins/ref/plugins/
-ADD blueocean-dashboard/target/blueocean-dashboard.hpi /usr/share/jenkins/ref/plugins/
-ADD blueocean-plugin/target/blueocean.hpi /usr/share/jenkins/ref/plugins/blueocean-plugin.hpi
-ADD blueocean-rest/target/blueocean-rest.hpi /usr/share/jenkins/ref/plugins/
-ADD blueocean-rest-impl/target/blueocean-rest-impl.hpi /usr/share/jenkins/ref/plugins/
-ADD blueocean-web/target/blueocean-web.hpi /usr/share/jenkins/ref/plugins/
+COPY blueocean-commons/target/blueocean-commons.hpi /usr/share/jenkins/ref/plugins/
+COPY blueocean-dashboard/target/blueocean-dashboard.hpi /usr/share/jenkins/ref/plugins/
+COPY blueocean-plugin/target/blueocean.hpi /usr/share/jenkins/ref/plugins/blueocean-plugin.hpi
+COPY blueocean-rest/target/blueocean-rest.hpi /usr/share/jenkins/ref/plugins/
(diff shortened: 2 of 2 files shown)


batmat:
Out of curiosity, can you provide details on why you say `COPY` is deprecated @bmustiata? And also references on that statement? 
For me, it was something floating around 0.10.0, but was rolled back a long time ago, and the docs doesn't seem to have that info.

In the current docs, it [seems to be actually the contrary](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#add-or-copy):
> Although `ADD` and `COPY` are functionally similar, generally speaking, `COPY` is preferred. 

Thanks

bmustiata:
Thank you Baptiste, I stand corrected. I had outdated information.
I'm going to update the pull request.

See the comments for: http://stackoverflow.com/a/24958548

batmat:
NP.

For the ignores, wondering if it shouldn't now be using wildcards as there're many things to unignore.

Didn't test it, but IIRC this should work:

`!*/target/*.hpi`

Does it work for you?
WDYT?

michaelneale:
this is looking good, sorry some edge case with building PRs is causing this to fail to checkout to validate, I might try a few things to see if I can fix this.

michaelneale:
@bmustiata you want to use the pattern as suggested by @batmat ? 

(BTW ignore the PR validator failures, there is something odd about this pull requests but I can't workout what, but it isn't a valid failure).

bmustiata:
For the following `.dockerignore`:
```
*
!*/target/*
!docker-demo/
```

I get:
```
docker build .
Sending build context to Docker daemon 49.66 kB
Step 1 : FROM jenkinsci/jenkins:latest
 ---> 437122afab9e
Step 2 : ADD blueocean-commons/target/blueocean-commons.hpi /usr/share/jenkins/ref/plugins/
lstat blueocean-commons/target/blueocean-commons.hpi: no such file or directory
```

Conforming to the [.dockerignore documentation](https://docs.docker.com/engine/reference/builder/#dockerignore-file) it should work, but it doesn't.

michaelneale:
@ndeloof - sending the bat signal - can you offer any tips here for docker?

ndeloof:
having same issue, but should work

quickly tested:
```
package main

import (
    "fmt"
    "github.com/docker/docker/pkg/fileutils"
)

func main() {	
	b,_ := fileutils.Matches("foo/target/foo.hpi", []string{"*", "!*/target/*.hpi"} )
	fmt.Printf("%s\n", b)
}
```

bmustiata:
@ndeloof it doesn't seem to work, the `import()` seems broken.

I changed the `import` with individual imports, but while I get no error while running `docker build .`, the full folders are being sent to the docker server.

batmat:
I confirm I'm unable to unignore by wildcard. So I propose to leave that apart for now, obviously. 

As I'm curious to get to the bottom of that one, I'll create a small reproducer project and file an issue in docker/docker.

bmustiata:
Hm, but I see these as two different issues.
1. The docker build for the blueocean-plugin is broken. Which should be fixed via this PR.
2. The docker itself has a bug so ignoring multiple files is not working correctly. I suggest closing this one, and creating a new issue for tracking the docker problem. I am willing to monitor the other issue, and make another PR, but atm the docker build for `blueocean-plugin` is not working.

blueoceandeploy:

michaelneale:
ok let me try it out myself. I have NFI why the PR is not being validated (it wouldn't help with this anyway).

michaelneale:
I was unable to get this to build: 

`
download plugin : variant from https://updates.jenkins.io/latest/variant.hpi
download plugin : variant-plugin from https://updates.jenkins.io/latest/variant-plugin.hpi
failed to download plugin variant-plugin
`

with a few tries - are others seeing this? Seems it isn't quite right. 
cc @ydubreuil @ndeloof

michaelneale:
I can't make this work - variant plugin is missing: https://groups.google.com/forum/#!topic/jenkinsci-dev/4G-QFb2ANME

hopefully that shows up soon. I tried pinning the tagger version of Jenkins to 2.2, but then it was missing the install-script.sh

so someone should try in a day when the variant plugin actually exists in an UC refresh.

michaelneale:
yay, this works now. I added a wiki page for variant plugin and it showed up. 

I tested this locally on OSX and it seems good. 

🐝 

LGTM

i386:
@michaelneale is there a mention of the docker image in the read me?

michaelneale:
@i386 yoann was going to add one, once publishing is sorted. There is/will be a "build-in-docker.sh" that will run build in docker, and then produce an image (easier for reproducing of problems in a known environment).

i386:
@michaelneale also @rtyler was asking on IRC if we were planning to automate it. Not sure if there are tickets for this?

michaelneale:
@i386 no there aren't - are you able to ? should be done from the off-net jenkins.io infra to publish IMO (can be weekly)

PR #247, opened 2016-06-14
Remove external reference to material icons

i386:
This is provided by react-material-icons now
@reviewbybees


scherler:
@i386 you missed on reference but https://github.com/jenkinsci/blueocean-plugin/pull/248/files is good so I guess you can close this one

PR #254, opened 2016-06-15
remove-deprecated-jdl-consts - just a small cleanup that's a bit overdue

sophistifunk:
Small cleanup that's been TODO for a while, I just remembered about it due to working on the related code in JDL :)

@reviewbybees


cliffmeyers:
LGTM!  🐝

PR #259, opened 2016-06-16
Set modal to use expand-in animation

kzantow:
Use a better animation when displaying the 'modal' view

@reviewbybees esp. @cliffmeyers @i386


cliffmeyers:
🐝

scherler:
:bee:

PR #260, opened 2016-06-16
[JENKINS-35993] Fix Blue Ocean button for when no context root

tfennelly:
Related to issue [JENKINS-35993](https://issues.jenkins-ci.org/browse/JENKINS-35993). 

### DIFF
--- a/blueocean-web/src/main/js/try.js
+++ b/blueocean-web/src/main/js/try.js
@@ -1,9 +1,7 @@
 var $ = require('jquery-detached').getJQuery();
 
 function getAppUrl() {
-    const headElement = document.getElementsByTagName("head")[0];
-    const rootUrl = headElement.getAttribute("data-rooturl");
-    
+    const rootUrl = $('head').attr('data-rooturl');
     if (!rootUrl) {
         return '';
     } else {
--- a/blueocean-web/src/main/js/try.js
+++ b/blueocean-web/src/main/js/try.js
@@ -1,6 +1,19 @@
 var $ = require('jquery-detached').getJQuery();
 var jsModules = require('@jenkins-cd/js-modules');
 
+function getAppUrl() {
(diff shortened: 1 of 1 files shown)


blueocean-web/src/main/js/try.js
```diff
@@ -1,6 +1,19 @@
 var $ = require('jquery-detached').getJQuery();
 var jsModules = require('@jenkins-cd/js-modules');
 
+function getAppUrl() {
+    var rootUrl = jsModules.getRootURL();
+    
+    // If there's no prefix, rooturl is blank, resulting in js-modules
+    // returning the static url. We don't want to use that when
+    // switching to blue ocean, so return '' instead.
+    if (rootUrl.indexOf('/static/') === 0) {
```
kzantow: why would this ever be?

kzantow:
:bee: still wish for this just being an <a href...

blueocean-web/src/main/js/try.js
tfennelly: Read the comment.

kzantow: ... I mean, why would js-modules ever return the static URL as the root URL? this isn't normal behavior. handling it here is the wrong place, fix js-modules to not do this... if you need that static resource url, create a method for it if there isn't one.

tfennelly: @kzantow We don't actually want the static url ... js-modules just falls back to giving that when the root url attribute is not set for some reason e.g. when there's no context path, or on older versions of Jenkins that just don't set it at all. This is ok for what js-modules does, but I'll just remove it and access the DOM directly. Gimme a min 😄

kzantow:
🐝  seems better

kzantow:
stilla :bee:

PR #262, opened 2016-06-16
need scm connection

michaelneale:
Related to issue #JENKINS-35995 

Summary of this pull request: 
Add some details so it can be released to the update center. 

Question - what should the version string be set to? Does it need to be set in all plugins or just the parent pom.xml? 

@reviewbybees


pom.xml
```diff
@@ -22,6 +24,23 @@
...
+  <developers>
+    <developer>
+      <id>michaelneale</id>
+      <name>Michael Neale</name>      
+    </developer>
+    <developer>
+      <id>jamesdumay</id>
+      <name>James Dumay</name>      
+    </developer>    
+  </developers>
```
scherler: ¿so you guys are the only devs? ;)

scherler:
:bee:

pom.xml
michaelneale: need to get a more complete list. I am doing this in an other PR soon.

PR #268, opened 2016-06-16
local build script to build the docker image

ydubreuil:
Simulate the Jenkins Pipeline docker step locally in a shell script so BlueOcean is built like in Jenkins

@reviewbybees


Jenkinsfile
```diff
@@ -1,7 +1,7 @@
 node {
   deleteDir()
   checkout scm
-  
+
```
michaelneale: whitespace changes for the Jenkinsfile are a little annoying

michaelneale:
Can you update this to master (next week) and try again?

ydubreuil:
@michaelneale rebased on master and removed the irrelevant `Jenkinsfile` modifications

michaelneale:
I like it. Is build.sh the right name? If so, 🐝  from me - otherwise a better name would be also good.

michaelneale:
@ydubreuil hrm... I think this should tie into also building a complete docker image - which is https://github.com/jenkinsci/blueocean-plugin/pull/244 - being fixed. Perhaps this script should also have an option to do that? WDYT?

ydubreuil:
@michaelneale yes, building an image in the end makes total sense, actually, I did it in the first place before removing it while pushing. I didn't want to build an image every time. I'll add an option to build the image on demand :)

tfennelly:
Cool stuff @ydubreuil !!! So this is building a docker image? Maybe the name of the script should reflect that? Also, a few comments would help too I think e.g. if there are command line params ... examples of how to run it etc etc

michaelneale:
yeah this is most useful - increasingly I like the idea of being able to reproduce things with simple scripts as much as possible. For some value of "simple script" ;)

seems some minor conflicts now

ydubreuil:
Addressed most comments I hope, there's a online help now which explains what this script is for. And rebased against master

I didn't try it on MacOS, and I'm afraid it can fails because it requires to share workspace folder with the build container (same requirement as the Pipeline Docker plugin btw).

ydubreuil:
Um, regarding the script name, it can probably be renamed `build-in-docker.sh`

michaelneale:
@ydubreuil it would likely have better success with the new "docker for OSX beta" which does proper bind mounting (unlike the virtual box based ones) - docker pipeline certainly works well on osx now with that. 

Yes, go ahead and rename it, why not!

ydubreuil:
@michaelneale script renamed, could you merge?

michaelneale:
🐝

michaelneale:
thanks @ydubreuil. Done.

PR #269, opened 2016-06-18
Feature/ux 68

vivek:
Related to issue # . 

Summary of this pull request: 

Adds following main features:
- Navigable API (implementation of HAL spec). Each generated resource response includes _link object that includes reference to this resource served as well as other navigable resources
- Class or capability discovery of resources served by BlueOcean API

@reviewbybees


michaelneale:
:bee:. A little nervous about something so big but it seems solid.

PR #278, opened 2016-06-19
Fixes JENKINS-36074

marcesher:
Related to issue # JENKINS-36074. 

Summary of this pull request: 

The change to `getCommitId()` resolves a null pointer exception for an
edge case where a job with an SCM configuration was run, but the run
failed because the SCM configuration was faulty (bad URL, unable to
connect to private repo, etc)

The change to `getActions()` is unrelated to JENKINS-36074, but I needed
to fix that after pulling the latest changes in order to see the
`getCommitId()` fix working. I do not understand why an action would be
`null`; this behavior seems to have been introduced in the past few days,
as it did not occur when I originally fixed `getCommitId()` a few days ago

@reviewbybees

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
@@ -202,7 +202,7 @@ public class AbstractRunImpl<T extends Run> extends BlueRun {
     public String getCommitId(){
         BuildData data = run.getAction(BuildData.class);
 
-        if (data == null
+        if(data == null
             || data.getLastBuiltRevision() == null
             || data.getLastBuiltRevision().getSha1String() == null) {
             return null;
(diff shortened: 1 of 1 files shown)


imeredith:
@marcesher Thanks for PR. 

I'm happy for the action null check to be here, but if its possible for you to attach a debugger and find out what is contributing the null please do so. I've looked at all impls I can find and they return empty collections rather than nulls. So its probably some specific plugin that is breaking the contract.

marcesher:
@imeredith thanks!

So I wasn't able to find the specific action, but I believe I've gotten to the bottom of it.

We have jobs in our Jenkins that are disabled and haven't been run in a very long time (a few years; bad hygiene that we'll be taking care of soon). I think what's happened is that since the last time those jobs ran, Jenkins configuration has changed a ton, and those old builds have actions that don't correspond to any currently installed plugins.

If I re-run those old jobs that fall into this category, then even if the job fails, blueocean works as expected.

I'm still uncertain whether the problem is in the generic case of a plugin/action being removed from Jenkins *since* the last build of a job, or whether it's that these jobs are so old that their build.xml for the last build is fundamentally different in some way that causes this null behavior.

Definitely a crazy edge case, but one I suspect I'd probably not be alone in having.

For what it's worth, it is curious to me that blueocean doesn't filter out disabled jobs (and, actually, non-pipeline jobs) from the pipeline screen, though if it's trying to emulate the current main jenkins screen, I can see that.

imeredith:
@marcesher Thanks for investigating. I agree that its likely to happy to others so having the null check here is the correct place. Never know what plugins/configuration users will have.

We are still in the early stages of BlueOcean so there are a number of things we have yet to do and/or missed. Also some of the less used features that Jenkins supports blueocean may never support, at least at a core level. I have created JENKINS-36081 to track this. Cheers.

PR #279, opened 2016-06-19
[JENKINS-35797] API for running pipelines

imeredith:
@vivek I can't get it use WebMethods on /. Get 404's

@reviewbybees

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
@@ -1,29 +1,18 @@
 package io.jenkins.blueocean.service.embedded.rest;
 
-import com.google.common.base.Predicate;
-import com.google.common.collect.Collections2;
-import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import hudson.model.BuildableItem;
-import hudson.model.Cause;
-import hudson.model.CauseAction;
 import hudson.model.Job;
-import hudson.model.ParametersAction;
 import hudson.model.Queue;
-import hudson.model.StringParameterValue;
-import hudson.model.queue.ScheduleResult;
 import io.jenkins.blueocean.commons.ServiceException;
 import io.jenkins.blueocean.rest.hal.Link;
 import io.jenkins.blueocean.rest.model.BluePipeline;
 import io.jenkins.blueocean.rest.model.BlueQueueContainer;
 import io.jenkins.blueocean.rest.model.BlueQueueItem;
 import jenkins.model.Jenkins;
-import org.jenkinsci.plugins.github.util.FluentIterableWrapper;
 
-import javax.annotation.Nullable;
 import java.util.Iterator;
 import java.util.List;
-import java.util.NoSuchElementException;
 
 /**
  * @author Ivan Meredith
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
@@ -1,17 +1,28 @@
 package io.jenkins.blueocean.service.embedded.rest;
 
+import com.google.common.base.Predicate;
+import com.google.common.collect.Collections2;
+import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import hudson.model.BuildableItem;
+import hudson.model.Cause;
+import hudson.model.CauseAction;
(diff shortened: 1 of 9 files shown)


vivek:
@imeredith as discussed on hangout, here is the list of things needed:

- Use @TreeResonse and remove @JsonRespone and @JsonResponseFilter. These bypass stapler processing and are meant for pure POJO json serialization
- s/@WebMethod(name = "new")/@WebMethod(name = "")
- Move new run creation from POST /queue/ to POST /runs/ API

I will leave other comments on the commits.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueQueueContainer.java
```diff
@@ -1,7 +1,46 @@
...
+
 /**
  * @author Ivan Meredith
  */
 public abstract class BlueQueueContainer extends Container<BlueQueueItem> {
+    @WebMethod(name = "new")
+    @JsonResponse
+    @JsonOutputFilter(excludes = "state")
+    @POST
+    public abstract BlueQueueItemCreateResponse createItem();
```
vivek: Following how we are naming get(String id) for getting an item, we should simply call it create(), Item is redundant.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueQueueItem.java
```diff
@@ -15,8 +15,7 @@
  *
  * @author Ivan Meredith
  */
-@ExportedBean
-public abstract class BlueQueueItem {
+public abstract class BlueQueueItem extends Resource {
 
     public static final String QUEUED_TIME = "qeueudTime";
```
vivek: s/qeueudTime/queuedTime.

You may like to change it in examples, tests etc.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
```diff
@@ -36,6 +47,37 @@ public BlueQueueItem get(String name) {
...
+               if (queueItem == null) {
+                   throw new ServiceException.UnexpectedErrorException("The queue item does not exist in the queue");
+               } else {
+                   return new BlueQueueItemCreateResponse(queueItem);
+               }
+           } else {
+               throw new ServiceException.UnexpectedErrorException("Queue item request was not accepted");
+           }
+        } else {
+            throw new ServiceException.NotImplementedException("This pipeline type does not support being queued.");
```
vivek: Would it better to give the name of job in the message?

vivek:
@imeredith Left few comments and with the discussed refactoring it should be good to go.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
```diff
@@ -35,6 +46,44 @@ public BlueQueueItem get(String name) {
         return null;
     }
 
+    /**
+     * Schedules a build. If build already exists in the queue and the pipeline does not
+     * support running multiple builds at the same time, return a reference to the existing
+     * build.
+     *
+     * @return Qeueu item.
```
tfennelly: typo

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueQueueContainer.java
```diff
@@ -1,7 +1,46 @@
...
+            return item.getPipeline();
+        }
+
+        public String getQueuedTime() {
+            return item.getQueuedTimeString();
+        }
+
+        public int getExpectedBuildNumber() {
+            return item.getExpectedBuildNumber();
+        }
```
tfennelly: Why do we return this? The `Run` instance now (since middle/late 2015) has the queueId on it for the queue item that caused the `Run`, allowing you to reliably figure out the corresponding `Run`. There's also the SSE events for the client side. So ... seeing as this seems a bit dubious (what happens if the "unexpected" happens?), then I'm not sure I get why would expose it.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueQueueItem.java
```diff
@@ -12,6 +17,8 @@
  */
 public abstract class BlueQueueItem extends Resource {
 
+    public static final String QUEUED_TIME = "qeueudTime";
```
tfennelly: typo

michaelneale:
@imeredith there is some feedback above

vivek:
@imeredith LGTM 🐝

michaelneale:
🐝

PR #280, opened 2016-06-20
[JENKINS-35813] Run a new build from the Branch tab and Pull request tab

tfennelly:
Related to issue [JENKINS-35813](https://issues.jenkins-ci.org/browse/JENKINS-35813). 

Working on Branches tab ... not yet working on PRs tab.

__TODO__:

- [x] Get it working on the PRs tab. Should be just a case of dropping in the same `RunPipeline` component.
- [x] Add toast popup. When job starts. Not sure this is really needed because the branch status icon changes anyway i.e. indicating that the build tas started.
- [x] Use proper scalable SVG/font icon. Currently just using some CSS.
- [x] Changing to using the proper REST API for starting the run (when [JENKINS-35797](https://issues.jenkins-ci.org/browse/JENKINS-35797) is done). Currently using a temp hack impl.
- [x] Demo video. See below.
- [x] Find out why `isomorphic-fetch` not working for `POST`s. Currently using `XMLHttpRequest` to post to the API.
- [x] Check other TODOs in the changeset and make sure none outstanding or JIRAs created if not doing in this PR.
- [x] Multibranch acceptance test. Doing as a separate task captured in [JENKINS-36177](https://issues.jenkins-ci.org/browse/JENKINS-36177).

__Demo video__ ...

[Run Pipeline](https://youtu.be/kdT02vPTxQI)


tfennelly:
Darn ... some linting errors I think ... battery dead though ... will fix later.

michaelneale:
looking gooooooooood!

blueocean-dashboard/src/main/js/components/Branches.jsx
```diff
@@ -37,17 +38,18 @@ export default class Branches extends Component {
...
             </td>
-            <td>{decodeURIComponent(name)}</td>
-            <td><CommitHash commitId={commitId} /></td>
-            <td>{msg || '-'}</td>
-            <td><ReadableDate date={endTime || ''} /></td>
+            <td onClick={open}>{decodeURIComponent(name)}</td>
+            <td onClick={open}><CommitHash commitId={commitId} /></td>
+            <td onClick={open}>{msg || '-'}</td>
+            <td onClick={open}><ReadableDate date={endTime || ''} /></td>
+            <td><RunPipeline organization={organization} pipeline={pipelineName} branch={name} /></td>
```
scherler: onClick all other for the last exception, bummer

tfennelly: Can't be avoided from what I can see.

scherler:
looks very good @tfennelly the only thing that concerns me that it bypassing the redux store. I would have based the toast notification onto a redux.state object but yeah need to dive more into your solution to understand.

tfennelly:
@scherler maybe elaborate on how you see the redux store being used here and why it "concerns" you that it's not used.

I thought about this and thought the redux store didn't make sense in this case. From what I have seen, it seems more appropriate to reading data from the server and managing it Vs posting "commands" to the server. How it's implemented at the moment using `fetch` is very simple and intuitive. I really would not like to see that getting lost in a mirage of difficult-to-follow redux boilerplate code split across multiple modules.

blueocean-dashboard/src/main/js/api/Branch.js
```diff
@@ -0,0 +1,94 @@
...
+        this.name = name;
+        this.sseListeners = [];
+    }
+
+    runDetailsRouteUrl(runId) {
+        if (runId === undefined) {
+            throw new Error('Branch.runDetailsRouteUrl must be supplied with a "runId" parameter.');
+        }
+        return `/organizations/${this.pipeline.organization}/${this.pipeline.name}/detail/${this.name}/${runId}/pipeline`;
+    }
```
cliffmeyers: I added UrlUtils.buildRunDetailsUrl in a past PR to handle this so it would be good to use it here too. I'm working on adding more methods to it so we can centralize URL building logic. Lack of URI encoding has caused a few bugs before.

blueocean-dashboard/src/main/js/components/Branches.jsx
cliffmeyers: Is the problem here that clicking on the RunPipeline button was also triggering an open?

cliffmeyers: If so, did you look into adding a `stopPropagation()` call inside the RunPipeline component? I haven't used React's event wrappers before but I think that would be a cleaner solution:
https://facebook.github.io/react/docs/events.html

blueocean-dashboard/src/main/js/components/RunPipeline.jsx
```diff
@@ -0,0 +1,99 @@
...
+
+export default class RunPipeline extends Component {
+
+    constructor(props) {
+        super(props);
+        const pipeline = new Pipeline(props.organization, props.pipeline);
+        this.branch = new Branch(pipeline, props.branch);
+        this.state = {
+            toast: undefined,
+        };
```
cliffmeyers: Small style thing but I usually prefer to initialize "empty" values to `null` so I can differentiate between it and a value which is truly undefined.

blueocean-dashboard/src/main/less/run-pipeline.less
```diff
@@ -0,0 +1,35 @@
...
+    // Need to set higher specificity rules for some of the
+    // Toast styles because they are being overridden by other
+    // styles in JDL (table styles).
+    .toast {
+      a.action {
+        color: @blueocean-blue;
+      }
+      .text {
+        white-space: normal;
+      }
```
cliffmeyers: Could you explain the changes in the styles here? I think we want to use the same blue color for all links in the Toast, so if the design spec changed, I think we should update the JDL. Similar as far as far the text, there were some specific rules in the JDL about max width of text.

tfennelly: @cliffmeyers absolutely agree. The reason for adding it is in the comment before it. Basically ... the `Toast` component rendering in the DOM inside a `<table>` and the `<table>` has JDL styles that are overriding the `Toast` styles. Simple as that.

I had a quick look at the JDL styles in `toast.less` and can see that it does not specify these styles at all so is just inheriting. I will add a JDL JIRA to have these added + a comment on this code in this PR pointing to the JIRA.

blueocean-dashboard/src/main/js/api/Branch.js
tfennelly: I'll have a look at this. Thanks for flagging.

blueocean-dashboard/src/main/js/components/Branches.jsx
tfennelly: Hmmm ... not sure ... personally, I prefer things to be explicit even at the expense of being a little more verbose. Having the `onClick` on each `<td>` is not something that bothers me as it is explicit.

That said ... it it's something people feel strongly about then I'll def use that.

blueocean-dashboard/src/main/js/components/RunPipeline.jsx
```diff
@@ -0,0 +1,99 @@
...
+                    toast: {
+                        text: `Started "${theBranch.name}" #${event.jenkins_object_id}`,
+                        action: {
+                            label: 'Open',
+                            callback: () => {
+                                const runDetailsUrl = theBranch.runDetailsRouteUrl(event.jenkins_object_id);
+                                reactContext.location.pathname = runDetailsUrl;
+                                reactContext.router.push(runDetailsUrl);
+                            },
+                        },
```
cliffmeyers: It feels a little strange to me that the Toast actually becomes a child of the button, although using `position: fixed` is a pretty easy way of getting it to the right spot on the screen. I know @i386 had some specific requirements about what happens if multiple toasts start to fire off that might complicate this approach. For now I think this ok, but we may need to refactor at some point if we need to handle multiple toasts or get fancy w/ transitions of them.

blueocean-dashboard/src/main/less/run-pipeline.less
cliffmeyers: Ah you're right, I missed that part specifically about **table styles**. Almost thinking we should just `!important` the toast styles from the JDL.

blueocean-dashboard/src/main/js/components/RunPipeline.jsx
tfennelly: Yeah I agree but didn't see an obvious alternative.

In reality it seems to me like the `Toaster` thing you talked about is the right way to go. Having to add render components to popup a message/alert seems wrong to me.

cliffmeyers: I think in the interest of getting the PR across the finish line this is okay but I do expect we'll need to refactor. I think ultimately we probably want that `Toaster` component to live in the main body of the application, and push toasts into it via a Redux action. The `Toaster` could handle automatically dismissing them after the specified time.

tfennelly: @cliffmeyers exactly, we can iterate towards it.  What's in this is fine for now until we "buy a toaster" 😉

cliffmeyers:
I think once we refactor toward the discussed `Toaster` impl it will address @scherler 's comment and all will be good. A lot of work here - nice job Tom. 🐝

scherler:
@cliffmeyers yes, what you have set that the toast are triggered by redux store and get in a central spot, what my comment was about. Thanks for the review!

PR #282, opened 2016-06-21
alpha version string, and some scm details

michaelneale:
Related to issue #JENKINS-36101 

I wish we didn't have to list contributors like this, it should be from scm. 

@reviewbybees


i386:
🐝 my unfortunate history with maven came in handy

michaelneale:
deploy on sandbox

tfennelly:
🐝

PR #289, opened 2016-06-22
[feature/JENKINS-35841] basic working version of growing logs - fetch…

scherler:
Related to issue # . /JENKINS-35841

Summary of this pull request: 
.
. working version of growing logs fetching every 3 sec. Maybe better to fetch based on sse.

@reviewbybees 

__Demo videos__ ...

[Tailing logs](https://youtu.be/vldz80GL3yw)
[Tailing logs](https://youtu.be/s5idUQ8I9JY)

…


blueoceandeploy:

michaelneale:
very exciting!

i386:
So much win!!

tfennelly:
Cool Thor. A bit of guesswork required wrt how it's working (in the absence of any comments ;) ).

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -210,7 +207,31 @@ exports.fetchJson = function fetchJson(url, onSuccess, onError) {
...
+    let refetchUrl;
+    if (start === null) {
+        refetchUrl = url;
+    } else {
+        refetchUrl = `${url}?start=${start}`;
+    }
+    fetch(refetchUrl, fetchOptions)
+        .then(checkStatus)
+        .then(response => {
+            if (response.headers.get('X-More-Data')) {
```
amuniz: Who is adding this header at server side?

scherler: I guess the LogRestImpl but @vivek or @imeredith can comment on that better

scherler:
@tfennelly cheers for the headsup,, added now some comments to explain the code.

scherler:
@michaelneale @i386  I implemented the scrollToBottom now, I activated it for running result and as well for failure, if a step has failed and has long log we scroll to the bottom (you can see it in the first movie), should I leave it in case of error or do you prefer I remove it again?

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -212,6 +209,50 @@ exports.fetchJson = function fetchJson(url, onSuccess, onError) {
...
+             * If X-More-Data is true, then client should repeat the request after some delay.
+             * In the repeated request it should use X-TEXT-SIZE header value with start query parameter.
+             */
+            if (response.headers.get('X-More-Data')) {
+                /*
+                 * X-TEXT-SIZE is the byte offset of the raw log file client should use in the next request
+                 * as value of start query parameter.
+                 */
+                const newStart = response.headers.get('X-TEXT-SIZE');
+                setTimeout(exports.fetchLogsReFetch(url, newStart, onSuccess, onError), 1000);
```
kzantow: Will this continue indefinitely, even if the user isn't watching the logs?

michaelneale: @scherler @kzantow yes this is a good point, it must only happen if they are on that screen. Obviously we can't tell if there eyes are on the screen (although there are apis for web cams, I am sure) ;)

michaelneale:
@amuniz the header is added by the blueocean server side, which is the same as what classic jenkins does (I believe it adds the same named header, as it is really just pass through).

blueocean-dashboard/src/main/js/redux/actions.js
scherler: @michaelneale yes it is a good point from @kzantow I am half way through with a solution but it is time to hit the hay. ;) ...anyway did you see the videos?

michaelneale: @scherler I will take a look. Great stuff, if you are aware of this that is fine with me. Looking good.

michaelneale:
deploy to sandbox

i386:
@scherler watched the videos - amazing stuff!

One thing I noticed in the second video with the larger log is the bottom of the window jumps around quite a lot. Could you make that smoother?

michaelneale:
videos look good. I have build this locally but spotted a problem with it following along this pipeline: 

```
stage "first"
  echo "r"
  node {
      sh "sleep 5s"
  }
  
stage "first2"
  echo "r"
  node {
      sh "sleep 5s"
  }
  
stage "first3"
  echo "r"
  node {
      sh "sleep 5s"
  }
  
stage "second"
node {
    sh "echo yeah"    
}
```

It never leaves the first stage, although header status is updated correctly.

michaelneale:
FYI for reviewers, this feature is specifically for log tailing of a single stage: https://issues.jenkins-ci.org/browse/JENKINS-35841

For following along to the next stage, aka "Karaoke mode" there is https://issues.jenkins-ci.org/browse/JENKINS-36211

scherler:
@i386 regarding the jumping around I need to look at this later on. I think the karaoke ticket is more important to start with now, so we have a basic experience to "sing along" a pipeline.

i386:
@scherler I'm ok with that but I would still like it addressed as part of the overall liveness work

michaelneale:
looking good - so this isn't fetching always unless it is open as per @kzantow's feedback?

blueocean-dashboard/src/main/js/components/LogConsole.jsx
```diff
@@ -14,50 +14,82 @@ export default class LogConsole extends Component {
...
+                this.updateScroll();
+            }, INITIAL_RENDER_DELAY);
+        }
+    }
 
-        if (data && data.split) {
-            lines = data.split('\n');
+    // Find the modal view container and adopt the scrollTop to focus the end
+    updateScroll() {
+        const nodes = document.getElementsByClassName('content');
```
kzantow: `.findDOMNode()` ?

scherler: would have use a @ref but that is not possible.  What is wrong with gEBCN?

scherler:
@i386 yeah but please enhancement, if possible. E.g. while I am doing the karaoke, someelse can fix that meanwhile I will not be on the logConsole compent.

@michaelneale yeah, it only fetches while visible and removes all timeouts on unMount.

Can I get a :bee: :bee:  https://pbs.twimg.com/media/Buc6whmCAAASCgs.jpg

michaelneale:
ok 🐝

i386:
@scherler I raised https://issues.jenkins-ci.org/browse/JENKINS-36231

blueocean-dashboard/src/main/js/components/LogConsole.jsx
kzantow: There could be multiple instances of things with that class name?

PR #290, opened 2016-06-22
[JENKINS-35863] Refactor log functions and add the ability to download it

imeredith:
@vivek 

@reviewbybees

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/LogResource.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/LogResource.java
@@ -1,5 +1,6 @@
 package io.jenkins.blueocean.service.embedded.rest;
 
+import com.google.common.base.Strings;
 import hudson.console.AnnotatedLargeText;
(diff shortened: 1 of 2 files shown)


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/LogResource.java
```diff
@@ -28,6 +28,20 @@ public void doIndex(StaplerRequest req, StaplerResponse rsp, @Header("Accept") A
         writeLog(req,rsp,accept);
     }
 
+    /**
+     * Downloads the log file in its entirety. It sets the Content-Disposition header so that browsers will download it.
+     */
+    public void doDownload(StaplerRequest req, StaplerResponse rsp) {
```
vivek: I would rather have it handled by a query parameter, for example: /log/?download=true. Any reason we want to not do this way? This will allow users to download any portion of the log file. For example, to download complete log, /log/?start=0&download=true. to download from 1MB offset: /log/?start=1000000&download=true

vivek:
@imeredith LGTM 🐝

PR #291, opened 2016-06-23
[FIX JENKINS-35990] refactor js extensions

kzantow:
Related to issue #JENKINS-35990 .

*NOTE: BUMP js-extensions version to 0.0.16 when finalizing*

Summary of this pull request: 
. Refactor Extension code to ES6 & add tests
. Add support for types & inheritance to extensions
. Add test result component, which uses type inheritance to display JUnit results (I can move this to another PR, but it's directly dependent on this)

Included here (as it is the driving example of how this extensibility feature works):
[Test Result Video](https://www.youtube.com/watch?v=QEeSOBCYY4o)

@reviewbybees esp. @tfennelly


blueoceandeploy:

blueocean-dashboard/src/main/js/components/Pipelines.jsx
```diff
@@ -56,7 +56,7 @@ export default class Pipelines extends Component {
                 </PageHeader>
                 <main>
                     <article>
-                        <ExtensionPoint name="jenkins.pipeline.list.top" />
+                        <Extensions.Renderer name="jenkins.pipeline.list.top" />
```
michaelneale: This is a bit of a huge change, I don't think anyone will like this. I don't.

blueocean-web/src/main/js/main.jsx
```diff
@@ -26,7 +26,7 @@ class App extends Component {
             <div className="Site">
                 <div id="outer">
                     <header className="global-header">
-                        <ExtensionPoint name="jenkins.logo.top"/>
+                        <Extensions.Renderer name="jenkins.logo.top"/>
```
michaelneale: This is a pretty huge naming change. The term "ExtensionPoint" is pretty well known. It is it still used in the yaml for example to specify where an extension plugs in. But it seems with this there is no more ExtensionPoint? just a Renderer (or whatever type it is). 

I don't think this will be very clear for people scanning the code. Perhaps a better name would be Extensions.RendererExtensionPoint or something? ExtensionPoint.Renderer or something? 

"Extensions.Renderer" is wrong - as Renderer is not an Extension (singular) - but a type of extension point right?

michaelneale:
"[ERROR] npm ERR! No compatible version found: @jenkins-cd/js-extensions@0.0.16" is failure. 

I like the typed extension point, but 🐛 on the exact terminology of "Extensions.Renderer" (it isn't an extension, it is an extension point). 

Looks like its in a good direction though, I like this, but needs someone else with more brain to review when ready.

kzantow:
@michaelneale it was recommended to change the _component_ name by @tfennelly and @kohsuke . The extension point itself is not necessarily just the component used to render it. For example, we might export a non-react object at some point and allow plugins to provide implementations to the UI, we actually do already for the stores. You can always name it something else, that's the beauty/ugliness of es6 import syntax (e.g. `import { Renderer as ExtensionRenderer } from '@jenkins-cd/js-extensions'`). Unclear what you don't like about the syntax, it's basically a namespace. We should be doing this for jdl, anyway, I think as it cleans up the imports a bit and makes things more consistent (since you can't rename the components during import), e.g.:

```javascript
import jdl from '@jenkins-cd/jenkins-design-language';
...
<jdl.Table ...>
<jdl.ResultRow ...>
```

Regarding the missing NPM dep, yes, that's an issue... the changes are in this PR, I'll work on separating it to get reviewed/published separately so this will pass.

blueocean-web/src/main/js/main.jsx
kzantow: Specifically `Extensions.Renderer` just refers to the component to render extensions. I wrote up docs based on the current set of changes: https://github.com/kzantow/blueocean-plugin/blob/01bf117c9f1bf792b13fa7b3be6e9f21ac5cbc68/js-extensions/EXTENSIONS.md#rendering-extension-points

kzantow: ... maybe you'd prefer the import just referred to the library as `ExtensionPoint`? And you'd have an `ExtensionPoint.Renderer` as the component to render them?

```javascript
import ExtensionPoint from '@jenkins-cd/js-extensions';
...
<ExtensionPoint.Renderer name="jenkins.logo.top" ... />
```

I'm sort of indifferent about this detail. Except for the `name` property which is inconsistent with `extensionPoint` in the yaml file.

michaelneale: @kzantow yes, that is really my only suggestion, that it is called out as an extension point (ie "things plugin here") - but its terminology. IF there was a group of your working on this and it made sense to you, then disregard my comment. The concept is great. And Extensions could be imported as ExtensionPoints or something like you say, via ES6, then that is ok. 

It is a little inconsistent with the naming in the config file as you say otherwise.

michaelneale:
looking good. 

Not sure why this hasn't triggered a build, unless it was cancelled.

michaelneale:
Getting close - now a test failure: io.jenkins.blueocean.jsextensions.JenkinsJSExtensionsTest.test

michaelneale:
@tfennelly I think this is good to go - 🐝  from me, but I think would be good to get your eyes on it.

michaelneale:
@kzantow oops looks like needs an update and re-apply (some changes in the meantime)

tfennelly:
@kzantow oops sorry I missed the notification for this ... will have a look ASAP. Need to help @scherler with something right now, but will def have a look later.

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -757,9 +766,37 @@ export const actions = {
...
     },
+    
+    fetchTestResults(config, runDetails) {
+        return (dispatch) => {
+            // const appUrlBase = config.getAppURLBase(); // TODO: Should use this instead
+            const appUrlBase = '/jenkins';
+            let url;
+            
+            if (runDetails.isMultiBranch) {
+                url = 'TODO'; // Construct different URL for multi-branch
```
scherler: That is not correct, right?

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -757,9 +766,37 @@ export const actions = {
...
+            } else {
+                // eslint-disable-next-line max-len
+                url = `${appUrlBase}/blue/rest/organizations/${runDetails.organization}/pipelines/${runDetails.branch}/runs/${runDetails.runId}/testReport/result`;
+            }
```
cliffmeyers: Check out the `buildRunDetailsUrl` function in `UrlUtils` for a utility to help build this URL so that it is fully escaped.

blueocean-dashboard/src/main/js/redux/actions.js
kzantow: Right, :bug: myself for this

kzantow: ... and fixed

js-extensions/src/ExtensionStore.js
```diff
@@ -0,0 +1,316 @@
...
+        var checkLoading = () => {
+            if (loadCountMonitor.counter === 0) {
+                onload(extensionPointMetadatas);
+            }
+        };
+    
+        // Iterate over each plugin in extensionPointMetadata, async loading
+        // the extension point .js bundle (if not already loaded) for each of the
+        // plugins that implement the specified extensionPointId.
+        for(var i1 = 0; i1 < this.extensionPointList.length; i1++) {
```
scherler: this.extensionPointList.map((extensionPoint, index)=>{//...} ?

kzantow: no?

michaelneale: I think a cosmetic thing: http://stackoverflow.com/questions/9338548/is-using-map-instead-of-a-for-loop-a-bad-idea-in-js

But a preference thing IMO.

michaelneale:
If no objections, 🐝 . This is an important change which needs to make its way in. 

@kzantow will also need to update https://github.com/cloudbees/blueocean-sample-pipeline-result-ext-plugin at some point if this goes in.

js-extensions/src/ExtensionStore.js
scherler: yes, agree, but we use es6 in all the rest of the PR and IMO we should be consistent with it (and all the good reasons you linked ;))

kzantow: es6 supports traditional for loops :)

tfennelly:
Mixing PRs makes it quite hard to see the `js-extensions` refactorings Vs the junit test feature 😭. I'll check out the code locally and try review via my IDE.

kzantow:
@tfennelly right, sorry about that, would have made a lot more unnecessary work to split it up as they sorta needed to be developed together.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
```diff
@@ -115,10 +115,10 @@ export class RunDetailsPipeline extends Component {
         }
         return (
             <div>
-                { nodes && nodes[nodeKey] && <ExtensionPoint
+                { nodes && nodes[nodeKey] && <Extensions.Renderer
```
tfennelly: I like the rename.

blueocean-dashboard/src/main/js/components/testing/TestResults.jsx
```diff
@@ -0,0 +1,126 @@
...
+            {duration}
+        </div>
+        <div className="test-console">
+            <h4>Error</h4>
+            <div className="error-message">
+                {t.errorDetails}
+            </div>
+            <h4>Output</h4>
+            <div className="stack-trace">
+                {t.errorStackTrace}
```
tfennelly: `expandable` already has this?

blueocean-dashboard/src/main/js/components/testing/TestResults.jsx
```diff
@@ -0,0 +1,126 @@
...
+                    status: 'PASSED',
+                }} />),
+            ];
+        }
+
+        if (newFailures.length > 0 || existingFailures.length > 0) {
+            if (newFailures.length === 0) {
+                newFailureBlock = [
+                    <h4>New failing - {newFailures.length}</h4>,
+                    <div className="">No new failures</div>,
```
tfennelly: `className` has no value?

blueocean-dashboard/src/main/js/components/testing/TestResults.jsx
```diff
@@ -0,0 +1,126 @@
...
+                suites.map((t, i) => <TestCaseResultRow key={i} testCase={{
+                    className: `${t.cases.filter(c => c.status === 'PASSED').length} Passing`, // this shows second
+                    name: t.name,
+                    duration: t.duration,
+                    status: 'PASSED',
+                }} />),
+            ];
+        }
+
+        if (newFailures.length > 0 || existingFailures.length > 0) {
```
tfennelly: `existingFailures` isn't used in this block, right?

kzantow: No, not used. But I want to show the `newFailures` block in the case there are new failures or existing failures, to make it clear if a given build has no _new_ failures.

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -772,16 +783,16 @@ export const actions = {
...
+                  stepBaseUrl,
+                  (json) => {
+                      const information = getNodesInformation(json);
+                      information.nodesBaseUrl = stepBaseUrl;
+                      return dispatch({
+                          type: ACTION_TYPES.SET_STEPS,
+                          payload: information,
+                      });
+                  },
+                  (error) => console.error('error', error) // eslint-disable-line no-console
```
tfennelly: Is `eslint-disable` needed for `console.error` ?

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -824,4 +835,31 @@ export const actions = {
...
+    fetchTestResults(config, runDetails) {
+        return (dispatch) => {
+            const baseUrl = `${config.getAppURLBase()}/rest/organizations/`;
+            let url;
+            if (runDetails.isMultiBranch) {
+                // eslint-disable-next-line max-len
+                url = `${baseUrl}${buildUrl(runDetails.organization, 'pipelines', runDetails.pipeline, 'branches', runDetails.branch, 'runs', runDetails.runId)}/testReport/result`;
+            } else {
+                // eslint-disable-next-line max-len
+                url = `${baseUrl}${buildUrl(runDetails.organization, 'pipelines', runDetails.branch, 'runs', runDetails.runId)}/testReport/result`;
```
tfennelly: You sure these URLs are encoded ? 😉

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/jsextensions/JenkinsJSExtensions.java
```diff
@@ -0,0 +1,203 @@
...
+        if (incoming instanceof BlueExtensionClass) {
+            BlueExtensionClass in = (BlueExtensionClass)incoming;
+            Map m = new HashMap();
+            Object value = mergeObjects(in.getClasses());
+            m.put("_classes", value);
+            return m;
+        }
+        return incoming;
+    }
+}
```
tfennelly: Was this moved or what?

kzantow: Yes, needed access to different classes.

blueocean-dashboard/src/main/js/redux/actions.js
kzantow: Yes, `buildUrl` ensures all components are URL encoded ;)

js-extensions/EXTENSIONS.md
```diff
@@ -0,0 +1,62 @@
...
+
+    import Extensions from '@jenkins-cd/js-extensions';
+    ...
+    <Extensions.Renderer extensionPoint="jenkins.navigation.top.menu" />
+
+For example, rendering the test results for a build may be scoped to the specific type of test results in this manner:
+
+    <Extensions.Renderer extensionPoint="test-results-view" dataType={data._class} testResults={data} />
+
+The `ExtensionRenderer` component optionally uses the [classes API](../blueocean-rest/README.md#classes_API) to look up an appropriate, specific set of views for the data being displayed. This should works seamlessly with other [capabilities](../blueocean-rest/README.md#capabilities).
```
tfennelly: `Extension.Renderer` ?

tfennelly: or `Extensions.Renderer ` ?

kzantow: ... in the code, there's an `ExtensionsRenderer` class, exported as `Renderer`. Typically someone should follow the `import Extensions from ...`, which makes usage of the component `Extensions.Renderer`

js-extensions/EXTENSIONS.md
```diff
@@ -0,0 +1,62 @@
...
+- `Renderer` - a React component to conveniently render extensions
+
+### Store API
+
+The `ExtensionStore` API is very simple, all public methods are asynchronous:
+
+- `getExtensions(extensionPointName, [type,] onload)`
+    This method will async load data and type information as needed, and call the onload handler with a list of extension exports, e.g. the React classes or otherwise exported references.
+
+- `getTypeInfo(type, onload)`
```
tfennelly: Does it really expose `getTypeInfo`? Does it need to? I'd think that should be a separate "Classes" api that extensions depends on.

js-extensions/spec/ExtensionStore-spec.js
```diff
@@ -0,0 +1,231 @@
...
+            expect(ep1.length).to.equal(3);
+            expect(ep1).to.include.members(["component-1.1","component-1.2","component-2.1"]);
+
+            expect(ep2.length).to.equal(3);
+            expect(ep2).to.include.members(["component-1.3","component-2.2","component-2.3"]);
+        });
+        
+        done();
+    });
+});
```
tfennelly: Was this moved from somewhere?

tfennelly: Oh changed to use mocha I think.

js-extensions/src/ExtensionRenderer.jsx
```diff
@@ -1,113 +1,111 @@
...
+    _setExtensions() {
+        ExtensionStore.instance.getExtensions(this.props.extensionPoint, this.props.dataType,
+            extensions => this.setState({extensions: extensions})
+        );
+    }
+    
     /**
      * This method renders the "leaf node" container divs, one for each registered extension, that live in the same
-     * react hierarchy as the &lt;ExtensionPoint&gt; instance itself. As far as our react is concerned, these are
+     * react hierarchy as the &lt;ExtensionRenderer&gt; instance itself. As far as our react is concerned, these are
```
tfennelly: `Extensions.Renderer` ?

js-extensions/src/ExtensionRenderer.jsx
```diff
@@ -1,113 +1,111 @@
...
             const children = el.children;
             if (children) {
-                const extensions = store.getExtensions(this.props.name);
+                const extensions = this.state.extensions;
 
                 // The number of children should be exactly the same as the number
                 // of extensions. See the render function for where these are added.
                 if (!extensions || extensions.length !== children.length) {
-                    console.error('Unexpected error in Jenkins ExtensionPoint rendering (' + this.props.name + '). Expecting a child DOM node for each extension point.');
+                    console.error('Unexpected error in Jenkins ExtensionRenderer rendering (' + this.props.extensionPoint + '). Expecting a child DOM node for each extension point.');
```
tfennelly: `Extensions.Renderer`?

I think you need to do a search and replace for this. It's name must have gone through a few iterations and old ones are still hanging around.

tfennelly: Ok sorry ... the type name is `ExtensionRenderer` but the react element is `Extensions.Renderer `. Hmmm.

js-extensions/src/ExtensionStore.js
tfennelly: es6 blah blah blah ... 400Mb npm download later

tfennelly: Anyway ... fairly sure that code was there before, so shouldn't really be up for review.

tfennelly:
🐝 but was hard to see the wood from the trees ... unit test, es6, name refactorings and then adding type support.

michaelneale:
:bee:

kzantow:
@reviewbybees done

PR #293, opened 2016-06-24
JENKINS-36207# Make BluePipeline extensible

vivek:
Related to issue # . 

@michaelneale @imeredith 

@reviewbybees

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineImpl.java
@@ -1,13 +1,18 @@
 package io.jenkins.blueocean.service.embedded.rest;
 
+import hudson.Extension;
+import hudson.model.Action;
 import hudson.model.BuildableItem;
 import hudson.model.Item;
 import hudson.model.ItemGroup;
 import hudson.model.Job;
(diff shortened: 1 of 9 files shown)


blueoceandeploy:

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineImpl.java
```diff
@@ -164,4 +177,28 @@ protected static String getRecursivePathFromFullName(BluePipeline pipeline){
         return pipelinePath.toString();
     }
 
+    @Extension(ordinal = 1)
+    public static class PipelineFactoryImpl extends BluePipelineFactory {
```
michaelneale: name could be maybe less redundant? DefaultPipelineFactory? Or just PipelineFactory?

vivek: Well, this is inner class, implementation name, does it matter to be so specific about naming? I would worry about it if it was an interface or contract like what goes in blueocean-rest module.

michaelneale: 👍🏻

michaelneale:
🐝

PR #294, opened 2016-06-24
JENKINS-36154# Link related fixes

vivek:
Related to issue # . 

Summary of this pull request: 
.
.
.

@reviewbybees

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/BranchImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/BranchImpl.java
@@ -1,5 +1,6 @@
 package io.jenkins.blueocean.service.embedded.rest;
 
+import hudson.Util;
 import hudson.model.Job;
 import io.jenkins.blueocean.rest.hal.Link;
 import io.jenkins.blueocean.rest.model.Resource;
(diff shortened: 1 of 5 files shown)


blueoceandeploy:

michaelneale:
:ant: failing test: io.jenkins.blueocean.service.embedded.MultiBranchTest.getMultiBranchPipelineRunChangeSets

vivek:
@michaelneale oops, there was typo, fixed.

scherler:
looks good, but you encode the branch name but a pipeline name could be "a/b" as well, or? If so, wouldn't you need the same for the name?

vivek:
@scherler No, you can't have folder with '/' in them. Only applies to branch name, which could have any char.

michaelneale:
🐝  I think this is good. Will any front end code need to change that was based on incorrect assumptions? if not - this is good to go.

vivek:
@michaelneale I will let @cliffmeyers @scherler confirm that.

cliffmeyers:
If this change requires that the branch name in the REST API be double URL encoded then yes it will have UI impact. I couldn't quite tell from the PR if that was the case, or if it's just updating the Link href's returned in the JSON.

vivek:
@cliffmeyers This PR just fixes Link to return properly encoded url path, UI can use it as it is. I thought this is what you wanted, no?

cliffmeyers:
@vivek yep, if we can follow the links then that's great.

cliffmeyers:
:bee:

michaelneale:
🐝

PR #296, opened 2016-06-27
Applied skeletal implementation to interface migration refactoring.

khatchad:
Related to no particular issue. This is a semantics-preserving refactoring.

Summary of this pull request: 
- We are evaluating a research prototype automated refactoring Eclipse plug-in called [Migrate Skeletal Implementation to Interface](https://github.com/khatchad/Migrate-Skeletal-Implementation-to-Interface-Refactoring). We have applied the tool to your project in the hopes of receiving feedback.
- The approach is very conservative and you should not observe many source code changes.
- The source code should be semantically equivalent to the original
- We have run tests prior to applying to the tool and following the application. All tests pass.

Thank you for your help in this evaluation!

@reviewbybees

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/BranchImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/BranchImpl.java
@@ -81,11 +81,6 @@ public class BranchImpl extends PipelineImpl {
         public String getAuthor() {
             return author;
         }
-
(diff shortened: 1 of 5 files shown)


i386:
@khatchad appreciate the contribution! However, unfortunately we can't use Java 8 in Blue Ocean due to backwards compatibility concerns. We need to maintain compatibility at a minimum with Java 7.

khatchad:
@i386 No problem! Thank you for the consideration.

PR #297, opened 2016-06-28
[FIXED JENKINS-36100] - fallback for DevelopmentFooter version string

kzantow:
Related to issue #JENKINS-36100. 

Summary of this pull request: 
. Add a fallback for DevelopmentFooter, when the json file can't be found

@reviewbybees 

image

### DIFF
--- a/blueocean-web/src/main/js/DevelopmentFooter.jsx
+++ b/blueocean-web/src/main/js/DevelopmentFooter.jsx
@@ -5,8 +5,13 @@ import revisionInfo from '../../../target/classes/io/jenkins/blueocean/revisionI
 
 export class DevelopmentFooter extends Component {
     render() {
-        if (!revisionInfo.name) {
-            return null;
+        if (!revisionInfo || !revisionInfo.name) {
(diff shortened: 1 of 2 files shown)


blueocean-web/src/main/js/DevelopmentFooter.jsx
```diff
@@ -5,8 +5,13 @@ import revisionInfo from '../../../target/classes/io/jenkins/blueocean/revisionI
 
 export class DevelopmentFooter extends Component {
     render() {
-        if (!revisionInfo.name) {
-            return null;
+        if (!revisionInfo || !revisionInfo.name) {
+            var blueOceanVersion = document.getElementsByTagName('head')[0].getAttribute('data-blue-ocean-version');
+            return (
+                <div className="development-footer">
+                    <span>Blue Ocean UI v{blueOceanVersion}</span>
```
kzantow: `{blueOceanVersion}` will potentially be '1.0' or '1.1', etc., it would be too barren without some additional text...

michaelneale:
🐝

PR #298, opened 2016-06-28
[JENKINS-36218] Pubsub event enrichment @ExtensionPoint

tfennelly:
Related to issue [JENKINS-36218](https://issues.jenkins-ci.org/browse/JENKINS-36218). 

Basic gist of the PR ... attach proper blueocean rest API urls to job events so the UI does not have to try manufacture them. The URL should come from REST API code.

Depends on:

* https://github.com/jenkinsci/pubsub-light-module/pull/2
* https://github.com/jenkinsci/sse-gateway-plugin/pull/4

__TODO__

- [x] Get help from @vivek or @imeredith to properly create job rest url link on job events. [See changes to `PipelineImpl.getLink` in this PR](https://github.com/jenkinsci/blueocean-plugin/blob/897e32a35f90b62d82f83535abd880bd413dba74/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineImpl.java#L208-L221).
- [x] Release `pubsub-light-module` and `sse-gateway-plugin` + update deps in this PR.
- [x] Get @michaelneale to do a local sanity check in case @tfennelly missed anything.


blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -551,17 +540,16 @@ export const actions = {
...
+                const jobs = branches[event.blueocean_job_pipeline_name] || [];
+                const branch = jobs.find(job => job.name === event.blueocean_job_branch_name);
 
                 if (!branch) {
                     return;
                 }
 
-                const url = `${config.getAppURLBase()}/rest/organizations/${branch.organization}` +
-                    `/pipelines/${event.blueocean_job_name}/branches/${branch.name}`;
+                const url = `${config.getAppURLBase()}${event.blueocean_job_rest_url}`;
```
kzantow: I assume these URLs are already encoded?

tfennelly: Yes, should already be encoded.

blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
```diff
@@ -0,0 +1,79 @@
...
+    enum BlueEventProps {
+        blueocean_job_rest_url,
+        blueocean_job_pipeline_name,
+        blueocean_job_branch_name,
+    }
+    
+    @Override
+    public void enrich(@Nonnull Message message) {
+        
+        message.set(EventProps.Jenkins.jenkins_org, OrganizationImpl.INSTANCE.getName());
```
kzantow: Should this be set by some ThreadLocal value?

blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
```diff
@@ -0,0 +1,79 @@
...
+        
+        String channelName = message.getChannelName();
+        if (channelName.equals(Events.JobChannel.NAME)) {
+            JobChannelMessage jobChannelMessage = (JobChannelMessage) message;
+            ParameterizedJobMixIn.ParameterizedJob job = jobChannelMessage.getJob();
+            Link jobUrl = PipelineImpl.getLink(job);
+            
+            jobChannelMessage.set(BlueEventProps.blueocean_job_rest_url, jobUrl.getHref());
```
kzantow: ... doesn't look like anything leading up to this is going to encode it

blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
tfennelly: I wouldn't think so, but we should ask @vivek if there is a "proper" way of getting this without dipping into the embedded rest API impl. Maybe the abstract rest API definition should have something for this? I just copied it from somewhere else.

blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
tfennelly: It's coming from a `Link` instance which should handle encoding of the URL tokens if needed. We need @vivek to look at this anyway.

kzantow:
:bee: although, I don't think the links are being encoded anywhere (I scanned the code briefly), but it shouldn't be handled by this set of changes, probably, rather in the `Link`. And I'm concerned about adding in more links to a somewhat hardcoded `jenkins` org, if we expect to support separate orgs, we should sort out a method to call to get it, asap.

blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
```diff
@@ -0,0 +1,79 @@
...
+public class BlueMessageEnricher extends MessageEnricher {
+    
+    enum BlueEventProps {
+        blueocean_job_rest_url,
+        blueocean_job_pipeline_name,
+        blueocean_job_branch_name,
+    }
+    
+    @Override
+    public void enrich(@Nonnull Message message) {
```
vivek: How is BlueMessageEnricher used? That is where does it get called from?

There are assumptions in the code that might not hold in all cases. 

I think it is probably going to work better if we have BO specific MessageEnricher, this way we can work with it cleanly:

```
BlueMessageEnricher extends MessageEnricher {
    public void enrich(@Nonnull Message message, BluePipeline pipeline);
}

// With pipeline, you get things like pipeline.getOrganization(), will work in all cases, like multi organization, pipeline instance of 
// BlueBranch or BlueFolderPipeline or BlueMultiMasterPipeline etc.
// There is code in there that are directly accessing Jenkins Job object, so better to put them in 
// impl to allow other kind of implementations. If we want to keep it default impl, we should try to see if BluePipeline can be used 
// instead of Jenkins Job
MessageEnricherImpl extends BlueMessageEnricher{

    public void enrich(@Nonnull Message message, BluePipeline pipeline){
        ...
    }

}

```

blueocean-events/pom.xml
```diff
@@ -0,0 +1,27 @@
...
+    <packaging>hpi</packaging>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>sse-gateway</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>blueocean-rest-impl</artifactId>
```
vivek: Better to use blueocean-rest than impl. Then it can work with any implementation of blueocean-rest.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineImpl.java
```diff
@@ -200,5 +204,19 @@ public BluePipeline getPipeline(Item item, Reachable parent) {
         return actionProxies;
 
     }
+    
+    // TODO: @vivek/@ivan please fix me the right way !!!
+    public static @Nonnull Link getLink(@Nonnull ParameterizedJobMixIn.ParameterizedJob job) {
```
vivek: I put my comments on BlueMessageEnricher. Need to understand uses of it bit better to further comment on how it works.

blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
tfennelly: @vivek As you can see, `BlueMessageEnricher` is an impl of `MessageEnricher`, which is part of the `pubsub-light` module, which is not Blue Ocean specific.  The `pubsub-light` module gives `MessageEnricher` instances/impls the opportunity to "enrich" event messages before they are published.

It seems reasonable to me that there would be an easy to use REST API way of getting a Blue Ocean URL for a given Jenkins domain model object instance without being in the context of a Stapler request. Are we saying this is not possible atm? Could we make it possible?

blueocean-events/pom.xml
tfennelly: Sure, it would def be better to use `blueocean-rest`, so long as it can give what we need.

tfennelly:
@vivek thanks for commenting. So my question atm is .... is there (or can we create) a way of constructing a BO domain model object from a classic Jenkins domain model object?

blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
vivek: @tfennelly There is nothing today that will give you Link for a given job, but something we can do what you are asking thru proper abstraction that works with any rest implementation. What I was saying is that, if you need it now, you can go ahead with the code sample I gave above to get instance of BluePipeline and get Link and Organization name from it. Maybe I should do it now, and you can remove dependency on rest-impl and do it right way?

tfennelly: @vivek yeah, I think it makes total sense to make this easy. I'd bet the house this is not the last time we'd need something like this so abstracting it properly makes total sense imo and is what I was expecting you'd say.

michaelneale:
@tfennelly that works great - tested with both multibranch and non multibranch. 

With freestyle, it worked, but there was a stacktrace on the server: 

```
WARNING: Error publishing Run queued event.
org.jenkins.pubsub.MessageException: Event enrichment failure due to unexpected exception in io.jenkins.blueocean.events.BlueMessageEnricher.
	at org.jenkins.pubsub.PubsubBus.publish(PubsubBus.java:97)
	at org.jenkins.pubsub.listeners.SyncQueueListener.publish(SyncQueueListener.java:88)
	at org.jenkins.pubsub.listeners.SyncQueueListener.publish(SyncQueueListener.java:82)
	at org.jenkins.pubsub.listeners.SyncQueueListener.onEnterBuildable(SyncQueueListener.java:64)
	at hudson.model.Queue$BuildableItem.enter(Queue.java:2531)
	at hudson.model.Queue$BuildableRunnable.run(Queue.java:2828)
	at hudson.model.Queue.maintain(Queue.java:1483)
	at hudson.model.Queue$1.call(Queue.java:295)
	at hudson.model.Queue$1.call(Queue.java:292)
	at jenkins.util.AtmostOneTaskExecutor$1.call(AtmostOneTaskExecutor.java:101)
	at jenkins.util.AtmostOneTaskExecutor$1.call(AtmostOneTaskExecutor.java:91)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at hudson.remoting.AtmostOneThreadExecutor$Worker.run(AtmostOneThreadExecutor.java:110)
	at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.NullPointerException
	at io.jenkins.blueocean.service.embedded.rest.PipelineImpl.getLink(PipelineImpl.java:220)
	at io.jenkins.blueocean.events.BlueMessageEnricher.enrich(BlueMessageEnricher.java:64)
	at org.jenkins.pubsub.PubsubBus.publish(PubsubBus.java:95)
	... 13 more
```

Not sure if it matters,

blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
tfennelly: @vivek Hey Vivek, I'm afraid that didn't work either. Same reason as before i.e. there's no active `StaplerRequest` instance so `ApiHead.INSTANCE()` throws an NPE.

```
org.jenkins.pubsub.MessageException: Event enrichment failure due to unexpected exception in io.jenkins.blueocean.events.BlueMessageEnricher.
	at org.jenkins.pubsub.PubsubBus.publish(PubsubBus.java:97)
	at org.jenkins.pubsub.listeners.SyncQueueListener.publish(SyncQueueListener.java:88)
	at org.jenkins.pubsub.listeners.SyncQueueListener.onLeft(SyncQueueListener.java:72)
	at hudson.model.Queue$LeftItem.enter(Queue.java:2609)
	at hudson.model.Queue.onStartExecuting(Queue.java:1126)
	at hudson.model.Executor$1.call(Executor.java:360)
	at hudson.model.Executor$1.call(Executor.java:346)
	at hudson.model.Queue._withLock(Queue.java:1371)
	at hudson.model.Queue.withLock(Queue.java:1232)
	at hudson.model.Executor.run(Executor.java:346)
Caused by: java.lang.NullPointerException
	at io.jenkins.blueocean.rest.ApiHead.INSTANCE(ApiHead.java:98)
	at io.jenkins.blueocean.service.embedded.rest.OrganizationImpl.getLink(OrganizationImpl.java:74)
	at io.jenkins.blueocean.service.embedded.rest.PipelineContainerImpl.getLink(PipelineContainerImpl.java:38)
	at io.jenkins.blueocean.service.embedded.rest.PipelineImpl$PipelineFactoryImpl.getPipeline(PipelineImpl.java:190)
	at io.jenkins.blueocean.service.embedded.rest.PipelineContainerImpl.get(PipelineContainerImpl.java:54)
	at io.jenkins.blueocean.service.embedded.rest.PipelineContainerImpl.get(PipelineContainerImpl.java:49)
	at io.jenkins.blueocean.events.BlueMessageEnricher.getLink(BlueMessageEnricher.java:85)
	at io.jenkins.blueocean.events.BlueMessageEnricher.enrich(BlueMessageEnricher.java:66)
	at org.jenkins.pubsub.PubsubBus.publish(PubsubBus.java:95)
	... 9 more
```

michaelneale: spider sense says race condition.

michaelneale:
:bee:

PR #299, opened 2016-06-28
Feature/jenkins 35866 handle folders

cliffmeyers:
Related to issue # JENKINS-35866. 

**NOTE: setting this to WIP since @scherler is going to help test some of the various Pipeline Graph / Log Console scenarios, since the logic for building up the REST URLs underwent changes**

Summary of this pull request: 
* Update router config so that ":pipeline" is now the pipeline's "fullName", URI-encoded
* Update screens to use new URL structure
* Fix a bug where deep-linking to Run Details would fail for jobs nested in folders most of the time (due to timing issue). This would trigger two REST fetches, one for "all pipelines" and one for "org pipelines". Since "org pipelines" does not return nested jobs, if the response arrived after "all pipelines", the selected pipeline would be missing in the "pipelines" store collection. This would cause rendering of PipelinePage to abort.

@reviewbybees


blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
```diff
@@ -38,11 +34,7 @@ class OrganizationPipelines extends Component {
         if (config) {
             const { organization } = this.context.params;
             this.props.fetchPipelinesIfNeeded(this.context.config, organization);
-            if (this.props.params.pipeline) {
-                const { pipeline } = this.props.params;
-                config.pipeline = pipeline;
-                this.props.setPipeline(config);
-            }
+
```
cliffmeyers: The above call to `setPipeline` was triggering a second fetch for pipelines - but only if the user deep-linked to a route with the :pipeline route param. These two calls were causing odd behavior depending on which one completed last. I decided to remove this, and instead let `componentWillReceiveProps()` execute this logic each time. This eliminates the timing issue and ensures the right pipeline is "selected" any time the props change.

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -108,7 +106,7 @@ export const actionHandlers = {
             return state.set('pipeline', null);
         }
         // [].slice(0) returns a clone, we do need it for uniqueness
-        const pipeline = pipelines.slice(0).filter(item => item.name === id);
+        const pipeline = pipelines.slice(0).filter(item => item.fullName === id);
```
cliffmeyers: Now we match based on fullName (which includes folder path).

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -302,13 +300,13 @@ export const actions = {
...
+    // eslint-disable-next-line no-unused-vars
     fetchPipelinesIfNeeded(config, organizationName) {
         return (dispatch, getState) => {
             const pipelines = getState().adminStore.pipelines;
             const baseUrl = config.getAppURLBase();
-            const url = !organizationName ?
-                `${baseUrl}/rest/search/?q=type:pipeline` :
-                `${baseUrl}/rest/organizations/${organizationName}/pipelines/`;
+            // TODO: update this code to call /search with organizationName once JENKINS-36273 is ready
+            const url = `${baseUrl}/rest/search/?q=type:pipeline`;
```
cliffmeyers: If we fetch the pipelines for an organization, right now it doesn't return any jobs that are nested in folders, which basically results in a null pipeline in the store. For now we'll use the full fetch for all pipelines, and eventually tweak this code back to add organizationName as a filter once JENKINS-32673 is complete.

blueocean-dashboard/src/main/js/components/RunDetails.jsx
```diff
@@ -34,32 +38,30 @@ class RunDetails extends Component {
             config.pipeline = pipeline;
 
             this.props.fetchRunsIfNeeded(config);
-            this.props.setPipeline(config);
```
cliffmeyers: This was another extra call to `setPipeline()` that could be problematic in firing off dup calls.

scherler:
@cliffmeyers I had a look on my configured projects and it seems to work just fine. However I did not created a folder yet, to test it with the folder.

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
michaelneale: Good catch!

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
```diff
@@ -108,6 +98,16 @@ class OrganizationPipelines extends Component {
             delete this.jobListener;
         }
     }
+
+    _selectPipeline(pipelines, params) {
```
michaelneale: _ convention for component private methods?

blueocean-dashboard/src/main/js/components/PipelineRowItem.jsx
```diff
@@ -37,7 +38,7 @@ export default class PipelineRowItem extends Component {
         const hasPullRequests = !simple && (
             numberOfSuccessfulPullRequests || numberOfFailingPullRequests);
 
-        const baseUrl = `/organizations/${organization}/${pipeline.name}`;
+        const baseUrl = buildPipelineUrl(organization, fullName);
```
michaelneale: nice

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -33,9 +32,8 @@ export function calculateNodeBaseUrl(config) {
 
 export function calculateStepsBaseUrl(config) {
     const { name, runId, branch, _appURLBase, isMultiBranch, node } = config;
-    let baseUrl =
-        `${_appURLBase}/rest/organizations/jenkins/` +
-        `pipelines/${uriString(name)}`;
+    // TODO: can't hard-code org
+    let baseUrl = `${_appURLBase}/rest/organizations/jenkins/pipelines/${name}`;
```
michaelneale: Maybe should open a ticket speciifcally for this? as it will bite (grepping for /jenkins/ shouldn't be too hard...

michaelneale:
:bee:

Code looks good, nicely documented with tests. And I checked it worked well with pipeline jobs in folders.

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
cliffmeyers: Yep, we aren't consistent with it but I use it in most of my code IIRC.

blueocean-dashboard/src/main/js/redux/actions.js
cliffmeyers: Sure, I'll add one.

PR #303, opened 2016-06-29
Fix issue with extension point bundle loading timing

kzantow:
Related to issue N/A . 

Summary of this pull request: 
. Fix issue with extension point bundle loading timing

@reviewbybees esp. @cliffmeyers 

- [ ] release new version (later)


cliffmeyers:
Tested this and confirmed it works. Nice! 🐝

michaelneale:
nice catch @kzantow :bee:

PR #304, opened 2016-06-29
feature/jenkins 36060 - personalization plugin

cliffmeyers:
Related to issue # JENKINS-36060. 

Summary of this pull request: 
* Create a new Maven module for the personalization plugin
* Demonstrate very simple POC of adding an item action to pipelines list. While testing this I uncovered an interesting bug in js-modules that Keith addressed in cloudbees/blueocean-plugin#303

@reviewbybees


blueocean-personalization/.storybook/config.js
```diff
@@ -0,0 +1,7 @@
+import { configure } from '@kadira/storybook';
```
kzantow: Could all of this somehow be automated with js-modules or another devDependency?

blueocean-personalization/src/main/js/components/ActionLink.jsx
```diff
@@ -0,0 +1,7 @@
+import React, { Component } from 'react';
+
+export default class ActionLink extends Component {
+    render() {
+        return <div>Test</div>;
```
kzantow: This part isn't actually ready for review, is it?

blueocean-personalization/src/main/resources/index.jelly
```diff
@@ -0,0 +1,4 @@
+<?jelly escape-by-default='true'?>
```
kzantow: What's this file for?

blueocean-plugin/pom.xml
```diff
@@ -24,6 +24,10 @@
         </dependency>
         <dependency>
             <groupId>${project.groupId}</groupId>
+            <artifactId>blueocean-personalization</artifactId>
```
kzantow: Need to add this to <modules> section, too?

blueocean-plugin/src/test/java/io/jenkins/blueocean/jsextensions/JenkinsJSExtensionsTest.java
```diff
@@ -40,18 +39,28 @@
     @Test
     public void test() {
         // Simple test of the rest endpoint. It should find the "blueocean-dashboard"
```
kzantow: NIT: 'blueocean-dashboard'

blueocean-plugin/pom.xml
cliffmeyers: Yes, it's added to `<modules>` as well.

kzantow: Ah, whoops, wrong pom :)

kzantow:
LGTM except the `ActionLink.jsx` for now

blueocean-personalization/src/main/js/components/ActionLink.jsx
cliffmeyers: You're right, not good to litter Test text all over the main dashboard, even for a day or two until the other tickets get completed. For now I updated it just to write a HTML comment in there so I can still easily prove / test the integration (see commit 01ffc01). Will that work?

kzantow:
:bee: from me, nice job

michaelneale:
can we defer merging this until multibranch is fixed? https://issues.jenkins-ci.org/browse/JENKINS-36338

blueocean-personalization/src/main/js/components/ActionLink.jsx
```diff
@@ -2,6 +2,6 @@ import React, { Component } from 'react';
 
 export default class ActionLink extends Component {
     render() {
-        return <div>Test</div>;
+        return <div dangerouslySetInnerHTML={{ __html: '<!--Test-->' }}></div>;
```
michaelneale: love that method name from react ;)

blueocean-personalization/src/main/resources/index.jelly
michaelneale: every plugin I think needs an index.jelly - I think...

michaelneale:
🐝  as code seems ok and keith seems happy, but not sure what am looking for (story book didn't show up anything for me)

cliffmeyers:
This is really just a shell of the plugin, since I have other JIRA tickets for the meaningful feature delivery. Still nice to keep this in a separate PR so we can get an idea of what needs to be added for new plugins. Going to merge now and will have a PR later today with some storybook for the Pipeline Card component.

PR #305, opened 2016-06-29
Bug/jenkins 36208

vivek:
Related to issue # . 

Summary of this pull request: 
As initially thought to implement a cache on QueueListener, which represents in mem structure of queue - this was for it being performant. While implementing, I discovered Jenkins keeps pretty similar and accurate in memory structure of queue state so duplicating the same in blueocean was unnecessary.

MultiBranch project's branches are pipeline projects and the queue item is in fact are represented as step so searching for queued items is not as straightforward as calling Jenkins.getInstance().getQueue().getItems(pipelineJob), its rather intrusive by looking for Workflow specific quieted items and returning those items in queue.

@imeredith I noticed QueueContainerImpl doesn't work with pipeline project. Please take a look at this PR, I made some minor changes to QueueItemImpl.

@reviewbybees


imeredith:
:bee: once you add the 2nd null check

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/MultiBranchPipelineQueueContainer.java
```diff
@@ -36,13 +36,19 @@ public BlueQueueItem get(String name) {
...
                 PipelineImpl pipeline = (PipelineImpl) multiBranchPipeline.getBranches().get(item.task.getOwnerTask().getName());
                 if(pipeline != null) {
-                    int runId=0;
+
                     if(item.task instanceof ExecutorStepExecution.PlaceholderTask) {
                         ExecutorStepExecution.PlaceholderTask task = (ExecutorStepExecution.PlaceholderTask) item.task;
-                        runId = task.run().getNumber();
+                        int runNumber;
+                        if(task.run() == null){
+                            runNumber = pipeline.job.getNextBuildNumber();
```
imeredith: Thats not quite right. What we actaully need to get the list of queued items and figure out what number it is for that job. But I think this is fine and I'll add this with my incoming replay PR

imeredith:
:bee: Build number isnt quite right when run == null, but I think its ok for now, I will fix it with changes i have in the works

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/MultiBranchPipelineQueueContainer.java
vivek: @imeredith I could do that but thats unnecessary expense (to iterate thru all queued items) and even with that you never know if you got the build number in sync with what jenkins thinks it is. Will be great if your replace PR can get us such info.

i386:
@vivek thanks so much for getting this one done 👍

vivek:
@i386 cheers 👍

PR #306, opened 2016-06-30
conditional assumptions for running problematic multibranch tests

michaelneale:
Lets us conditionally run problematic problems 

cc @vivek 
@reviewbybees


vivek:
Looks great, 🐝

PR #308, opened 2016-06-30
[FIX JENKINS-35860] - extension point component type filter

kzantow:
Related to issue # 35860 . 

Summary of this pull request: 
. Add component type filter for extensionPoints

@reviewbybees


js-extensions/EXTENSIONS.md
```diff
@@ -12,8 +12,8 @@ JSe `@jenkins-cd/js-extensions` module exports 2 things:
 
 The `ExtensionStore` API is very simple, all public methods are asynchronous:
 
-- `getExtensions(extensionPointName, [type,] onload)`
-    This method will async load data and type information as needed, and call the onload handler with a list of extension exports, e.g. the React classes or otherwise exported references.
+- `getExtensions(extensionPointName, [filter,] onload)`
+    This method will async load data and filter the information as needed, and call the onload handler with a list of extension exports, e.g. the React classes or otherwise exported references.
```
tfennelly: I'm imagining what this might mean here, but I think a concrete use case example would help. "filter as needed" is a bit ambiguous.

tfennelly: If the filtering is specific to the component `dataType` field, then I would prefer that to be explicit in this API i.e. `getExtensions(extensionPointName, [dataTypes,] onload)`.

js-extensions/EXTENSIONS.md
```diff
@@ -51,12 +51,18 @@ Extensions are defined in a `jenkins-js-extensions.yaml` file in the javascript
...
 
     export default class NavLink extends React.Component {
         ...
     }
 
+#### Enforcing specific component types
+
+In order to ensure a specific component is returned, an extension point may also use the `componentType` filter - it accepts an object prototype (e.g. an ES6 class), e.g.:
+
+    <Extensions.Renderer extensionPoint="test-view" componentType={TestResults} ... />
```
tfennelly: `componentType` or `dataType` ?

tfennelly:
@kzantow Any particular reason why the contents of `EXTENSIONS.md` are not in `README.md` ?

js-extensions/EXTENSIONS.md
kzantow: `componentType` and `dataType` are different filters...

tfennelly: @kzantow Is the difference explained somewhere? When would you use one over the other? When would you use both?

michaelneale:
I like it. As @tfennelly mentioned, perhaps some docs on what componentType vs dataType is (componentType is a react component name, right?)

kzantow:
@michaelneale `componentType` is a react component type or other javascript type (if it's not a React extension point). I've labeled it this way to be consistent with the `component` term in the yaml file.

michaelneale:
Looking good. So my take is that dataType is the type provided to as data to the extension, and componentType is the type helps filter things down if many things offer to implement that extension point? 

@tfennelly is this clear enough now? I'm not sure if I have it right yet.

michaelneale:
OK I think 🐝  from me. @tfennelly speak now or ... 

I am slightly confused that different componentTypes types could provide the same named extensionPoint, but I am sure there is a good reason. (I can contrive situations, but I assume there was a real need for this).

js-extensions/README.md
```diff
@@ -1,4 +1,80 @@
...
+    This method will async load data, filter the extensions based on the provided `filter`s, and call the onload handler with a list of extension exports, e.g. the React classes or otherwise exported references.
+    `filter` currently supports the following properties:
+    - `dataType` - to filter components that accept specific types of data from the server
+    - `componentType` to enforce a specific component type is returned (e.g. `Link`)
+
+- `getTypeInfo(type, onload)`
+    This will return a list of type information, from the [classes API](../blueocean-rest/README.md#classes_API), this method also handles caching results locally.
+
+- `init()`
+    Required to be called with `{ extensionDataProvider: ..., typeInfoProvider: }` see: [ExtensionStore.js](src/ExtensionStore.js#init) for details. This is currently done in [init.jsx](../blueocean-web/src/main/js/init.jsx), with methods to fetch extension data from `<jenkins-url>/blue/js-extensions/` and type information from `<jenkins-url>/blue/rest/classes/<class-name>`.
```
tfennelly: I think it should be made more explicit (it is somewhat implicit in "This is currently done in ...", but not sure that helps a lot) that this method is not something that users of this API should ever call (? in fact, is it idempotent ?). I'd nearly not publicise it's existence at all.

tfennelly:
I dunno ... I think I do get the theoretical difference between `dataType` and `componentType`, but I don't really like the impl from a usability perspective because I think it's going to be a bit confusing for people. It's just a bit too subtle. Perhaps that can be fix by more docs, but it just smells like this is not right to me and there might be a way of doing this that has less potential for confusing people.

Have we hit a concrete use case for needing both of these yet? If we have, can we articulate that somewhere and then eventually in the docs. A clear real-world example always helps.

kzantow:
> @tfennelly "Have we hit a concrete use case for needing both of these yet? If we have, can we articulate that somewhere and then eventually in the docs. A clear real-world example always helps."

No, we don't really have a concrete use-case.

`dataType` we have concrete use cases and it's modeled after Jenkins' current extensibility mechanism, with the `extensionPoint` essentially being the view to display. So, `<ExtensionClass>/<view>` is analogous to `<dataType>/<extensionPoint>`. In Java, this ensures the data is of a certain type and only returns views pertinent to the specific data available, but enforces nothing about the view.

The `componentType` enforces the returned view component matches an expected type - e.g. a list of links might require a `Link` type.

js-extensions/README.md
kzantow: Fair enough, I'll move this to some docs in the code, I guess..

kzantow:
@imeredith @michaelneale - NPM fetch is broken on the PR builder, ideas?

michaelneale:
@kzantow yeah failure was due to npm infra, It was passing before. 
Currently having other (unrelated issues)

michaelneale:
I am ok with this 🐝  - @tfennelly is this go or no go? @cliffmeyers ?

kzantow:
@michaelneale i'm 🐛 'ing this, talked with @tfennelly and we have a slightly different approach, should have it wrapped up today

michaelneale:
@kzantow ok, good to know.

i386:
@tfennelly, @kzantow said this one is ready for review.

michaelneale:
yeah @kzantow didn't look like a temporary error, some small problem still

kzantow:
@michaelneale need to republish `js-extensions` beta

js-extensions/src/ExtensionRenderer.jsx
```diff
@@ -1,6 +1,6 @@
 var React = require('react');
 var ReactDOM = require('react-dom');
-var ExtensionStore = require('./ExtensionStore.js');
+var ExtensionStore = require('./ExtensionStore.js').instance;
```
sophistifunk: How come we have a .instance property, rather than just exporting a single object, or a .getInstance() helper if we have a singleton but need some future-proofing?

js-extensions/src/ExtensionStore.js
```diff
@@ -53,7 +53,7 @@ export class ExtensionStore {
             extension.instance = instance;
             return;
         }
-        throw `Unable to locate plugin for ${extensionPointId} / ${pluginId} / ${component}`;
+        throw new Error(`Unable to locate plugin for ${extensionPointId} / ${pluginId} / ${component}`);
```
sophistifunk: 👍

sophistifunk:
LGTM

michaelneale:
:bee:

js-extensions/src/ExtensionRenderer.jsx
kzantow: @sophistifunk I'm all ears for a better way to do this, I would love to have some sort of DI to be able to hook this together rather than relying on `init()` method calls

michaelneale:
This is pending review from @tfennelly

js-extensions/src/ExtensionRenderer.jsx
tfennelly: I'm looking at this `instance` export and I don't really see the need for it Vs just using statics (same on `ClassMetadataStore`) e.g. `ExtensionStore.getExtensions(...)` or `ClassMetadataStore.componentType( ..)`?

I actually see how in the tests you have ...

```javascript
var ClassMetadataStore = require('../dist/ClassMetadataStore').instance;
```

And then make the usage look as though it's a `static` e.g. `ClassMetadataStore.componentType(...)`.

js-extensions/src/ClassMetadataStore.js
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Returns a filtering function to only return untyped extensions
+     */
+    untyped() {
+        return (extensions, onload) => {
+            // exclude typed extensions when types not requested
+            extensions = extensions.filter(m => !('dataType' in m));
+            onload(extensions);
+        };
+    }
```
tfennelly: Do `componentType`, `dataType` and `untyped` have anything to do with `ClassMetadataStore` ? Seems like they belong more on `ExtensionStore`, seeing as though they are for filtering the extensions in their.

Is there scope here to formalise the `Filter` mechanism into a type and extend it to create `ComponentTypeFilter`, `DataTypeFilter` etc in class files of their own?

tfennelly:
🐝 

Though I don't really like the `instance` approach on `ExtensionStore` and `ClassMetadataStore` + some of the code seems to be in the wrong'ish place imo + I'd like to see the `Filter` mechanism formalised a bit more Vs bolting it onto the side of `ClassMetadataStore` (which seems like the wrong place to me anyway).

kzantow:
@tfennelly the filter is about as 'formalized' as it should be, I think? The specific filters are dependent on the type information, therefore located with the class that provides it (well, except the componentType, which should be moved, probably)

js-extensions/src/ExtensionRenderer.jsx
kzantow: Well, I am morally opposed to statics in almost all cases. I need to have a particular instance configured by the user (e.g. blueocean-web), though, so we don't have another mechanism to do this. I think we should investigate a DI system. The tight coupling of some things has me already feeling worried about the future of this codebase...

tfennelly:
@kzantow okioki ... I'd prob try formalize it more, but maybe that's just personal taste.

🐝 good to go !!

PR #311, opened 2016-06-30
JENKINS-35884# Capability annotation support

vivek:
Adds supports for [Capability annotation](https://github.com/jenkinsci/capability-annotation) with /rest/classes API. 

@reviewbybees

### DIFF
--- a/blueocean-rest-impl/pom.xml
+++ b/blueocean-rest-impl/pom.xml
@@ -69,6 +69,11 @@
             <artifactId>scm-api</artifactId>
             <version>1.1</version>
         </dependency>
+        <dependency>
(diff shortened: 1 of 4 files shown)


i386:
@vivek is there a plan to add the annotation to known parts of the codebase?

vivek:
@i386 Yes, once this is in, good time to come up with reasonable capability and it to our code base.

imeredith:
:bee:

i386:
Can that be done as part of JENKINS-35884?

vivek:
@i386 We could do that but putting annotation on API model resources is orthogonal, this is core functionality. I think we should merge this one and annotating our model classes can be done in another ticket. I will open a ticket for that.

i386:
Sounds good

PR #314, opened 2016-07-04
update centre requires url to wiki to deploy

michaelneale:
Alpha release is not showing up in experimental UC

@reviewbybees


stephenc:
:bee:

PR #315, opened 2016-07-05
[JENKINS-35843] Support PR's from origin

imeredith:
Related to issue # . 

Summary of this pull request: 
.
.
.

@reviewbybees


michaelneale:
🐝  seems good. Fairly routine change. 

Interesting that it built 2 things to validate this, I thought that is what this change would do, but it seems to be doing it already (the branch, and the origin PR).

PR #316, opened 2016-07-05
[JENKINS-35881] replay a run

imeredith:
This PR allows a run to be replayed

### DIFF
--- a/blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/AbstractRunImplTest.java
+++ b/blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/AbstractRunImplTest.java
@@ -4,16 +4,24 @@ import hudson.model.Label;
 import io.jenkins.blueocean.service.embedded.rest.PipelineImpl;
 import io.jenkins.blueocean.service.embedded.rest.PipelineRunImpl;
 import io.jenkins.blueocean.service.embedded.scm.GitSampleRepoRule;
+import jenkins.branch.BranchProperty;
+import jenkins.branch.BranchSource;
+import jenkins.branch.DefaultBranchPropertyStrategy;
+import jenkins.plugins.git.GitSCMSource;
+import jenkins.scm.api.SCMSource;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.replay.ReplayAction;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.jenkinsci.plugins.workflow.multibranch.WorkflowMultiBranchProject;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 
 import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
 
 /**
  * @author Ivan Meredith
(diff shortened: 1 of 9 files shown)


michaelneale:
pretty much all the multibranch tests fail with this change

imeredith:
@michaelneale should be fixed now

imeredith:
Still need to document, doing that now

blueocean-rest/README.md
```diff
@@ -337,6 +337,22 @@ Pipelines can be nested inside folder.
       "qeueudTime" : "2016-06-22T11:05:41.309+1200"
     }
 
+## Replay a pipeline run
```
michaelneale: Can it mention here if it runs the exact commit or HEAD? (I have already forgotten)

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/AbstractRunImplTest.java
```diff
@@ -0,0 +1,67 @@
...
+    @Rule
+    public GitSampleRepoRule sampleRepo = new GitSampleRepoRule();
+
+    @Before
+    public void setup() throws Exception{
+        super.setup();
+        sampleRepo.init();
+    }
+    //Disabled, see JENKINS-36453
+    //@Test
```
michaelneale: can we have the test use a multibranch (as we know that works?)

michaelneale:
Looks good, but would like to see the test enabled with multibranch if possible (as we know that works).

michaelneale:
I got a "no crumbs included" 🐛  when trying: 

```
curl -XPOST http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/bee/branches/master/runs/1/replay
```

blueoceandeploy:
Ah it needs to be Content-Type: application json

michaelneale:
that test run failed (odd error, looks like process spawning).

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
```diff
@@ -193,8 +193,14 @@ public final String getEndTimeString(){
     @Navigable
     public abstract Object getLog();
 
+    /**
+     * Replays a pipeline. Sam SCM commit/revision should be used. in the new run.
```
vivek: Typo? s/Sam/Same/

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/AbstractRunImplTest.java
```diff
@@ -60,8 +68,53 @@ public void replayRunTest() throws Exception {
...
+
+        Assert.assertNotEquals(new PipelineRunImpl(b1, null).getCommitId(), new PipelineRunImpl(b2, null).getCommitId());
+
+        request().post("/organizations/jenkins/pipelines/p/branches/master/runs/1/replay").build(String.class);
+
+        j.waitForCompletion(job1.getLastBuild());
 
+        request().get("/organizations/jenkins/pipelines/p/branches/master/runs/3/").build(Map.class);
+        request().get("/organizations/jenkins/pipelines/p/branches/master/runs/3/").build(Map.class);
+        Map r = request().get("/organizations/jenkins/pipelines/p/branches/master/runs/3/").build(Map.class);
```
vivek: There are similar looking /runs/3 GET calls, is that intended?

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
```diff
@@ -193,8 +193,14 @@ public final String getEndTimeString(){
...
+     *
+     * @return The queued item.
+     */
     @POST @TreeResponse @WebMethod(name = "replay")
```
vivek: From REST API perspective: I see stop, replay verbs appearing in the API URL. Could we not use POST .../runs/:id with payload to indicate action (stop, replay)? Also it fits more of PUT than POST. PUT is considered idempotent, stop and replay both are in fact idempotent. 

```
POST /.../runs/1
{
  "action":"replay"
}
```

vivek:
@imeredith Looks good in general, left some comments.

vivek:
LGTM 🐝 As per discussion with @imeredith, we are going to defer addressing verb in URL for later time when there is yet another verb for run needed.

PR #317, opened 2016-07-08
[JENKINS-36513] Use TreeResponse for return of stop()

imeredith:
Changed stop() to return BlueRun. Uses TreeResponse annotation

@reviewbybees


vivek:
@imeredith LGTM 🐝

PR #318, opened 2016-07-08
[JENKINS-36514] durationInMillis can be null so use Long instead of long

imeredith:
Related to issue # . 

Summary of this pull request: 
.
.
.

@reviewbybees


i386:
LGTM after tests pass.

PR #320, opened 2016-07-08
JENKINS-36376# Annotate model classes with @Capability

vivek:
Related to issue # . 

Summary of this pull request: 
.
.
.

@reviewbybees

### DIFF
--- a/blueocean-rest-impl/pom.xml
+++ b/blueocean-rest-impl/pom.xml
@@ -75,12 +75,7 @@
             <artifactId>workflow-durable-task-step</artifactId>
             <version>2.0</version>
         </dependency>
-        <dependency>
-            <groupId>io.jenkins.blueocean.rest.annotation</groupId>
-            <artifactId>capability-annotation</artifactId>
-            <version>1.2</version>
-        </dependency>
-        
+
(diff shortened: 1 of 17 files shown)


cliffmeyers:
@vivek could you provide an example of the REST API in the README for this? I wasn't able to follow whether the endpoint needs to be called with a :id argument, or if the UI can simply call `/rest/classes/` to pull down all metadata information.

vivek:
@cliffmeyers Its documented in this section: https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-rest/README.md#class-of-resource-discovery

cliffmeyers:
@vivek can we have a way to batch requests for multiple classes, or even pull down all of the known classes / capabilities that the backend supports in a single call?

I'm thinking ahead to the Pipelines screen, and we have at least three different types that will show up:  multi-branch, pipeline, and folder. As it stands now, I think we'd need to make three separate REST calls to fetch the capabilities for each, and then proceed with folder filtering. That seems a little heavy to me.

Since the metadata isn't something that will change, it seems sensible that we fetch it all once on initial app load. What do you think?

cliffmeyers:
Discussed w/ Vivek and cut a new ticket for enhancement request. LGTM 🐝

PR #321, opened 2016-07-08
JENKINS-36286# LinkResolver to resolve link of a jenkins model object…

vivek:
Related to issue # . 

Summary of this pull request: 
.
.
.

@reviewbybees 

… mapped to BlueOcean API object

Supported Jenkins models, Job, ItemGroup, FlowNode (stage, prallel, step)


vivek:
@tfennelly I have modified MessageEnricher to use LinkResolver to give an idea how it can be done. I have hardcoded "jenkins" organization for now. This will get addressed once we have multi-organization implemented, this will give us good idea on how we can provide abstraction on providing multiple organization in generic way. You might like to make such changes in MessageEnricher yourself, in that case I can back out that change.

imeredith:
LGTM 🐝

PR #323, opened 2016-07-11
React versions were out of sync with jdl

michaelneale:
Related to issue #JENKINS-36555 . 

Summary of this pull request: 
React needs to be up to date between JDL and blue ocean. 

@reviewbybees 

Would be nice to tool our way to detect this?


sophistifunk:
The check-dependencies script in the root would flag this if we do it within this checkout, but it doesn't currently compare the dependencies in JDL any more because we extracted it to a different repo.

Also, I think I may be the only person who runs that script ;-)

sophistifunk:
But anyway, 
bbm-bueno

i386:
@sophistifunk could the script be added to the pipeline and fail the build?

sophistifunk:
notonyourlife


..


Wait... I mean "yes"

PR #326, opened 2016-07-11
JENKINS-36488# Use LinkResolver to get favorite item

vivek:
Related to issue # . 

Summary of this pull request: 

cc: @cliffmeyers @imeredith 
@reviewbybees

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/FavoriteImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/FavoriteImpl.java
@@ -2,12 +2,12 @@ package io.jenkins.blueocean.service.embedded.rest;
 
 import hudson.Util;
 import hudson.model.Item;
-import hudson.model.Job;
 import io.jenkins.blueocean.rest.Reachable;
 import io.jenkins.blueocean.rest.hal.Link;
+import io.jenkins.blueocean.rest.hal.LinkResolver;
 import io.jenkins.blueocean.rest.model.BlueFavorite;
-import jenkins.branch.MultiBranchProject;
-import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import io.jenkins.blueocean.rest.model.BluePipeline;
+import jenkins.model.Jenkins;
 
 /**
  * @author Vivek Pandey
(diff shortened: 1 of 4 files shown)


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/FavoriteImpl.java
```diff
@@ -16,27 +16,30 @@
 
     private final Object item;
     private final Link self;
+    private final LinkResolver linkResolver;
 
     public FavoriteImpl(Item item, Reachable parent) {
         this.self = parent.getLink().rel(Util.rawEncode(item.getFullName()));
+        this.linkResolver = Jenkins.getActiveInstance().getInjector().getInstance(LinkResolver.class);
```
kzantow: any way to `@Inject LinkResolver` instead?

kzantow: ... also `Jenkins.getInstance()` is the right way as of 2.0.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/hal/Link.java
```diff
@@ -28,4 +29,27 @@ public String getHref(){
...
+
+
+    /**
+     * Gives ancestor Link, for example, for link("/a/b/c/d"), ancestor is "a/b/c/".
+     */
+    public Link ancestor(){
+        int i = href.lastIndexOf("/");
+
+        if(i>0) {
+            int j = href.substring(0, i).lastIndexOf("/");
```
kzantow: Will these links ever end with a slash? Maybe get the last index before the last character?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/FavoriteImpl.java
```diff
@@ -16,27 +16,30 @@
...
-        //TODO: Needs https://issues.jenkins-ci.org/browse/JENKINS-36286 to be fixed,
-        //      it should use LinkResolver for it to work in all cases
-        Object obj = null;
-        if(item instanceof WorkflowJob){
-            if(item.getParent() instanceof MultiBranchProject){
-                Link s = OrganizationImpl.INSTANCE.getLink().rel(String.format("pipelines/%s/branches/",
-                    ((MultiBranchProject) item.getParent()).getName()));
-                obj = new BranchImpl((Job) item, s);
+        BluePipeline pipeline = null;
+        for(BluePipelineFactory factory:BluePipelineFactory.all()){
```
kzantow: Could this loop be moved up somehow to a utility method? Seems it would be a common pattern to find a pipeline based on an `item`

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/FavoriteImpl.java
vivek: >any way to @Inject LinkResolver instead?

No we can't since FavoriteImpl is not guice managed.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/hal/Link.java
vivek: constructor has assertion on it. 

`assert this.href.endsWith("/");`

Adding the similar check might add to redundancy.

kzantow: hm. `assert` doesn't typically execute at runtime. but this sounds good, I assume most of the cases are covered by tests. 👍

kzantow:
:bee: LGTM based on feedback, minor nitpick about `getActiveInstance()`

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/FavoriteImpl.java
vivek: Good point. Let me do that.

kzantow:
:bee: LGTM :)

PR #327, opened 2016-07-11
[FIXED JENKINS-35725] 404 page when pipeline not found

kzantow:
Related to issue #JENKINS-35725. 

Summary of this pull request: 
. Add `FetchStatus` to pass to components to indicate a fetch is in progress or failed
. Display stylized 404 page when an invalid pipeline URL is accessed

image

@reviewbybees


i386:
Screenshot looks great 👍

michaelneale:
some gulp errors @kzantow it looks like

kzantow:
@michaelneale yep, lint issues now fixed

michaelneale:
@kzantow I like this, but its a bigger change to the fetcher. I would like @scherler and @cliffmeyers to weigh in (ideally @tfennelly too) as I get the impression this may be the place that other fetch logic is placed?

michaelneale:
Any tests possible?

michaelneale:
need to have some kind of tests for this to be merged.

kzantow:
@michaelneale yep, will get to this tonight

cliffmeyers:
LGTM! Nice. 🐝

kzantow:
@cliffmeyers would appreciate a re-bee if you have a sec (added some basic test cases / slight code modification)

kzantow:
@cliffmeyers yes, it doesn't seem to be doing what I expect :-/

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
```diff
@@ -11,8 +13,6 @@ import {
 } from '@jenkins-cd/design-language';
 import { buildOrganizationUrl, buildPipelineUrl } from '../util/UrlUtils';
 
-const { object } = PropTypes;
```
scherler: why?

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
```diff
@@ -44,17 +52,18 @@ export default class PipelinePage extends Component {
...
-                {React.cloneElement(this.props.children, { pipeline })}
+                {this.props.children}
             </Page>
         );
     }
 }
 
 PipelinePage.propTypes = {
-    children: object,
+    children: PropTypes.any,
```
scherler: why not add above any? IMO the code you refactor here is much cleaner then the one you deliver instead. It is obvious that object/any is PropTypes since it is use exclusively within XXX.propTypes. I drop the redundant PropTypes

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -1,6 +1,6 @@
 import keymirror from 'keymirror';
 
-import fetch from 'isomorphic-fetch';
+import fetch from '../util/fetch';
```
scherler: IMO the whole fetch refactor does not belong in this PR

blueocean-dashboard/src/main/js/util/fetch.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+    onStart(id) {
+        console.log('onStart');
+        this._active.push(id);
+    }
+    onComplete(id) {
+        const idx = this._active.indexOf(id);
+        if (idx >= 0) {
+            this._active.splice(idx, 1);
+            console.log('onComplete');
```
scherler: I guess you will remove that, right?

blueocean-dashboard/src/main/js/components/NotFound.jsx
```diff
@@ -0,0 +1,16 @@
...
+import { Link } from 'react-router';
+
+/**
+ * Simple component to render a fullscreen 404 page
+ */
+export default () => (
+    <Fullscreen className="not-found">
+        <div className="message-box">
+            <h3>Page not found (404)</h3>
+            <div className="message">Jenkins could not find the page you were looking for. Check the URL for errors or press the back button.</div>
```
scherler: Which back button? The only button is "open Dashboard"

blueocean-dashboard/src/main/js/redux/actions.js
scherler: hmm, where do you expose fetch in util? You expose fetcher as default but that is not equal fetch. Did you try to see a run detail log? that should IMO fail with that code, since it will throw an exception converting txt to json

kzantow: @scherler that's a good point, it was a long running work in progress, I'll scale back the fetch refactor, it's not really necessary

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
kzantow: Hm, this is definitely a code style difference of opinion. Our files are currently mixed, which is bad. Defining the vars this way into these common english terms includes `object` and `array` which are only a case differences from the Javascript `Object` and `Array` types and alone should be enough to not do it, in my opinion. On top of it, changing the prop types often requires a 2 line change, instead of a 1-line change. It will also lead to lint issues when people forget to remove unused `PropType` definitions. It's also poor code practice to have the `const` definition so far away from the usage, if that is the preferred way to do it. So, I strongly disagree that the code before is in any way 'cleaner'.

blueocean-dashboard/src/main/js/components/NotFound.jsx
kzantow: ... the browser back button.

blueocean-dashboard/src/main/js/util/fetch.js
kzantow: yep, good catch. :'( for missing that

kzantow:
@scherler your comments addressed, please have a re-look

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
cliffmeyers: I believe @sophistifunk had made a point as well that using the PropTypes directly via destructuring will cause problems with Flow, which we may end up integrating into our plugins once we sort out any remaining kinks in the JDL.

scherler: as you said "code style difference of opinion". As stated it is the context, but I agree to move it just before the usage makes a lot of sense. 

Most of you points are covered by eslint rules and I am against using React.PropTypes.<type> there is no benefit in it You will repeat React.PropTypes meaningless times. 

Anyway ....

blueocean-dashboard/src/main/js/components/NotFound.jsx
scherler: so may be put "browser"  before the "back button" so it is clear for stupid people like me.

scherler:
besides the browser back button ;) :bee:

PR #329, opened 2016-07-11
Josh/checkdeps 2016 07 12

sophistifunk:
Related to issue #JENKINS-36555 . 

Summary of this pull request: 
Update checkdeps script to only look at production / runtime versions, and to also inspect some "blessed" first-level deps, such as JDL.

Coincidental small fix to Activity screen's blank state.

@reviewbybees

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,11 +1,12 @@
 node {
   deleteDir()
   checkout scm
-  
+
   docker.image('cloudbees/java-build-tools').inside {
     withEnv(['GIT_COMMITTER_EMAIL=me@hatescake.com','GIT_COMMITTER_NAME=Hates','GIT_AUTHOR_NAME=Cake','GIT_AUTHOR_EMAIL=hates@cake.com']) {
       try {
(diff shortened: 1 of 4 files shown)


sophistifunk:
This doesn't yet run checkdeps in the pipeline.

kzantow:
:bee: don't like the node_module needs, but seems to be the common way to do things

sophistifunk:
@kzantow the only other method would be to go out to npm.org to check, or to make assumptions about where on the filesystem those dependencies might be installed, unless I'm missing something?

kzantow:
@sophistifunk I know, there's not a much better way AFAIK. More of a general comment, I'm not a fan of hardcoding `node_modules` paths, but it's a pretty common thing to do...

sophistifunk:
Attempting to run in CI :D

sophistifunk:
I ams teh pipeline mastah*

sophistifunk:
mastah *
asterisk

michaelneale:
Looks good @sophistifunk

PR #330, opened 2016-07-12
[feature/JENKINS-35886] Implement expand on hash reference 

scherler:
Related to issue # . JENKINS-35886

Summary of this pull request: 

 - [X] Developer can click on a logs line number and set the current location to that line
 - [X] Developer can right click and copy a link to the logs line number and give it to another developer. When the developer visits, it expands the correct step and takes them to the linked log line.
 - [x] Select the right node in the graph -> covers Jenkins 36565 further if "this" is merged then merge https://github.com/jenkinsci/blueocean-acceptance-test/pull/5

**Depends on https://github.com/jenkinsci/jenkins-design-language/pull/76 to get through.**

@reviewbybees


i386:
@scherler does this select the right node in the graph too?

blueocean-dashboard/src/main/js/components/ScrollToHash.jsx
```diff
@@ -5,7 +5,7 @@ export const scrollToHash = ComposedComponent => class extends Component {
     componentDidMount() {
         setTimeout(() => {
             this.scrollToAnchor();
-        });
+        }, 100);
```
sophistifunk: Why do we need more than one layout's delay?

michaelneale:
eslint passing, but now a test failing

blueocean-dashboard/src/main/js/components/ScrollToHash.jsx
scherler: I changed it to 1, but basically I need to give time for ResultItem to expand. further setTimeout should have the second parameter anyway

scherler:
@i386 yes I just fixed that (so it will now as well highlight the node in case of overview url .../pipeline)

michaelneale:
small issues with eshint etc 🐜  but getting there

michaelneale:
@scherler @i386 is this one ok now?

i386:
Sounds like it but I've been unable to verify this myself

scherler:
@michaelneale for me yeah

michaelneale:
@scherler I read that in your accent. 

Manual inspection reveals this appears to work (local testing). 
Would be nice to have some unit test to cover it, but 🐝  otherwise.

PR #333, opened 2016-07-12
bug/jenkins 36618 - dup job name

cliffmeyers:
Related to issue # JENKINS-36618. 

Summary of this pull request: 
* Fix a bug where multiple jobs with the same name in different folders would produce an error in the pipeline list and only display one job.
* This was caused by a duplicate key value. In order to repro this in a test, you must use **Enzyme.mount** rather than Enzyme.shallow. This is because shallow doesn't actually render the nested child components that have the key applied, so React won't trigger the error in `flattenChildren` unless a full mount happens. This requires bringing JSDOM along the for the ride, and
* Requires Extensions.Renderer not to throw an error complaining about a missing 'extensionsDataProvider'. I kinda hacked my way around this but I suspect there's a better solution... adding Tom and Keith for input.

cc @tfennelly @kzantow 

@reviewbybees


blueocean-dashboard/src/main/js/components/Pipelines.jsx
```diff
@@ -62,10 +62,15 @@ export default class Pipelines extends Component {
...
                             { pipelineRecords
-                                .map(pipeline => <PipelineRowItem
-                                  key={pipeline.name} pipeline={pipeline}
-                                  showOrganization={!organization}
-                                />)
+                                .map(pipeline => {
+                                    const key = pipeline._links.self.href;
+                                    return (
+                                        <PipelineRowItem
+                                          key={key} pipeline={pipeline}
```
kzantow: why not just change to `pipeline._links...`

blueocean-dashboard/src/test/js/util/EnzymeUtils.js
```diff
@@ -0,0 +1,36 @@
...
+
+    global.navigator = {
+        userAgent: 'node.js',
+    };
+
+    // Extensions.Renderer will fail during Enzyme.mount without an
+    // 'extensionDataProvider' function being provided
+    // FIXME: there's probably a much cleaner way to do this.
+    ExtensionStore.init({
+        extensionDataProvider: () => undefined,
```
kzantow: this should be fine

kzantow:
:bee: wow, lots of code for testing a 1-line change but nice work :)

PR #334, opened 2016-07-12
[FIXED JENKINS-36299] add 'all passing' view to test display

kzantow:
Related to issue # JENKINS-36299. 

Summary of this pull request: 
. Add encouraging 'all passing' view for test results

image

Also needs: https://github.com/jenkinsci/jenkins-design-language/pull/77

@reviewbybees


cliffmeyers:
This LGTM to me. Is there any chance of adding a unit test for TestResults.jsx that asserts the correct state is rendered?

cliffmeyers:
Adding a 🐝  anyways, will let @kzantow make the call as to whether the unit test is too much a hassle.

blueocean-dashboard/src/test/js/testResult-spec.js
```diff
@@ -6,12 +6,9 @@ import { shallow } from 'enzyme';
 import TestResults from '../../main/js/components/testing/TestResults.jsx';
 
 describe("TestResults", () => {
-  let wrapper;
-  const
-    renderer = createRenderer();
+  const renderer = createRenderer();
```
cliffmeyers: I think you can remove this from the test if you like.

cliffmeyers:
Great to see more tests! :bee

sophistifunk:
Is this component only shown when there are test results? What I mean is, it's never used to render a (no-fail + no-pass) situation?

kzantow:
@sophistifunk yes, only used for the 'all passing', there's another thing that shows up if there are 0 tests.

kzantow:
@michaelneale I'd like to get this merged, could do a quick review of this one?

michaelneale:
:bee: I did test this wiht the kzantow/failure-project (I created an all passing branch).

PR #335, opened 2016-07-12
JENKINS-36489# NPE fix

vivek:
Related to issue # . 

Summary of this pull request: 
.
.
.

@reviewbybees

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/MultiBranchPipelineImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/MultiBranchPipelineImpl.java
@@ -133,24 +133,39 @@ public class MultiBranchPipelineImpl extends BlueMultiBranchPipeline {
         /**
          * TODO: this code need cleanup once MultiBranchProject exposes default branch. At present
          *
-         * At present we look for master as primary branch, if not found we find the latest build and return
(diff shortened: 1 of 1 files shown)


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/MultiBranchPipelineImpl.java
```diff
@@ -133,24 +133,39 @@ public Integer getWeatherScore(){
...
          *
          * If there are no builds taken place 0 score is returned.
          */
 
-        Job j = mbp.getBranch("master");
+        Job j = mbp.getItem("master");
         if(j == null) {
-            j = mbp.getBranch("production");
-            if(j == null){ //get latest
+            j = mbp.getItem("production");
```
kzantow: not directly related to the change, but what is `"production"` ?

kzantow:
:bee:

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/MultiBranchPipelineImpl.java
vivek: 'production' branch. if not master then production, heuristic we came up with in absence of default branch concept available in underlying model.

PR #336, opened 2016-07-12
bug/url encoding bugs

cliffmeyers:
Related to issue # JENKINS-36613, JENKINS-36616

Summary of this pull request: 
* Combined a couple of these together since they all kind of stack on top of one another and it's not wise to test them all in isolation.

@reviewbybees


cliffmeyers:
@michaelneale  would you be willing to give me a hand testing this branch? It incorporates fixes for JENKINS-36613, JENKINS-36616 and JENKINS-36618.

blueocean-dashboard/src/main/js/components/Branches.jsx
```diff
@@ -22,30 +23,34 @@ export default class Branches extends Component {
...
         const {
             latestRun: { id, result, startTime, endTime, changeSet, state, commitId, estimatedDurationInMillis },
             weatherScore,
-            name,
+            name: branchName,
         } = data;
-        const url = `/organizations/${organization}/${pipelineName}/detail/${name}/${id}/pipeline`;
+
+        const cleanBranchName = decodeURIComponent(branchName);
+        const url = buildRunDetailsUrl(organization, fullName, cleanBranchName, id, 'pipeline');
```
scherler: you decode the branchName to later on encode it again?

blueocean-dashboard/src/main/js/components/RunDetails.jsx
```diff
@@ -85,8 +85,11 @@ class RunDetails extends Component {
 
         const baseUrl = buildRunDetailsUrl(organization, name, branch, runId);
 
-        const currentRun = this.props.runs.filter(
-            (run) => run.id === runId && decodeURIComponent(run.pipeline) === branch)[0];
+        /* eslint-disable arrow-body-style */
```
scherler: why do you disable it and not fix it?

```
this.props.runs.filter((run) => run.id === runId &&  decodeURIComponent(run.pipeline) === branch)[0];
```

michaelneale:
Is there a test case for UrlUtils at least? It seems like it would be fairly reasonable to unit test.

cliffmeyers:
@michaelneale yep, myself and @scherler have added unit tests for the UrlUtils methods.

blueocean-dashboard/src/main/js/components/RunDetails.jsx
cliffmeyers: I find those one liners a real pain because I can't easily drop a `debugger` statement in there to debug. I can switch it back if you prefer.

blueocean-dashboard/src/main/js/components/Branches.jsx
cliffmeyers: Yes because we don't necessarily want to pass the encoded value around, in my opinion at least. Once we start to leverage the HAL stuff we shouldn't ever have to do any manual encoding, maybe decoding in select places.

michaelneale:
:bee: it fixes the above ones. 

It would be nice to see tests for this if possible but understand if not. Go ahead.

michaelneale:
Ignore the PR-336 failure @cliffmeyers - is old PR job not needed now.

cliffmeyers:
I created https://issues.jenkins-ci.org/browse/JENKINS-36660 to track things we need covered in acceptance tests. Going to merge this now.

PR #344, opened 2016-07-13
bug/jenkins-36619 - close modal 404

cliffmeyers:
**Decription**
* Fix the bug by encoding the project name correctly in the fallbackUrl
* I invested several hours into trying to write a unit test for this scenario. The idea was to expose a property "navigateToLocation:func" on RunDetails that would be called with the correct URL, then use sinon.spy to spy on it and assert that the URL was correct. Unfortunately this won't work because the callback is only invoked after the modal closes, which is requires some React state transitions to fire off first. The test execution won't wait, so it just shows the callback as not having been invoked. Maybe there's a way to delay it but I'm at a loss. I will push up a feature branch of the testing work so we have it for later in case someone else wants to take a look. Adding this bug to https://issues.jenkins-ci.org/browse/JENKINS-36660 so we can catch it in an acceptance test.

**Submitter checklist**

**Reviewer checklist**
- [x] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-dashboard/src/main/js/components/RunDetails.jsx
+++ b/blueocean-dashboard/src/main/js/components/RunDetails.jsx
@@ -96,9 +96,9 @@ class RunDetails extends Component {
         const status = currentRun.result === 'UNKNOWN' ? currentRun.state : currentRun.result;
 
         const afterClose = () => {
-            const fallback = `/organizations/${organization}/${name}/`;
+            const fallbackUrl = buildPipelineUrl(organization, name);
 
-            location.pathname = this.opener || fallback;
+            location.pathname = this.opener || fallbackUrl;
             location.hash = `#${branch}-${runId}`;
 
             router.push(location);


cliffmeyers:
Flagged as WIP because I am waiting for clarification from @i386 on the JIRA ticket as to whether we need to do more than fix the 404 in this bug.

scherler:
Off the top of my head you could timeout the router push, but not sure whether that would do you any good

tfennelly:
What would you try test if you could? Maybe you could move code around to at leasy make the function unit testable (outside the react lifecycle etc).

cliffmeyers:
@tfennelly I tried to describe it above. Basically I wanted to test that the component would trigger navigation to the correct URL when the modal was closed.

cliffmeyers:
Going to let this PR just address the bug fix. I have a separate ticket https://issues.jenkins-ci.org/browse/JENKINS-36276 that I will use to track the other item @i386 wanted changed. I also added a reference to this bug in https://issues.jenkins-ci.org/browse/JENKINS-36660 which will add some acceptance tests to cover this and related scenarios. Anyone willing to give me a bee if that takes the place of a unit test?

kzantow:
:bee: simple 'nuff, it looks like

PR #346, opened 2016-07-14
Fix to launch multibranch

michaelneale:
**Decription**
This is a regression fix, a few different things. cc @cliffmeyers @vivek 

Related to https://issues.jenkins-ci.org/browse/JENKINS-36675

**Submitter checklist**

**Reviewer checklist**
- [x] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
+++ b/blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
@@ -38,31 +38,32 @@ import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.multibranch.WorkflowMultiBranchProject;
 
 import javax.annotation.Nonnull;
+import java.net.URLEncoder;
 
 /**
  * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
  */
 @Extension
 public class BlueMessageEnricher extends MessageEnricher {
-    
+
     enum BlueEventProps {
         blueocean_job_rest_url,
         blueocean_job_pipeline_name,
         blueocean_job_branch_name,
     }
-    
+
     @Override
     public void enrich(@Nonnull Message message) {
-        
+
         // TODO: Replace once https://issues.jenkins-ci.org/browse/JENKINS-36286 is done
         message.set(EventProps.Jenkins.jenkins_org, OrganizationImpl.INSTANCE.getName());
-        
+
         String channelName = message.getChannelName();
         if (channelName.equals(Events.JobChannel.NAME)) {
             JobChannelMessage jobChannelMessage = (JobChannelMessage) message;
             ParameterizedJobMixIn.ParameterizedJob job = jobChannelMessage.getJob();
             Link jobUrl = getLink(job);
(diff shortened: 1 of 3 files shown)


vivek:
@michaelneale Have not run the PR but the changes looks good to me 🐝

cliffmeyers:
Afraid I don't have a chance to run this myself but was just in the code and the changes look sensible to me. RunPipeline is gonna get refactored soon in another ticket and may obviate this change anyway, so good to get the fix in now IMO. :bee

michaelneale:
@cliffmeyers @vivek  this is actually 3 fixes:
1) The actual regression that was stopping running even "master" branch
2) I added an encoding call to ensure it can run branches like "feature/blah"
3) I changed the blue message enricher to correctly encode the branch name so that SSE would pick it up correctly.
The 1st one was really a regression

@cliffmeyers we will need another PR/issue for fixing PullRequests.jsx - are you able to open that? I figured you wanted to refactor it anyway to be more in common. It certainly won't cope with names like this the way it is.

michaelneale:
@cliffmeyers as I suspected, this doesn't work in folders (obviously neither would it have before). 

I will leave it up to you if you want to merge this in (makes it a bit better, possibly back where it was) or leave it out, for a bigger fix for paths/branches/folders and the ensuing confusion.

michaelneale:
@cliffmeyers ok did some snooping, using fullName seems to help, needed to tweak the SSE a bit to ensure freestyle works too in a folder and so on. I am not sure if this helps, but I will leave this call up to you if this moves it forward, or gets in the way of your plans. 

The SSE isn't quite right for things in folders or with "funny" names, but there aren't errors, and you can launch things at least.

blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
```diff
@@ -38,37 +38,38 @@
...
-    
+
     @Override
     public void enrich(@Nonnull Message message) {
-        
+
         // TODO: Replace once https://issues.jenkins-ci.org/browse/JENKINS-36286 is done
         message.set(EventProps.Jenkins.jenkins_org, OrganizationImpl.INSTANCE.getName());
-        
+
```
michaelneale: this was gulp --fixLint - honest. Scouts honour.

scherler: or your ide, my is configured to remove empty spaces on line endings ;)

scherler:
Have not test it but :bee: for the code

blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
michaelneale: I use the best: atom. I do think this was gulp though, as I ran it once to clean things up.

cliffmeyers:
Tested this locally and the following scenarios all worked:

# MBP nested in folder, from Branches tab, running both "master" and "feature/XYZ" type branches
# MBP at top-level, same scenarios as above
# Running Pipeline nested in folder, from Activity tab
# Running top-level Pipeline, from Activity tab

Looks good here :bee . I'm merging it now.

PR #347, opened 2016-07-14
this is needed to work with a fresh workspace

michaelneale:
This is needed as the scm api required by branch source is newer (we think...)

by @vivek & me

@reviewbybees

### DIFF
--- a/blueocean-rest-impl/pom.xml
+++ b/blueocean-rest-impl/pom.xml
@@ -67,7 +67,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>scm-api</artifactId>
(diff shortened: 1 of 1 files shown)


michaelneale:
I believe @sophistifunk  was seeing this from time to time.

i386:
Hmm could we use a maven import scope on the another Jenkins package to pull in all the right dependencies for this module? That way when we bump the maven version all of this stays in sync

michaelneale:
@i386 I think that is just pushing the problem around. Depending on github branch source I guess would probably do it... but then we don't want a hard dependency on that. scm api is the one that is wanted AFAICT

vivek:
LGTM, tested this fix 🐝

PR #348, opened 2016-07-14
[JENKINS-35868] changeset url

imeredith:
**Decription**

Adds urls for changeset commits. ie to github.

**Submitter checklist**

No extra tests for this as I can't make it generate a url when doing it locally.

**Reviewer checklist**
- [x] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ChangeSetResource.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ChangeSetResource.java
@@ -1,7 +1,5 @@
 package io.jenkins.blueocean.service.embedded.rest;
 
-import com.strangeberry.jmdns.tools.Browser;
-import hudson.model.AbstractBuild;
 import hudson.scm.ChangeLogSet;
 import hudson.scm.ChangeLogSet.Entry;
 import hudson.scm.RepositoryBrowser;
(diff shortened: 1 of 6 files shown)


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ChangeSetResource.java
```diff
@@ -50,6 +53,20 @@ public String getTimestamp(){
         }
     }
 
+    @Exported
+    public String getUrl() {
```
i386: Does this mean its only ever available in the embedded driver?

imeredith: Yes. All of that data is only in the embedded driver.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ChangeSetResource.java
```diff
@@ -50,9 +50,38 @@ public String getTimestamp(){
         }
     }
 
+    @Override
+    public String getUrl() {
```
i386: @imeredith is there a unit test worth adding here?

michaelneale:
yeah I think a unit test would be nice here, but not sure if there is precedent for PRs in the multibranch test suite

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ChangeSetResource.java
imeredith: I don't know how to unit test this. It's using a local repo only.

michaelneale: fair enough

michaelneale:
haven't had time to test this out but code looks good

michaelneale:
🐝 tested out and works. Accept explanation. Nice work.

PR #353, opened 2016-07-15
[JENKINS-36067] Hide branch col for non-multibranch activity view

imeredith:
**Decription**

Hides the branch column for regular pipelines and freestyle jobs.

To test just visit the activity page for both multibranch and non multibranch jobs

screen shot 2016-07-15 at 12 19 12 pm

screen shot 2016-07-15 at 12 19 00 pm

**Submitter checklist**
- [ ] Apppppropriate unit or acceptance tests or explaination to why this change has no tests

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


i386:
🐛 the "Completed" column should be pinned to the left hand side.

imeredith:
@i386 The Completed column has not moved. The reason its not butted up to the right is htere is an unlabled column called  "actions" now

i386:
@imeredith ahhhhh

imeredith:
@i386 As for the other columns, all but the message are a fixed width.

tfennelly:
LGTM so far ... what can we do about an automated test. We might need to get a more general multi-branch test in place first and test this as part of that.

kzantow:
We could probably generalize this type of thing more easily with the `ClassMetadataStore` and a component. Something like a wrapper component:

```
<IfCapability data={thing_with._class} capability="some.capabilty.or.class.name">
  ... contnents
</IfCapability>
```

I can add that to my prototype type handling prototype.

tfennelly:
@kzantow yeah, that sounds good but I think I'd have something specific for common tests like "isMultiBranch". Asking users of these components to enter class names etc would just piss people off.

Actually ... I think the impl of `Pipeline.isMultiBranch` should use the new stuff you added. I think it can, right? Atm it's doing a very specific class name check, while it should be able to do a more general type check.

kzantow:
@tfennelly the problem is classes API is async

kzantow:
@tfennelly but if we start using the 'capabilities' instead of class names, the capability could be just `multibranch` or something

tfennelly:
@kzantow ah right ... we'd need to mount it in the component and only set it's state once we have all this info.

imeredith:
@i386 do you want me to change anything with this?

i386:
@imeredith I think @kzantow's comment was relevant here before getting a 🐝

michaelneale:
@imeredith you interested in the IfCapability react component? That would be a super thing to have

blueocean-dashboard/src/main/js/components/Activity.jsx
```diff
@@ -58,6 +58,8 @@ export class Activity extends Component {
 
             config.pipeline = pipeline;
             this.props.fetchRunsIfNeeded(config);
+
+            this.props.fetchCapabilitiesIfNeeded(this.props.pipeline._class);
```
tfennelly: Seems wrong to be fetching this here?

blueocean-dashboard/src/main/js/components/Runs.jsx
```diff
@@ -74,7 +75,9 @@ export default class Runs extends Component {
                 {id}
             </td>
             <td><CommitHash commitId={commitId} /></td>
-            {showBranchCol && <td>{decodeURIComponent(pipeline)}</td>}
+            <IfCapibility dataClass={pipelineClass} capability="io.jenkins.blueocean.rest.model.BlueMultiBranchPipeline" >
+                <td>{decodeURIComponent(pipeline)}</td>
+            </IfCapibility>
```
tfennelly: I know this was what was suggested to you, but I'd be dead against us having Java class names (e.g. "io.jenkins.blueocean.rest.model.BlueMultiBranchPipeline") littered through the JavaScript code. Things like "isMultiBranch" etc should (imo) be encapsulated in some way, with class checks etc pushed down and out of site from the app code.

blueocean-dashboard/src/main/js/components/Activity.jsx
scherler: Why do you mean @tfennelly? ...but I agree that should be done in PipelinePage since it maybe needed for all children in the component tree.

...later more

blueocean-dashboard/src/main/js/components/IfCapability.jsx
```diff
@@ -0,0 +1,42 @@
...
+        } else {
+            return null;
+        }
+    }
+};
+
+IfCapability.propTypes = {
+    dataClass: string,
+    capability: string,
+};
```
scherler: fetchCapabilitiesIfNeeded: func,
capabilities: array,

is missing, I wonder that eslint did not complain on that

blueocean-dashboard/src/main/js/components/IfCapability.jsx
```diff
@@ -0,0 +1,42 @@
...
+    createSelector,
+    connect,
+} from '../redux';
+const { object, array, func, string , bool} = PropTypes;
+
+export default class IfCapability extends Component {
+    constructor(props) {
+        super(props);
+        const { dataClass } = this.props;
+        this.props.fetchCapabilitiesIfNeeded(dataClass);
```
scherler: You cannot do this here. That needs to go into componentWillMount.

blueocean-dashboard/src/main/js/components/IfCapability.jsx
```diff
@@ -0,0 +1,42 @@
...
+    }
+};
+
+IfCapability.propTypes = {
+    dataClass: string,
+    capability: string,
+};
+
+
+const selectors = createSelector([capabilitiesSelector], (capabilities) => ({ capabilities }));
```
scherler: We have started to not connect certain component to redux on that level. It is more efficient to do this line and and the following in the PipelinePage and then in that page you need to switch to

```
{React.cloneElement(this.props.children, { pipeline,  capabilities})}
```

You have an early out already, so your component should still work the same, just get injected the capabilities from somewhere else.

blueocean-dashboard/src/main/js/components/IfCapability.jsx
scherler: https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/JENKINS-36067-hide-branch/3/pipeline#step-12-log-2301 

it did ;)

scherler:
Needs some tweaks and 

```
npm run lint:fix
```

is your friend ;)

...but looks promising @imeredith

tfennelly:
@imeredith sorry for the slow reply ... was out and about for a bit.

Just checked out the code and ran it ... now I'm wondering why not use the `ClassMetadataStore` from `js-extensions` (Vs making rest calls .. redux etc).

tfennelly:
Hey @imeredith ... take a look at https://github.com/jenkinsci/blueocean-plugin/pull/366

blueocean-dashboard/src/main/js/components/Runs.jsx
imeredith: Right, it was my plan to do this, its just a wip that i commited so i could get some help

imeredith:
Cheers i'll look more tomorrow. I couldnt find the class meta data store when i was looking

imeredith:
@tfennelly I added the caps stuff to Activity too. But we need a better way than replicating setState didMount etc in every component we need caps in. One option I guess would be to create create a superclass that does it.

blueocean-dashboard/src/main/js/components/Activity.jsx
```diff
@@ -47,6 +48,14 @@ RunNonMultiBranchPipeline.propTypes = {
 };
 
 export class Activity extends Component {
+    constructor(props) {
```
scherler: Activity is not the place to do it, see my last comment, that has to go into pipelinePage and passed as children component.

As for now it only works on the activity tab but that should work as well in branch and pr.

blueocean-dashboard/src/main/js/components/records.jsx
```diff
@@ -92,4 +92,5 @@ export const State = Record({
     steps: null,
     currentBranches: null,
     testResults: null,
+    capabilities: null,
```
scherler: I see you forgot to remove that, however IMO you should re-insert the setCapabilities in the actions (just drop the fetch) and invoke the setCapabilities when you receive them from the store.

kzantow: No, not here. The component keeps track of the capabilities for each object type. The component maintains it's own render state based on whether or not it's been fetched.

js-extensions/src/ClassMetadataStore.js
```diff
@@ -8,13 +8,20 @@ export class ClassMetadataStore {
...
          * Fetch function for the classMetadata
          */
         this.classMetadataProvider = classMetadataProvider;
+
+        /**
+         * Onload callbacks cache. Used to ensure we don't
+         * issue multiple in-parallel requests for the same
+         * class metadata.
+         */
+        this.classMetadataOnloadCallbacks = {};
```
kzantow: This is becoming too common of a pattern, we should generalize it at the `fetch` level for a given URL.

michaelneale:
Looking close, some updating with master and this may be nearly done

imeredith:
@michaelneale nah we have specs failing, trying to find a soln to that now.

blueocean-dashboard/src/main/js/components/Capability.jsx
```diff
@@ -0,0 +1,76 @@
+import React, { Component } from 'react';
+import update from 'react-addons-update';
```
tfennelly: It would be interesting to know how much weight this adds to the bundles and if it's worth it Vs manually handling the update (shallow clones etc).

blueocean-dashboard/src/main/js/components/Capability.jsx
```diff
@@ -0,0 +1,76 @@
...
+import { classMetadataStore } from '@jenkins-cd/js-extensions';
+
+export const capabilityStore = classes => ComposedComponent => class extends Component {
```
tfennelly: I think a comment on this to explain the syntax etc would be good.

tfennelly: In fact, maybe a general link to some docs explaining the composition pattern (Vs mixins e.g. https://facebook.github.io/react/blog/2016/07/13/mixins-considered-harmful.html).

blueocean-dashboard/src/main/js/components/Capability.jsx
```diff
@@ -0,0 +1,76 @@
...
+    _classesToObj(_classes) {
+        if (!_classes) {
+            return {
+                has: () => false,
+            };
+        }
+
+        return {
+            _classes,
+            has: capability => _classes.find(_class => _class === capability) !== undefined,
```
tfennelly: It's interesting ... some of these new JavaScript syntax enhancements ... I actually find them to be harder to read.

blueocean-dashboard/src/main/js/components/Capability.jsx
```diff
@@ -0,0 +1,76 @@
...
+        };
+    }
+    _setState(key, value) {
+        // Block calls to setState for components that are
+        // not in a mounted state.
+        if (!this.unmounted) {
+            const newData = { capabilities: {} };
+            newData.capabilities[key] = { $set: value };
+            this.setState(previousState => update(previousState, newData));
```
tfennelly: `previousState`? I don't see it defined anywhere. Maybe some comments here as to what's happening. No obvious to me anyway.

js-extensions/src/ClassMetadataStore.js
tfennelly: Yeah, @ivan and I were saying that. No need to do in this PR though imo.

cliffmeyers: Perhaps we could create a ticket to track the work? Otherwise it's just tech debt we can't easily measure.

blueocean-dashboard/src/main/js/components/IfCapability.jsx
```diff
@@ -0,0 +1,26 @@
...
+        if (capabilities[_class].has(capability)) {
+            return this.props.children;
+        }
+
+        return null;
+    }
+}
+
+IfCapability.propTypes = {
+    _class: string,
```
cliffmeyers: I usually reserve prefacing with an underscore for "pseudo private" methods or properties. Perhaps "className" (which you can feel free to repurpose in any React component) or something like "modelClass"? Just a nit.

blueocean-dashboard/src/main/js/components/Capability.jsx
cliffmeyers: Seconded, I've actually never seen this kind of class declaration before and have no idea what it does :/ Some thorough jsdocs here would be great.

js-extensions/src/ClassMetadataStore.js
kzantow: @tfennelly agree, unless it's excessively slow, we should not make this change only to revert it later (soon) when we fix it at a lower level

blueocean-dashboard/src/main/js/components/Capability.jsx
imeredith: previousState is the arguemoent to the function that is being passed into setState

blueocean-dashboard/src/main/js/components/Capability.jsx
imeredith: () => false 

function () { return false; }

capability => _classes.find(_class => _class === capability) !== undefined,

function (capability) {
    return classes.find(_class => _class === capability) !== undefined;
}

I think once you get used to it it is just as readable.

As for how I implemented the returned object, im not so happy about.

blueocean-dashboard/src/main/js/components/Capability.jsx
kzantow: @imeredith the convention adopted seems to be always using parens around lambda parameters, though there's no lint rule enforcing it. people complained when i did the same thing ;)

imeredith: I'm not against using parenthesis but I can't anyone against not using them on the general web too. I suspect people will get used to it as ES6 gets more adoption. 

scala and java8 but support similar styles and I generally haven't seen parenthesis used for single args unless you need to specify a type but generally you don.t.

blueocean-dashboard/src/main/js/components/Capability.jsx
imeredith: It is fairly lightweight It has no deps and is 150lines of code. Its a good library i would use again I think.

imeredith:
@tfennelly @kzantow @cliffmeyers @scherler This is ready to review i think

blueocean-dashboard/src/main/js/components/Activity.jsx
```diff
@@ -68,16 +69,20 @@ export class Activity extends Component {
             return null;
         }
 
+        const { capabilities } = this.props;
+        const isMultiBranchPipeline = capabilities[pipeline._class].contains(MULTIBRANCH_PIPELINE);
```
tfennelly: Ohhh this makes me a little sad tbh .... referencing `_class` directly ... the `MULTIBRANCH_PIPELINE` constant is a bit ugly 😢 

What was there (`Pipeline.isMultibranch(pipeline)`) def seems better to me, but doesn't use the capabilities so I accept that that might not work as is, but surely it could be morphed into something like `Pipeline.isMultibranch(capabilities)` and centralize/hide logic like `capabilities[pipeline._class].contains(MULTIBRANCH_PIPELINE)`.

blueocean-dashboard/src/main/js/components/Activity.jsx
```diff
@@ -124,9 +137,10 @@ Activity.contextTypes = {
...
     runs: array,
     pipeline: object,
+    capabilities: object,
     fetchRunsIfNeeded: func,
 };
 
 const selectors = createSelector([runsSelector], (runs) => ({ runs }));
 
-export default connect(selectors, actions)(Activity);
+export default connect(selectors, actions)(capabilityStore(props => props.pipeline._class)(Activity));
```
tfennelly: `_class` again 😢

blueocean-dashboard/src/main/js/components/Capability.jsx
```diff
@@ -0,0 +1,111 @@
...
+
+        return false;
+    }
+}
+
+/**
+ * capabilityStore is a enchances components to inject capabilities for
+ * a given class.
+ *
+ * Usage: capabilityStore(classFunction)(React.Component) where
```
tfennelly: What is `capabilityStore`? Is `Capability.jsx === capabilityStore` ?

tfennelly: Ah ok sorry ... looks like one does `import capabilityStore from 'Capability.jsx'` I think

blueocean-dashboard/src/main/js/components/Capability.jsx
tfennelly: Well, whatever the trends are in scala etc, I def find the parenthesis help me to "see" the different code elements more easily. Sure I'd get used to it, but I'd get used to anything if I have to.

tfennelly: I keep feeling like all of these JavaScript language "enhancements" are going to result in me writing/reading [code like this](http://www.jsfuck.com/) some day.

tfennelly: I'm sure I've read people saying that these kinds of enhancements are a big part of the reason why Scala never really made it mainstream ... normal people can't maintain the code after.

blueocean-dashboard/src/main/js/components/IfCapability.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+IfCapability.propTypes = {
+    className: string,
+    capability: string,
+    capabilities: object,
+    children: React.PropTypes.node,
+};
+
+export default capabilityStore(props => props.className)(IfCapability);
+
```
tfennelly: Nice job Ivan !!

tfennelly:
🐝 assuming the ATH tests all pass.

Nice work Ivan !!

imeredith:
https://ci.blueocean.io/job/Acceptance%20Tests%20Param/8 was ath :)

PR #360, opened 2016-07-18
Jenkins 36700

scherler:
Needs to be merged **BEFORE** https://github.com/jenkinsci/blueocean-acceptance-test/pull/12

**Decription**
- [x] Addressing point 2. from https://issues.jenkins-ci.org/browse/JENKINS-36700
- [x] Acceptance test see above.
- [x] JENKINS-36684 - Remove "scroll to item" feature in Activity, Branches and PR tabs
- [x] Fix bug to link in long logs -> https://youtu.be/KtPjfrqAVA4
- [x] Fix bug with scrollDown vs. linking log lines   -> see https://youtu.be/nAsYp4s3pE0 for the problem

For the last two points I grabbed a video, which shows that we now will wait until the sub-render in logConsole has processed all lines and then scroll where we have to scroll. As soon https://issues.jenkins-ci.org/browse/JENKINS-36771 is through we will have that while we wait for the render to have finished. -> linking working after the pr: https://youtu.be/u2YR4dqrSNM 

As requested by @i386, if there are no steps, there should be no logHeader.

screenshot from 2016-07-19 11-51-51

Test are not looking for an empty state. If they should do I will adopt a karaoke test.

**Submitter checklist**

**Reviewer checklist**
- [x] Reviewed the code

@reviewbybees


blueocean-dashboard/src/main/js/components/LogConsole.jsx
```diff
@@ -1,7 +1,5 @@
 import React, { Component, PropTypes } from 'react';
-import { scrollToHash } from './ScrollToHash';
-
-const { array, bool, string } = PropTypes;
+import { scrollHelper } from './ScrollHelper';
 
 const INITIAL_RENDER_CHUNK_SIZE = 100;
 const INITIAL_RENDER_DELAY = 300;
```
sophistifunk: Where does this number come from? A 300ms pause in the UI will be pretty seriously noticeable to users. And will it cease to be enough if more components are instantiated, or the user happens to have YouTube running in another tab? Will it be enough if the user isn't looking at this tab and it has been de-prioritised? Surely there's something deterministic we can hook into from React?

michaelneale: that looks like it has been there for a while though? not related to this PR

sophistifunk: It's used in this PR tho... I dunno. Just seems to me like one of those things that will bite us in the rear with a mystery-meat environment-specific failure sometime in the future ¯\_(ツ)_/¯

sophistifunk:
I haven't run it, but 🐝 on the code.

michaelneale:
:bee: 

@scherler I am still seeing in some cases it jumps to finish node when I launch and open it immediately (but that doesn't seem anything to do with this PR). 

I would prefer this empty state to say "no steps yet" if it knows it is executing, would be my only suggestion

blueocean-dashboard/src/main/js/components/LogConsole.jsx
scherler: @michaelneale yes, that is coming from @cliffmeyers when he implemented the "buffered rendering". @sophistifunk I have no better idea ATM how to do the "increment rendering" so did not touched it, only adopted the code to fix the bugs it causes.

scherler:
@michaelneale sorry did not chaged it before merge, but will do it in another pr

PR #369, opened 2016-07-22
JENKINS-36481# Rollbar and it.pluginVersion() removed

vivek:
**Decription**

**Submitter checklist**

Check the footer of dashboard, should show commitId instead of version

**Reviewer checklist**
- [x] Reviewed the code

@reviewbybees 

BlueOceanUI is a bootstrap and core infrastructure class, we do want to
keep it fiarly abstract and indepedent of direct Jenkins dependency.

Rollbar or any such analytics tool integration should be done out of blueocean-web
possibly inside a separate plugin or some place else that makes sense.

Footer version to display the commitId/branch

### DIFF
--- a/blueocean-web/src/main/js/DevelopmentFooter.jsx
+++ b/blueocean-web/src/main/js/DevelopmentFooter.jsx
@@ -5,14 +5,6 @@ import revisionInfo from '../../../target/classes/io/jenkins/blueocean/revisionI
 
 export class DevelopmentFooter extends Component {
     render() {
-        if (!revisionInfo || !revisionInfo.name) {
-            var blueOceanVersion = document.getElementsByTagName('head')[0].getAttribute('data-blue-ocean-version');
-            return (
-                <div className="development-footer">
-                    <span>Blue Ocean UI v{blueOceanVersion}</span>
-                </div>
-            );
-        }
         return (
           <div className="development-footer">
               <span>Built at {moment(revisionInfo.timestamp).format('Do MMMM YYYY hh:mm A')}&nbsp;</span>
--- a/blueocean-web/src/main/js/DevelopmentFooter.jsx
+++ b/blueocean-web/src/main/js/DevelopmentFooter.jsx
@@ -5,6 +5,12 @@ import revisionInfo from '../../../target/classes/io/jenkins/blueocean/revisionI
 
 export class DevelopmentFooter extends Component {
     render() {
+        if (!revisionInfo || !revisionInfo.name) {
+            // TODO: At minimum we should return Jenkins version. Jenkins version is always present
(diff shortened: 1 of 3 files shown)


blueocean-web/src/main/resources/io/jenkins/blueocean/BlueOceanUI/index.jelly
```diff
@@ -30,37 +29,6 @@
             <link rel="stylesheet"
                   href="${resURL}/plugin/blueocean-web/assets/css/jenkins-design-language.css"
                   type="text/css"/>
-
-            <j:if test="${it.includeRollbar()}">
```
kzantow: 👍 👍

blueocean-web/src/main/js/DevelopmentFooter.jsx
```diff
@@ -5,14 +5,6 @@ import revisionInfo from '../../../target/classes/io/jenkins/blueocean/revisionI
 
 export class DevelopmentFooter extends Component {
     render() {
-        if (!revisionInfo || !revisionInfo.name) {
-            var blueOceanVersion = document.getElementsByTagName('head')[0].getAttribute('data-blue-ocean-version');
-            return (
-                <div className="development-footer">
-                    <span>Blue Ocean UI v{blueOceanVersion}</span>
```
kzantow: But now we won't see the revision number on screen, right? From the ticket, am I reading right that there will be a follow-on to restore it?

blueocean-web/src/main/js/DevelopmentFooter.jsx
```diff
@@ -5,14 +5,6 @@ import revisionInfo from '../../../target/classes/io/jenkins/blueocean/revisionI
 
 export class DevelopmentFooter extends Component {
     render() {
-        if (!revisionInfo || !revisionInfo.name) {
-            var blueOceanVersion = document.getElementsByTagName('head')[0].getAttribute('data-blue-ocean-version');
```
kzantow: Maybe this should have `if (!revisionInfo) { return null; }` instead?

vivek: @kzantow Is there going to be a case when revisionInfo might not be present?

kzantow: It's generated at build time but I've seen it missing before... there must have been some reason for the check in the first place.

vivek: @kzantow Ok, I will put the condition back and return null as suggested.

kzantow:
:bee: thanks, @vivek !

michaelneale:
🐝

PR #375, opened 2016-07-26
[JENKINS-36903] test for things that we later use and if not present …

scherler:
**Decription**

The name as in name: '(basename $(git rev-parse --show-toplevel))', are not resolved during build time,
Anyway we do not use the name anywhere in the component, I changed it to sha, which we actually use.

**Submitter checklist**

**Reviewer checklist**
- [x] Reviewed the code

@reviewbybees 

…would throw an exception


michaelneale:
nice catch ! 🐝

PR #378, opened 2016-07-27
Jenkins 36131 "Show all" link required anywhere a log is displayed

scherler:
**Decription**

- [X] works for freestyle
- [X] works for pipeline
- [X] links "Full log" shown only if needed (e.g. when fetchAll was true we do not show the link)
- [X] in follow along the Full Log button should not never be shown, since you see everything already
- [X] you can link in a "full log"
- [X] in freestyle we really show the full log, however the "show all" link only apply to that step, so the user can look back if needed
- [X] loading the content using ?start=0 will trigger loading of all

AT: https://github.com/jenkinsci/blueocean-acceptance-test/pull/21

**Submitter checklist**

**Reviewer checklist**
- [x] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-dashboard/src/main/js/components/LogConsole.jsx
+++ b/blueocean-dashboard/src/main/js/components/LogConsole.jsx
@@ -116,7 +116,7 @@ export class LogConsole extends Component {
         return (<code
           className="block"
         >
-            { hasMore && <div key={0} id={`${prefix}log-${0}`}>
+            { hasMore && <div key={0} id={`${prefix}log-${0}`} className='fullLog'>
                 <a
                   key={0}
                   href={`?start=0#${prefix || ''}log-${1}`}
--- a/blueocean-dashboard/src/main/js/components/LogConsole.jsx
+++ b/blueocean-dashboard/src/main/js/components/LogConsole.jsx
@@ -108,7 +108,7 @@ export class LogConsole extends Component {
 
     render() {
         const lines = this.state.lines;
-        const { prefix = '' } = this.props;
+        const { prefix = '', hasMore = false } = this.props;
         if (!lines) {
             return null;
         }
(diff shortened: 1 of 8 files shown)


michaelneale:
nice. 

It seems to work. 

2 minor things: 
1) the "full log" isn't quite clear it is a link/call to action (just blends in with the text)
2) when the full (large) log is loading (I used the examples from the ATH), it kind of chugs a bit, and there isn't a visual indication that it is busy fetching stuff (and hard to cancel). 

I think #1 could be easily done (is it possible to make it a link style somehow @i386 ?)

#2 may be another ticket/future thing to address 

so partial :bee:

i386:
How about this? This uses a slightly smaller button style (height is 20px and font is 12px). Clicking this button will open the log in a new window.

Its important that the button does not use up the line numbers and should look like the mockup below.

upload-2
How about this? This uses a slightly smaller button style (height is 20px and font is 12px). Clicking this button will open the log in a new window.

Its important that the button does not use up the line numbers and should look like the mockup below.

failing result

michaelneale:
@i386 that is a design change though, we did agree that it would fill in the lines above (in https://issues.jenkins-ci.org/browse/JENKINS-36131) as per classic, so it will take longer to change possibly. 

I do like that button though.

i386:
@michaelneale so we should raise another ticket?

michaelneale:
@i386 I think if the style could be tightened up (possibly to use that button, if Thor has time) then this can be merged, and then we can review/another ticket to "open in new tab".

michaelneale:
Actually if the button is more work (I am aware Thorsten is taking vacation, yay!) then having it as a clear hyperlink would be enough to merge this in, and we can follow up changes as needed.

scherler:
@michaelneale cheers mate! Yeah I am loving the AT, it helped me to develop the testing scripts/pipeline in a more meaningful way. I as well use them on my local box for developing, good to hear that you like them. 

Regarding the loading (#2), I am starting https://issues.jenkins-ci.org/browse/JENKINS-36770  which blocks https://issues.jenkins-ci.org/browse/JENKINS-36771. The last ticket will bring that loading bar and yes the whole rendering 10k logs is hitting my box hard and if you have two 10k logs open the browser is having REAL problems.

Will implement the button now.

scherler:
screenshot from 2016-07-28 10-53-28

I changed as well "in follow along the Full Log button should not be shown, since you see everything already" I saw that while I was in the AT cinema watching "nightwatch"  - great movie, enjoying it every time. ;)  I added a test for that as well

i386:
Hey @scherler if you add the classes `btn-secondary inverse` to the anchor for "Show complete log" it will be styled appropriately 

jenkins blue ocean 2016-07-29 09-20-13

PR #379, opened 2016-07-27
[JENKINS-36238] New SSE client gateway with store and forward support

tfennelly:
# Description

See [JENKINS-36238](https://issues.jenkins-ci.org/browse/JENKINS-36238)

Depends on

* https://github.com/jenkinsci/sse-gateway-plugin/pull/7 and
* https://github.com/jenkinsci/pubsub-light-module/pull/3

Two possible tests ... minimum of 1:

1. **Min**: Verify that there are no regressions i.e. that SSE based notifications still work on the browser e.g. run a build and toasts etc popup .. build run animations appear.
1. **Bonus**: Setup a http proxy server that redirects requests to the Jenkins instance you are testing against. Connect the browser to Jenkins via this proxy - the activity page of a job ... establish a session etc. Kill the proxy and then, in another browser window/tab, connect to the activity page for the same job (not via the proxy of course), start a run and let it finish ... switch to the activity page connecting via the browser and then restart the proxy ... the events missed should come though and result in the activity page coming back up-to-date.

For a http proxy I used the `http-proxy` package, installing it globally and then running a simple script.

Install:

```
npm install -g http-proxy
```

The script (`httpproxy.js`) contents ...

```javascript
var httpProxy = require('http-proxy');
 
httpProxy.createProxyServer({target:'http://localhost:8080'}).listen(18080);
console.log('Proxying port 18080 to localhost:8080');
```

And then running the script:

```
node httpproxy.js
```

**Of course**, if you have a second machine you can do the same test from there, using it as the client, turning off its network (forcing a disconnect) and then turning it back on.

# Submitter checklist

# Reviewer checklist

- [x] Reviewed the code

### DIFF
--- a/blueocean-dashboard/package.json
+++ b/blueocean-dashboard/package.json
@@ -38,7 +38,7 @@
     "@jenkins-cd/design-language": "0.0.65",
     "@jenkins-cd/js-extensions": "0.0.19",
     "@jenkins-cd/js-modules": "0.0.5",
-    "@jenkins-cd/sse-gateway": "0.0.7-beta2",
+    "@jenkins-cd/sse-gateway": "0.0.7",
     "immutable": "3.8.1",
     "isomorphic-fetch": "2.2.1",
     "keymirror": "0.1.1",
--- a/blueocean-dashboard/package.json
+++ b/blueocean-dashboard/package.json
@@ -38,7 +38,7 @@
     "@jenkins-cd/design-language": "0.0.65",
     "@jenkins-cd/js-extensions": "0.0.19",
     "@jenkins-cd/js-modules": "0.0.5",
-    "@jenkins-cd/sse-gateway": "0.0.6",
+    "@jenkins-cd/sse-gateway": "0.0.7-beta2",
     "immutable": "3.8.1",
     "isomorphic-fetch": "2.2.1",
     "keymirror": "0.1.1",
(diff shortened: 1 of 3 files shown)


michaelneale:
WOW

michaelneale:
OMG this works! I had to use npm install http-proxy with a package.json (installing via -g didn't seem to pick it up correctly). 

but i was able to stop the proxy, see errors, start, and run a build again. And it Just Worked. 

WHAT DEVILRY IS THIS!!!!

🐝 

However, @vivek had some questions on the linked PR around thread safety on the server side, so this can't be merged until they are addressed in https://github.com/jenkinsci/sse-gateway-plugin/pull/7

one that is ok - then GO GO GO

vivek:
@michaelneale totally! SSE brings liveliness to the whole experience. @tfennelly makes it looks easy:) but lots of work behind the scene. Great stuff!

i386:
Dude this is awesome 💯 👍

tfennelly:
Hey @vivek, check out my last mods to `EventHistoryStore` in https://github.com/jenkinsci/sse-gateway-plugin/pull/7 ... I've locked down write access to the methods that do the init, so I think we're ok with this now.

As I said on the PR .... also remember that those methods are package private + the class itself is `@Restricted(NoExternalUse.class)`, so I think there's really no risk of rogue calls to those setters. In any case ... they'll be rejected/ignored now, so we're double-safe :)

And another "also remember" ... there was never any locking/synchronization on any of those variables, so the worse thing that could have happened was for things to be start behaving unexpectedly i.e. storing events somewhere random, or not cleaning up on a regular enough basis (or too often). There was never going to be classic MT race conditions or anything.

kzantow:
:bee: on this, of course but see my comments in: https://github.com/jenkinsci/sse-gateway-plugin/pull/7

i386:
@tfennelly 💥boom 💥  good to see this land 👍 👍

PR #386, opened 2016-07-29
Feature/jenkins 37023 pipeline card linking

cliffmeyers:
**Description**
* Note: I have temporarily duplicated StopPropagation (and added small behavior) inside of "personalization" until we create a "js-commons" package for Blue Ocean to share code between dashboard and personalization. I don't think this really makes sense in the JDL.
* To test, make sure you have some favorites. Click on the card title to view activity, click elsewhere on card to view run details. Clicking "run again" should do nothing, clicking "Favorite" show remove favorite. (lots of stopProp code needed here for this all to work right)
* I could not really find a fair way to unit test this since it deals with navigating to new routes. This can be covered in acceptance tests.

**Submitter checklist**

**Reviewer checklist**
- [x] Reviewed the code

@reviewbybees


i386:
@cliffmeyers nice dude 👍

blueocean-personalization/src/main/js/components/StopPropagation.js
```diff
@@ -0,0 +1,44 @@
+/**
```
sophistifunk: Why is this a component, rather than a single line inside event handlers that need it?

sophistifunk:
The code looks fine, but it seems to be an epic overkill. Am I missing something?

michaelneale:
I am seeing a fair chunk of errors in the console: 

```
jenkins-js-extension.js:22707 Error: addComponentAsRefTo(...): Only a ReactOwner can have refs. You might be adding a ref to a component that was not created inside a component's `render` method, or you have multiple copies of React loaded (details: https://fb.me/react-refs-must-have-owner).(…)(anonymous function) @ jenkins-js-extension.js:22707
jenkins-js-extension.js:16895 Uncaught (in promise) TypeError: Cannot read property 'componentWillLeave' of undefined(…)
```

when there are favourites. 

I also have seen `Error rendering Connect: TypeError: Cannot read property 'item' of undefined` appear in place of favorites (I unfav'ed everything) and refreshed the page. 

So there seem to be a few 🐛 's around this. I did a mvn -DcleanNode install as well... so I am not sure what is going on.

cliffmeyers:
It might be that this branch is using the wrong version of js-extensions. Doubtful I can look at it before leaving for the airport tomorrow, but I will be bringing my work laptop with me to help move a couple things along

cliffmeyers:
Yeah, may have been a merge error on my part. Try ticking to 0.0.20

michaelneale:
@cliffmeyers aren't you on vacation ;) leave it alone! hahaha

blueocean-personalization/src/main/js/components/StopPropagation.js
tfennelly: Good question ... seems strange to have this inside the markup.

tfennelly:
@cliffmeyers I too am curious why `StopPropagation` needs to be a react component.

cliffmeyers:
@sophistifunk @tfennelly you gents are right, I don't think we need a component for this specific issue. I initially started thinking ahead to other events we'd want to stop prop on - maybe this would be better than having 2-3 event handlers? But for what we need to do here yeah, this is probably overkill. Will get this fixed up when I get a chance.

sophistifunk:
That makes sense, but IMO components *shouldnt* be calling `stopPropagation` on events they don't already handle themselves.

cliffmeyers:
@sophistifunk right, so initially I tried to stopPropagation within the Favorite component itself because it makes sense for that component to handle the event. However, I ran into an issue on the Branches tab with this approach. Branches has an onClick listener registered on the TR element in the table, and it also has an ExtensionPoint within that TR where the Favorite component is added by the personalization plugin. In that scenario, it made no difference what I did within the Favorite component  (stopProp, stopImmediate, or calling these on "nativeEvent") -> the onClick on the TR still received the event. I think it's because these two pieces of DOM are created via two separate ReactDOM.render calls, each getting their own document listener because of how React doesn't its synthetic event magic.

Hopefully that gives a little insight into the madness that spawned this code.

All that said, I think in this scenario it should work fine suppressing within Favorite, so I'll give that a try here first and report back.

Thanks, it's actually really helpful to be asked to explain this and reason out the best fix.

sophistifunk:
I think this is another "two-reacts" problem. AFAIK, react does all its event handling with a single listener on `document`, so a separate render() tree shouldn't cause this issue. Maybe we should reach out to the Nylas guys to see if they've had this issue.

michaelneale:
This works as advertised. 🐝 

If you want to explore other ways to not need a wrapped component, also fine, but this is functional, even if it appears to be more code than ideally needed (and fact there is a dirty big "stop" component, if it is merged, can serve as a waypoint for future work as it can be made clear that it could be simplified).

cliffmeyers:
I have made changes based on @sophistifunk 's totally appropriate feedback in commit 7da4051. We should still axe the use of StopPropagation in Dashboard, although would be nice if I could cut a separate ticket for that so as to not weigh down this PR (that ok, @i386  / @michaelneale  ?)

cliffmeyers:
Created https://issues.jenkins-ci.org/browse/JENKINS-37281 and going to merge this feature now.

PR #387, opened 2016-07-29
Bug/jenkins 36904 running queued anim

cliffmeyers:
**Description**
* Change styles for pulse
* @386 I attached a video of the anim, want to make sure it's right before we merge. "running" now uses a solid white pulse that matches what we had for "queued". I also noticed that previously the outer circle for "queued" was a solid white, so it looked more like a "running" job that was 100% done. I changed the outer circle to the lighter color, which I think is visually appropriate. If you'd rather we stick with the more solid white just lmk.

**Submitter checklist**

**Reviewer checklist**
- [ ] Reviewed the code

@reviewbybees


cliffmeyers:
[running-queued-anim.mov.zip](https://github.com/jenkinsci/blueocean-plugin/files/391128/running-queued-anim.mov.zip)

i386:
@cliffmeyers that LGTM 👍

PR #394, opened 2016-08-08
[JENKINS-37200] Set NODE_ENV via js-extensions js-builder plugin

tfennelly:
# Description

An attempt to address [JENKINS-37200](https://issues.jenkins-ci.org/browse/JENKINS-37200).

It definitely seems to work better on FF when we build everything with `NODE_ENV=production` (i.e. doesn't hang on my machine, but did hang before), but not totally convinced this is the only issue. Still, I think this is worth doing.

There are no unit tests for this and imo it doesn't make sense to create any. I did run the ATH, but it is failing. I do not think it is relating to these changes. Seems like karaoke mode is broken. Looking into that now.

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

... defaulting to "production" mode so as to disable react dev tools, which seems to grind Firefox to a crawl

### DIFF
--- a/js-extensions/package.json
+++ b/js-extensions/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jenkins-cd/js-extensions",
-  "version": "0.0.20",
+  "version": "0.0.21-beta4",
   "description": "Jenkins Extension Store",
   "main": "index.js",
   "files": [
@@ -39,6 +39,7 @@
   },
   "dependencies": {
     "js-yaml": "^3.6.0",
+    "envify": "3.4.1",
     "@jenkins-cd/js-modules": "0.0.5"
   }
 }
--- a/blueocean-dashboard/package.json
+++ b/blueocean-dashboard/package.json
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "@jenkins-cd/design-language": "0.0.67",
-    "@jenkins-cd/js-extensions": "0.0.20",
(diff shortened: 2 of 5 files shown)


tfennelly:
__NOTE__: ATH fails do to changes in https://github.com/jenkinsci/blueocean-plugin/pull/389, so we need to fix those first.

js-extensions/package.json
```diff
@@ -39,6 +39,7 @@
   },
   "dependencies": {
     "js-yaml": "^3.6.0",
+    "envify": "3.4.1",
```
michaelneale: what is this one required for?

michaelneale:
@tfennelly I took the liberty of updating this to master to fix that other problem that was introduced... will now try ATH against this.

michaelneale:
ATH running now: https://ci.blueocean.io/job/Acceptance%20Tests%20Param/5/

michaelneale:
I think this is ok. Passes tests, and ATH. 

What I am not sure about is that extra dep in js-extensions, but if it is justified, 🐝  from me. Nice work @tfennelly

js-extensions/package.json
tfennelly: @michaelneale yeah, is needed to apply the process env settings, which are used in the react code to enable/disable dev mode.

Boring detail (ignore at will) .... it's a browserify transform that looks in the code for `process.env.NODE_ENV` and simply replaces them with strings so that it works in the browser.

PR #393, opened 2016-08-05
[JENKINS-36828] fix BlueOcean Dockerfile relies on plugins deployed to UC

ndeloof:

PR #396, opened 2016-08-08
[JENKINS-37141] rework Dockerfile for development image

ydubreuil:
# Description

See [JENKINS-37141](https://issues.jenkins-ci.org/browse/JENKINS-37141).

This is currently a work in progress to discuss the approach. Dependency resolution of plugins doesn't work correctly unfortunately. It's based on `install-plugins.sh` script for the official Jenkins image, but it seems this script suffers from some issues (see [JENKINS-36828](https://issues.jenkins-ci.org/browse/JENKINS-36828))

The main idea is to use Maven Assembly plugin to gather all BlueOcean plugins files in a folder used as a build context for Docker. Maven can't be used to resolve plugin dependencies from the POM because POM depends on JAR artifacts of plugins but not HPI one. When running Jenkins locally, dependency resolution is done by maven-hpi-plugin. Could be an option to add a new target to this plugin to copy dependent plugin to a folder used by the assembly.

@michaelneale @ndeloof

### DIFF
--- a/blueocean-docker-dev-image/pom.xml
+++ b/blueocean-docker-dev-image/pom.xml
@@ -27,7 +27,7 @@
     <!--
         If you need to debug this and doesn't want to wait for a full BlueOcean build, you can run
 
-             mvn -Pdocker-dev-image -pl .,blueocean-docker-dev-image package
+             mvn -pl .,blueocean-docker-dev-image package
     -->
 
     <build>
@@ -67,6 +67,14 @@
                         </configuration>
                     </execution>
                 </executions>
+            </plugin>
+            <!-- We don't want this module to be deployed, it doesn't make sense -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-deploy-plugin</artifactId>
+                <configuration>
+                    <skip>true</skip>
+                </configuration>
             </plugin>
         </plugins>
     </build>
--- a/blueocean-docker-dev-image/pom.xml
+++ b/blueocean-docker-dev-image/pom.xml
@@ -1,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.jenkins.blueocean</groupId>
+        <artifactId>blueocean-parent</artifactId>
+        <version>1.0.0-b07-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>blueocean-docker-dev-image</artifactId>
+    <packaging>pom</packaging>
+
+    <name>Module :: BlueOcean :: Docker development image</name>
+    <url>https://wiki.jenkins-ci.org/display/JENKINS/Blue+Ocean+Plugin</url>
+
+    <dependencies>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>blueocean</artifactId>
+            <version>${project.version}</version>
+            <type>hpi</type>
+            <scope>provided</scope>
+        </dependency>
+    </dependencies>
+
+    <!--
+        If you need to debug this and doesn't want to wait for a full BlueOcean build, you can run
+
+             mvn -Pdocker-dev-image -pl .,blueocean-docker-dev-image package
+    -->
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.jenkins-ci.tools</groupId>
+                <artifactId>maven-hpi-plugin</artifactId>
+                <version>1.119</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>bundle-plugins</goal>
+                        </goals>
+                        <configuration>
+                            <outputDirectory>${project.build.directory}/docker-dev-image/plugins</outputDirectory>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-assembly-plugin</artifactId>
+                <version>2.6</version>
+                <executions>
+                    <execution>
+                        <id>distro-assembly</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>single</goal>
+                        </goals>
+                        <configuration>
+                            <finalName>docker-dev-image</finalName>
+                            <appendAssemblyId>false</appendAssemblyId>
+                            <descriptors>
+                                <descriptor>src/assembly/docker-dev-image.xml</descriptor>
+                            </descriptors>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+</project>
--- a/blueocean-pipeline-api-impl/pom.xml
+++ b/blueocean-pipeline-api-impl/pom.xml
@@ -67,7 +67,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-durable-task-step</artifactId>
-            <version>2.0</version>
+            <version>2.4</version>
         </dependency>
 
         <!-- Not needed by blueocean runtime but adds to blueocean experience -->
--- a/build-in-docker.sh
+++ b/build-in-docker.sh
@@ -99,12 +99,12 @@ build_inside() {
 }
 
 make_image() {
-  echo "${yellow}=> ${normal}Building BlueOcean docker image ${tag_name}"
-  (cd "$HERE" && docker build -t "$tag_name" . )
+  echo "${yellow}=> ${normal}Building BlueOcean docker development image ${tag_name}"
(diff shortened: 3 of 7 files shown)


ndeloof:
moving this to maven-hpi-plugin is imho the best option, as we get there best understanding about maven dependency resolution, optional stuff, version conflict resolution as well as jenkins specific packaging. As a result this can produce a set of artifacts ID and/or handle copying them to some directory (can maybe delegate / reuse code from dependency:copy-dependencies)

michaelneale:
@ydubreuil @ndeloof Is https://github.com/jenkinsci/blueocean-plugin/pull/393 still valid? ie 2 different efforts? (the latter doesn't work with the experiemental update center when I tried)

ydubreuil:
@michaelneale we can try to fix #393 to use the URL of the experimental UC but I'm having doubts that dependency resolution will work as expected. Could be worth the try.

ydubreuil:
Actually, maven-hpi-plugin has a [`bundle-plugins`](https://github.com/jenkinsci/maven-hpi-plugin/blob/master/src/main/java/org/jenkinsci/maven/plugins/hpi/BundlePluginsMojo.java#L48) goal, not documented in the plugin site, which is why I missed it. I'll try it.

pom.xml
```diff
@@ -181,7 +182,13 @@
 
         <dependency>
             <groupId>${project.groupId}</groupId>
-            <artifactId>blueocean-plugin</artifactId>
```
michaelneale: why was this changed?

blueocean-dist/pom.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
michaelneale: this module really only exists for docker - perhaps a better name than "dist"? Or will it have other uses?

pom.xml
ydubreuil: `blueocean-plugin` is only the name of the folder module, the artifact name is `blueocean`, see [pom.xml](https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-plugin/pom.xml#L11)

blueocean-dist/pom.xml
ydubreuil: > Or will it have other uses?

That's what I was wondering when choosing dist, but I don't find any other use right now, and given we will not publish the artifact, it'll not be an issue the rename it latter if needed. So, I'm also in favor pick up a more explicit name. `blueocean-docker-dev-image` could be a good feet, but maybe getting rid of this module would be even better?

I created this module to cleanly separate our 2  Dockerfile, but if we choose to move #403 to a separate repository, we could move this back into `blueocean-plugin/pom.xml` and a Dockerfile in the root folder.

ydubreuil:
Changed the labels, ready to be reviewed, only the newly introduced module might have it's name changed

pom.xml
michaelneale: hrm... I am not sure how it works currently then?

michaelneale:
Given the suggestion to move #403 to another repo - perhaps this could be part of the aggregator? I'll start an email chain with vivek to see, as this is a bit fiddly to explain to people why it is here. 

In theory the Dockerfile in the root of a project could mean the environment to build it in, or how to build the distribution of it, but isn't clear here...

blueocean-dist/pom.xml
michaelneale: I am inclined to the latter - this may make more sense in the aggregator module. 

A dockerfile in the root would make sense for this IMO. (the only downside is, in some cases, we use the Dockerfile to specify the environment needed for building the project itself, not how to distribute, so it isn't quite as clear cut to me).

michaelneale:
@ydubreuil I think @vivek is ok with this, so looks good to me if you think this is the best way to go.

ydubreuil:
@michaelneale  I can't move this to the aggregator module and get rid of `blueocean-dist` module. maven-hpi-plugin requires HPI dependencies like this to work:

```
        <dependency>
            <artifactId>blueocean</artifactId>
            <version>${project.version}</version>
            <type>hpi</type>
            <scope>provided</scope>
        </dependency>
```
and I can't add that to `blueocean-plugin`, it fails to build. I can move this into a submodule of  `blueocean-plugin`, it will be less disturbing, and call it `blueocean-plugins-all-dist`.

michaelneale:
@ydubreuil ok moving to a submodule (looks like this needs some updating anyway, when you get around to it) looks sensible.

michaelneale:
Closing this for now until yoann gets a chance to revisit it

ydubreuil:
@michaelneale Back on this, it will help to have this finished for the PR auto deployment.  I was not able to move the code to a sub-module, so I created a Maven profile to isolate it from a standard build.

I squashed commits and rebased it from master to make it easier to review. Doing some tests now, I will push shortly.

michaelneale:
ok thanks @ydubreuil 

This is looking better now I think. There are a few deps version I see you bumped - presumably required to make it actually work? (we probably don't hit them with UC installs as it likely already has a newer version) - but might want to check them off with @vivek 

So this is a profile which runs a submodule, but otherwise will stay out of the way?

build-in-docker.sh
```diff
@@ -99,12 +99,12 @@ build_inside() {
 }
 
 make_image() {
```
michaelneale: can chance this file could go in bin with others?

blueocean-pipeline-api-impl/pom.xml
```diff
@@ -67,7 +67,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-durable-task-step</artifactId>
-            <version>2.0</version>
+            <version>2.4</version>
```
michaelneale: Look ok @vivek ?

blueocean-docker-dev-image/pom.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.jenkins.blueocean</groupId>
+        <artifactId>blueocean-parent</artifactId>
+        <version>1.0.0-b07-SNAPSHOT</version>
```
michaelneale: hrm... will this be auto updated with the maven release plugin ? (as it isn't the same as other modules).

build-in-docker.sh
ydubreuil: I can move it yep

blueocean-docker-dev-image/pom.xml
michaelneale: Also - will this be deployed as a plugin as part of a release? if to - an entry will need to exist in: https://github.com/jenkins-infra/repository-permissions-updater/tree/master/permissions

Ideally this is NOT deployed (as it is only for dev time use) right?

blueocean-pipeline-api-impl/pom.xml
ydubreuil: I had to bump it to be able to start all plugins correctly. (This makes the docker image a great test to ensure all plugins dependencies are coherent)

blueocean-docker-dev-image/pom.xml
ydubreuil: Didn't think about this, might be better to not isolate that in a module then... given that building this module only takes a few seconds.

blueocean-pipeline-api-impl/pom.xml
michaelneale: +2

blueocean-docker-dev-image/pom.xml
michaelneale: If you have an alternative way, may be better

ydubreuil: um, I don't have an alternative right now, I'll try to come up with something, it doesn't make sense to have this deployed on Maven.

pom.xml
```diff
@@ -107,6 +107,7 @@
     <module>blueocean-config</module>
     <module>blueocean</module>
     <module>blueocean-jwt</module>
+    <module>blueocean-docker-dev-image</module>
```
michaelneale: so for folks building blue ocean on their desktop - this won't add significant overhead?

ydubreuil: it's just a matter of copying files to a target directory, so 3 seconds lost maybe

michaelneale:
I think this is ok. 
Adds about 20 seconds to local build - which isn't too bad.. 

ideally this would be in a separate code base, as it seems kind of unrelated, but if this is where it has to be, then so be it. I would like @tfennelly to double check but :bee: from me

tfennelly:
🐝 LGTM insofar as I get all that's happening. Nice job @ydubreuil !!!! 🍻 

I see no reason not to run with something like this and see how it works and evolve it over time. Even if it doesn't work in the end, it's not like it's a product feature that we can'd just remove.

tfennelly:
@ydubreuil btw ... the ATH uses hpi assembly and it seems to work fine.

tfennelly:
> @ydubreuil btw ... the ATH uses hpi assembly and it seems to work fine.

Before some smart-ass says something ... I mean the hpi assembly part of it seems to work fine ... the ATH itself has serious issues that I'm struggling to resolve, mostly because docker is wrecking my head ... @ydubreuil , I might ping yu about that and see can you help me there 😄

ydubreuil:
Great, let's squash and merge then!

PR #398, opened 2016-08-09
[FIXED JENKINS-35826] pagination

kzantow:
# Description

See [JENKINS-35788](https://issues.jenkins-ci.org/browse/JENKINS-35788).

To review: add > page size item lists (can set it locally to a small value [here](https://github.com/jenkinsci/blueocean-plugin/pull/398/files#diff-e77e4253f1c689109685078aa2887065R7), if necessary). Validate that only the default page size is retrieved first, validate clicking to load more works as expected. Validate direct linking to downstream pages works correctly and all downstream pages also continue to work. Also check that the loading animations aren't horribly screwed up (I hope to make some changes later, maybe to make them more attractive but shouldn't hold this PR up). To sum up: test the whole app (yes, sorry, but the changes were that extensive due to the way a lot of it was structured).

# Submitter checklist
- [x] validate SSE events are still working as expected
- [x] Fix issue causing page sizes too small when getting pipelines list
- [x] Branches list
- [x] PR list
- [x] Change default fetch size to 25

# Reviewer checklist
- [ ] Reviewed the code


blueocean-dashboard/.vscode/launch.json
```diff
@@ -0,0 +1,46 @@
+{
```
kzantow: I hate Eclipse. I tell it NOT to commit this, yet...

blueocean-dashboard/src/main/js/util/smart-fetch.js
```diff
@@ -0,0 +1,277 @@
...
+         * class metadata.
+         */
+        this.fetchCallbacks = {};
+        this.promises = {};
+    }
+
+    /**
+     * Generalization of consolidation of duplicate requests
+     */
+    execute(key, promiseProvider, onComplete) {
```
tfennelly: What is the key? The url?

tfennelly: And what's the `promiseProvider`?

blueocean-dashboard/src/main/js/util/smart-fetch.js
```diff
@@ -0,0 +1,277 @@
...
+                });
+        } else {
+            // We already have an in-flight request, just store the onComplete callback
+            callbacks.push(onComplete);
+        }
+    }
+
+
+    /**
+     * Generalization of consolidation of duplicate requests
```
tfennelly: "Generalization of consolidation of duplicate requests" ... duplicated on this function and the previous. In both cases, it doesn't really help me understand what the functions are really doing. Sounds good though :)

blueocean-dashboard/src/main/js/util/smart-fetch.js
```diff
@@ -0,0 +1,277 @@
...
+            try {
+                onComplete(data);
+            } catch (e) {
+                // TODO how best to handle failures
+                console.error(e); // eslint-disable-line no-console
+            }
+            return data; // for downstream .then
+        });
+        return holder.promise;
+    }
```
tfennelly: So it uses one promise instance to handle all the callers, stringing them together on the `then`? If so, that seems baaaadddddd to me.

Would also explain why there's a TODO wrt handling the errors i.e. if each caller was given their own `Promise` then passing back the errors would be trivial.

blueocean-dashboard/src/main/js/util/smart-fetch.js
```diff
@@ -0,0 +1,277 @@
...
+
+/**
+ * Static fetch options used for every request
+ */
+const fetchOptions = { credentials: 'same-origin' };
+
+/**
+ * DuplicateCallTracker maintains active fetches against a particular key
+ */
+export class DuplicateCallTracker {
```
tfennelly: Does it need to be `export`ed?

blueocean-dashboard/src/main/js/util/smart-fetch.js
```diff
@@ -0,0 +1,277 @@
...
+            const hasMore = data.length > limit;
+            const outData = assignObj(concatenator(pager, existingData, data));
+            outData.$success = true;
+            outData.$pager = Object.assign(pager, {
+                current: first,
+                hasMore,
+                currentData: outData,
+                startIndex: existingData.length > 0 ? pager.startIndex : first,
+            });
+            Object.freeze(outData); // already deep frozen
```
tfennelly: What does the "already deep frozen" comment here mean? If it's already frozen then why call it again?

blueocean-dashboard/src/main/js/util/smart-fetch.js
```diff
@@ -0,0 +1,277 @@
...
+        applyFetchMarkers(out, this);
+        return out;
+    };
+    if (vals) {
+        Object.assign(obj, vals);
+    }
+    return obj;
+}
+
+const fetchPagedData = (pager, concatenator, onData, existingData, first, limit) => {
```
tfennelly: Does this function not belong as part of `Pager`?

blueocean-dashboard/src/main/js/util/smart-fetch.js
```diff
@@ -0,0 +1,277 @@
...
+    getTotalPages() {
+        return Math.floor((this.startIndex + this.currentData.length) / this.pageSize);
+    }
+}
+
+/**
+ * For data returned as a JSON array (rather than nested in some unknown object structure)
+ * this will work to concatenate pages properly, and is the default
+ */
+export function arrayConcatenator(pager, existing, incoming) {
```
tfennelly: Again, belong on `Pager`?

blueocean-dashboard/src/main/js/util/smart-fetch.js
kzantow: The problem with that... is that each function will be called multiple times, and promises are supposed to resolve or reject exactly once.

kzantow: Hmm, you might be right about this part, just use the promise and resolve or reject all of them, I'll try to make that work.

blueocean-dashboard/src/main/js/util/smart-fetch.js
kzantow: *deep* frozen, the holder object is not, it's a shallow freeze @tfennelly

blueocean-dashboard/src/main/js/util/smart-fetch.js
tfennelly: @kzantow right ... have one promise on the actual fetch request, and then a promise for each caller, resolving/rejecting all based on the resolve reject of the fetch promise.

blueocean-dashboard/src/main/js/util/smart-fetch.js
tfennelly: Sounds like the comment needs to be made clearer. In fact ... even with your above comment I'm not sure I know what's what :)

blueocean-dashboard/src/main/js/util/smart-fetch.js
kzantow: @tfennelly right exactly, turns out it's actually easier than that, since that behavior is built into promises

michaelneale:
Hrm when I try to open any job from the dashboard I get the beautiful 404 screen... but can't open anything (show more works for dashboard, btu not loading anything) 🐛

kzantow:
@michaelneale aha, yes auth... fail on my part, I'll get that fixed up tonight.

michaelneale:
IIRC there was something that had to be added to fetch to pass along cookies with the request (added by @scherler)

michaelneale:
Close, but still no 🍌. When things are in folders they don't seem to load. Eg I created a folder, put a freestyle job in it: 🐜 

```
(anonymous function) @ jenkins-js-extension.js:16763
http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/domainscopes%2Fblueocean-demo-project Failed to load resource: the server responded with a status of 404 (Not Found)
```

Also, when I use github org folder, same thing happens (as it effectively creates an additional layer):

```
jenkins-js-extension.js:17633 GET http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/domainscopes%2Fblueocean-demo-project 404 (Not Found)(anonymous function) @ jenkins-js-extension.js:17633dedupe @ jenkins-js-extension.js:17390dedupe @ jenkins-js-extension.js:17413fetch @ jenkins-js-extension.js:17632(anonymous function) @ jenkins-js-extension.js:16196(anonymous function) @ redux-thunk-2.0.1.js:1124(anonymous function) @ react-redux-4.4.5.js:6091componentWillMount @ jenkins-js-extension.js:12104performInitialMount @ blueocean.js:23415mountComponent @ blueocean.js:23323mountComponent @ blueocean.js:30622performInitialMount @ blueocean.js:23436mountComponent @ blueocean.js:23323mountComponent @ blueocean.js:30622_updateRenderedComponent @ blueocean.js:23862_performComponentUpdate @ blueocean.js:23832updateComponent @ blueocean.js:23751receiveComponent @ blueocean.js:23650receiveComponent @ blueocean.js:30703_updateRenderedComponent @ blueocean.js:23854_performComponentUpdate @ blueocean.js:23832updateComponent @ blueocean.js:23751receiveComponent @ blueocean.js:23650receiveComponent @ blueocean.js:30703_updateRenderedComponent @ blueocean.js:23854_performComponentUpdate @ blueocean.js:23832updateComponent @ blueocean.js:23751receiveComponent @ blueocean.js:23650receiveComponent @ blueocean.js:30703updateChildren @ blueocean.js:21751_reconcilerUpdateChildren @ blueocean.js:29343_updateChildren @ blueocean.js:29442updateChildren @ blueocean.js:29430_updateDOMChildren @ blueocean.js:25049updateComponent @ blueocean.js:24878receiveComponent @ blueocean.js:24834receiveComponent @ blueocean.js:30703updateChildren @ blueocean.js:21751_reconcilerUpdateChildren @ blueocean.js:29343_updateChildren @ blueocean.js:29442updateChildren @ blueocean.js:29430_updateDOMChildren @ blueocean.js:25049updateComponent @ blueocean.js:24878receiveComponent @ blueocean.js:24834receiveComponent @ blueocean.js:30703updateChildren @ blueocean.js:21751_reconcilerUpdateChildren @ blueocean.js:29343_updateChildren @ blueocean.js:29442updateChildren @ blueocean.js:29430_updateDOMChildren @ blueocean.js:25049updateComponent @ blueocean.js:24878receiveComponent @ blueocean.js:24834receiveComponent @ blueocean.js:30703_updateRenderedComponent @ blueocean.js:23854_performComponentUpdate @ blueocean.js:23832updateComponent @ blueocean.js:23751receiveComponent @ blueocean.js:23650receiveComponent @ blueocean.js:30703_updateRenderedComponent @ blueocean.js:23854_performComponentUpdate @ blueocean.js:23832updateComponent @ blueocean.js:23751receiveComponent @ blueocean.js:23650receiveComponent @ blueocean.js:30703_updateRenderedComponent @ blueocean.js:23854_performComponentUpdate @ blueocean.js:23832updateComponent @ blueocean.js:23751performUpdateIfNecessary @ blueocean.js:23664performUpdateIfNecessary @ blueocean.js:30736runBatchedUpdates @ blueocean.js:31656perform @ blueocean.js:33897perform @ blueocean.js:33897perform @ blueocean.js:31595flushBatchedUpdates @ blueocean.js:31682closeAll @ blueocean.js:33963perform @ blueocean.js:33910batchedUpdates @ blueocean.js:27324batchedUpdates @ blueocean.js:31603dispatchEvent @ blueocean.js:28309
jenkins-js-extension.js:17583 ERROR:  Response {type: "basic", url: "http://localhost:8080/jenkins/blue/rest/organizati…s/pipelines/domainscopes%2Fblueocean-demo-project", status: 404, ok: false, statusText: "Not Found"…}
```

Multibranch works, as does freestyle at the top level.

kzantow:
@michaelneale thanks much for the testing/feedback glad you can think of these semi-obvious cases I missed :)

michaelneale:
OK - will try it again later. Need to also check cases where the branch name is "foo/bar" etc... @cliffmeyers  has a good test set of pathological cases... 

Has this been run over the ATH yet?

michaelneale:
OK close, I can open all the things now, but 


1) karaoke doesn't work for multibranch correctly. It never starts or follows along. There are no console errors but on the server I see: 

"Aug 10, 2016 5:35:19 PM org.jenkinsci.plugins.ssegateway.sse.EventDispatcher unsubscribe
WARNING: Invalid SSE unsubscribe configuration. No active subscription matching filter: "

Implying that it isn't matching the correct SSE event. 

2) When I follow along a non multibranch one, it follows the stages, but the header never changes colour (ie it thinks it is running when when finished). 

To reproduce: 

1) use kzantow/failure-project, try building the michael branch and following along (or any really)
2) use this pipeline: 
```
node {
    stage "start"
    sleep 4
    stage "finish"
    sh "echo yeah"    
}
```
screen shot 2016-08-10 at 5 37 17 pm
screen shot 2016-08-10 at 5 37 45 pm





Run it and follow along, you will see that it will follow along but never show as finished.

kzantow:
@michaelneale re: seeing less than 2 items, there's a problem with filtering/pagination (I think with folders) server side causing that, I'll include a fix here

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultiBranchPipelineImpl.java
```diff
@@ -206,6 +211,16 @@ public BluePipelineContainer getBranches() {
     }
 
     @Override
+    @Navigable
+    public BluePipelineContainer getFindBranches(final String type) {
```
kzantow: @vivek would appreciate taking a glance at these server-side changes to validate they make sense. I need to paginate PRs separate from branches on the server side.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ExtensionClassContainerImpl.java
```diff
@@ -67,11 +69,9 @@ public Link getLink() {
 
     private Class getClazz(String name){
         try {
-            return this.getClass().getClassLoader().loadClass(name);
+            return Jenkins.getInstance().getPluginManager().uberClassLoader.loadClass(name);
```
kzantow: @vivek @imeredith I'm not sure how this was working before (my entire BO app was broken), it shouldn't have been AFAICT, please let me know if this is not the right thing to do for some reason.

vivek: @kzantow 👍

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueMultiBranchPipeline.java
```diff
@@ -65,6 +65,11 @@
     public abstract BluePipelineContainer getBranches();
 
     /**
+     * @return Gives {@link BluePipelineContainer}
+     */
+    public abstract BluePipelineContainer getFindBranches(String type);
```
vivek: @kzantow I believe it results in to findBranches API, to keep it restful we should avoid adding such APIs, this should really go as search API, see OmniSearch or maybe pass query parameter in getBranches() to return filtered collection of branches.

Maybe we do it as separate PR?

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/Containers.java
```diff
@@ -36,7 +37,14 @@ public T get(String name) {
 
             @Override
             public Iterator<T> iterator(int start, int limit) {
-                return base.subList(start,start+limit).iterator();
+                if (start >= base.size()) {
```
vivek: @kzantow This change looks good. I think there might be other places similar logic needs  to be applied, if so use a common mechanism or List wrapper to return subList with check for stat/end?

michaelneale:
No this still isn't right. It isn't tracking the fact that the job finished, if it is a regular pipeline, and a multibranch pipeline, when running, never follows along, it sits in the queued/grey state for ever. 🐛

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ExtensionClassContainerImpl.java
```diff
@@ -74,5 +74,3 @@ private Class getClazz(String name){
         }
     }
 }
-
```
kzantow: oops!

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/ContainerFilter.java
```diff
@@ -0,0 +1,80 @@
...
+import hudson.ExtensionPoint;
+import hudson.model.Item;
+
+/**
+ * Simple extension point to allow filtering item types by a specific key
+ * This can be used anywhere a list of items are returned, and it will examine
+ * the current request for a ?filter=filter1(param:value),filter2 parameter,
+ * okay the last bit is a future addition once/if OmniSearch and this are consolidated
+ */
+public abstract class ContainerFilter implements ExtensionPoint {
```
vivek: @kzantow this should be moved to rest-impl module, we are avoiding adding hudson.model.* reference to blueocean-rest. This is to keep it purely abstract - only thing it depends on is stapler. I know it being plugin gives core classes but we are really trying to keep it independent of that, see BluePipelineFactory or BlueRunFactory, they are in rest-impl.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineJobFilters.java
```diff
@@ -0,0 +1,85 @@
...
+            public boolean apply(Item job) {
+                if (!isPullRequest(job)) {
+                    return true;
+                }
+                return false;
+            }
+        };
+        @Override
+        public String getName() {
+            return "origin";
```
vivek: What is this 'origin', why hard coded?

vivek:
@kzantow ContainerFilter looks great. Looking at how you are using, it brings filtering capability for any container. Only 🐛 I have is to move it out of blueocean-rest to blueocean-rest-impl module. Otherwise its awesome. Nice job!

michaelneale:
Marked as work in progress as the SSE events are still WIP (last I heard...) and last I checked. 

GETTING CLOSE!

cliffmeyers:
Just going to list a couple of issues I found while testing:
- Progress bar overlays very top of screen when pressing "Load More" repeatedly and scrolling main header off screen
- Make sure you have a dozen or so pipelines in the main list. Press "Load More" repeatedly, and click through to one from the bottom of the list. Then click its most recent Run. Copy/paste the URL of the Run Details into a new browser window. Close the modal, then navigate back to Pipelines from main nav. Click "load More" - note that the button becomes disabled prematurely, user is unable to load the entire list.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/ContainerFilter.java
kzantow: @vivek the only thing with moving that to the `impl` is plugins would have to depend on `blueocean-rest-impl` in order to add filter types for this, do you think that's fine?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineJobFilters.java
kzantow: @vivek - just needed some sort of identifier, any better ideas? 'origin' as in the typical git term for the main repo rather than a different remote

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/ContainerFilter.java
vivek: @kzantow yes this is fine as this is indeed single instance jenkins specific extension. We are keeping such extension points (e.g. BluePipelineFactory, BlueRunFactory) inside rest-impl.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineJobFilters.java
vivek: @kzantow This is fine, looking at ContainerFilter, its ok as the name is specific to a filter.

blueocean-dashboard/src/main/js/util/dedupe-calls.js
```diff
@@ -0,0 +1,48 @@
...
+        // get active or create
+        return this.promises[key] || (this.promises[key] =
+            promiseCreator()
+            .then((data) => {
+                delete this.promises[key];
+                return data;
+            })
+            .catch((err) => {
+                delete this.promises[key];
+                return Promise.reject(err);
```
cliffmeyers: Will this fail in IE?

blueocean-dashboard/src/main/js/util/UrlUtils.js
```diff
@@ -140,10 +140,90 @@ export function calculateRunLogURLObject(config) {
...
+            url += '/activity';
+        }
+        url += '/detail' +
+        (branch ? ('/branch/' + uriString(branch)) : '') +
+        '/' + encodeURIComponent(runId) +
+        (tab ? ('/' + tab) : '');
+    }
+    /* eslint-enable prefer-template */
+    return url;
+}
```
cliffmeyers: As we add more routes to the app, we will probably need to split this into multiple functions otherwise this logic may get unwieldy. Should we have a function for each route so that each one is small / simple?

blueocean-dashboard/src/main/js/util/smart-fetch.js
```diff
@@ -0,0 +1,211 @@
...
+            })
+            .catch(err => {
+                _onData({ $failed: err });
+            });
+    }
+    // return a fake promise, a thenable
+    // so it can be resolved multiple times
+    return {
+        then: (fn) => fetch(url, _onData, fn),
+    };
```
cliffmeyers: Is the intent of this `fetch` method that the user should provide a callback function in all cases? So we won't support returning a Promise it?

blueocean-dashboard/src/main/js/util/UrlUtils.js
kzantow: @cliffmeyers  I don't think so, we should modify this function to examine the routes & params and build URLs based on that, but I don't think it's necessary yet. Besides, this is specific to this module, it accounts for only the routes defined here, other modules have to come up with their own way of linking to their own routes.

cliffmeyers: @kzantow right, I guess we'll have to see how things like the pipeline editor and organization management get structured.

blueocean-dashboard/src/main/js/util/smart-fetch.js
kzantow: Yes, it's a fake promise as noted in the comment, so you can use it like:

`fetch(url, data => { ... })`

or

`fetch(url).then(data => { ... }).catch()`, which will work, as what is returned from `.then` is a real `Promise`. The issue is that the method is called multiple times, and we don't want a promise resolving and using a cached result to prevent multiple invocations of the callback...

blueocean-dashboard/src/main/js/util/dedupe-calls.js
kzantow: @cliffmeyers if this fails in IE, `isomorphic-fetch` will also fail, I think and we have bigger problems.

cliffmeyers: Per https://github.com/matthew-andrews/isomorphic-fetch we should probably add the "es6-promise"; just jumped out at me because it's first time I saw a ref to native Promise in our code.

blueocean-dashboard/src/main/js/util/UrlUtils.js
cliffmeyers: Note, the ability to click on a Favorite card in main Dashboard is broken since the URL structure to run details has changed. It'd be good if we could promote this code so that it could be shared between the plugins. I started the Commons JS project in #404 - we could possibly get that merged to master, then refactor this func into the shared lib? Otherwise maybe just dup the code in "blueocean-personalization" temporarily.

cliffmeyers:
@kzantow am sure you're aware that with the URL change to run details that the ATH has numerous failures.

kzantow:
@cliffmeyers no, but it's not surprising, it's next on my list to check out and address :)

cliffmeyers:
@kzantow sure thing :)

blueocean-dashboard/src/main/js/util/dedupe-calls.js
kzantow: @cliffmeyers good call, I didn't realize support wasn't even in IE 11, only Edge

blueocean-dashboard/src/main/js/util/UrlUtils.js
kzantow: @cliffmeyers right, also, keep in mind this is just for these basic pages. maybe we can simplify a lot of this restructuring the router/context but for now, I just want things to "work" if i pass in a pipeline to get a link to the right thing.

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
```diff
@@ -12,37 +12,40 @@ import * as pushEventUtil from './util/push-event-util';
 const { object, array, func, node, string } = PropTypes;
 
 class OrganizationPipelines extends Component {
-
-    // FIXME: IMO the following should be dropped
+    constructor(props) {
+        super(props);
+        this.state = {};
```
scherler: You can drop the whole constructor, since you do not set anything on the state as default.

kzantow: Umm... state is null without this?

scherler: and? this.setState will not care.

Anyway I just looked for the usage of this.state and you only use it for one compare, but IMO you could do without, only comparing the orgaName for current props and nextProps.

No need for state at all.

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
```diff
@@ -11,24 +16,37 @@ import {
...
+        };
+    }
+    
+    componentWillMount() {
+        if (this.props.params) {
+            this.props.fetchPipeline(this.props.params.organization, this.props.params.pipeline);
+        }
+    }
+    
+    _componentWillReceiveProps(nextProps) {
```
scherler: hmm _componentWillReceiveProps is highly confusing with the API method, can you not choose another name?

michaelneale:
ping @kzantow

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
scherler: state is evil, rule of thumb do not use it. ;) 

seriously, state is tricky since you cannot persist state of a component. Props you can persist!

i386:
Where are we at with this?

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -227,6 +327,26 @@ exports.fetchLogsInjectStart = function fetchLogsInjectStart(url, start, onSucce
             }
         });
 };
+
+/**
+ * Determines if the provided object is a "run" type
+ */
+function isRun(obj) {
+    return obj._class === 'io.jenkins.blueocean.rest.impl.pipeline.PipelineRunImpl'
```
imeredith: This should be using capabilites,

kzantow: Is there a capability for the run type? Are the capabilities documented anywhere, btw? @imeredith

scherler: https://github.com/jenkinsci/blueocean-plugin/tree/master/blueocean-rest#classes-api

kzantow: I know that, I need to know specifically what capability I should be looking for with this, I can check out the Java code, was just wondering if there was anything documented

michaelneale: cc @vivek - is there a list that can be looked at, or is it best to either look at the code or hit the api and see what it returns and guess from the names what the capabilities return look like?

michaelneale: @kzantow from vivek: 

"Scan blueocean-rest-impl and blueocean-rest module for @capabilities annotation on the relevant Java classes that corresponds to relevant json object. Basically knowledge of backend model. If ui folks can tell what generic capabilities they are looking for I can help them create a doc with corresponding capabilities name. 

Few known things come to my mind are, run, branch, types of runs, test types, log... 
"

He is on the road this evening, but is considering adding common ones to README if that will help.

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
tfennelly: +1 on -1 for `_componentWillReceiveProps` ... don't name things **nearly the same** as something else that's very fundamental.

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
```diff
@@ -100,25 +105,28 @@ class OrganizationPipelines extends Component {
...
+    
+    _getOrganizationName(nextProps) {
+        if (nextProps && nextProps.params) {
+            return nextProps.params.organization;
+        }
+        if (this.props && this.props.params && this.props.params.organization) {
+            return this.props.params.organization;
+        }
+        if (this.context && this.context.params && this.context.params.organization) {
+            return this.context.params.organization;
```
tfennelly: I'm sure it's naivety wrt react on my behalf, but the fact that you needed to write code like the above highlights to me why react props/context is a shity way of passing data around. 😢

blueocean-dashboard/src/main/js/components/MultiBranch.jsx
```diff
@@ -118,7 +124,13 @@ export class MultiBranch extends Component {
                         })
                         }
                     </Table>
+                    {branches.$pager &&
+                        <button disabled={!branches.$pager.hasMore} className="btn-show-more btn-secondary" onClick={() => branches.$pager.fetchMore()}>
```
tfennelly: Is there a reason for the `$` prefix on some of these property names?

blueocean-dashboard/src/main/js/components/PageLoading.jsx
```diff
@@ -0,0 +1,56 @@
...
+    constructor(props) {
+        super(props);
+        this.state = {
+            percentage: undefined,
+            running: false,
+        };
+    }
+    
+    componentWillMount() {
+        if (this.props.duration) {
```
tfennelly: What exactly is `duration`? The name suggests it's the total length of time for which the progress bar remains visible (the 100%), but I'm not sure from the code.

blueocean-dashboard/src/main/js/components/Pipelines.jsx
```diff
@@ -78,6 +68,12 @@ export default class Pipelines extends Component {
                                 })
                             }
                         </Table>
+                        
+                        { pipelines && pipelines.$pager &&
+                            <button disabled={!pipelines.$pager.hasMore} className="btn-show-more btn-secondary" onClick={() => pipelines.$pager.fetchMore()}>
+                                {pipelines.$pending ? 'Loading...' : 'Show More'}
```
tfennelly: Maybe the button `disabled` attribute should also be influenced by `pipelines.$pending`? Seems like it should also be disabled when "Loading ..." ?

tfennelly:
Did basic test on it and I see that it loses the set of visible records when I move between pages/routes e.g. it shows 2 pipelines on the top level page ... click show more to see 2+ ... click on one of the pipelines ... go back to top level page and it just shows 2 again.

kzantow:
@tfennelly yes, I intentionally removed caching here. Keep in mind, the default fetch size will be much larger, so it'll fill the page. I'm planning on addressing caching later, probably not in this PR. I think it's important to determine that it's actually a problem before actually solving it.

blueocean-dashboard/src/main/js/components/MultiBranch.jsx
kzantow: @tfennelly no, do you have a better idea? I'm happy to replace them with underscores or something else, this is to not conflict with other properties.

tfennelly:
@kzantow ah ok ... I'd have thought that data was stored in the redux store and would just have been "there" the next time you go back.

scherler:
@kzantow since you just mentioned it. My "personal problem" with the PR is that you removed the caching and nuked the current way to work in redux (all actions have been changed). I would have tried to not implement the feature so intrusive.  I am not fan of "have removed it but will implement it later again", later never comes!

kzantow:
@scherler you realize, though, that this was necessary. I had to change the entire way things were being fetched from top to bottom. Without this change, direct links would not work.

i386:
This looks great BTW!

Quick question, on Chrome I get a focus ring around the "Show More" button - is that intentional? many of our buttons don't have one.

jenkins blue ocean 2016-08-25 09-07-58

i386:
🐛 clicking on the card does not open the modal (404)

aug -25-2016 10-20-19

kzantow:
@i386 yup, I hadn't pushed the fix. I had to undo the change to fix the background flash when you open modals, it changed the URLs. better to do that in a different PR, anyway as part of the other ticket

kzantow:
@i386 also, regarding the blue outline, that's a default bootstrap focus style, we should probably see that on other things or we probably have some accessibility issues

i386:
@kzantow OK I am fine with the outline. Ill take your most recent changes for a whirl.

i386:
@kzantow OK can confirm the cards work. However, there is a 🐛 . If you click the card on the dashboard then close the model you do not end up on the dashboard again.

aug -25-2016 11-53-07

scherler:
@kzantow I can think about a solution that would not had that big impact on the underlying code. You could have added pagination on top of what we had. Anyway now you have implemented fetching your way let us see whether that is better then what we had before. Loosing the full working cache feature for pagination because "we do it later better" is for me no reason to remove it!

michaelneale:
I have found a :bug: that navigating from the dashboard into a run result leaves me with a blank activity screen when dismissed, as per: 



(doesn't matter what type of job).

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
michaelneale: Is this a case where (ugh) destructuring would work? There is lodash "get" (but its stringy, as are many), or other weird suggestions: http://stackoverflow.com/questions/2631001/javascript-test-for-existence-of-nested-object-key

blueocean-dashboard/src/main/js/PipelineRoutes.jsx
```diff
@@ -14,34 +12,72 @@ import {
...
+    return {
+        onEnter: ({ location, params }, replace) => {
+            replace(relativeUrl(location, next));
+        },
+    };
+}
+
+// multibranch-pr-github/detail/test-pass-fail-stage/1/pipeline
+// :pipeline/detail/:branch/:runId/pipeline
+const runDetailRoutes = { path: ':runId', component: RunDetails,
```
michaelneale: Does this subsume some of what was sketched in https://github.com/jenkinsci/blueocean-plugin/pull/362 ?

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
```diff
@@ -12,37 +12,34 @@ import * as pushEventUtil from './util/push-event-util';
...
         const config = this.context.config;
         if (config) {
-            const { organization } = this.context.params;
-            this.props.fetchPipelinesIfNeeded(this.context.config, organization);
-
-            const _this = this;
-
+            const organizationName = this._getOrganizationName();
+            
+            if (organizationName) {
```
michaelneale: this pattern appears a few times, would it not make sense for fetchOrg pipelines to take a null org name or something like that so it all collapses to one line in all cases?

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
```diff
@@ -59,25 +56,29 @@ class OrganizationPipelines extends Component {
                     // crud operations are relative low frequency, so not much
                     // benefit to be got from optimizing things here.
                     // TODO: fix https://issues.jenkins-ci.org/browse/JENKINS-35153 for delete
-                    _this.props.fetchPipelines(_this.context.config, _this.props.organization);
-                    _this.props.updateBranchList(eventCopy, _this.context.config);
+                    if (this._getOrganizationName()) {
```
michaelneale: similar pattern again. Isn't this already assigned a variable above? is `this` something else now?

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
```diff
@@ -88,10 +89,14 @@ class OrganizationPipelines extends Component {
     }
 
     componentWillReceiveProps(nextProps) {
-        const config = this.context.config;
-        const { pipeline } = nextProps.params;
-        config.pipeline = pipeline;
-        this.props.setPipeline(config);
+        const organizationName = this._getOrganizationName(nextProps);
+        if (this._getOrganizationName(this.props) !== organizationName) {
```
michaelneale: ditto, same pattern.

michaelneale: similar pattern again

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
```diff
@@ -140,15 +147,15 @@ OrganizationPipelines.propTypes = {
...
 
 OrganizationPipelines.childContextTypes = {
     pipelines: array,
-    pipeline: object,
 };
 
-const selectors = createSelector([pipelinesSelector],
-    (pipelines) => ({ pipelines }));
+const selectors = createSelector([allPipelinesSelector, organizationPipelinesSelector],
+    (allPipelines, organizationPipelines) => ({ allPipelines, organizationPipelines }));
```
michaelneale: how come this is now split in twain, org pipelines vs all? (I remember some discussion about it, just not sure how it relates to this... remind me?)

blueocean-dashboard/src/main/js/components/MultiBranch.jsx
michaelneale: tasty perl influence. He loves the perl.

blueocean-dashboard/src/main/js/components/PageLoading.jsx
michaelneale: not sure, but it looks real prrty

blueocean-dashboard/src/main/js/components/PageLoading.jsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
```
michaelneale: this looks so nice to use... it really heklps the experience.

blueocean-dashboard/src/main/js/components/Pipelines.jsx
```diff
@@ -66,7 +57,7 @@ export default class Pipelines extends Component {
                           className="pipelines-table fixed"
                           headers={headers}
                         >
-                            { pipelineRecords
+                            { pipelines && pipelines
```
michaelneale: wow that looks odd when the map is on the next line (i assume null check?)

blueocean-dashboard/src/main/js/components/Pipelines.jsx
michaelneale: yeah testing locally its too fast to load, so I guess, do you want to let people just click it a bunch of times to batch it up? or disable while loading... (either... could be argued for if I get the meaning)

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
scherler: @tfennelly the problem is that we use both instead to focus on one.

blueocean-dashboard/src/main/js/components/PullRequest.jsx
```diff
@@ -2,7 +2,7 @@ import React, { Component, PropTypes } from 'react';
 import { LiveStatusIndicator, ReadableDate } from '@jenkins-cd/design-language';
 import Extensions from '@jenkins-cd/js-extensions';
 import RunPipeline from './RunPipeline.jsx';
-import { buildRunDetailsUrl } from '../util/UrlUtils';
+import { getLocation } from '../util/UrlUtils';
```
michaelneale: is this the one getLocation to rule them all? or does it just replace run details url?

blueocean-dashboard/src/main/js/components/PullRequests.jsx
```diff
@@ -70,27 +79,32 @@ export class PullRequests extends Component {
...
-
         if (this.state.unsupportedJob) {
             return (<NotSupported />);
         }
 
+        const { pullRequests } = this.props;
+
         // early out
-        if (!branches) {
+        if (!pullRequests) {
```
michaelneale: can't really follow the changes here as it seems noisy. I guess eventually this will be capabilities based rather than data? (or not, as it may not have any branches but could?)

blueocean-dashboard/src/main/js/components/PullRequests.jsx
```diff
@@ -113,7 +127,13 @@ export class PullRequests extends Component {
                             />);
                         })}
                     </Table>
+                    {pullRequests.$pager &&
```
michaelneale: would it not have a $pager?

blueocean-dashboard/src/main/js/components/RunDetails.jsx
```diff
@@ -4,104 +4,87 @@ import {
...
-        const organizationUrl = buildOrganizationUrl(organization);
+        const organizationUrl = getLocation({
+            organization: this.props.params.organization,
+        });
         this.context.router.push(organizationUrl);
     }
     navigateToPipeline() {
-        const { organization, fullName } = this.props.pipeline;
-        const pipelineUrl = buildPipelineUrl(organization, fullName);
+        const pipelineUrl = getLocation({
```
michaelneale: kind of still wants to be a one liner...

blueocean-dashboard/src/main/js/components/MultiBranch.jsx
tfennelly: Just looks a bit weird when some of the properties have a `$` prefix and others don't. Makes me wonder why.

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
tfennelly: Yeah ... they cry out to be abused and that's what happens.

tfennelly:
Tbh, I tried yesterday to get my head around the substantive changes in this PR, but failed. It just seemed like there were so many changes in there that were unrelated to the stated goal of the PR, making it so hard for me to review. I came to the conclusion that we'd just have to run with these changes and fix issues later if there are any. Do we think we could do that? @scherler @michaelneale ? Not ideal, but better than people getting frustrated with each other I think. Obviously it needs to functionally work and the ATH tests all need to pass.

blueocean-dashboard/src/main/js/components/PullRequests.jsx
kzantow: @michaelneale , this is because pull requests are now fetched separately from branches to be more clear, I suppose

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
kzantow: @michaelneale well, the dashboard would, in theory show you all pipelines you have access to, while the org view would only show you pipelines within the org. The fetches and paging would be separate. Since I had to change this around due to paging, anyway, I figured best to go ahead and build it in a way that makes the most sense. Though, right now with the "jenkins" org, it's very awkward ever seeing the org page at all and there are probably still enough things hardcoded that it will not work (I tried to replace the /jenkins references in the rest calls with org where I found them, but I don't think I got them all), anyway, just one thing we don't have to completely redo when we get to making orgs functional

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
kzantow: @michaelneale not really sure, could be fine. seems like the main dashboard should just be a search or most recent activity, which could make it a different purpose than simply getting all the pipelines for an org

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
kzantow: no, just changing this to use ES6 syntax (within the closure `this` is outer scope, same as the `_this` variable was doing)

blueocean-dashboard/src/main/js/components/PageLoading.jsx
kzantow: yes, total length of _estimated_ time. it still goes away when not needed. rather than showing the unknown time animation, which looks bad

michaelneale:
Still some :bug: 
1 Running a job (like the "passing" branch of the kzantow/failure-project) sometimes yields a finished run but the steps show up as blue
2 If I am following a long a build, and scroll to the top, it yanks me back down to the bottom (that mode seems broken)
3 If I try to run a build with a broken connection (so it can't checkout scm) then I get an error with no flownodes returned (this one may not be related but I will confirm)

(will confirm these rae on this branch only). 

This is the bad state after running:

screen shot 2016-08-26 at 10 32 05 am

i386:
@michaelneale  #2 is occurring on master.

blueocean-dashboard/src/main/js/redux/actions.js
kzantow: This I must have run synchronously, I can't really adapt it to an async pattern, and I don't have an async hook to load up capabilities in this case, no react component, etc... need to revisit this and capabilities as a whole, as we've discussed

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
kzantow: This should also be revisited, but I don't think it warrants a change at the moment

kzantow:
NOTE: at this point, the PR passes ATH locally without issue, there still seems to be an issue running ATH in general (including against the current master) from _ci_.

michaelneale:
I agree with @tfennelly, need to get this in and move one. Hopefully not such long lived and broad PRs come along too oftne, but this LGTM. 

:bee: (if I find any thing, follow on ticket).

i386:
Boom! Congratulations on landing this 🎉

PR #399, opened 2016-08-09
JENKINS-36976# stop API with blocking call feature

vivek:
# Description

See https://issues.jenkins-ci.org/browse/JENKINS-36976

# Submitter checklist
Not sure if it applies to API only changes but anyways, provide link to how to run test harness. 

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
@@ -20,8 +20,10 @@ import io.jenkins.blueocean.service.embedded.rest.AbstractRunImpl;
 import io.jenkins.blueocean.service.embedded.rest.BlueRunFactory;
 import io.jenkins.blueocean.service.embedded.rest.ChangeSetResource;
 import io.jenkins.blueocean.service.embedded.rest.QueueContainerImpl;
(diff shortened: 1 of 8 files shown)


i386:
@cliffmeyers can you please review this one? It will effect tickets assigned to you.

imeredith:
@vivek This PR only covers the blocking feature, it does not have a hard stop as far as I can see?

vivek:
@imeredith yes and no. There is no reliable way to generically provide hard kill for all kind of Runs. Workflow run does better job where it does some cleanup  after executor is taken offline.  Basically as far as I understand there is no way to hard kill that takes executor offline so as part of blocking call we interrupt the running build and wait for executor to go offline and in some cases like workflow run there may be extra cleanup.

cliffmeyers:
REST API looks good to me. Would probably make sense for @imeredith or @michaelneale to weigh in as to whether the server code should get a bee. I don't have enough depth on the Java side to critique it.

michaelneale:
code looks good to me. It is clear that stopping is a "fire and pray" sort of thing, so we can only do what we can only do, but as long as the UX "feels right" (ie we acknowledged the user has requested it). 

There is no "stopping state" for a run unfortunately, so it is a bit fiddly, but @cliffmeyers seems happy with the api, so 🐝

michaelneale:
FYI a change like this should have the ATH run against it...

vivek:
@michaelneale sorry about that, for the next time, add how to run link to the reviewer's template, there isn't any, no idea how its run.

PR #402, opened 2016-08-11
[WiP] Execute query from a specific item root

kzantow:
For discussion.

I ran into the need to add filtering to arbitrary results from the BO API. Unfortunately, the current query API wouldn't be particularly useful/easy to add that sort of thing to. Also, it's not currently composable, from what I can tell. With a few modifications, maybe we could consolidate things:

1. execute a query from an item root rather than always global Jenkins
1. add/change a way to *filter* items rather than only finding them (this doesn't work for all query results - e.g. Users)

@vivek PTAL

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineSearch.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineSearch.java
@@ -1,9 +1,9 @@
 package io.jenkins.blueocean.service.embedded.rest;
 
 import hudson.Extension;
-import hudson.Plugin;
 import hudson.model.Item;
 import hudson.model.ItemGroup;
+import hudson.model.Items;
 import io.jenkins.blueocean.commons.ServiceException;
 import io.jenkins.blueocean.rest.OmniSearch;
 import io.jenkins.blueocean.rest.Query;
@@ -49,7 +49,7 @@ public class PipelineSearch extends OmniSearch<BluePipeline>{
     }
 
     @Override
-    public Pageable<BluePipeline> search(Query q) {
+    public Pageable<BluePipeline> search(Query q, ItemGroup<?> root) {
         String s = q.param(EXCLUDED_FROM_FLATTENING_PARAM);
         String org = q.param(ORGANIZATION_PARAM);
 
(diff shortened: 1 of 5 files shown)


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineSearch.java
```diff
@@ -65,13 +65,7 @@ public String getType() {
...
                 } catch (ClassNotFoundException e) {
                     try {
-                        //TODO: There should be better ways to find a class from a plugin.
-                        Plugin p = Jenkins.getInstance().getPlugin("blueocean-pipeline-api-impl");
-                        if(p != null){
-                            c = p.getWrapper().classLoader.loadClass(s1);
-                        }else{
-                            logger.error("blueocean-pipeline-api-impl plugin not found!");
-                        }
+                        c = Jenkins.getInstance().getPluginManager().uberClassLoader.loadClass(s1);
```
vivek: @kzantow Thanks, didn't know about this oberClassLoader 🐝

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/OmniSearch.java
```diff
@@ -15,9 +21,21 @@
 
     public abstract String getType();
 
-    public abstract Pageable<T> search(Query q);
+    public abstract Pageable<T> search(Query q, ItemGroup<?> root);
```
vivek: @kzantow No reference to hudson.model.* from blueocean-rest 🐛  Maybe some other abstract to define a context under which search needs to happen, maybe a Container?

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/ApiHead.java
```diff
@@ -50,11 +52,7 @@ public ApiHead() {
      */
     @WebMethod(name="search") @GET @PagedResponse
     public Pageable<?> search(@QueryParameter("q") Query query) {
-        for (OmniSearch os : OmniSearch.all()) {
-            if (os.getType().equals(query.type))
-                return os.search(query);
-        }
-        return Pageables.empty();
+        return OmniSearch.query(query, Jenkins.getInstance());
```
vivek: @kzantow Sorry, Same as other comment, we do not want to define Jenkins dependency directly on blueocean-rest module 🐛

PR #403, opened 2016-08-11
Dockerfile to build an image bundling BlueOcean

ydubreuil:
# Description

See [JENKINS-36828](https://issues.jenkins-ci.org/browse/JENKINS-36828).

This PR proposes a different approach to fix JENKINS-36828 than the one in #393. It uses Maven to resolve plugins dependencies instead of `install-plugins.sh` script available in the Jenkins official image.

Maven has a better knowledge of the plugins dependency tree than `install-script.sh` script which doesn't take into account optional dependencies correctly. See #396 for more explanations on the script's limitation.

@ndeloof @michaelneale

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,2 @@
 *
-!blueocean-commons/target/blueocean-commons.hpi
-!blueocean-dashboard/target/blueocean-dashboard.hpi
-!blueocean-plugin/target/blueocean.hpi
-!blueocean-rest/target/blueocean-rest.hpi
-!blueocean-rest-impl/target/blueocean-rest-impl.hpi
-!blueocean-web/target/blueocean-web.hpi
-!docker-demo/
+!download-plugins.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,33 +1,10 @@
-FROM jenkinsci/jenkins:latest
+# Build a Docker image bundling BlueOcean plugins and dependencies
 
-COPY blueocean-commons/target/blueocean-commons.hpi /usr/share/jenkins/ref/plugins/
-COPY blueocean-dashboard/target/blueocean-dashboard.hpi /usr/share/jenkins/ref/plugins/
-COPY blueocean-plugin/target/blueocean.hpi /usr/share/jenkins/ref/plugins/blueocean-plugin.hpi
-COPY blueocean-rest/target/blueocean-rest.hpi /usr/share/jenkins/ref/plugins/
-COPY blueocean-rest-impl/target/blueocean-rest-impl.hpi /usr/share/jenkins/ref/plugins/
-COPY blueocean-web/target/blueocean-web.hpi /usr/share/jenkins/ref/plugins/
+FROM jenkinsci/jenkins:2.17
+ENV BLUEOCEAN_VERSION 1.0-alpha-6
 
 USER root
-
-RUN cd /usr/share/jenkins/ref/plugins/; \
-	install-plugins.sh blueocean-commons \
-                       blueocean-dashboard \
-                       blueocean-plugin \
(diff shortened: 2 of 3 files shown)


ndeloof:
totally hack-ish, but admittedly make it clear a shell script can't compete with maven to handle transitive dependencies and version conflicts :P

ydubreuil:
Note that the image doesn't work because there is some dependency inconsistency in BlueOcean POM in version 1.0-alpha-6

Those inconsistencies are fixed in #396

michaelneale:
@ydubreuil so is there any reason why #396 isn't marked as ready for review? its main impact is that it adds a "dist" module (which may end up confusing) purely for docker, right? 

Should this happen in a separate repo? Given we want to produce docker images for general use (using UC) and ones using the current state of the world (for dogfood) they generally aren't interesting to the day to day developers of blue ocean, right? Does it makes sense to have it in this repo?

michaelneale:
should #393 be closed in favour of this? #393 looks like it should work, as isn't that how other "flavoured" images are built? (do we even want to have a blueocean flavoured image with it pre-installed, when longer term it may end up in recommended plugins anyway?)

A development image IS required of course

ydubreuil:
@michaelneale yes, this supersedes #393. It could be moved into a separate repository, given it's not using anything from this repo. I don't know of any other images bundling plugins like this. The approach taken in #393 unfortunately doesn't work, as `install-plugins.sh` is not able to correctly resolve plugin dependencies. This is why I went to Maven instead.

michaelneale:
@ydubreuil I propose this goes to a different repo, unless you strongly think it belongs here... I am not sure if we want to encourage plugins to have their own special docker flavoured image necessarily (although blue ocean has enough interest it may be justified)

michaelneale:
Or perhaps not another repo, but a subdir of this? Just something to keep it "to the side"?

michaelneale:
@vivek has suggested it is normal for "dist" functionality to live with a plugin, even for docker images, so this could go either way. There is less pressure for this now that blue ocean is in experimental update center, so I am still inclined to keep this hidden away somewhere

michaelneale:
Just closing this for now - re-open when have time to revisit

SalahAdDin:
Please, re open it, i have a problem with this.

michaelneale:
@SalahAdDin ??

There is another PR in progress that does this a different way

SalahAdDin:
I can't build myt proper Dockerfile for this plugin, and the official docker image haven't a public Dockerfile :/

michaelneale:
@SalahAdDin still don't follow - you can use the official jenkins docker image, and add the blueocean plugin as a dependency - https://hub.docker.com/_/jenkins/ like any other plugin 

Official docker image is here: https://github.com/jenkinsci/docker/blob/6eaa9b15926232310317490a3b9975ef61be763c/Dockerfile

in any case: https://github.com/jenkinsci/blueocean-plugin/pull/513 is WIP

SalahAdDin:
Yes, i tried extend the official jenkins image, but, after it i don't know how install this plugin :/

michaelneale:
Please use JIRA, or a maililng list for support, not an old closed ticket. I am not sure how to answer. Blue ocean is a plugin, so you would run the image, and install the plugin into it from the GUI, or it could be via that mechanism described on the official image if you want to "pre bake" it in. Every plugin isn't going to have its own docker image (we did look at one point at publishing a pre-made blueocean image, but it is hard to support at this stage when most use it as a plugin via the normal installation means).

SalahAdDin:
@michaelneale Well, i test this: https://hub.docker.com/r/jenkinsci/blueocean/ 
And i like it, i want make some like it but i don't understand how do it.

PR #404, opened 2016-08-11
stub out Blue Ocean Commons JS

cliffmeyers:
# Description

This is an empty project shell with basic gulpfile and karma config based off the JDL.
cc @imeredith @kzantow 

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
so this is "common" as in a shared library, not the similarly named "commonjs" module/convention system right? ;)

cliffmeyers:
That's right. If you have a name you think would be less confusing, by all means let me know. I modeled the name after the Java module ;)

michaelneale:
@cliffmeyers make sure there are bad security flaws to make it like the apache library ;) (SORRY I just couldn't help...). 

I think name is fine.

michaelneale:
is the expectation that every module that uses js extensions for plugins also uses this?

kzantow:
`blueocean-core` or `blueocean-js-util` maybe?

imeredith:
I really need this for the JWT stuff. At the moment I have an impl and personalization and dashboard.

cliffmeyers:
How about blueocean-core-js ?

kzantow:
`blueocean-core-js` is what I originally had, then removed, but now see it's probably the lest name conflicting. 👍

cliffmeyers:
@michaelneale perhaps we can get this merged to unblock Ivan and look into merging with js-extensions later? Just not sure yet if the two should be coupled.

michaelneale:
I am ok with what you agree on - if renaming to "core" makes sense - why not. 
Having another moudule just means another thing to release, but then again, js-extensions doesn't change very much so its probably fine.

PR #405, opened 2016-08-12
Task/jenkins 36229

vivek:
# Description

See https://issues.jenkins-ci.org/browse/JENKINS-36229.

# Submitter checklist

To test user permission for a pipeline:

* See https://github.com/jenkinsci/blueocean-plugin/tree/task/JENKINS-36229/blueocean-rest
* On local jenkins, enabled security with no anonymous read access
* GET http://localhost:8080/organizations/jenkins/pipelines/p1/ should give permissions of logged in users all true for read, create, start and stop
* go back to global security, allow anonymous read access and call above pipeline api, log out and as anonymous user call above API, the permission object in pipeline details will be true only for read and false for everything else.

Triggered https://ci.blueocean.io/job/Acceptance%20Tests%20Param/11/

# Reviewer checklist
- [x] Reviewed the code


michaelneale:
@i386 @vivek what is expected behavior for this when running an authed jenkins that allows anon access? what sort of manual testing to do we need to do to verify this?

i386:
@michaelneale well thats the config we run it on .io

- When auth is enabled but anon access is off we should redirect to login
- When auth is enabled but anon access is on there is no redirect, sees login button
- When auth is disabled then the user does not get redirected, does not see login button

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/PipelineApiTest.java
```diff
@@ -51,7 +53,7 @@ public void getFolderPipelineTest() throws IOException {
         MockFolder folder = j.createFolder("folder1");
         Project p = folder.createProject(FreeStyleProject.class, "test1");
 
-        Map response = get("/organizations/jenkins/pipelines/folder1/test1");
+        Map response = get("/organizations/jenkins/pipelines/folder1/pipelines/test1");
```
michaelneale: is there a reason why this changed in this PR?

michaelneale:
ATH runs ok, but I have not checked this out locally. @i386 my quesiton is that this is only about the API. If this goes in before corresponding front end changes will the behavior be "acceptable" until front end is changed to accomodate it - or does it all need to be validated together?

i386:
@michaelneale I think its acceptable - theres a slight chance we might need to tweak but the bulk of it is done IMO

michaelneale:
OK, seems ok. 

When I ran in non authed mode with anon access, I saw errors in the GUI console but I think they are legit: 

```
http://localhost:8080/jenkins/blue/rest/organizations/jenkins/user/ Failed to load resource: the server responded with a status of 404 (Not Found)
```

which probably could be expected? 

My one 🐜  is to know why that test case line changed (commented above), but otherwise provisional 🐝  from be.

blueocean-config/src/main/resources/io/jenkins/blueocean/config/BlueOceanConfig/header.jelly
```diff
@@ -0,0 +1,18 @@
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler">
+  <script>
+    (function () {
+      window.$$blueoceanConfig = JSON.parse('${it.blueOceanConfig}');
```
tfennelly: Seems like this plugin is no longer just a dogfood thing, right? If so, then maybe we should (at a min) store this inside the `jenkinsCIGlobal` object, alongside other Jenkins related stuff (where `js-modules` stores stuff).

In reality though, I'd really like to discourage all but `js-modules` from storing things in the global scope (and for it to just store refs to commonjs modules). We could add this in the page markup and have a javascript module for reading it. And to the js-extensions stuff in the same way (kill 2 birds).

tfennelly:
@vivek @michaelneale I added a comment above.

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/PipelineApiTest.java
vivek: @michaelneale Noticed this was not the right URL, it was working by accident as folder API allows such call in limited sort of way. Why do you ask this question?

blueocean-config/src/main/resources/io/jenkins/blueocean/config/BlueOceanConfig/header.jelly
vivek: @tfennelly I don't have context on jenkinsCIGlobal. So how does other plugin config is added to it? Please post pointer to doc where other plugins add data to it.

>We could add this in the page markup and have a javascript module for reading it. And to the js-extensions stuff in the same way (kill 2 birds).

Sorry, not sure what you mean by this. You mean blueocean-config injects a <div> element with some id and it gets picked up by js-extensions (?)?

michaelneale:
@tfennelly can you provide a snippet/example of what you would rather vivek does?

blueocean-config/src/main/resources/io/jenkins/blueocean/config/BlueOceanConfig/header.jelly
tfennelly: > @tfennelly I don't have context on jenkinsCIGlobal. So how does other plugin config is added to it? Please post pointer to doc where other plugins add data to it.

The only example is in `js-modules`. None of the other plugins put anything into the global scope.

> Sorry, not sure what you mean by this. You mean blueocean-config injects a element with some id and it gets picked up by js-extensions (?)?

Right.

tfennelly:
Comment responded to above.

I think, for now, lets just go with what we have here and come back and clean it all up as part of [JENKINS-35851](https://issues.jenkins-ci.org/browse/JENKINS-35851). I'll add a comment to that ticket.

vivek:
@tfennelly Thanks. I have fixed the merge conflicts and its ready to go now.

PR #407, opened 2016-08-16
JENKINS-36376# Adding missing capabilities

vivek:
# Description

See https://issues.jenkins-ci.org/browse/JENKINS-36376.

# Submitter checklist

Call classes API: 

GET http://localhost:8080/jenkins/blue/rest/classes/?q=io.jenkins.blueocean.service.embedded.rest.AbstractPipelineImpl,io.jenkins.blueocean.rest.impl.pipeline.PipelineImpl,io.jenkins.blueocean.rest.impl.pipeline.BranchImpl

- For class io.jenkins.blueocean.service.embedded.rest.AbstractPipelineImpl, it's capabilities should include hudson.model.Job but not io.jenkins.blueocean.rest.model.BlueBranch and org.jenkinsci.plugins.workflow.job.WorkflowJob

- For class io.jenkins.blueocean.rest.impl.pipeline.PipelineImpl, it's capabilities should include hudson.model.Job, org.jenkinsci.plugins.workflow.job.WorkflowJob but not io.jenkins.blueocean.rest.model.BlueBranch.

- For class io.jenkins.blueocean.rest.impl.pipeline.BranchImpl, it's capabilities should include hudson.model.Job, io.jenkins.blueocean.rest.model.BlueBranch and org.jenkinsci.plugins.workflow.job.WorkflowJob

@cliffmeyers  Please review. Note: there is capability to indicate it's a pipeline job (org.jenkinsci.plugins.workflow.job.WorkflowJob), capability to indicate a pipeline is branch (io.jenkins.blueocean.rest.model.BlueBranch). pipeline with neither is some other jenkins job type.

https://ci.blueocean.io/job/Acceptance%20Tests%20Param/15/

# Reviewer checklist
- [x] Reviewed the code

- Refactored pipeline factory to resolve WorkflowJob and moved to pipeline module
- Capability added to branch, pipeline, non-pipeline jobs etc.

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineFolderImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineFolderImpl.java
@@ -97,7 +97,7 @@ public class PipelineFolderImpl extends BluePipelineFolder {
 
     @Override
     public Link getLink() {
-        return OrganizationImpl.INSTANCE.getLink().rel("pipelines").rel(PipelineImpl.getRecursivePathFromFullName(this));
+        return OrganizationImpl.INSTANCE.getLink().rel("pipelines").rel(AbstractPipelineImpl.getRecursivePathFromFullName(this));
     }
 
     @Extension(ordinal = 0)
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/BranchImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/BranchImpl.java
@@ -5,12 +5,12 @@ import hudson.Util;
 import hudson.model.Item;
 import hudson.model.Job;
 import io.jenkins.blueocean.rest.Reachable;
+import io.jenkins.blueocean.rest.annotation.Capability;
 import io.jenkins.blueocean.rest.hal.Link;
 import io.jenkins.blueocean.rest.model.BluePipeline;
 import io.jenkins.blueocean.rest.model.Resource;
 import jenkins.branch.MultiBranchProject;
 import io.jenkins.blueocean.service.embedded.rest.BluePipelineFactory;
-import io.jenkins.blueocean.service.embedded.rest.PipelineImpl;
 import jenkins.scm.api.SCMHead;
 import jenkins.scm.api.actions.ChangeRequestAction;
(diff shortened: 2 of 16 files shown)


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineFolderImpl.java
```diff
@@ -97,7 +97,7 @@ public BlueFavorite favorite(@JsonBody BlueFavoriteAction favoriteAction) {
 
     @Override
     public Link getLink() {
-        return OrganizationImpl.INSTANCE.getLink().rel("pipelines").rel(PipelineImpl.getRecursivePathFromFullName(this));
+        return OrganizationImpl.INSTANCE.getLink().rel("pipelines").rel(AbstractPipelineImpl.getRecursivePathFromFullName(this));
```
michaelneale: I am sure I have asked before, but how can something be both abstract and a concrete impl at the same time?

michaelneale:
Code looks nice, and tests, I'll let @cliffmeyers eyeball this to ensure it seems right functionally 

"AbstractPipelineImpl" still reads funny to me as a name ;) hehehe

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineFolderImpl.java
vivek: @michaelneale Abstract in the class name is different from Java 'abstract' keyword. Abstract in class name means its not specific to particular implementation and serves most generic jobs.

michaelneale: Gotcha. "GenericPipelineImpl" may have also worked? (irrelevant to this PR, but just curious). I find naming things really hard.

vivek: Maybe, I find it hard too, we have precedence on doing it with AbstractRunImpl as well. I have seen such usage elsewhere, such names are really opinion oriented, no hard and fast. For the same reason, you could have an 'abstract' class without Abstract name in it:)

cliffmeyers:
@vivek so should I expect that a Freestyle job would be returned as "AbstractPipelineImpl" which when fetched via Capabilities API would yield only the "Job" capability?

PR #410, opened 2016-08-16
[JENKINS-36171] Steps with no logs should not be expandable

scherler:
# Description

Steps with no logs should not be expandable

See [JENKINS-36171](https://issues.jenkins-ci.org/browse/JENKINS-36171) especially the comment about steps that can have logs but do not have them. The PR covers those that do not provide the possibility to have logs (no more 404 for those logs urls)

TESTS: https://github.com/jenkinsci/blueocean-acceptance-test/pull/24

screenshot from 2016-08-16 17-39-18

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

…not do not make the resultItem expandable.

### DIFF
--- a/blueocean-dashboard/src/main/js/util/logDisplayHelper.js
+++ b/blueocean-dashboard/src/main/js/util/logDisplayHelper.js
@@ -31,6 +31,9 @@ export const getNodesInformation = (nodes) => {
         const isFailingNode = errorNodes.indexOf(item.id) > -1;
         const isRunningNode = runningNodes.indexOf(item.id) > -1;
 
+        // FIXME: TS I need to talk to cliffMeyers how we can refactor the following code to use capabilities
+        // the problem I see ATM is that we would need to ask the c-API everytime for each action, whether this
+        // action has the capability for logging
         const hasLogs = item.actions ? item.actions
                 .filter(action => action._class === 'org.jenkinsci.plugins.workflow.support.actions.LogActionImpl').length > 0
             : false;
--- a/blueocean-dashboard/src/main/js/util/logDisplayHelper.js
+++ b/blueocean-dashboard/src/main/js/util/logDisplayHelper.js
@@ -31,6 +31,9 @@ export const getNodesInformation = (nodes) => {
         const isFailingNode = errorNodes.indexOf(item.id) > -1;
         const isRunningNode = runningNodes.indexOf(item.id) > -1;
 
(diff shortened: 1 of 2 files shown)


michaelneale:
Running ATH here now: https://ci.blueocean.io/job/Acceptance%20Tests%20Param/18/

i386:
🐛 Not sure if introduced in this change or not but it looks like the alignment is slightly off on the caret and the text. 

This is what it should look like:
jenkins blue ocean 2016-08-17 10-58-12

blueocean-dashboard/src/main/js/util/logDisplayHelper.js
```diff
@@ -31,6 +31,9 @@ export const getNodesInformation = (nodes) => {
         const isFailingNode = errorNodes.indexOf(item.id) > -1;
         const isRunningNode = runningNodes.indexOf(item.id) > -1;
 
+        const hasLogs = item.actions ? item.actions
+                .filter(action => action._class === 'org.jenkinsci.plugins.workflow.support.actions.LogActionImpl').length > 0
```
michaelneale: Should this be using capabilities cc @vivek @cliffmeyers ? (if so, can you suggest what ones to use?). 
In theory we should be looking up the capabilities list from the _class vs doing === on it directly (assuming there is a log capability).

michaelneale:
Works as advertised. I can't see the issue @i386 is pointing out when I ran it locally. 
Provisional 🐝  from me - however the one :ant: is that capabilities should be looked up (I mentioned in the code) but would need an example snippet maybe from @cliffmeyers to show how to do that.

blueocean-dashboard/src/main/js/util/logDisplayHelper.js
scherler: I agree with you @michaelneale I said that to @tfennelly yesterday. I followed https://issues.jenkins-ci.org/browse/JENKINS-36170 but I will search for an example of the capabilities.

michaelneale: If the failure mode was to show up all log expanders, then it would be less of an issue...

scherler:
@michaelneale as announced in IRC I added:

```
        // FIXME: TS I need to talk to cliffMeyers how we can refactor the following code to use capabilities
        // the problem I see ATM is that we would need to ask the c-API everytime for each action, whether this
        // action has the capability for logging
```

PR #413, opened 2016-08-17
[JENKINS-36642] Implement alternative view, so you can always use the…

scherler:
# Description

I create a small video to show what I came up with https://youtu.be/kUVlkZncbKE 

You can see that I created now a hook with which you can see in the pipelineDetails the logconsole as you would do in freestyle projects. ATM it is triggered by adding view=0 as query parameter.

See [JENKINS-36642](https://issues.jenkins-ci.org/browse/JENKINS-36642).

Please @i386 review whether that is what you are looking for. If so I will create AT tests for it. 

# Submitter checklist
- [x] Ran Acceptance Test Harness against PR changes. -> https://ci.blueocean.io/job/Acceptance%20Tests%20Param/32/console

# Reviewer checklist
- [x] Reviewed the code

… logConsole view as in freestyle, even if you are looking at steps


michaelneale:
YES, I like this. This seems a good "escape to log" hatch that I find I am missing from time to time.

Worth continuing IMO

i386:
🐛 While I like the mode switch (we can do something with that a bit later 🏆 ) this ticket was to solve the problem where a pipeline has been put in the queue then stopped before it started executing. At the moment, the screen would be empty - no pipeline graph and no steps. What we would like to see done for this ticket is all the available logs in a freestyle like log component.

i386:
@scherler the use case you have in this PR - Pipeline is running or failed, nodes in the graph have not built yet - is covered by [JENKINS-36105](https://issues.jenkins-ci.org/browse/JENKINS-36105). For this we just want to display "There are no steps"

i386:
@scherler perhaps repurpose this PR so that it retains the param to jump between the full log and logs broken up step-by-step (but remove the changes to the empty state). Ill come up with a design that allows you to switch between the two more gracefully (You've given me a good idea on how to do this)

scherler:
@i386 the video shows a job that has been cancelled in the queue (never got any build executors). 

To simulate this I configured my local jenkins to have 0 build runners. Then started a job and stopped it again after a bit.

That already shows the No steps message, the only thing I added were the button to see the full log (switch to console view). IMO  JENKINS-36105 is a duplicate from another ticket I did before my vacation which introduced the box with the no steps message. 

Meaning what you write in your first comment should not be true anymore "shows empty screen". Did you see it recently?

I remove the "hook" link in the box and leave the functionality as you said, glad you like it. Can you give a bee then?

scherler:
@i386 btw https://issues.jenkins-ci.org/browse/JENKINS-36642?focusedCommentId=267091&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-267091 contains the state you desire, correct?

scherler:
@i386 for me that is ready to merge, so ping for the BBBBBEEEEEE ;)

michaelneale:
Running ATH: https://ci.blueocean.io/job/Acceptance%20Tests%20Param/22/

i386:
@scherler lets hit the reset button here.

> To simulate this I configured my local jenkins to have 0 build runners. Then started a job and stopped it again after a bit.

Right, exactly this scenario :) On your branch this is what gets displayed:

jenkins blue ocean 2016-08-19 16-29-32

What we want to show is the full available log below:
state

However, the blue bar should still show in the scenario where the pipeline actually started and for whatever reason the steps did not run.

scherler:
https://youtu.be/IZ9giy6QysU I think it is like you expect it @i386

scherler:
Just did validated: 
screenshot from 2016-08-19 12-54-22

scherler:
@i386 I need to use this as base for https://issues.jenkins-ci.org/browse/JENKINS-36133

michaelneale:
@scherler what is status of this and ATH - should it be able to run with it now?

i386:
@scherler close. We only want to display the log when the run is stopped and there is no stage info. This does not apply to cases where it is queued or running.

michaelneale:
@i386 when I set executors to zero I see this: 

screen shot 2016-08-22 at 8 03 46 pm

So there is no queued state that I can see. 

The API returns (I have hidden some): 

```
{
"_class": "io.jenkins.blueocean.rest.impl.pipeline.PipelineRunImpl",
"_links": {},
"actions": [],
"artifacts": [],
"changeSet": [],
"durationInMillis": 0,
"enQueueTime": "2016-08-22T20:06:30.545+1000",
"endTime": null,
"estimatedDurationInMillis": 36596,
"id": "13",
"organization": "jenkins",
"pipeline": "pipe",
"result": "UNKNOWN",
"runSummary": "?",
"startTime": "2016-08-22T20:06:30.547+1000",
"state": "RUNNING",
"type": "WorkflowRun",
"commitId": null
}
```

This shows it as RUNNING but UNKNOWN - so I am not sure what else @scherler can do. The API should probably be returning Queued, but this may be a result of unifying things previously so it shows up in the one place.

scherler:
screenshot from 2016-08-22 12-27-15

If you have a "known" pipeline, you will see above screenshot. As @michaelneale pointed out, the result is UNKOWN and the state is running. The "nodes/" for this will return each node with result=null and state=null

scherler:
I found the change which introduced the desired box https://github.com/jenkinsci/blueocean-plugin/pull/389 however AFAIS that should never had worked. I recall that 4-5 month ago I pointed out in some of the first tickets. that the backend never return state === 'QUEUED'

BTW the code is still in there and I found a bug 

```
componentDidMount() {
        const { result } = this.props;
        if (!result.isQueued()) {
            // determine scroll area
            const domNode = ReactDOM.findDOMNode(this.refs.scrollArea);
            // add both listemer, one to the scroll area and another to the whole document
            domNode.addEventListener('wheel', this.onScrollHandler, false);
            document.addEventListener('keydown', this._handleKeys, false);
        }
    }
```

In case we look at the details in queued state we will not be able to stop karaoke mode anymore since the didMount only get executed once and we do not add the eventListener later on again, in case the run had been queued when we first looked at it.

Anyway that said the problem is the backend that never returns queued state, if it would you would see the the following image, however I "faked" it now by always returning this.

screenshot from 2016-08-22 15-12-59

scherler:
should I merge or should I go? https://www.youtube.com/watch?v=oGIFublvDes

scherler:
One observation that I made: I created anther job trying to play around with our app and it seems that a second job will show up in state === RUNNING "however" the question is why the first time we do not get the correct information

michaelneale:
@scherler unfortunately needs updating to master again.. but that behavior may be expected. The crtical point for this was the commit you made to only show if the fun is finished...

i386:
@scherler @michaelneale from my testing this looks correct now.

michaelneale:
ok great - once @scherler updates this to master, and checks it is ok with ATH - then good to go

PR #416, opened 2016-08-18
[JENKINS-37006] Improved module resolution/loading/sharing across browser bundles

tfennelly:
# Description

See [JENKINS-37006](https://issues.jenkins-ci.org/browse/JENKINS-37006).

Depends on these PRs (most of the changes are there):

* https://github.com/jenkinsci/js-modules/pull/2
* https://github.com/jenkinsci/js-builder/pull/5

`js-modules` support for version range/wildcard module `import` + support for `export` of internal deps (e.g. exporting `react` and `react-dom` from the main `blueocean-web` bundle).

So, the explicit export code in the `init.jsx` in `blueocean-web` is gone now. Exports (and imports) are now done in the `package.json` i.e.

```javascript
  "jenkinscd": {
    "import": [
      "history",
      "immutable",
      "keymirror",
      "react-redux",
      "react-router",
      "redux",
      "redux-thunk"
    ],
    "export": [
      "@jenkins-cd/js-extensions",
      "@jenkins-cd/design-language",
      "react",
      "react-dom"
    ]
  }
```

> Of course, for more fine-grained control,  you can also call `import` and `export` on the `bundle` instance in the `gulpfile.js`.

If the bundle can use any version of a package, just specify the package import with a version number of "any" on it (using the same format as when running `npm install` on the command line) e.g. `immutable@any`. This also tells `js-modules` not to load the bundle if it's not loaded i.e. to wait for something else to load the bundle i.e. it's "provided". We use this for importing `react` and `react-dom` into all other bundles generated outside `blueocean-web` e.g. in dashboard.

It also supports imports of version ranges e.g. `immutable@3.8.x` or `immutable@3.x`, where `x` can be any number. OR, you can specify an ordered list of compatible versions e.g. `immutable@3.8.x,3.8.1`. What the later translates to in `js-modules` is ... "if another bundle has already loaded any 3.8.* version of `immutable` that can be used (no need to load anything), otherwise load 3.8.1 and use it".

# Submitter checklist
- [x] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below. N/A
- [x] Test @kzantow branch ... https://github.com/kzantow/blueocean-plugin/tree/victory-failure and make necessary tweaks to make sure it works without the dev jumping through a load of hoops. __Verified__: Worked without any tweaks. [Branch here](https://github.com/tfennelly/blueocean-plugin/tree/JENKINS-37006-with-kzantow-victory). [Screenshot here](https://www.dropbox.com/s/jjfdgvy9h6ybj2f/Screenshot%202016-08-18%2015.40.16.png?dl=0). 

# Reviewer checklist
- [x] Reviewed the code

### DIFF
--- a/blueocean-web/gulpfile.js
+++ b/blueocean-web/gulpfile.js
@@ -41,7 +41,6 @@ builder.bundle('src/main/js/blueocean.js')
     .export('react')
     .export('react-dom')
     .export('redux')
-    .export('react-addons-css-transition-group') // Have to export this because it dips down into the react package internals grrr
     .generateNoImportsBundle();
 
 //
--- a/blueocean-web/gulpfile.js
+++ b/blueocean-web/gulpfile.js
@@ -1,6 +1,9 @@
 //
 // See https://github.com/jenkinsci/js-builder
 //
+
+process.env.SKIP_BLUE_IMPORTS = 'YES';
+
 var gi = require('giti');
 var fs = require('fs');
 var builder = require('@jenkins-cd/js-builder');
@@ -43,7 +46,7 @@ builder.bundle('src/main/js/blueocean.js')
 //
 builder.bundle('src/main/js/try.js')
     .inDir('target/classes/io/jenkins/blueocean')
-    .withExternalModuleMapping('jquery-detached', 'core-assets/jquery-detached:jquery2') // Bundled in Jenkins 2.x 
+    .import('jquery-detached', 'core-assets/jquery-detached:jquery2') // Bundled in Jenkins 2.x
     .less('src/main/less/try.less');
 
 // 
--- a/blueocean-web/gulpfile.js
+++ b/blueocean-web/gulpfile.js
@@ -36,6 +36,12 @@ builder.src(['src/main/js', 'src/main/less', 'node_modules/@jenkins-cd/design-la
 builder.bundle('src/main/js/blueocean.js')
     .inDir('target/classes/io/jenkins/blueocean')
     .less('src/main/less/blueocean.less')
+    .export("@jenkins-cd/js-extensions")
+    .export("@jenkins-cd/design-language")
+    .export('react')
+    .export('react-dom')
+    .export('redux')
+    .export('react-addons-css-transition-group') // Have to export this because it dips down into the react package internals grrr
     .generateNoImportsBundle();
 
 //
--- a/blueocean-web/package.json
+++ b/blueocean-web/package.json
@@ -11,8 +11,8 @@
     "bundle:watch": "gulp bundle:watch"
   },
   "devDependencies": {
-    "@jenkins-cd/js-builder": "0.0.36-beta22",
-    "@jenkins-cd/js-test": "1.1.1",
+    "@jenkins-cd/js-builder": "0.0.36-beta24",
+    "@jenkins-cd/js-test": "1.2.1",
     "babel-eslint": "^6.0.2",
     "babel-preset-es2015": "^6.6.0",
     "babel-preset-react": "^6.5.0",
@@ -26,8 +26,8 @@
   },
   "dependencies": {
     "@jenkins-cd/design-language": "0.0.68",
-    "@jenkins-cd/js-extensions": "0.0.21-beta12",
-    "@jenkins-cd/js-modules": "0.0.6-beta9",
+    "@jenkins-cd/js-extensions": "0.0.21-beta13",
+    "@jenkins-cd/js-modules": "0.0.6-beta10",
     "history": "2.0.2",
     "immutable": "3.8.1",
     "keymirror": "0.1.1",
@@ -38,24 +38,6 @@
     "react-redux": "4.4.5",
     "react-router": "2.3.0",
     "redux": "3.5.2",
-    "redux-thunk": "2.0.1",
-    "window-handle": "1.0.0"
-  },
-  "jenkinscd": {
-    "import": [
-      "history",
-      "immutable",
-      "keymirror",
-      "react-redux",
-      "react-router",
-      "redux",
-      "redux-thunk"
-    ],
-    "export": [
-      "@jenkins-cd/js-extensions",
-      "@jenkins-cd/design-language",
-      "react",
-      "react-dom"
-    ]
+    "redux-thunk": "2.0.1"
   }
 }
--- a/blueocean-web/package.json
+++ b/blueocean-web/package.json
@@ -11,7 +11,7 @@
     "bundle:watch": "gulp bundle:watch"
   },
   "devDependencies": {
-    "@jenkins-cd/js-builder": "0.0.36-beta8",
+    "@jenkins-cd/js-builder": "0.0.36-beta21",
     "@jenkins-cd/js-test": "1.1.1",
     "babel-eslint": "^6.0.2",
     "babel-preset-es2015": "^6.6.0",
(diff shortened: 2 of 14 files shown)


tfennelly:
__Note__ build failure is due to `checkdeps` checking `sse-gateway`, which is shouldn't be doing.

js-extensions/spec/ExtensionStore-spec.js
```diff
@@ -129,50 +129,50 @@ describe("ExtensionStore.js", function () {
...
+    //        expect(extensions[0]).to.equal('typed-component-1.1');
+    //    });
+    //
+    //    extensionStore.getExtensions('ept-2', [ClassMetadataStore.dataType('type-2')], function(extensions) {
+    //        expect(extensions.length).to.equal(1);
+    //        expect(extensions).to.include.members(["typed-component-1.2"]);
+    //
+    //        done();
+    //    });
+    //});
```
tfennelly: Idisable these tests because they're failing (on master too).

tfennelly: I created a JIRA for this ... https://issues.jenkins-ci.org/browse/JENKINS-37516

blueocean-web/gulpfile.js
```diff
@@ -43,7 +46,7 @@ builder.bundle('src/main/js/blueocean.js')
 //
 builder.bundle('src/main/js/try.js')
     .inDir('target/classes/io/jenkins/blueocean')
-    .withExternalModuleMapping('jquery-detached', 'core-assets/jquery-detached:jquery2') // Bundled in Jenkins 2.x 
+    .import('jquery-detached', 'core-assets/jquery-detached:jquery2') // Bundled in Jenkins 2.x
```
michaelneale: didn't we get rid of the one place we used jquery? I forget.

js-extensions/spec/ExtensionStore-spec.js
michaelneale: I presume we aren'r running things as part of the main ci pipeline?

sophistifunk:
I'm just wondering why -web has a bunch of things in the import list? I figured being the "top-most" in the JS hierarchy it would be the one doing all the exporting but not the importing?

michaelneale:
Wow, big drop tom! nice!



However, I see errors in the console: 

```
jenkins-js-extension.js:21258 Error: Minified exception occurred; use the non-minified dev environment for the full error message and additional helpful warnings.(…)(anonymous function) @ jenkins-js-extension.js:21258
immutable-3-8-1.js:1800 Uncaught (in promise) TypeError: Expected Array or iterable object of values: Error: Minified exception occurred; use the non-minified dev environment for the full error message and additional helpful warnings.(…)
```

that appeared when I try this branch. I have just 2 jobs (one pipeline, one freestyle in a folder).

tfennelly:
@michaelneale what exactly did you do to produce that error and where do you see it? In the browser? Did you do -DcleanNode?

blueocean-web/gulpfile.js
tfennelly: `try.js` is used in classic Jenkins and doesn't use `react`. It's the blue button on the top bar.

blueocean-web/package.json
```diff
@@ -42,15 +42,20 @@
...
+    "import": [
       "history",
       "immutable",
-      "react-router",
       "keymirror",
       "react-redux",
       "react-router",
       "redux",
       "redux-thunk"
+    ],
```
tfennelly: @sophistifunk wrt your comment on these imports, yes, we could remove these and just bundle all of this stuff inside the main/bootstrap bundle. I did it this way originally thinking it might be better doing multiple concurrent async loads of smaller bundles Vs one load of a monster bundle. A bit unscientific so it might make sense to remove these imports and see if a monster bootstrap bundle works better. However, not planning to do that in this PR.

js-extensions/spec/ExtensionStore-spec.js
tfennelly: @michaelneale right ... not run as part of main CI job. I changed the `package.json` to run the tests as part of the prepublish script i.e. prior to that it was the job of the person making changes and doing the publish to try remember to run the tests before publishing.

michaelneale: ok, well good catch!

blueocean-web/gulpfile.js
michaelneale: 👍🏻

michaelneale:
@tfennelly yes, I did a -DcleanNode (mvn clean -DcleanNode install -DskipTests is what I run most times) - I will try it again if you aren't seeing it. Just saw it on dashboard - first page.

blueocean-web/package.json
michaelneale: that sounds like a worthy experiment, maybe note it for later and can do some A/B with just those changes in future?

michaelneale:
as discussed, errors are due to personalization (if you have a one or more favourites it will show up)

tfennelly:
To clarify @michaelneale last statement ... the error is def due to the changes in this PR, but appears when using the personalization stuff. Just say'n in case @cliffmeyers thinks it's something he needs to spend time looking at.

I'm looking into it now ... seems like one of those issues where I might spend a day looking for a 1 word fix grrrrr

blueocean-web/package.json
tfennelly: Actually I was thinking about it more after I wrote the comment above ... other bundles are using most of these packages too, so I think it actually does make sense to externalize them from all bundles and `import` them everywhere (just as we're doing atm).

michaelneale:
@tfennelly I guess it was one of those more than 1 day things :/

tfennelly:
@michaelneale eventually found the culprit. What a pain in the *****x.

`react-addons-css-transition-group` dips down into the internals of react (and so drags it in through the back door ... grrr), which means we have to do an explicit export from the main bootstrap bundle (in `blueocean-web`) and then an explicit `import` into personalization.

michaelneale:
this will need attention of @cliffmeyers and @scherler (to try and also to review).

js-extensions/spec/ResourceLoadTracker-spec.js
```diff
@@ -40,8 +40,8 @@ describe("ResourceLoadTracker.js", function () {
             ResourceLoadTracker.onMount('ep-1');
             cssElements = document.getElementsByTagName('link');
             expect(cssElements.length).to.equal(2);
-            expect(cssElements[0].getAttribute('href')).to.equal('/adjuncts/908d75c1/org/jenkins/ui/jsmodules/plugin-1/extensions.css');
-            expect(cssElements[1].getAttribute('href')).to.equal('/adjuncts/908d75c1/org/jenkins/ui/jsmodules/plugin-2/extensions.css');
+            expect(cssElements[0].getAttribute('href')).to.equal('/jenkins/adjuncts/908d75c1/org/jenkins/ui/jsmodules/plugin-1/extensions.css');
```
michaelneale: how come this path is appended in the test cases now? (curious)

blueocean-web/gulpfile.js
```diff
@@ -36,6 +36,12 @@ builder.src(['src/main/js', 'src/main/less', 'node_modules/@jenkins-cd/design-la
 builder.bundle('src/main/js/blueocean.js')
     .inDir('target/classes/io/jenkins/blueocean')
     .less('src/main/less/blueocean.less')
+    .export("@jenkins-cd/js-extensions")
+    .export("@jenkins-cd/design-language")
+    .export('react')
+    .export('react-dom')
+    .export('redux')
+    .export('react-addons-css-transition-group') // Have to export this because it dips down into the react package internals grrr
```
michaelneale: Nice catch - this has bitten before. I think you were right, it was this culprit in other cases.

js-extensions/spec/ResourceLoadTracker-spec.js
tfennelly: Coz of an update to the `js-test` module. It's good.

tfennelly:
Okay, I've closed off the loophole that was allowing `react-addons-css-transition-group` to require react through the back door. No longer a need to explicitly export/import this module between web and personalization.

scherler:
:bee: LGTM

michaelneale:
trying ATH here: https://ci.blueocean.io/job/Acceptance%20Tests%20Param/40/ just checking a few other things.

PR #427, opened 2016-08-23
Feature/jenkins 36199 toaster

cliffmeyers:
# Description
- See [JENKINS-36199](https://issues.jenkins-ci.org/browse/JENKINS-36199).
- This preserves the existing "toast" functionality when running a pipeline but provides a new simpler API to hook into it. There are some minor changes to how multiple toasts are handled. They now stack up vertically from bottom to top; before this change triggering multiple toasts would actually sit one top of one another.

# Submitter checklist

# Reviewer checklist
 - Click on a simple Pipeline or Freestyle job from the main list and click the "Run" button, verify that toasts display for "queued" and "running"; verify the open link opens Run Details correctly.
 - Click on a Multibranch Pipeline from the main list and navigate to "Branches" tab; click the "Run" button alongside one or more branches, verify toasts display and Run Details opens. Bonus points: try this with gitflow branch names, and with projects nested in a folder.
- [x] Reviewed the code


cliffmeyers:
cc @tfennelly since I made some changes to the `RunPipeline` code you worked on.

i386:
@cliffmeyers nitpick: why are there no toasts when running from the dashboard?

cliffmeyers:
@i386 just an oversight really! I can add them in this PR or in the one for stopping a build from Dashboard which will go up tomorrow.

i386:
🐛 something seems a bit off in this branch - perhaps some CSS wasn't merged correctly? 

jenkins blue ocean 2016-08-24 09-18-59

i386:
@cliffmeyers I think we should put toasts everywhere except if a run is triggered from the modal

i386:
@cliffmeyers any reason the queued toast goes away quicker than the other toasts?

aug -24-2016 09-20-54

michaelneale:
@i386 if it is a pipeline thing won't actually be in the queued for more than a few milliseconds even if there are parts waiting on an executor (may be related)

michaelneale:
Yes main 🐛  that I see is that the styles seem wrong with favourites. 

The queue popping up quickly looks odd (queue makes very little sense for pipeline, as it doesn't sit in queued generally for a  human scale of time) too, but not sure what to do about that.

cliffmeyers:
@i386 re: margins, rats, that slipped in! I saw you'd made a change eliminating top margin in the JDL and it got pulled in when I created Toaster. Here's the commit: https://github.com/jenkinsci/jenkins-design-language/commit/a57c972be5f0cc1856620e5b0961961e4aae2229   should we use 1 rem instead?

cliffmeyers:
@i386 @michaelneale yeah, in retrospect the queue thing does look odd. I was programmatically removing it after the run came back so there would not be two Toasts for the same job. I could just leave them both up for the regular duration if that works.

michaelneale:
@cliffmeyers worth a try (both spacing and the queue item) - would be odd looking to leave it there I suspect... but it would fade anyway... and means we aren't special casing pipeline job types

i386:
@cliffmeyers @michaelneale actually don't worry about the Queued toast in this PR. In https://issues.jenkins-ci.org/browse/JENKINS-37425 we will kill it so there is one "started" toast only (this is in your queue Cliff)

i386:
@cliffmeyers ahh that's my fault then. I was trying to reduce the white space on the top of the dashboard. Could you handle it the way you think it should be?

michaelneale:
@cliffmeyers code looks good BTW, kudos on test cases. 
As part of it, could the displayed name in the toast be changed to be either the full name or display name, or *something* or *anything* that isn't the encoded path (double encoded, technically)? as it looks bad with git flow like names (maybe that is a separate ticket that @i386 will open) but thought I would suggest here... given context, it is probably ok just to show the job name in most cases.

cliffmeyers:
@michaelneale sure I can make that change along with the others.

blueocean-core-js/test/js/ToastService-spec.js
```diff
@@ -0,0 +1,61 @@
...
+        assert.equal(toastService.count, 1);
+
+        toastService.removeToast({
+            id: 12345,
+        });
+
+        assert.equal(toastService.count, 0);
+    });
+
+});
```
tfennelly: Would it make sense to add a test where there are multiple toasts?

michaelneale: fancy toaster - 2 slices at once?

blueocean-dashboard/src/main/js/api/Branch.js
```diff
@@ -29,4 +29,18 @@ export default class Branch extends Pipeline {
...
+     * @param branchName2
+     * @returns {boolean}
+     * @private
+     */
+    _compareBranchNames(branchName1, branchName2) {
+        return branchName1 === branchName2 ||
+            encodeURIComponent(branchName1) === branchName2 ||
+            branchName1 === encodeURIComponent(branchName2);
+    }
 }
```
tfennelly: Seems like the `Branch` and `Pipeline` classes should move to the new `blueocean-core-js` package?

blueocean-dashboard/src/main/js/api/Pipeline.js
```diff
@@ -36,10 +36,17 @@ export default class Pipeline {
         if (runId === undefined) {
             throw new Error('Branch.runDetailsRouteUrl must be supplied with a "runId" parameter.');
         }
+
+        // clean the branch name before passing it through to URL builder
+        let branchName = this.branchName;
+        while (branchName.indexOf('%2') >= 0) {
```
tfennelly: Is that a safe assumption to make?

tfennelly: NIT ... I've seen others do it too ... all docs for `string.indexOf` that've read state something like "_or -1 if it never occurs_", yet people code that into conditions as `>= 0` (as opposed to `!== -1`) 😸

blueocean-web/.babelrc
```diff
@@ -0,0 +1,8 @@
+{
+    "presets": [
+        "es2015", "react", "stage-0"
+    ],
+    "plugins": [
+        "transform-decorators-legacy"
+    ]
+}
```
tfennelly: My imagination, or did I see another `.babelrc` that looked the very same as this? Can we put this in the top level folder and not duplicate?

michaelneale: these are all plugins right? so potentially stand alone?

blueocean-core-js/test/js/ToastService-spec.js
tfennelly: @michaelneale can it take thick slices ... it's very annoying when the bread sticks in the toaster !!

tfennelly: Or gets burned if the "delay" is too long

michaelneale: well cliff did mention he wanted to have a component called "toast" so the metaphors will continue!

tfennelly:
LGTM ... added a few comments ... safety of the assumption wrt checking if a branch name was encoded or not ... I might be wrong but I'd have a feeling you'd need to do more than just check if it contains `%2`.

Also .. since you're making changes to `Branch` and `Pipeline`, can we move them to the new commons package too?

blueocean-core-js/src/js/ToastService.js
```diff
@@ -0,0 +1,56 @@
...
+ * Created by cmeyers on 8/18/16.
+ */
+import { observable, computed } from 'mobx';
+
+/**
+ * Holds one or more toasts in state for display in UI.
+ */
+export class ToastService {
+
+    @observable toasts = [];
```
scherler:
```
 @observable toasts = Immutable.Map([])
```

blueocean-core-js/src/js/ToastService.js
```diff
@@ -0,0 +1,56 @@
...
+     *  id: unique identifier (optional, will be autogenerated if ommitted)
+     * }
+     * @returns {number} unique ID of toast
+     */
+    newToast(toast) {
+        if (!toast.id) {
+            toast.id = Math.random() * Math.pow(10, 16);
+        }
+
+        // TODO: determine why it's necessary to re-set the "toasts" field to trigger the UI update
```
scherler: With immutable that should not happen, I reckon react detects that you are passing around the same object and does not update UI for that reason. 

```
this.toasts = this.toasts.set(toast.id, toast)
```

blueocean-core-js/src/js/ToastService.js
```diff
@@ -0,0 +1,56 @@
...
+        return toast.id;
+    }
+
+    /**
+     * Removes a toast with the matching value of toast.id.
+     *
+     * @param toast
+     */
+    removeToast(toast) {
+        this.toasts = this.toasts.filter((item) => {
```
scherler:
```
this.toasts = this.toasts.delete(toast.id)
```
https://facebook.github.io/immutable-js/docs/#/Map/delete

blueocean-core-js/src/js/ToastService.js
```diff
@@ -0,0 +1,56 @@
...
+            return toast.id !== item.id;
+        });
+    }
+
+    @computed
+    get count() {
+        return this.toasts ? this.toasts.length : 0;
```
scherler:
```
return this.toasts.size;
```

blueocean-core-js/src/js/ToastService.js
```diff
@@ -0,0 +1,56 @@
...
+     *  onActionClick: function, callback to invoke when action link is clicked
+     *  onDismiss: function, callback to invoke when toast is dismissed (immediately, or after timeout)
+     *  dismissDelay: number, duration in millis after which to auto-dismiss this Toast
+     *  id: unique identifier (optional, will be autogenerated if ommitted)
+     * }
+     * @returns {number} unique ID of toast
+     */
+    newToast(toast) {
+        if (!toast.id) {
+            toast.id = Math.random() * Math.pow(10, 16);
```
scherler:
```
toast.id = this.toasts.max().id + 1;
```

blueocean-web/src/main/js/init.jsx
```diff
@@ -43,6 +43,11 @@ exports.initialize = function (oncomplete) {
     const Extensions = require('@jenkins-cd/js-extensions');
     jenkinsMods.export('jenkins-cd', 'js-extensions', Extensions);
 
+    // Create and export a shared instance of the core
+    // js module
+    const corejs = require('@jenkins-cd/blueocean-core-js');
+    jenkinsMods.export('jenkins-cd', 'blueocean-core-js', corejs);
+
```
tfennelly: Done differently now. Web bundle exports now done [on the `bundle` instance in the gulpfile](https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-web/gulpfile.js#L39-L43).

So, add `export("@jenkins-cd/blueocean-core-js")` there and `import("@jenkins-cd/blueocean-core-js@any")` wherever it's used. We could also [add a "default" `import` in js-extensions](https://github.com/jenkinsci/blueocean-plugin/blob/master/js-extensions/%40jenkins-cd/js-builder.js#L47-L54).

blueocean-core-js/test/js/ToastService-spec.js
cliffmeyers: Fixed in cab6402. Glad you guys are enjoying the name ;)

blueocean-dashboard/src/main/js/api/Branch.js
cliffmeyers: I agree we should move them as soon as they are used in two different plugins or we are supremely confident that they will be widely used by developers. We need to do an extra "gulp bundle:watch + slink, npm publish" dance for any code in there, and it can be a bit of productivity hit.

blueocean-web/.babelrc
cliffmeyers: I think in practice there won't be many projects that use our Maven multi-module approach to organization. Beyond that, please see points 1 and 2 from this comment here:  https://github.com/jenkinsci/js-builder/pull/4#issuecomment-239001170

I think trying to implement our own behavior that's close to how babel works OOTB but slightly different is a bad path to go down, and I'd hate to see us have to add another package.json in this project that needs to be managed.

blueocean-web/src/main/js/init.jsx
cliffmeyers: Fixed in 61f0b7c.  @tfennelly if you could double-check the changes there I would appreciate it. It did restore the ToastService to proper working order.

One more question too: when building the new `js-extensions` I noticed that the test "js-builder plugin test -> test yamlToJSON" is failing. I tried adjusting the destination path to "../target..." but I still got an error when requiring the file, even though it was there. Have you seen this, perhaps know the fix? It's broken on master too.

cliffmeyers:
@scherler re: Immutable, I'd like to experiment with MobX a bit more to understand whether there is a need for immutable data structures. If we could get by without needing them, that'd be a win, as long as it doesn't result in stale views or poor UI performance. Since this is used in exactly one field I am hoping we can let it slide for now?

blueocean-dashboard/src/main/js/api/Pipeline.js
cliffmeyers: Used what is hopefully safer logic in bd743c0

cliffmeyers:
screen shot 2016-08-24 at 9 39 42 am  Revised spacing around favorites cards.

cliffmeyers:
@michaelneale ran the ATH locally and all 120 assertions passed.

cliffmeyers:
@michaelneale with your blessing and bee I can merge this in later tonight.

scherler:
@cliffmeyers above code with immutable is MUCH cleaner and much faster in performance. You can use it perfectly in mbox AFAIK. 

As you have experience the need of cloning and a handling of state, that has to be done with immutable, as I am never tiered to say, further in that service.

...but I feel like Sancho Panza telling you guys not to fight immutable.
dq

michaelneale:
:bee: (just check if this needs updating to master, and if so, if ATH needs to run over it again as I see that small change for the encoding).

PR #438, opened 2016-08-26
[FIX JENKINS-36054] Lengthen page reload time tolerance from 2 to 20 seconds

tfennelly:
# Description

See [JENKINS-36054](https://issues.jenkins-ci.org/browse/JENKINS-36054).

A rerun of earlier attempt to handle machine hibernation ... reverting the revert and lengthening the time tolerance from 2 to 20 seconds.

The risk of having it too long is that someone could, in theory, bring a machine into and out of hibernation fast enough for us to ignore it (< 20s ?).

If this does not work then we're going to have try a more complex solution, pinging the server with a request that contains a UUID of some sort, asking the server to try send it back to us over SSE and if we don't get that we then reload, assuming SSE `EventSource` is clobbered and can't be recovered. There are risks that that might not work perfectly either because it too will depend on a time tolerance.

### DIFF
--- a/blueocean-web/src/main/js/main.jsx
+++ b/blueocean-web/src/main/js/main.jsx
@@ -149,3 +149,6 @@ function startApp(routes, stores) {
 Extensions.store.getExtensions(['jenkins.main.routes', 'jenkins.main.stores'], (routes = [], stores = []) => {
     startApp(routes, stores);
 });
+
+// Enable page reload.
+require('./reload');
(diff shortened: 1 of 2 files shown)


scherler:
I wonder whether we should return the timeout in the reload and whether we should place the requiered in the App component in componentWill(Un)Mount methods, since ATM that can result in leaking timeouts.

michaelneale:
how does this behave with the ATH? worth a try again?

i386:
@tfennelly not seeing any flashing so I suppose thats good? :)

michaelneale:
well I guess that is a :bee: as long as it doesn't upset the ATH?

tfennelly:
@michaelneale yeah, ATH passes fine https://ci.blueocean.io/job/Acceptance%20Test%20PR/32/

PR #439, opened 2016-08-26
[FIX JENKINS-37712] - branch encoding was wrong

kzantow:
# Description

See [JENKINS-37712](https://issues.jenkins-ci.org/browse/JENKINS-37712).

Tests covered in pending ATH test cases.

To test: create a multibranch project and in the repo, create branch with a slash in the name - e.g. `hotfix/JENKINS-37712`, verify you can access the run details for that branch.

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

@jenkinsci/code-reviewers @reviewbybees esp. @scherler

### DIFF
--- a/blueocean-dashboard/src/main/js/util/UrlUtils.js
+++ b/blueocean-dashboard/src/main/js/util/UrlUtils.js
@@ -173,11 +173,12 @@ export function getRestUrl({ organization, pipeline, branch, runId }) {
     const jenkinsUrl = require('../config').getJenkinsRootURL();
     let url = `${jenkinsUrl}/blue/rest/organizations/${encodeURIComponent(organizationName)}`;
     if (pipelineName) {
-        // pipelineName might include a folder path, don't escape it
+        // pipelineName might include a folder path, don't encode it
         url += `/pipelines/${pipelineName}`;
     }
     if (branch) {
-        url += `/branches/${encodeURIComponent(branch)}`;
+        // JENKINS-37712 branch needs to be double-encoded for some reason
+        url += `/branches/${encodeURIComponent(encodeURIComponent(branch))}`;
     }
     if (runId) {
         url += `/runs/${encodeURIComponent(runId)}`;


blueocean-dashboard/src/main/js/util/UrlUtils.js
```diff
@@ -173,11 +173,12 @@ export function getRestUrl({ organization, pipeline, branch, runId }) {
...
     let url = `${jenkinsUrl}/blue/rest/organizations/${encodeURIComponent(organizationName)}`;
     if (pipelineName) {
-        // pipelineName might include a folder path, don't escape it
+        // pipelineName might include a folder path, don't encode it
         url += `/pipelines/${pipelineName}`;
     }
     if (branch) {
-        url += `/branches/${encodeURIComponent(branch)}`;
+        // JENKINS-37712 branch needs to be double-encoded for some reason
+        url += `/branches/${encodeURIComponent(encodeURIComponent(branch))}`;
```
vivek: This is going to work as long as branch name is not already encoded. For example, for multi-branch project /pipelines/:id/ API will give branch names as an array:

    "branchNames" : [ "master", "feature2", "feature%2Fux-1" ]

Above 'feature/ux-1' appears encoded. If UI code uses this name and calls getResstUrl() its going to break. Maybe these branch names are coming from SSE events and are not encoded? Maybe then documenting this method that branch must not be encoded?

kzantow: Good points. Why does it have to be double encoded in the first place? Could we just fix that?

vivek: @kzantow More of jenkins core question but I see why its this way. 

Basically, '/' in particular is used by Stapler to dispatch/route. So if you have it singly encoded then Stapler will see /feature/ux-1 as two separate tokens, feature and ux-1 so might try getFeature("ux-1"). For this reason, any url unfriendly names are encoded and stored in model. When such encoded names are used in URL, they are doubly encoded so that 'feature/ux-1' is treaded as one token with single encoding. Branch names are such example, where a branch name could have any url unfriendly characters and are encoded in core and kept as such.

michaelneale: yes in core they are double encoded. In classic URIs you will also see they are double encoded. Much confusing. 

cc @cliffmeyers can you give this a sniff test? , as I know this has driven you insane plenty of times, perhaps there is a better way?

cliffmeyers: In some parts of the code, I decode the parameter in a while loop until last value equals just decoded value. Then I encode once or twice, depending on whether it's the fullName or branchName, respectively.

In newer code, I just pass the HAL self href to a function that parses out the right bits and returns the UI URL. That code is in blueocean-core-js but pending review I think.

cliffmeyers: This latter style requires us to pass around objects (for example, pass down the full Run object rather than its constituent props) but that makes way more sense to me, at least.

michaelneale:
Appears to work as advertised. 

I am inclined to :bee: as this I think is how it was  ( @cliffmeyers  will know for sure though).

scherler:
@michaelneale most cases are now covered in https://github.com/jenkinsci/blueocean-acceptance-test/pull/26 Today I will add test for queued state and then you can play with it and see whether majority of the cases are covered.

PR #440, opened 2016-08-26
Task/jenkins 36687 run pipeline refactor

cliffmeyers:
# Description
- This is just the first part of this refactor; trying to keep changes small and mergeable.
  - Updated build in blueocean-core-js to support styles and assets
  - Updated blueocean-web build so it can pull in both JDL and CoreJS
  - Started work on utility functions to convert a HAL href to an application HREF (for Run Details, our favorite)
- See [JENKINS-36687](https://issues.jenkins-ci.org/browse/JENKINS-36687).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-dashboard/package.json
```diff
@@ -35,7 +35,7 @@
     "skin-deep": "^0.16.0"
   },
   "dependencies": {
-    "@jenkins-cd/blueocean-core-js": "0.0.1-beta5",
+    "@jenkins-cd/blueocean-core-js": "0.0.2",
```
tfennelly: So have we already published these changes?

tfennelly:
So we have JDL type stuff in `blueocean-core-js` now? That smells wrong to me and seems like it's going to lead to confusion both for developers ("where should I __put__ X ... jdl or core-js?") and for "users" ("where should I __find__ X ... jdl or core-js?")

+1 on putting stuff like the new URL building utils in core-js.

cliffmeyers:
@tfennelly I was unsure exactly where this should go, but after batting it around with Keith we felt that only "dumb" purely visual components should go in the JDL. RunPipeline will be a smart component (integrating with REST API, depending on new fetch libs) which didn't seem right for the JDL, and would force a JDL to CoreJS dependency as well. I had also considered creating a pure visual component in JDL and then wrapping it with a new smart component in CoreJs... but that would leave a visual component of the "run" and "stop" icon that was basically useless.

So with all that said, do you still feel the same?

sophistifunk:
I definitely like the idea of keeping JDL free of networking and "business logic" stuff. It's supposed to be mostly about the visual styling and UI component "building blocks" - not necessarily atomic in nature, but doing nothing more than interacting with the user and dispatching events that code in other repos knows how to answer. We had some network stuff creep in there a while ago, and it was kind of a pain to get it back out again :D

michaelneale:
yes the JDL went well beyond its "bailiwick" with the amount of code in it (vs actual design related assets), so I think this is ok at a high level to have a core-js (there is a separate ticket to make it and jdl use standard gulp build I think I saw....).

sophistifunk:
bailiwick

michaelneale:
There are is a fair grab bag of changes in this, not sure what to review. Things seem to work, but it is based on a master from a few days ago I can tell (seeing it re-applied with the pagination changes would be good). Not sure what there is to functionally review, code wise, Tom had a few questions. Most things seem like a good idea I am sure.

cliffmeyers:
@michaelneale you're right, there is nothing functional here to review... just some infrastructure stuff that isn't being actively used. Given the merge hell some people have been trapped in, I figured I would open this up sooner rather than later, and try to get UrlBuilder into folks hands sooner as well.

tfennelly:
@cliffmeyers @sophistifunk @michaelneale totally agree on JDL not having shit in there that will create deps on jenkins core etc, but the question/issue still stands re this approach creating multiple homes for reusable visual components and what to do about that. I think multiple homes is a very bad idea.

Is there no way to define just the visual parts in the JDL, with them having the right action hooks on them that you can plug in an instance of the [`Pipeline`](https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-dashboard/src/main/js/api/Pipeline.js) class (which should also move to `core-js`), which handles the actual running of the pipeline etc.

So something like (rough):

```jsx
<RunPipelineXXXWidget pipeline={{pipelineInst}} ... />
```

Or ...

```jsx
<RunPipelineYYYWidget pipeline={{pipelineInst}} ... />
```

Where `RunPipelineXXXWidget` and `RunPipelineYYYWidget` are pure visual components defined in the JDL and `pipelineInst` is an instance of [`Pipeline`](https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-dashboard/src/main/js/api/Pipeline.js) (after it moves `blueocean-core-js`).

kzantow:
@tfennelly definitely don't think components that specific should go in JDL (although, it's not really my call exactly what it should encompass). My feeling is it would have basic UI components, much like Bootstrap provides, with the Jenkins styling and behavior necessary to use them and create a consistent UI as a plugin developer. However, things like a button to run a pipeline is not really a UI component but a functional thing and belongs ideally in a blueocean-pipeline-ui module or something similar. I think the crux of the problem is we don't have a great way to share components between plugins (aside from making more npm modules - which has already hit critical mass of being a pain in the butt - where a plugin would have to maintain the npm module and the Jenkins plugin itself). Can you think of any reasons we wouldn't want to do this?

PULL_REQUEST_TEMPLATE
```diff
@@ -5,13 +5,13 @@ See [JENKINS-XXXXX](https://issues.jenkins-ci.org/browse/JENKINS-XXXXX).
 # Submitter checklist
 - [ ] Link to JIRA ticket in description, if appropriate.
 - [ ] Change is code complete and matches issue description
-- [ ] Apppropriate unit or acceptance tests or explaination to why this change has no tests
+- [ ] Appropriate unit or acceptance tests or explanation to why this change has no tests
```
kzantow: :bee: :bee: :bee: ! :)

blueocean-web/gulpfile.js
```diff
@@ -55,36 +55,45 @@ builder.bundle('src/main/js/try.js')
...
-                if (err) {
-                    return logger.logError(err);
+/**
+ * Link (or copy) the specified module's subdir to a dir within /src/main/webapp/assets
+ * @param {string} dirName name of directory link
+ * @param {string} modulePath path within module, e.g. '@org-name/module/name/some/path
+ */
+function linkAssets(dirName, modulePath) {
+    var isWindows = /^win/.test(process.platform);
+    var assetsDstPath = './src/main/webapp/assets/' + dirName;
```
kzantow: :( I know this isn't a change, but not a fan of generating things into the source dir

kzantow:
Seems like a reasonable change, but seems like there are a lot of random things that could be improved to make situations like this work easier. As it stands, seems like a :bee: along the way to grand changes of grandeur

tfennelly:
@kzantow 

> @tfennelly definitely don't think components that specific should go in JDL (although, it's not really my call exactly what it should encompass). My feeling is it would have basic UI components, much like Bootstrap provides

Fair enough, that's true.

My original point still stands though ... creating "Yet Another Widget Library" isn't a good idea either. I'd prefer to broaden the definition of the JDL to be more than "jenkins-bootstrap" if we were to do something like that.

cliffmeyers:
@tfennelly @kzantow @michaelneale: I'm not opposed to the idea of combining the JDL and CoreJS into a more general-purpose library but that's kind of separate discussion and not really feasible to do before beta IMO. I do think it's worth talking about, since I'm not sure the mandate of "only visual stuff in JDL" makes sense since most folks writing a BlueOcean plugin are going to need stuff in CoreJS too and I can't see a JDL-only use case either.

I would however hate to see us be forced to write a "dumb" component that provides no value on its own just so we can decorate it with smarts later in CoreJS. That would be the case here: a button that toggles between "Play" and "Stop" icons is not useful on its own. This approach would also establish a dependency from CoreJS on the JDL, more watch complexity for local dev, etc. To me, being able to structure the code in a sensible manner is really the most important concern.

blueocean-web/gulpfile.js
sophistifunk: Agreed 100%; I hate this with the fire of a thousand suns, and I think after the demo we *really* need to find a way to not do it.

michaelneale:
Question for @i386: there is a bunch of blue ocena stuff in JDL - and now there is core-js. Does it make sense to have JDL, and slowly move components out of it so that it becomes more a styles centric thing?

i386:
JDL is supposed to be the bootstrap of Jenkins - it shouldn't contain any controls that relate to specific functionality (e.g. things specific to the run modal or the run buttons) or networking code. If core-js is supposed to allow sharing across modules that don't fit that definition then thats up to you guys.

tfennelly:
@i386 @cliffmeyers @michaelneale by that definition then, it seems like the pipeline graph should not be in the JDL?

I still think putting all of these reusable view components into `blueocean-core-js` alongside other non view components is a mistake. It feels to me like `blueocean-core-js` is already becoming a bit of dumping ground for anything of any kind that might possibly be reusable. Of course reusable components are important, but I think we need some logical separation here. Maybe we can look at doing this after the beta?

What would also help is for us to have some top level docs for people e.g. a static GitHub pages site for JDL that explains clearly what it is and also points people to the "blueocean-widgets" repo (excuse made-up name) if they are looking for higher level functional widgets that are not found in the JDL etc (and visa versa for people that are thinking of creating reusable widgets ... pointing them to the JDL for styles etc).

cliffmeyers:
I am closing this PR in favor of of #448 which includes all of these changes anyways. I don't want to have to merge master multiple times since there were a number of conflicts resulting from the pagination and JWT work.

PR #441, opened 2016-08-26
[FIX JENKINS-37708] queued item regression

kzantow:
# Description

See [JENKINS-37708](https://issues.jenkins-ci.org/browse/JENKINS-37708).

This needs an ATH test, JS tests are insufficient.

To test: set executors to 0, queue a freestyle build, click the queued item validate the 'item is waiting to run' page shows.

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/redux/actions.js
+++ b/blueocean-dashboard/src/main/js/redux/actions.js
@@ -366,6 +366,7 @@ export const actions = {
 
             // Only interested in the event if we have already loaded the runs for that job.
             if (eventJobRuns && event.job_run_queueId) {
+                let nextId = 0; // this is a terrible hack
                 for (let i = 0; i < eventJobRuns.length; i++) {
                     const run = eventJobRuns[i];
                     if (event.job_ismultibranch
@@ -379,11 +380,14 @@ export const actions = {
                         // run. No need to create another i.e. ignore this event.
                         return;
                     }
+                    if (parseInt(run.id) > nextId) { // need the estimated next id
+                         nextId = parseInt(run.id);
+                    }
                 }
 
                 // Create a new "dummy" entry in the runs list for the
                 // run that's been queued.
-                const newRun = {};
+                const newRun = { id: `${nextId+1}` };
 
                 // We keep the queueId so we can cross reference it with the actual
                 // run once it has been started.
--- a/blueocean-dashboard/src/main/js/redux/actions.js
+++ b/blueocean-dashboard/src/main/js/redux/actions.js
@@ -49,10 +49,6 @@ function tryToFixRunState(run, pipelineRuns) {
                 job_run_queueId: found.job_run_queueId,
             });
         }
-        return Object.assign({}, run, {
-            state: 'RUNNING',
-            result: 'UNKNOWN',
-        });
     }
     return run;
 }
@@ -438,8 +434,8 @@ export const actions = {
                             if (event.jenkins_event !== 'job_run_ended') {
                                 return { ...data,
                                     id: event.jenkins_object_id, // make sure the runId is set so we can find it later
-                                    state: 'RUNNING', // This is a horrible hack due to issues with QUEUED status
-                                    result: event.job_run_status,
+                                    state: event.job_run_status, // This is a horrible hack due to issues with QUEUED status
+                                    result: 'UNKNOWN',
                                 };
                             }
                             return data;
@@ -532,10 +528,22 @@ export const actions = {
 
     fetchRun(config) {
         return (dispatch, getState) => {
+            const runs = getState().adminStore && getState().adminStore.runs ? getState().adminStore.runs[config.pipeline] : [];
             smartFetch(
                 getRestUrl(config),
                 data => {
-                    const runs = getState().adminStore && getState().adminStore.runs ? getState().adminStore.runs[config.pipeline] : [];
+                    if (data.$failed) { // might be a queued item...
+                        const found = runs.filter(r => r.id === config.runId)[0];
+                        if (found) {
+                            found.$success = true;
+                            dispatch({
+                                id: config.pipeline,
+                                type: ACTION_TYPES.SET_CURRENT_RUN,
+                                payload: found,
+                            });
+                            return; // skip the next dispatch
+                        }
+                    }
                     dispatch({
                         id: config.pipeline,
                         type: ACTION_TYPES.SET_CURRENT_RUN,
--- a/blueocean-dashboard/src/main/js/redux/actions.js
+++ b/blueocean-dashboard/src/main/js/redux/actions.js
@@ -447,7 +447,13 @@ export const actions = {
                             if (event.jenkins_event !== 'job_run_ended') {
                                 return { ...data,
                                     id: event.jenkins_object_id, // make sure the runId is set so we can find it later
-                                    state: 'RUNNING', // This is a horrible hack due to issues with QUEUED status
+                                    // here, we explicitly set to running. this is because
+                                    // pipeline runs get removed from the queue, a running event is sent
+                                    // but they are still queued in Jenkins, as they may not actually
+                                    // be executing anything. However, pipeline doesn't send any further
+                                    // events so we're never notified when it actually does start, so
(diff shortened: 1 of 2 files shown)


blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -383,11 +384,14 @@ export const actions = {
                         // run. No need to create another i.e. ignore this event.
                         return;
                     }
+                    if (parseInt(run.id, 10) > nextId) { // need the estimated next id
+                        nextId = parseInt(run.id, 10);
+                    }
                 }
```
tfennelly: The above changes need some explanation. What's trying to be achieved etc? Why do we need to manufacture a fake run Id?

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -439,7 +443,7 @@ export const actions = {
                                 return { ...data,
                                     id: event.jenkins_object_id, // make sure the runId is set so we can find it later
                                     state: 'RUNNING', // This is a horrible hack due to issues with QUEUED status
-                                    result: event.job_run_status,
+                                    result: 'UNKNOWN',
```
tfennelly: Why are we changing this Vs using the state from the event? The state is not unknown afaik (it's [`QUEUED`](https://github.com/jenkinsci/pubsub-light-module/blob/master/src/main/java/org/jenkins/pubsub/listeners/SyncQueueListener.java#L82), I think).

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -532,10 +536,22 @@ export const actions = {
 
     fetchRun(config) {
         return (dispatch, getState) => {
+            const runs = getState().adminStore && getState().adminStore.runs ? getState().adminStore.runs[config.pipeline] : [];
             smartFetch(
                 getRestUrl(config),
                 data => {
-                    const runs = getState().adminStore && getState().adminStore.runs ? getState().adminStore.runs[config.pipeline] : [];
+                    if (data.$failed) { // might be a queued item...
```
tfennelly: Are we still using jquery? 😜

blueocean-dashboard/src/main/js/util/UrlUtils.js
```diff
@@ -173,11 +173,12 @@ export function getRestUrl({ organization, pipeline, branch, runId }) {
...
     let url = `${jenkinsUrl}/blue/rest/organizations/${encodeURIComponent(organizationName)}`;
     if (pipelineName) {
-        // pipelineName might include a folder path, don't escape it
+        // pipelineName might include a folder path, don't encode it
         url += `/pipelines/${pipelineName}`;
     }
     if (branch) {
-        url += `/branches/${encodeURIComponent(branch)}`;
+        // JENKINS-37712 branch needs to be double-encoded for some reason
+        url += `/branches/${encodeURIComponent(encodeURIComponent(branch))}`;
```
tfennelly: Nothing to do with this PR, but the fact that we are needing to do this screams to me that something is f&cked up somewhere. 😿

michaelneale: @tfennelly it has been in there since the start, stapler requires double encoding (!) in paths (the URIs in blue ocean do not, which in retrospect seems a mistake to be different) - this was talked about in another PR. @cliffmeyers I believe has some WIP to centralise the URL calculation so this doesn't have to happen.

blueocean-dashboard/src/main/js/redux/actions.js
michaelneale: Jake Weary.

blueocean-dashboard/src/main/js/redux/actions.js
michaelneale: yes I would like to know. The result should NOT however be "queued", but the state should be if it really was queued (see horrible hack above)

blueocean-dashboard/src/main/js/redux/actions.js
michaelneale: yes how did this work in the past? I thought it used the queue id (cc @imeredith).

blueocean-dashboard/src/main/js/util/UrlUtils.js
michaelneale: ie the "some reason" is stapler.

michaelneale:
Some questions raised by @tfennelly on the code, also when I tried this, styles are not loading due to some :bug: I can't work out: 

```
http://localhost:8080/jenkins/static/5fd8e9cf/plugin/blueocean-web/assets/octicons/octicons.css Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/jenkins/static/5fd8e9cf/plugin/blueocean-web/assets/css/jenkins-design-language.css Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/jenkins/static/5fd8e9cf/plugin/blueocean-web/assets/css/normalize.css Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/jenkins/static/5fd8e9cf/plugin/blueocean-web/assets/css/latofonts.css Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/jenkins/static/5fd8e9cf/plugin/blueocean-web/assets/css/normalize.css Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:8080/jenkins/static/5fd8e9cf/plugin/blueocean-web/assets/css/latofonts.css Failed to load resource: the server responded with a status of 404 (Not Found)
```

for example.

And it ends up amusingly looking like: 

screen shot 2016-08-29 at 3 37 50 pm

michaelneale:
This seems to work for queues, however, I am still seeing this weird state of things: 

screen shot 2016-08-29 at 5 36 57 pm

for the kzantow/failure-project "passing" branch. I run it, follow along, and then finished nodes are showing up as unfinished, so still a 🐛  it seems?

blueocean-dashboard/src/main/js/redux/actions.js
tfennelly: The queueId has always been on all events and was they way in which we correlated between queued and running runs (the queueId would always be the same).

It really depends on why the above change was made. There may be a reason but would don't know without a comment.

blueocean-dashboard/src/main/js/redux/actions.js
tfennelly: Ah, ignore my last comment. I was mixing up `state` and `result`. However, I see the state is set to `RUNNING` with an interestingly teasing, yet ultimately useless comment. Was prob me .... bad boy. It's be nice to know what the "issues with QUEUED status" were.

blueocean-dashboard/src/main/js/util/UrlUtils.js
tfennelly: I know it's been like that for a while, but it still scares me :)

blueocean-dashboard/src/main/js/redux/actions.js
kzantow: The issue is now we fetch a run when the RunDetails page is loaded, before we were fetching all runs and passing a run object along. If an item is actually queued, without a runId, it results in an invalid route. We need a `runId` for the route to work properly, and then fetch-or-not based on what we see in the cache. There is still a problem that refreshing the page would not work because you can't fetch a queued item by runId (even though it shows up in the activity list with an estimated run id, calculated similarly to the code here).

blueocean-dashboard/src/main/js/redux/actions.js
kzantow: We aren't using the state from the result because it often says `QUEUED` when we want `RUNNING`. The pipeline job moves from the queue, but isn't currently executing anything and still returns `QUEUED`. But we never get further updates when it actually *is* `RUNNING`.

blueocean-dashboard/src/main/js/redux/actions.js
kzantow: ❤️  jQuery

michaelneale:
:bee: I think those comments are clearer... still unpleasant

PR #462, opened 2016-09-01
[JENKINS-37837] JWT feature flag

imeredith:
# Description

This PR disables JWT by default and lets developers enable it with DFEATURE_BLUEOCEAN_JWT_AUTHENTICATION system property.

See [JENKINS-37837](https://issues.jenkins-ci.org/browse/JENKINS-37837).

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

### DIFF
--- a/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
+++ b/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
@@ -4,6 +4,7 @@ import hudson.Extension;
 import hudson.security.AuthorizationStrategy;
 import hudson.security.FullControlOnceLoggedInAuthorizationStrategy;
 import io.jenkins.blueocean.BluePageDecorator;
+import io.jenkins.blueocean.commons.BlueOceanConfigProperties;
 import jenkins.model.Jenkins;
 import net.sf.json.util.JSONBuilder;
 
@@ -16,10 +17,8 @@ import java.io.StringWriter;
 public class BlueOceanConfig extends BluePageDecorator {
 
     public boolean isRollBarEnabled(){
-        return Boolean.getBoolean("BLUEOCEAN_ROLLBAR_ENABLED");
+        return BlueOceanConfigProperties.ROLLBAR_ENABLED
     }
-    public final boolean FEATURE_JWT = Boolean.getBoolean("FEATURE_BLUEOCEAN_JWT_AUTHENTICATION");
-
 
     public String getBlueOceanConfig(){
         return createConfig();
@@ -48,7 +47,7 @@ public class BlueOceanConfig extends BluePageDecorator {
                         .key("authorizationStrategy").object()
                             .key("allowAnonymousRead").value(allowAnonymousRead)
                             .endObject()
-                        .key("enableJWT").value(FEATURE_JWT)
+                        .key("enableJWT").value(BlueOceanConfigProperties.BLUEOCEAN_FEATURE_JWT_AUTHENTICATION)
                         .endObject()
                     .endObject()
                 .endObject();
--- a/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
+++ b/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
@@ -18,6 +18,7 @@ public class BlueOceanConfig extends BluePageDecorator {
     public boolean isRollBarEnabled(){
         return Boolean.getBoolean("BLUEOCEAN_ROLLBAR_ENABLED");
     }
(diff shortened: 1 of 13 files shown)


blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
```diff
@@ -18,6 +18,7 @@
     public boolean isRollBarEnabled(){
         return Boolean.getBoolean("BLUEOCEAN_ROLLBAR_ENABLED");
     }
+    public final boolean FEATURE_JWT = Boolean.getBoolean("FEATURE_BLUEOCEAN_JWT_AUTHENTICATION");
```
i386: Make that `BLUEOCEAN_FEATURE_JWT_AUTHENTICATION`

i386: Good to see this config object being used 👍

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/BlueOceanRootAction.java
```diff
@@ -21,7 +21,7 @@
 public class BlueOceanRootAction implements UnprotectedRootAction, StaplerProxy {
     private static final String URL_BASE="blue";
 
-    private final boolean disableJWT = Boolean.getBoolean("DISABLE_BLUEOCEAN_JWT_AUTHENTICATION");
+    private final boolean enableJWT = Boolean.getBoolean("FEATURE_BLUEOCEAN_JWT_AUTHENTICATION");
```
i386: Update this property name

blueocean-rest/README.md
```diff
@@ -110,18 +110,17 @@ To avoid needing a crumb to POST data, the header `Content-Type: application/jso
 
 # Security
 
+NOTE: JWT is disabled by default for now. to enable JWT authentication use FEATURE_BLUEOCEAN_JWT_AUTHENTICATION=true system property.
```
i386: Can you document two things here:
- what happens when the flag is on
- what happens when the flag is off

i386: It would be nice if our system properties were in a single place

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
```diff
@@ -44,13 +45,13 @@ private String createConfig() {
                     .key("security")
                     .object()
                         .key("enabled").value(jenkins.isUseSecurity())
-                        .key("authorizationStrategy")
-                        .object()
+                        .key("authorizationStrategy").object()
                             .key("allowAnonymousRead").value(allowAnonymousRead)
+                            .endObject()
+                        .key("enableJWT").value(FEATURE_JWT)
```
vivek: Does this property accessible as security.jenkinsConfig.enableJwt? Is it not more of BO specific feature so maybe it should be security.enableJwt? or maybe just top level enableJwt?

blueocean-rest/README.md
michaelneale: I think there is a standard Jenkins way of doing sys props.. just not sure what.

kzantow: Semi-standard way is to use the fully qualified class name dot property name, which is a reasonable thing to do.

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
michaelneale: it is in a system property as a top level thing I think, this is just in the client side config object.

blueocean-rest/README.md
michaelneale: gotcha.. right.

michaelneale:
looks like 🐛  as test cases may need to be changed: 

```
Failed tests: 
  ProfileApiTest.badTokenTest1:338 expected:<401> but was:<404>
  ProfileApiTest.badTokenTest2:350 expected:<401> but was:<404>
  ProfileApiTest.createUserFavouriteFolderTest:190 expected:<200> but was:<403>
  ProfileApiTest.createUserFavouriteTest:138 expected:<200> but was:<403>
  ProfileApiTest.getAuthenticatedUser:306 expected:<200> but was:<404>
  ProfileApiTest.getUserDetailsTest:113 expected:<alice@jenkins-ci.org> but was:<null>
```

(local as well as CI environment)

blueocean-core-js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@jenkins-cd/blueocean-core-js",
-  "version": "0.0.5",
+  "version": "0.0.8",
```
michaelneale: why such a big jump?

PR #463, opened 2016-09-02
JENKINS-37830# API to remove a queued item

vivek:
# Description

See https://issues.jenkins-ci.org/browse/JENKINS-37830.

To test manually:

* Create a queued job
* curl  http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/:id/queue and get queue id from here then
* curl -X DELETE http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/:id/queue/:id/

# Submitter checklist

ATH passes locally

# Reviewer checklist
- [x] Reviewed the code


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueItemImpl.java
```diff
@@ -55,6 +57,15 @@ public int getExpectedBuildNumber() {
     }
 
     @Override
+    public void delete(){
+        if(!item.hasCancelPermission()){
+            throw new ServiceException.ForbiddenException(String.format("Not authorized to stop queue: %s", getId()));
+        }
+
+        Jenkins.getInstance().getQueue().cancel(item);
```
michaelneale: does this happen on behalf of the user that requested it or is it a system level operation? (I am guessing it has no concept of user?).

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/TreeResponse.java
```diff
@@ -31,14 +34,42 @@
         public Object invoke(StaplerRequest request, StaplerResponse response, Object instance, Object[] arguments)
                 throws IllegalAccessException, InvocationTargetException, ServletException {
 
-            final Object resp = target.invoke(request, response, instance, arguments);
+            /**
+             *  If request.method and HTTP verb annotations {@link GET}, {@link POST}, {@link PUT} and {@link DELETE}
+             *  do not match it skips invoking this target. If there no such annotations present then GET as default is
+             *  assumed and request is dispatched to target.
+             */
+            if (matches(request)) {
```
michaelneale: how come it is necessary to do this?

michaelneale:
this has a change on top of the queue stop - wouldn't it be better to keep it just about the queue at this stage?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueItemImpl.java
vivek: @michaelneale Depends on the authorization strategy. For example admin can set a user to only cancel specific projects. Basically authorization is applicable at any item level, be it job or folder or run and it could be applicable system wide or per user or a group.

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/TreeResponse.java
vivek: @michaelneale It was a bug that was hiding, adding a DELETE operation exposed it. Without this if you have multiple methods operating at '/' routing level can be selected arbitrarily. This enforces method level routing based on HTTP method annotation, thats the intent anyways.

michaelneale:
as suspected, this breaks running builds from favorties as PUT doesn't work. So this is a 🐛  until that can be chased down with @cliffmeyers. Perhaps punt until next week.

cliffmeyers:
@vivek we should touch base on Tues to talk about the PUT vs. POST issue. I was never able to repro it locally though. @michaelneale perhaps it had to do with the fact that anonymous access was disabled on your local Jenkins?

vivek:
@cliffmeyers sure, lets talk about that POST vs PUT thing. What they got to do with this ticket?

vivek:
@michaelneale, @cliffmeyers is going to try out this branch and make the POST/PUT call according to whats documented and commit to this PR after it all works. I guess when you are back Monday you can test it out and see it works well for you.

cliffmeyers:
@michaelneale @vivek I confirmed the the backend changes did produce and error when trying to call "run" and "replay" via a PUT method. I switched things over to POST and it works fine now.

vivek:
@cliffmeyers thanks! @michaelneale when you get chance test it out.

michaelneale:
I was able to run the API, and also run things from favourites etc, so seems good. 

However, the versions for core js need to be the proper ones to merge right @cliffmeyers ? So provisional "LGTM" if cliff is happy.

vivek:
@cliffmeyers I merged latest master to resolve conflicts, it sets core version to whats at master. Please review and if all looks good we can merge it.

vivek:
@cliffmeyers Ok, so I investigated ATH failure further, for replay and POST of multibranch projects) UI is sending PUT and it fails. Looks like this is different code path where PUT for replay and run needs to be changed to POST?

cliffmeyers:
@vivek the problem is that the version of blueocean-core-js was wrong after the last merge. Not your fault since a beta version landed in master (which we try to avoid) but reverting to 0.0.9 means we lost the PUT -> POST change. Fixing this now, and getting everything to a non-beta version as well.

cliffmeyers:
Ran latest changes against ATH locally: passed.

`2016-09-06 17:24:22,571 [Thread-3] INFO   (InputStreamHandler.java:47) com.github.eirslett.maven.plugins.frontend.lib.DefaultGulpRunner - OK. 253  total assertions passed. (3m 5s)`

vivek:
@cliffmeyers Thanks. Once CI job passes, I will merge these changes.

PR #468, opened 2016-09-03
JENKINS-37763# numberOfRunningPipelines and numberOfQueuedPipelines a…

vivek:
# Description

See https://issues.jenkins-ci.org/browse/JENKINS-37763.

* Create pipeline or freestyle job with sleep, do couple of them
* call GET http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/:id, and see number of running jobs is equal to numberOfRunningPipelines field in pipeline object.
* make sure there are jobs in queue and call above GET URL and check numberOfQueuedPipelines is same as queued items for that pipeline builds
* For multi-branch project, run few branches and compare using above logic
> Please note for Pipeline and MBP queue API is not implemented correctly - its mostly going to be 0, once https://issues.jenkins-ci.org/browse/JENKINS-37337 is implemented, numberOfQueuedPipelines element can be tested correctly.

# Submitter checklist

Added tests for non-multibranch pipeline job. Could not get a queued/running multi branch test to write, was tested manually. Possibly such tests should be added in ATH.

# Reviewer checklist
- [ ] Reviewed the code

…dded to pipeline object


vivek:
@cliffmeyers this needs your review.

michaelneale:
Code LGTM - but I would like @cliffmeyers to give this a bee after he tries it as he will be consuming it.

cliffmeyers:
@vivek I pulled this down to test it and here is what I see:

# I ran three different freestyle jobs and the number of running and queued jobs for each is accurate. 
# I noticed that if I POST'd repeatedly to the same freestyle job to create a new run, each time the same id and expectedBuilderNumber was returned - like you could only queue the freestyle job once. Is that expected behavior?
# If I POST'd to run the same branch a few times, it created multiple runs for it immediately and number of runs is accurate.

So I think functionality-wise this is a :bee from me but perhaps you could shed some light on the second point there. Thanks!

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractPipelineImpl.java
```diff
@@ -206,6 +212,20 @@ public Resource resolve(Item context, Reachable parent, Item target) {
...
 
     @Override
+    public int getNumberOfRunningPipelines(){
+        RunList runningJobs =  job.getBuilds().filter(isRunning);
+        return Iterators.size(runningJobs.iterator());
+    }
+
+    @Override
+    public int getNumberOfQueuedPipelines() {
+        if(job instanceof BuildableItem) {
```
kzantow: Could we have `BuildableItem`s that are random other types unsupported in BO that would throw these numbers off?

vivek: @cliffmeyers 

>I noticed that if I POST'd repeatedly to the same freestyle job to create a new run, each time the same id and expectedBuilderNumber was returned - like you could only queue the freestyle job once. Is that expected behavior?

to be clear this behavior is different issue not specific to this PR. However @imeredith can answer this question as he has written this part of the code.

vivek: @kzantow no, I don't think so. BuildableItem can be scheduled and queued so thats right contract to check against.

imeredith: Freestyle jobs only have one queued item unless they are paramaterized.

blueocean-rest/README.md
```diff
@@ -529,7 +535,9 @@ Each branch in the repo with Jenkins file will appear as a branch in this pipeli
         "numberOfSuccessfulBranches": 0,
         "numberOfSuccessfulPullRequests": 0,
         "totalNumberOfBranches": 3,
-        "totalNumberOfPullRequests": 0
+        "totalNumberOfPullRequests": 0,
+        "numberOfQueuedPipelines" : 0,
+        "numberOfRunningPipelines" : 2
```
tfennelly: Is there a typo here .... shouldn't `totalNumberOfPullRequests` be `2` ? (`numberOfQueuedPipelines + numberOfRunningPipelines `)

blueocean-rest/README.md
```diff
@@ -565,7 +573,9 @@ Each branch in the repo with Jenkins file will appear as a branch in this pipeli
             "name": "feature2",
             "organization": "jenkins",
             "weatherScore": 100,
-            "pullRequest": null
+            "pullRequest": null,
+            "numberOfQueuedPipelines" : 0,
+            "numberOfRunningPipelines" : 2
```
tfennelly: `totalNumberOfPullRequests` not on this?

blueocean-rest/README.md
```diff
@@ -601,7 +611,9 @@ Each branch in the repo with Jenkins file will appear as a branch in this pipeli
             "name": "master",
             "organization": "jenkins",
             "weatherScore": 100,
-            "pullRequest": null
+            "pullRequest": null,
+            "numberOfQueuedPipelines" : 0,
+            "numberOfRunningPipelines" : 2
```
tfennelly: `totalNumberOfPullRequests` ?

tfennelly:
LGTM insofar as I get it.

cliffmeyers:
Functionally this looks good to me. Thanks @vivek !

blueocean-rest/README.md
vivek: No, this is fine, totalNumberOfPullRequests can be zero as not all running/queued jobs are result of pull requests.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractPipelineImpl.java
```diff
@@ -206,6 +212,20 @@ public Resource resolve(Item context, Reachable parent, Item target) {
...
+            return Iterables.size(Jenkins.getInstance().getQueue().getItems((BuildableItem)job));
```
kzantow: I guess the real question is this is an `AbstractPipelineImpl` ? but it's getting all the `BuildableItems` in Jenkins? Should it be restricted just to items related to this pipeline? Otherwise should this API be elsewhere?

kzantow:
:bee: I totally misread a bit there, @vivek sorry! :)

PR #469, opened 2016-09-03
Task/jenkins 37685 toaster mobx fixes

cliffmeyers:
# Description
- See [JENKINS-37685](https://issues.jenkins-ci.org/browse/JENKINS-37685).

So the original problem here is that if `ToastService` just used the `push()` method to add the toast to the `toasts` property, the UI was *NOT* displaying the toasts. The reason for this is that MobX is aware of whether any `@observer`s have subscribed to an object's `@observable` properties and will only trigger an update if there is at least one subscriber (for performance). Since `ToastDrawer` was only passing the `toasts` array down to the JDL `Toaster` it was not itself actually referencing any of the observable props, and therefore not receiving an update to trigger a pass of the new props.

The fix was just to add a reference to `toastService.count` which is a `@computed` property on the `ToastService`. This creates the subscription which ensures there s a re-render of `ToastDrawer` when the toasts are added and removed.

In practice I don't think we'll see this kind of bug in many other places since `ToastDrawer` is sort of a contrived component that bridges MobX and the pure `Toaster` JDL component. Another fix would have been to make `Toaster` an `@observer` but since it's in the JDL I don't think we want to add the `mobx-react` dependency.

In practice this shouldn't come up very often since most `@observer` components will reference at least some of the `@observable` props and life will be good.

# Submitter checklist

# Reviewer checklist
   - [x] Execute some runs and ensure that Toasts still display as expected.
- [x] Reviewed the code


michaelneale:
I think a :bee: from me (not sure if you checked it with ATH, should be ok I as small change)

blueocean-core-js/src/js/ToastService.js
```diff
@@ -37,10 +38,7 @@ export class ToastService {
             newToast.id = Math.random() * Math.pow(10, 16);
         }
 
-        // TODO: determine why it's necessary to re-set the "toasts" field to trigger the UI update
-        const toasts = this.toasts.slice();
```
kzantow: @michaelneale will be disappointed we are no longer `slicing` the `toasts`

blueocean-web/src/main/js/components/ToastDrawer.jsx
```diff
@@ -19,6 +19,9 @@ export class ToastDrawer extends Component {
     }
 
     render() {
+        // need to reference observable prop in render to trigger subscription
+        toastService.count;
```
kzantow: this is very weird...

kzantow:
:bee: not happy with the random property reference needed in render, though

cliffmeyers:
ATH passed locally.

`2016-09-06 15:10:24,846 [Thread-3] INFO   (InputStreamHandler.java:47) com.github.eirslett.maven.plugins.frontend.lib.DefaultGulpRunner - OK. 330  total assertions passed. (3m 25s)`

cliffmeyers:
ATH passed locally (ran again after publish of Core JS).

`2016-09-06 15:29:50,017 [Thread-3] INFO   (InputStreamHandler.java:47) com.github.eirslett.maven.plugins.frontend.lib.DefaultGulpRunner - OK. 253  total assertions passed. (3m 36s)`

PR #471, opened 2016-09-05
[JENKINS-37954] Running job has incorrect stage selected before steps start executing

scherler:
# Description

See [JENKINS-37954](https://issues.jenkins-ci.org/browse/JENKINS-37954) why I removed the pipelineGraph @i386 if you do not agree I will look into the stage thing but IMO it is not worth it.

screenshot from 2016-09-05 18-40-47

# Submitter checklist
https://github.com/jenkinsci/blueocean-acceptance-test/pull/33

# Reviewer checklist
- [x] Reviewed the code


michaelneale:
@scherler nice - but some lint errors on this failed the build

i386:
🐛 we do want to show the Pipeline graph - please see @michaelneale's comments on the JIRA ticket.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
```diff
@@ -24,15 +24,15 @@ import { calculateNode } from '../util/KaraokeHelper';
...
-const queuedState = () => (
+const QueuedState = () => (
     <EmptyStateView tightSpacing>
         <p>
             <Icon {...{
                 size: 20,
                 icon: 'timer',
                 style: { fill: '#fff' },
             }} />
-            <span>Waiting for run to start.</span>
```
michaelneale: I'm not sure if this ever worked, I have never seen this copy show up with pipeline... is this the right thing to change?

scherler: It worked with freestyle (and other job types that have a real queued state). We had a ticket around 2 weeks ago where we covered that part. However the queued state for that is ATM broken.

I am now reusing this and added a hook for AT to be able to identified the content.

scherler:
@michaelneale doh, the testing file :( BTW that file is huge!

@i386 ok will trick the pipeline graph then. Me still thinks that without it is much cleaner, since the graph does not provide any benefit nor does it reflect the reality of things, since we are not in the first node of the graph but rather waiting to get there.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
michaelneale: gotcha, thanks, good to know.

michaelneale:
@scherler often times there will be no stage graph to show, so it will only show up if it has seen it run before (in which case it is reasonably confident that it hasn't changed, but it could change once it starts running, just like in karaoke today). 

Really, the current behavior before this PR wasn't foo *far* from wrong. The main thing wrong was showing the wrong empty state message/icon, and selecting the last item. If it said the correct wording, with theicon, and selected the first node in the stage graph, would be spot on behavior.

michaelneale:
Looks good. 

Assuming you run through ATH then :bee:

PR #480, opened 2016-09-08
[JENKINS-36906 & JENKINS-37832] Login related things

imeredith:
# Description

This PR covers 2 tickets. It adds login/logout buttons. It also detects if Jenkins has been restarted  or if the user has changed and reloads the page.

See [JENKINS-36906](https://issues.jenkins-ci.org/browse/JENKINS-36906).
See [JENKINS-37832](https://issues.jenkins-ci.org/browse/JENKINS-37832).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-core-js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@jenkins-cd/blueocean-core-js",
-  "version": "0.0.12",
+  "version": "0.0.13-beta1",
```
vivek: beta dependency?

i386:
Nice 👍

vivek:
@imeredith LGTM 🐝  except beta dependencies introduced. Guess you will release them before merge?

imeredith:
@vivek its normal for us to do beta releases and then do proper ones before we merge. Unforunatly changes to npm modules wont be built without that.

blueocean-core-js/src/js/fetch.js
```diff
@@ -4,8 +4,29 @@ import isoFetch from 'isomorphic-fetch';
 import utils from './utils';
 import config from './config';
 
+let refreshToken = null;
 export const FetchFunctions = {
-        /**
+    checkRefreshHeader(response) {
+        const _refreshToken = response.headers.get('X-Blueocean-Refresher');
```
michaelneale: that's a bit of an odd name for a header - wouldn't normally use a verb in a header.

vivek: @imeredith All good then. Looks nice 🐝

i386:
@vivek @imeredith I think the refresher is good for catching when Jenkins is restarted, upgraded or Blue Ocean version changes - we should continue to do this when we switch on JWT 👍

vivek:
@i386 probably not, lets decide on this when we enable JWT.

michaelneale:
@imeredith this will need updating (probably with @cliffmeyers  bump of js lib versions caused a conflict)

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/BlueOceanRootAction.java
```diff
@@ -52,20 +59,37 @@ public Object getTarget() {
...
+                SecurityContext securityContext = new SecurityContextImpl();
+                securityContext.setAuthentication(tokenAuthentication);
+                SecurityContextHolder.setContext(securityContext);
+
+                //TODO: implement this as filter, see PluginServletFilter to clear the context
+            } else {
+                // Add X-Blueocean-Refresher to response, so that we can detect when the user changes. Wont be needed with JWT
+                MessageDigest md = null;
+                try {
+                    md = MessageDigest.getInstance("SHA-1");
```
michaelneale: yeah this is a bit absurd isn't it, is there no way to statically pick one that is known to be installed? 

cc @i386 is there a standard jenkins way of hashing (I heard you mentioned)

i386: @michaelneale only if you want to hash and sign. Its a bit of work to use.

michaelneale: We don't want to add any more guava - it is notorious (for being an unversioned hot mess)

i386: Its only shitty if their types are exposed in public api

michaelneale:
So tried a few things: 

* Created multibranch, restarted server while looking at activity page
* Leave open on dashboard restart server

and so far, so good. 

Some other scenarios to test: 
* Anon read access
* Github oauth with restarting server
* Allowing session to expire (how? just wait?)

Also, this needs to be scraped over with the ATH but is looking good.

imeredith:
ath passed

michaelneale:
@imeredith looks good to me - I did a wide but shallow test. And @vivek likes code, so why not :bee: if @i386 is ok?

PR #486, opened 2016-09-08
JENKINS-37202# Set pagination by default

vivek:
# Description

See https://issues.jenkins-ci.org/browse/JENKINS-37202.

* Create 120 jobs (?)
* Call URL localhost:8080/jenkins/blue/rest/search/?q=type:pipeline
* You should get only 100 records

# Submitter checklist
Ran ATH locally, all tests passed successfully.

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest/src/main/java/io/jenkins/blueocean/rest/pageable/PagedResponse.java
+++ b/blueocean-rest/src/main/java/io/jenkins/blueocean/rest/pageable/PagedResponse.java
@@ -28,6 +28,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 @InterceptorAnnotation(PagedResponse.Processor.class)
 public @interface PagedResponse {
     class Processor extends Interceptor {
+        private static final int DEFAULT_LIMIT=100;
         @Override
         public Object invoke(StaplerRequest request, StaplerResponse response, Object instance, Object[] arguments)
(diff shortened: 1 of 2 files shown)


imeredith:
:bee: lgtm

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/pageable/PagedResponse.java
```diff
@@ -28,6 +28,7 @@
 @InterceptorAnnotation(PagedResponse.Processor.class)
 public @interface PagedResponse {
     class Processor extends Interceptor {
+        private static final int DEFAULT_LIMIT=100;
```
tfennelly: Hey Vivek, is this default configurable in any way? Would it make sense to allow people set a startup prop (with an upper and lower bound)?

tfennelly:
🐝 LGTM

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/pageable/PagedResponse.java
vivek: @tfennelly Not at the moment. We can set it via property if need be but please note that limit query parameter lets you override it as well. What makes it complicated is that with https://issues.jenkins-ci.org/browse/JENKINS-38055, we are going to revisit it to keep per API limit default.

PR #488, opened 2016-09-08
Task/jenkins 37519 fetch capabilities

cliffmeyers:
# Description
- See [JENKINS-37519](https://issues.jenkins-ci.org/browse/JENKINS-37519).
- Relies on changes in #494; can't merge this until it lands.
- The idea here is that we use `CapabilityAugmenter` (happy to rename to something less cryptic) to attach to a Promise returned from a REST data call. The "augmenter" will:
   - Walk the entire object graph and find any objects with a `_class` property,
   - Make a single call to `/blue/rest/classes` endpoint to load capabilities for all classes found,
   - Update the original data set by adding a `_capabilities` property alongside each `class` property which is an array of all that class's capabilities.
   - Capability data is cached in the instance, so it will not be reloaded unless a user kills the app.
- Possible enhancements
   - We can enhance the caching to use local storage if desired. For people who leave the app open all day it won't make much of a difference but if folks are opening lots of different windows it will cause some extra hits to the server. We could also look at adjusting the HTTP caching headers on the `/blue/rest/classes` endpoint.
   - ~~We could enhance the Augmenter using something like Strategy pattern to determine how the graph is walked, in case we want to optimize. I'm going to benchmark the walker now to see if this is worthwhile.~~ Probably a waste of time unless there is a reason we need to selectively walk the graph. In all my tests the parse took no longer than 1ms, and in many cases reported as 0ms. With actions enabled, it was between 0 and 2ms.

# Submitter checklist

# Reviewer checklist
   - Changes to dashboard / personalization code will happen in a separate PR; the additions here won't actually impact anything in the app right now.
- [ ] Reviewed the code


blueocean-core-js/src/js/capability/CapabilityAugmenter.js
```diff
@@ -0,0 +1,99 @@
...
+        classMap[className] = [];
+    }
+
+    classMap[className].push(clazz);
+};
+
+const canWalk = (item) => item && (typeof item === 'object' || Array.isArray(item));
+
+// TODO: can we really filter out 'actions' in all cases? should it use a leading underscore?
+const IGNORED_PROPS = ['_links', 'actions'];
```
scherler: https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-dashboard/src/main/js/util/logDisplayHelper.js#L35

```
        // FIXME: TS I need to talk to cliffMeyers how we can refactor the following code to use capabilities
        // the problem I see ATM is that we would need to ask the c-API every time for each action, whether this
        // action has the capability for logging
        const hasLogs = item.actions ? item.actions
                .filter(action => action._class === 'org.jenkinsci.plugins.workflow.support.actions.LogActionImpl').length > 0
            : false;
```

If we filter this action we lose knowledge as the above, or?

cliffmeyers: @scherler I will add an option to fetch data for "actions" too. It's very large, so that's why I was hoping we could disable it. Which REST API calls are you examining the actions for?

cliffmeyers: Fixed in 4b9b230

cliffmeyers:
I would like to keep this PR as just the new API, and then open a second PR that introduces it throughout the app - just to limit annoyances with npm publish and version conflicts.

blueocean-core-js/src/js/capability/Capability.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Created by cmeyers on 9/8/16.
+ */
+export const CAPABILITY = {
+    SIMPLE_PIPELINE: 'org.jenkinsci.plugins.workflow.job.WorkflowJob',
```
kzantow: Actual capability references should not be defined in this core module. We should come up with an easier way to reference them... my suggestion, which isn't perfect, is to add a `<obj>._isa.<ShortClassName/CapabilityName>` or `_can` so we don't have to define a bunch of random stuff all over.

For example, if we had a capability that was `displayTestResults`, it would be mapped to pertinent objects as `<obj>._can.displayTestResults` or... `<obj>._can_displayTestResults` so we don't have to test for `_can` all over the place. Similarly classes might be mapped by short class name, so the above example would map to: `<obj>._isa.WorkflowJob` or `<obj>._isa_WorkflowJob`. We could optimize some property adders to closures and reuse JS strings for this stuff, if necessary.

It would be nice to have the class info and capability info separate for `_isa` and `_can` references, respectively.

blueocean-core-js/src/js/capability/CapabilityApi.js
```diff
@@ -0,0 +1,25 @@
...
+     * Fetch the capabilities for one or more class names.
+     *
+     * @param classNames
+     * @returns {Promise} with fulfilled {object} keyed by className, with an array of string capability names.
+     * @private
+     */
+    fetchCapabilities(...classNames) {
+        const path = config.getJenkinsRootURL();
+        const classList = classNames.join(',');
+        const classesUrl = utils.cleanSlashes(`${path}/blue/rest/classes/?q=${classList}`);
```
kzantow: Should be able to just convert the `q=${classList}` to a post body

blueocean-core-js/src/js/capability/CapabilityStore.js
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Fetch the capabilities for one or more class names.
+     *
+     * @param classNames
+     * @returns {Promise} with fulfilled {object} keyed by className, with an array of string capability names.
+     * @private
+     */
+    _fetchCapabilities(classNames) {
+        return this._capabilityApi.fetchCapabilities(classNames)
```
kzantow: Might be some duplicate capabilities fetched but the batching should make it not really a problem like it was when fetching individually.

blueocean-core-js/src/js/capability/CapabilityAugmenter.js
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /**
+     * Add "_capabilities" data or all objects with a "_class" property.
+     *
+     * @param {object|Array} data
+     * @param {boolean} includeActions whether to fetch capabilities for items in the 'actions' property
+     * @returns {Promise}
+     */
+    augmentCapabilities(data, includeActions = false) {
```
kzantow: I would always include everything, just make it consistent.

blueocean-core-js/src/js/capability/CapabilityAugmenter.js
```diff
@@ -0,0 +1,144 @@
...
+        if (!includeActions) {
+            ignoredProps.push(ACTIONS_PROP_NAME);
+        }
+
+        const started = new Date().getTime();
+
+        let node = nodesToWalk.shift();
+
+        while (node) {
+            nodesAlreadyWalked.push(node);
```
kzantow: Wouldn't really need this in a JSON response since there's no circular data references, but probably good to do anyway if it was reused in a different context.

kzantow: Ah, good point. Should probably just omit that field and other known metadata field like `_links`, no?

blueocean-core-js/src/js/capability/Capability.js
cliffmeyers: I removed this is lieu of changes in commit 	73d091a. I am not sure if this is exactly what you had in mind or if it's workable. We should probably define some of the core capabilities around pipelines in a place where dashboard and personalization can get at them to avoid "magic string" bugs.

blueocean-core-js/src/js/capability/CapabilityStore.js
cliffmeyers: These were already de-duped upstream but good to do it here too. Added in commit e03ecff

kzantow: I mean, there's a de-duper for identical get fetches, but if this has some random number and order of class names, that won't match.

cliffmeyers: This is just making sure that duplicate class names aren't passed to the REST API itself. e.g.

.../classes/?q=a.A,b.B,C.C,b.B,a.A

becomes

.../classes/?q=a.A,b.B,C.C

In order to handle multiple concurrent fetches in a really smart way it'd need a bit more work. It'd probably make more sense just to do the localStorage caching if that's a problem, since it would kill two birds w/ one stone.

blueocean-core-js/src/js/capability/CapabilityAugmenter.js
cliffmeyers: yep, `_links` is already omitted by default.

blueocean-core-js/src/js/capability/CapabilityAugmenter.js
cliffmeyers: Fixed in commit d9e7b55

blueocean-core-js/src/js/capability/CapabilityApi.js
cliffmeyers: Fixed in commit 47aa141

cliffmeyers:
@kzantow at this point I believe I've addressed all the comments in the PR. You think this is ready for a bee?

cliffmeyers:
There is a test failure here but unrelated to my changes; appears to be introduced in #494

cliffmeyers:
@kzantow In a few spots in the app we use a `clone` method that uses the `JSON.stringify/parse` trick. Unfortunately these means any instances of the "can" method get tossed away. As I see it we have two options:

1. Introduce a special `clone` method that handles this (although then everyone has to use it), or
2. Just rework the API a bit so this

`pipeline.can('org.jenkinsci.plugins.workflow.job.WorkflowJob')`

becomes something like

```
import { capable } from '@jenkins-cd/blueocean-core-js;
capable(pipeline, 'org.jenkinsci.plugins.workflow.job.WorkflowJob')
```

Feels like the latter is less error prone to me. What do you think?

cliffmeyers:
@kzantow I added the "capable" function in 94fe278 to help move this along. I think it's the most flexible solution and the code is not appreciably more verbose than before.

blueocean-core-js/src/js/capability/CapabilityAugmenter.js
```diff
@@ -0,0 +1,151 @@
...
+        // eslint-disable-next-line no-param-reassign
+        classMap[className] = [];
+    }
+
+    classMap[className].push(clazz);
+};
+
+const canWalk = (item) => item && (typeof item === 'object' || Array.isArray(item));
+
+const DEFAULT_IGNORED_PROPS = ['_links'];
```
kzantow: NIT: would add any other known things like `$pager`, though it shouldn't be fetching through the pagination fetch... but still

blueocean-core-js/src/js/capability/CapabilityAugmenter.js
```diff
@@ -0,0 +1,151 @@
...
+
+                target._capabilities = capabilities || [];
+            }
+        }
+
+        if (this._perfLoggingEnabled) {
+            console.debug(`augmenter.inject: ${new Date().getTime() - started}ms`);
+        }
+
+        if (this._warnLoggingEnabled) {
```
kzantow: Side note: I found the `debug` package useful for the conditional logging stuff

kzantow: ... but this is probably better for performance reasons.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/BranchImpl.java
```diff
@@ -16,10 +16,14 @@
...
 
+import static io.jenkins.blueocean.rest.model.KnownCapabilities.BLUE_BRANCH;
+import static io.jenkins.blueocean.rest.model.KnownCapabilities.PULL_REQUEST;
+import static io.jenkins.blueocean.rest.model.KnownCapabilities.JENKINS_WORKFLOW_JOB;
+
 /**
  * @author Vivek Pandey
  */
-@Capability({"io.jenkins.blueocean.rest.model.BlueBranch","org.jenkinsci.plugins.workflow.job.WorkflowJob", "io.jenkins.blueocean.rest.impl.pipeline.PullReuqest"})
+@Capability({BLUE_BRANCH, JENKINS_WORKFLOW_JOB, PULL_REQUEST})
```
kzantow: Not a problem with this PR, but I'd love some standardization guidelines for capabilities.

PR #502, opened 2016-09-15
Task/jenkins 37791 run favorite reordering

cliffmeyers:
# Description
- See [JENKINS-37791](https://issues.jenkins-ci.org/browse/JENKINS-37791).
- Note: the logic that I settled on here is that the "standard sort" (see below) is applied whenever the favorites are loaded, any time a favorite or added or removed, or any time a job finishes (failed, aborted, successful, unstable). If the UI receives an update that a job was queued or is running, the stack does not resort. Navigating away and back will resort the stack.

# Submitter checklist

# Reviewer checklist
   - Note that the "standard order" here is basically red cards first, then blue, gray, then green. Slight oversimplification - refer to the sort helper to see exact logic).
   - Run several favorites and note the list is not sorted - the cards stay in position after being run.
   - Wait for one of the jobs to finish; note that the stack is re-sorted into the standard order.
   - Stopping a running job should also card the cause to sort in standard order.
   - Add or remove some favorites and note that each time the cards are sorted in standard order.
   - Run a pipeline, navigate away, then back: sorted in standard order.
- [x] Reviewed the code


michaelneale:
Seems good to me - worthy of a :bee:. Kudos for tests. 

The functionality seems right (I am sure I missed a combination) - I think @i386 could verify but don't let that block things. 

Would also, as always, be nice to drag ATH over it when in a good state but this is likely not a cross cutting change.

i386:
@cliffmeyers I like this solution BTW 👍

PR #503, opened 2016-09-15
JENKINS-37878: Display url provider

imeredith:
# Description
WIP, probably isnt the best place for this stuff.

See [JENKINS-37878](https://issues.jenkins-ci.org/browse/JENKINS-37878).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanDisplayURLImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanDisplayURLImpl.java
@@ -1,0 +1,85 @@
+package io.jenkins.blueocean;
+
+import hudson.Extension;
+import hudson.model.AbstractBuild;
+import hudson.model.Job;
+import hudson.model.Run;
+import hudson.tasks.test.TestResult;
+import io.jenkins.blueocean.commons.ServiceException;
+import jenkins.branch.MultiBranchProject;
+import jenkins.model.Jenkins;
+import org.jenkinsci.plugins.displayurlapi.DisplayURLProvider;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
+
+/**
+ * Created by ivan on 13/09/16.
+ */
+@Extension
+public class BlueOceanDisplayURLImpl extends DisplayURLProvider {
+    @Override
+    public String getRoot() {
+        Jenkins jenkins = Jenkins.getInstance();
+        if (jenkins == null) {
+            throw new IllegalStateException("Jenkins has not started");
+        }
+        String root = jenkins.getRootUrl();
+        if (root == null) {
+            root = "http://unconfigured-jenkins-location/";
+        }
+        return root + "blue/";
+    }
+
+    @Override
+    public String getRunURL(Run<?, ?> run) {
+        if(run instanceof WorkflowRun) {
+            WorkflowJob job =  ((WorkflowRun) run).getParent();
+            if(job.getParent() instanceof MultiBranchProject) {
+                return getJobURL(((MultiBranchProject) job.getParent()))+ "detail/" +  encode(job.getDisplayName()) + "/" + run.getNumber() + "/";
+            }
+        }
+        if(run.getParent() instanceof Job) {
+            Job job = ((Job) run.getParent());
+            return getJobURL(job) + "detail/" + encode(job.getDisplayName()) + "/" + run.getNumber() + "/";
+        }
+
+        return null;
+    }
+
+    @Override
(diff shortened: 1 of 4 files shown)


imeredith:
@vivek where would the best place for this to go be?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanDisplayURLImpl.java
```diff
@@ -0,0 +1,85 @@
...
+@Extension
+public class BlueOceanDisplayURLImpl extends DisplayURLProvider {
+    @Override
+    public String getRoot() {
+        Jenkins jenkins = Jenkins.getInstance();
+        if (jenkins == null) {
+            throw new IllegalStateException("Jenkins has not started");
+        }
+        String root = jenkins.getRootUrl();
+        if (root == null) {
```
i386: Perhaps we should make upstream DisplayURLProvider do this for you if you call `super.getRoot()`?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanDisplayURLImpl.java
```diff
@@ -0,0 +1,85 @@
...
+        } else {
+            jobPath = encode(project.getFullName());
+        }
+
+        return getRoot() + "organizations/jenkins/pipelines/" + jobPath + "/";
+    }
+
+    private String encode(String path) {
+        try {
+            return URLEncoder.encode(path, "UTF-8");
```
i386: BTW Jenkins uses `Util.encode` for this. Should we be using this instead?

i386:
@imeredith WDYT of just moving this to its own module?

vivek:
@imeredith rest api implementations are not the best place for display url provider implementation to be. Maybe move it in separate plugin or make it part of blueocean-config or blueocean-dashboard?

i386:
How about a new module 'blueocean-services' or something like that? Then again config is probably just fine

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanDisplayURLImpl.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Created by ivan on 13/09/16.
+ */
+@Extension
+public class BlueOceanDisplayURLImpl extends DisplayURLProvider {
+    @Override
+    public String getRoot() {
+        Jenkins jenkins = Jenkins.getInstance();
+        if (jenkins == null) {
```
tfennelly: I don't think this check is required. `Jenkins.getInstance()` is `@Nonnull` return.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanDisplayURLImpl.java
```diff
@@ -0,0 +1,85 @@
...
+            Job job = ((Job) run.getParent());
+            return getJobURL(job) + "detail/" + encode(job.getDisplayName()) + "/" + run.getNumber() + "/";
+        }
+
+        return null;
+    }
+
+    @Override
+    public String getChangesURL(Run<?, ?> run) {
+        return getRunURL(run) + "changes";
```
tfennelly: Looks like `getRunURL` can return `null`?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanDisplayURLImpl.java
```diff
@@ -0,0 +1,85 @@
...
+    private String encode(String path) {
+        try {
+            return URLEncoder.encode(path, "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            throw new ServiceException.UnexpectedErrorException("Error encoding url");
+        }
+    }
+    @Override
+    public String getTestUrl(TestResult result) {
+        return getRunURL(result.getRun()) + "/tests";
```
tfennelly: Ditto earlier wrt `getRunUrl` returning `null`.

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/BlueOceanDisplayURLImplTest.java
```diff
@@ -0,0 +1,193 @@
...
+            file += "echo '"+msg+"'";
+            return this;
+        }
+
+        public String getFileContents() {
+            return file;
+        }
+    }
+
+}
```
tfennelly: Kudos on the tests here man !!

michaelneale:
no please no more modules ;)

i386:
@michaelneale we are using maven 🍘

PR #505, opened 2016-09-18
Add a very simple feature flag system

i386:
# Description

# Submitter checklist
# Reviewer checklist
- [ ] Reviewed the code


i386:
@vivek I wrote a simple feature toggle system so that @sophistifunk can merge the editor into master and have it disabled. It was fun.

blueocean-config/src/test/java/io/jenkins/blueocean/config/FeaturesTest.java
```diff
@@ -0,0 +1,21 @@
+package io.jenkins.blueocean.config;
+
+import com.google.common.collect.ImmutableMap;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class FeaturesTest {
```
i386: @vivek these tests currently fail because JUnit isn't picking up the `features.properties` in the test resources. Any ideas?

kzantow: add a test/resources feature.properties?

i386:
@kzantow yeah I updated with a change. Should be working now

sophistifunk:
Two things:

1) This seems to be entirely server-side? We would need it to be client-side, unless we were disabling the entire plugin, but I don't think there's a way to have a plugin installed-yet-disabled?

2) Last time I spoke to @michaelneale the plan was to extract the create/edit pipeline functionality into a seperate repo, which would make feature flags unnecessary (for the editor specifically, they're probably still a good idea).

i386:
There client side - check window.$blueoceanConfig.features for a map of flags and their values

blueocean-config/src/main/resources/features.properties
```diff
@@ -0,0 +1,10 @@
+# This file defines the enabled features
+# Any of these features can be overriden on the command line via environment variable or java system property.
+#
+# e.g. if you have a feature such as "my.cool.feature=value" you can pass "-Dfeature.my.cool.feature=anothervalue"
+# or set the environment variable "FEATURE_MY_COOL_FEATURE=ANOTHERVALUE"
+#
+# Features are looked up first by environment variable and then by system property. If no value is found, then
+# features will be resolved to the value as specified in this file
+
+pipeline.editor.enabled=false
```
tfennelly: If this is a real property maybe just name it `pipeline.editor`. Surely the `enabled` part is implicit.

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
```diff
@@ -55,14 +56,23 @@ private String createConfig() {
                             .key("allowAnonymousRead").value(allowAnonymousRead)
                             .endObject()
                         .key("enableJWT").value(BlueOceanConfigProperties.BLUEOCEAN_FEATURE_JWT_AUTHENTICATION)
```
vivek: Don't we want to move enableJWT to the same feature set, inside properties file?

blueocean-config/src/main/java/io/jenkins/blueocean/config/Features.java
```diff
@@ -0,0 +1,53 @@
...
+                    properties.load(inputStream);
+                }
+            } catch (IOException e) {
+                throw new IllegalStateException("Could not load " + BLUEOCEAN_FEATURES_PROPERTIES);
+            }
+            // Allow features to be enabled via system property or environment variable
+            return ImmutableMap.copyOf(Maps.transformEntries(Maps.fromProperties(properties), new EntryTransformer<String, String, String>() {
+                @Override
+                public String transformEntry(String key, String value) {
+                    String featureKey = "feature." + key;
```
vivek: @i386 we should probably keep one property name be it coming from env or jam system property. Maybe all capital with underscore (env property) or all small with dots as separator. Otherwise there is always this confusion where the property names in properties file is different from the one coming from ENV?

blueocean-config/src/test/java/io/jenkins/blueocean/config/FeaturesTest.java
```diff
@@ -0,0 +1,21 @@
...
+    @Test
+    public void testLookupFeature() {
+        Features features = new Features();
+        assertEquals("false", features.getFeature("my.cool.feature"));
+    }
+
+    @Test
+    public void testSystemPropertyOverride() {
+        Features features = new Features();
+        System.setProperty("feature.my.cool.feature", "true");
```
vivek: Ah, this is overwriting my.cool.feature property to true and so the other test fails as it finds it true. setProperty() returns existing value (if present) that test could have set in finally block but it won't work since the property name  in internal map is not 'feature.my.cool.feature' but my.cool.feature. I think at least wee should either keep 'feature.' in all places or keep the names same al places.

vivek: @i386 features.properties should have been loaded from main source package correctly. But maybe you want to test custom properties.

i386:
@vivek I think the only thing missing here is that the Booleans get serialized to the JSON as a string.

vivek:
@i386 Maybe but this is not what I see and the way tests are written, one of the test is overriding the property and other test fails. 

Larger point is that it's confusing where for every property defined in properties file, there is going to be corresponding JVM system property with 'feature.' prefix and then there is capitalized version with '_' for env property. Can we keep just one variable name for all?

PR #508, opened 2016-09-19
[FIX JENKINS-38345] Wrong property to reset loading indicator classes

kzantow:
# Description

See [JENKINS-38345](https://issues.jenkins-ci.org/browse/JENKINS-38345).

To test: use Safari, get the loading bar to show up, validate there is no javascript error when the loading bar disappears.

This is a simpler fix than https://github.com/jenkinsci/blueocean-plugin/pull/507

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/LoadingIndicator.js
+++ b/blueocean-dashboard/src/main/js/LoadingIndicator.js
@@ -49,7 +49,7 @@ export default {
             clearTimeouts();
             setLoaderClass('complete', 10);
             timeouts.push(setTimeout(() => {
-                document.getElementById('loadbar').classList = '';
+                document.getElementById('loadbar').className = '';
             }, 500));
         }
     },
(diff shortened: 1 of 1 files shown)


scherler:
I am not fan of your solution http://stackoverflow.com/questions/15040297/clear-element-classlist 

> I recommend not using className as classList could result in faster DOM updates.

kzantow:
@scherler completely disagree, don't solve performance problems that don't exist. This is an infrequent change, performance isn't a concern, less code is better for maintenance.

scherler:
see the edit.

scherler:
Further @kzantow I do not understand why you would oben a new PR instead of commenting it in the other

kzantow:
@scheler I did comment on the other PR, figured you were busy and it's a simple enough change

PR #509, opened 2016-09-19
docs for shrinkwrap

cliffmeyers:
# Description
- Just some docs for working with npm and shrinkwrap. Figured some feedback would be good before merging to master.


cliffmeyers:
cc @scherler @tfennelly @sophistifunk @imeredith @michaelneale @vivek you might want to peek at the updates to the docs here; should hopefully be concise instructions for life in the post-shrinkwrap era.

PR #511, opened 2016-09-20
Task/jenkins 37715 validate package shrinkwrap jsons

cliffmeyers:
# Description
- Code is final yet but curious for some feedback on the approach taken here.
- Experiment with validating the package.json and npm-shrinkwrap.json  (similar to jenkinsci/jenkins-design-language#96  and calling it from the Jenkinsfile so we can fail the CI build if they're out of sync.
- See [JENKINS-37715](https://issues.jenkins-ci.org/browse/JENKINS-37715).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


validatedeps.js
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env node
```
michaelneale: can this go in checkdeps.js - as separate function? seems to make more sense than add another file at the root

cliffmeyers: Yes, I can combine them and enhance checkdeps a little. You think the command args should be a list of projects to validate?

./validatedeps.js blueocean-dashboard blueocean-personalization weird-folder/oddly-nested-project-for-some-odd-reason ?

michaelneale: yes I think so - optional args passed to it perhaps? 

Perhaps @sophistifunk can suggest how to do it, given he first created checkdeps.js IIRC

validatedeps.js
```diff
@@ -0,0 +1,106 @@
...
+}
+
+const workingPath = process.cwd();
+const resolvedPath = buildPath(`${workingPath}/${projectPath}`);
+console.log(`validating dependencies in ${resolvedPath}`);
+const packageJsonPath = buildPath(`${resolvedPath}/package.json`);
+const shrinkwrapJsonPath = buildPath(`${resolvedPath}/npm-shrinkwrap.json`);
+
+// match x, tilde, gt, lt, star, carat or whitespace in version
+const IMPRECISE_VERSION_CHARS_PATTERN = /[x~><*|\^\s]+/;
```
sophistifunk: Why don't we make sure we have `^\d+\.\d+\.\d+$` instead of trying to find imprecise ones? It looks like this wouldn't flag a version of "latest" or "3" as bad?

validatedeps.js
sophistifunk: Checkdeps is pretty ungainly code and hard enough to follow as it is, probably better to keep them separate since they do separate things. We might want to rename them to make it clearer what they do, since they're both "checking" and "validating" dependencies.

michaelneale: ok - thats fine, just don't like a proliferation of files at the root - bin dir can be used for odds and ends and scripts

sophistifunk: "pull requests welcome"

/me *hides*

validatedeps.js
tfennelly: Any reason not to use the [semver](https://www.npmjs.com/package/semver) package?

sophistifunk: It's my understanding the whole reason for this (and the shrinkwrapping) is that semver is a bedtime story.

validatedeps.js
cliffmeyers: Relocated both scripts to "bin" in 19d8d58, b85f367 and bd0b6d1

validatedeps.js
cliffmeyers: @tfennelly I wanted to keep this as a standalone script; using the `semver` package would require a package.json and an install process. Trying to avoid making this more complex than it needs to be.

@sophistifunk that pattern would fail the "-beta1" version format we use for `@jenkins-cd` libs which would cause our PR branch builds to fail. I can certainly write a regex for that pattern too.

cliffmeyers: Fixed in 7e057f7

cliffmeyers:
This one should be ready for merge now I think. cc @michaelneale or @sophistifunk if you'd be so kind as to add a bee / approve.

validatedeps.js
sophistifunk: That makes sense to me.

sophistifunk:
bee

michaelneale:
:bee: - nice work.

PR #512, opened 2016-09-21
[JENKINS-38252] Updated SSE Gateway plugin to 1.10

tfennelly:
# Description

Depends on https://github.com/jenkinsci/sse-gateway-plugin/pull/11

See [JENKINS-38252](https://issues.jenkins-ci.org/browse/JENKINS-38252).

Test would be to just make sure e.g. the run button pops up the STARTED toast. Other than that, we need to see if SSE config still appears in the list of blocked requests when BO comes under load.

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -198,7 +198,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>sse-gateway</artifactId>
-            <version>1.9</version>
+            <version>1.10-SNAPSHOT</version>
         </dependency>
 
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -198,7 +198,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>sse-gateway</artifactId>
-            <version>1.10-SNAPSHOT</version>
+            <version>1.10</version>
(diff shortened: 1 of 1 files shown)


michaelneale:
OK after much pain (unrelated to this) did get to try it - but wasn't able to see the OPEN popup with a multibranch (I ran the "passing" branch of the kzantow/failure-project multibranch project).

tfennelly:
@michaelneale Great catch. Thanks a lot. [This fixes it for me](https://github.com/jenkinsci/sse-gateway-plugin/pull/11/commits/c5e5e5772e9951b05891bcbac87fae8e7e523475).

The issue was that because I made the config async it was then happening outside the context of a request and so had no user authentication other than anonymous. Then, the run events were being filtered out (not pushed to the user from the pubsub layer) because of permissions. I suppose it proves that the pubsub permission checks work right :)

michaelneale:
OK this seems nice - I tried with anon and non anon, kicking things off - seems good. Tried with multibranch too. 

Not all combinations (perhasp ATH could inform) but I think this is good. 

@tfennelly worth publishing the SSE change and changing this to be non snapshot? tentative :bee: from me - nice work.

PR #513, opened 2016-09-21
[JENKINS-37141] rework Docker build for development image - take 2

ydubreuil:
See [JENKINS-37141](https://issues.jenkins-ci.org/browse/JENKINS-37141).

I just discovered `hpi:assemble-dependencies` ( @tfennelly gave me the hint) , which is exactly what is
needed to bundle all plugins into a folder which can be used for the Docker build. A dedicated module is not needed at all (it was required by `hpi:bundle-plugins`). It's a bit unfortunate to discover
this now, but better late than never..

This PR just roll backs adding a new module and takes advantage of `hpi:assemble-dependencies`. This could also be used by ATH test suite runner script to avoid running `mvn install` before calling it.

One question left for the `Dockerfile`, do we really want to depend on `jenkinsci/jenkins:latest` ?. It seems better to depend on a fixed version. Personally, I would use the Jenkins version declared in `pom.xml` to be consistent.


tfennelly:
🐝 LGTM

@ydubreuil not sure what you mean by "this could also be used by ATH test suite runner ...". This is exactly what the ATH is using and why I suggested it to you. Am I missing something?

ydubreuil:
I meant that with this, it is not necessary to run `hpi:assemble-dependencies` again before launching ATH tests. We could simplify `run.sh` script to just point to the `plugins` folder generated in BlueOcean. With the added benefit that `mvn install` on BlueOcean source would not be a prerequisite to run ATH, `mvn package` would be enough.

michaelneale:
I think depend on a fixed version (ideally what we put in the pom.xml as the jenkins version - just keep it in sync). 

@tfennelly does this change how the ATH has to run?

.dockerignore
```diff
@@ -1,8 +1,3 @@
 *
-!blueocean-commons/target/blueocean-commons.hpi
-!blueocean-dashboard/target/blueocean-dashboard.hpi
-!blueocean-plugin/target/blueocean.hpi
-!blueocean-rest/target/blueocean-rest.hpi
-!blueocean-rest-impl/target/blueocean-rest-impl.hpi
-!blueocean-web/target/blueocean-web.hpi
+!blueocean/target/plugins
```
michaelneale: how come this change?

Dockerfile
```diff
@@ -1,33 +1,17 @@
 FROM jenkinsci/jenkins:latest
```
michaelneale: yeah better to use a specific version IMO

michaelneale: and the jenkinsci one - not the "official" one? (ie the one in the root namespace?)

michaelneale:
I think this looks good, and @tfennelly agrees - had a few questions (in coments) but a :bee: from me. Better that it isn't a submodule I think

.dockerignore
SalahAdDin: Where is these folder? Have we create it?

ydubreuil: @michaelneale now that dependencies are assembled in the aggregator module, it's not needed to list them here

Dockerfile
ydubreuil: yeah, I had this in mind too

ydubreuil: @michaelneale dogfood is using this image, basing the image on the LTS would install an older Jenkins version on dogfood. I'll keep it as is for now.

PR #516, opened 2016-09-21
feature/JENKINS-38321-js-extensions

sophistifunk:
# Description

Updates ExtensionRenderer to show children as a default when no extensions are present for the specified ExtensionPoint. 

Also includes a small fix to checkdeps to make it a little more strict, but only when checking dependencies on jenkins-cd projects

*NB* This is not yet published to npm (hence the version bump and -beta)

See [JENKINS-38321](https://issues.jenkins-ci.org/browse/JENKINS-38321).

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

 * Update ExtensionRenderer to show children as placeholders if no extensions are present
 * Tests for ExtensionRenderer, and a few tweaks for testability because JS
 * Improvements to checkdeps script backported from editor branch

### DIFF
--- a/js-extensions/spec/ExtensionRenderer-spec.js
+++ b/js-extensions/spec/ExtensionRenderer-spec.js
@@ -1,0 +1,110 @@
+'use strict';
+
+const React = require('react');
+const jsTest = require('@jenkins-cd/js-test');
+const assert = require('chai').assert;
+const mount = require('enzyme').mount;
+const jsdom = require('jsdom');
+const ExtensionRenderer = require('../dist/ExtensionRenderer.js').ExtensionRenderer;
+const extensionStoreInstance = require('../dist/ExtensionStore.js').instance;
+
+const $ = React.createElement;
+
+function mockExtension(props) {
+    return (
+        $('h1', {}, 'Extension is a H1')
+    );
+}
+
+const extensionPointMetadata = {
+    ep1: [
+        mockExtension
+    ]
+};
+
+const mockExtensionStore = {
+    getExtensions: function(extensionPoint, filters, onLoad) {
+        onLoad = onLoad || filters;
+        onLoad(extensionPointMetadata[extensionPoint] || []);
+    }
+};
+
+const mockResourceLoadTracker = {
+    onMount: function() {
+        // Don't care
+    },
+
+    onUnmount: function() {
+        // Don't care
+    }
+};
+
+describe('ExtensionRenderer', function () {
+
+    var oldDocument;
+    var oldWindow;
+
+    var oldExtensionStore;
+    var oldResourceLoadTracker;
+
+    before(function() {
+
+        oldDocument = global.document;
+        oldWindow = global.window;
+
+        const doc = jsdom.jsdom('<!doctype html><html><body></body></html>');
+        global.document = doc;
+        global.window = doc.defaultView;
+    });
+
+    beforeEach(function () {
+        oldExtensionStore = ExtensionRenderer.ExtensionStore;
+        oldResourceLoadTracker = ExtensionRenderer.ResourceLoadTracker;
+
+        ExtensionRenderer.ExtensionStore = mockExtensionStore;
+        ExtensionRenderer.ResourceLoadTracker = mockResourceLoadTracker;
+    });
+
+    afterEach(function () {
+        ExtensionRenderer.ExtensionStore = oldExtensionStore;
+        ExtensionRenderer.ResourceLoadTracker = oldResourceLoadTracker;
+    });
+
+    after(function () {
+        global.document = oldDocument;
+        global.window = oldWindow;
+    });
+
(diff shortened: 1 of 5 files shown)


michaelneale:
I assume @kzantow and probably @cliffmeyers should have a review of this? 

kudos on adding the tests ;)

kzantow:
... argh, can't edit reviews!? Meant to put: "Could you _explain_ the use case somewhere" :)

sophistifunk:
The use case I have for this is the "New Pipeline" button:

```
<Extensions.Renderer extensionPoint="jenkins.pipeline.create.action">
    <a target="_blank" className="btn-secondary inverse" href={newJobUrl}>
        New Pipeline
    </a>
</Extensions.Renderer>
```

I'm not sure I can see a reason we'd want to stop people from using extentionPoints to replace things, but I'm all ears. If we don't provide a nice way to do it, people can just force-hide things with CSS.

js-extensions/spec/ExtensionRenderer-spec.js
```diff
@@ -0,0 +1,110 @@
...
+const mount = require('enzyme').mount;
+const jsdom = require('jsdom');
+const ExtensionRenderer = require('../dist/ExtensionRenderer.js').ExtensionRenderer;
+const extensionStoreInstance = require('../dist/ExtensionStore.js').instance;
+
+const $ = React.createElement;
+
+function mockExtension(props) {
+    return (
+        $('h1', {}, 'Extension is a H1')
```
tfennelly: Oh man ... I got so excited there. For a brief moment, I thought we were using jQuery 🚬

js-extensions/src/ExtensionRenderer.jsx
```diff
@@ -1,7 +1,7 @@
 var React = require('react');
 var ReactDOM = require('react-dom');
-var ExtensionStore = require('./ExtensionStore.js').instance;
-var ResourceLoadTracker = require('./ResourceLoadTracker').instance;
+var importedExtensionStore = require('./ExtensionStore.js').instance;
+var importedResourceLoadTracker = require('./ResourceLoadTracker').instance;
```
tfennelly: Nothing to do with this PR, but I vomit into my mouth every time I see that `.instance` thing on those `js-extensions` types. Just say'n.

michaelneale:
C.R.E.A.M 

BTW ignore the job report from "continuous-integration/jenkins/pr-head" that was a wayward PR deployer attempt.

scherler:
@michaelneale https://www.youtube.com/watch?v=PBwAxmrE194 :)

michaelneale:
@scherler that is the way to @sophistifunk's heart.

scherler:
C.R.E.A.M :laughing:

js-extensions/src/ExtensionRenderer.jsx
kzantow: ... that's fair, though conversely I feel ill every time I see things instantiated in JS modules and exported essentially as static references. At least this way, you can get a reference to the class for testing/etc. All ears for a better solution that doesn't look like worst practice Java code from 1995.

cliffmeyers: I've been thinking about this too. I wonder if we could shift toward individual files simply exporting a func or class, and letting another file ("index"?) be responsible for creating instances and wiring things together. This would give us some flexibility in tests to set things up with mocks. The Capability API in blueocean-core-js does this, and it really helped from a testability perspective.

tfennelly: > ... conversely I feel ill every time I see things instantiated in JS modules and exported essentially as static references.

In what code is anyone using `js-modules` directly? Just curious.

> At least this way, you can get a reference to the class for testing/etc. All ears for a better solution that doesn't look like worst practice Java code from 1995.

Everything I see in `ExtensionStore` is effectively static, so I don't get why most of the functions on it are not static and so used ala `ExtensionStore.get(blah)`, as opposed to `ExtensionStore.instance.get(blah)`. They can easily be duck-typed/mocked for testing ... couldn't be easier than with JavaScript (and there are libs for helping).

Anyway ... prob a topic for another day.

kzantow: yes, no more on this, just a clarification: _common JS_ modules, not `js-modules`.

js-extensions/spec/ExtensionRenderer-spec.js
sophistifunk: Nope, just avoiding a complete build-process rewrite :D

sophistifunk:
Wu Tang references I'm your man, but Harry Potter jokes I don't even get. I was so disappointed by that movie. From all the hype I thought it was going to be something awesome like Labyrinth, but instead it was dark and disturbing, like those Police Academy movies.

PR #517, opened 2016-09-22
Topic/display url review

i386:
# Description

Builds on top of #503.
* We needed to split the provider impl into its own plugin (did not belong to rest or config)
* display-url-provider 0.5 simplified calculating the base URL
* Never returns null
* Since multiple modules depend on pipeline and SCM APIs their maven module versions are defined in the root POM, as is maven best practice.
* Bumped github branch source to 2.10 as it has support for display url provider 💅 

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
So the aim of this is to provide URLs for notifications that support it to link to blue ocean right? 

I have a few problems with this approach: 

1) its a big big change for a small but important thing
2) it shouldn't be on by default, ie people installing blue ocean shouldn't have URLs changed just yet (maybe via a switch?)
3) it probably belongs in blueocean-pipeline-api-impl - if it has a dependency on pipeline
4) could it be deployed separately? 

Modules are like puppies, modules that are deployed as plugins, doubly so.

blueocean/pom.xml
```diff
@@ -9,7 +9,7 @@
     </parent>
 
     <artifactId>blueocean</artifactId>
-    <packaging>hpi</packaging>
+    <packaging>pom</packaging>
```
michaelneale: ???

i386: oh thanks IDEA

michaelneale: Thanks Obama

blueocean-testing/pom.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
michaelneale: IF we really want a discrete testing utils module - probably best done separately as a PR

blueocean-displayurl-provider/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+          <scope>test</scope>
+      </dependency>
+      <dependency>
+          <groupId>org.jenkins-ci.plugins</groupId>
+          <artifactId>git</artifactId>
+          <scope>test</scope>
+      </dependency>
+      <dependency>
+          <groupId>${project.groupId}</groupId>
+          <artifactId>blueocean-testing</artifactId>
```
michaelneale: shouldn't this be test scope?

blueocean-pipeline-api-impl/pom.xml
```diff
@@ -19,95 +19,86 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>blueocean-rest-impl</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>blueocean-testing</artifactId>
```
michaelneale: also - scope?

blueocean-pipeline-api-impl/pom.xml
```diff
@@ -19,95 +19,86 @@
...
+        </dependency>
+
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.6</version>
```
michaelneale: any reason why these are being dropped as part of this PR?

michaelneale:
lets park this until ivan is back. I think this should be in another module (and these other refactorings done separately)

michaelneale:
@imeredith can we close this? (given you are working on it another way)

michaelneale:
@imeredith we can chop this now right?

michaelneale:
closing as not needed

PR #519, opened 2016-09-22
Jenkins can abort but take a while for state to be finalised, so no u…

michaelneale:
# Description

A build can be aborted but still be running for a bit. It seems pointless to test the running status. 

cc @cliffmeyers  (just FYI in case you didn't know this fact). 

This test line frequently breaks various builds for no good reason (and nothing to do with blue ocean).

@reviewbybees


i386:
🐝

PR #520, opened 2016-09-22
[FIX JENKINS-37753] REGRESSION: Steps showing up as incomplete when they are in fact complete

scherler:
# Description

See [JENKINS-37753](https://issues.jenkins-ci.org/browse/JENKINS-37753).

# Submitter checklist
https://github.com/jenkinsci/blueocean-acceptance-test/pull/43 -> not tested due to problems with updatecenter mirror in spain, @recena is looking into it. -> NOW tested since update center is back

# Reviewer checklist
- [x] Reviewed the code

…inished with the run, this makes sure that they will be refetched after the run ended.

### DIFF
--- a/blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
+++ b/blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
@@ -225,7 +225,7 @@ export class RunDetailsPipeline extends Component {
      * @returns {boolean} true when parallel, false otherwise
      */
     isParallel(nodeInfo) {
-        // in case we have edges arrays we compare the first edge, if not we know we are not in parallel modeq
+        // in case we have edges arrays we compare the first edge, if not we know we are not in parallel mode
         return this.mergedConfig.nodeReducer.edges[0] && nodeInfo.edges[0] ?
         this.mergedConfig.nodeReducer.edges[0].id === nodeInfo.edges[0].id : false;
     }
--- a/blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
+++ b/blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
@@ -139,7 +139,7 @@ export class RunDetailsPipeline extends Component {
     // Listen for pipeline flow node events.
     // We filter them only for steps and the end event all other we let pass
     _onSseEvent(event) {
-        const { fetchNodes, fetchSteps } = this.props;
+        const { fetchNodes, fetchSteps, removeStep, removeLogs } = this.props;
         const jenkinsEvent = event.jenkins_event;
         // we are using try/catch to throw an early out error
         try {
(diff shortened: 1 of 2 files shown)


recena:
The [mirror](http://mirrors.clinkerhq.com/jenkins/) is back again.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
```diff
@@ -176,6 +176,31 @@ export class RunDetailsPipeline extends Component {
...
+                    if (keyArray[keyArray.length - 1] === 'nodes') {
+                        keyArray.pop();
+                    }
+                    const searchString = keyArray.join('/');
+                    // fire to remove all logs from cache
+                    removeLogs(searchString);
+                    /** we have to check now if all steps have been declared as finished,
+                     * if not we will remove them from the cache so we get the final information*/
+                    const notFinishedSteps = Object.keys(this.props.steps)
+                        .filter((item) => item.indexOf(searchString) !== -1)
```
kzantow: NIT: isn't this condition usually written as `>= 0` ? or should I adopt a different standard?

scherler: @kzantow talk to @tfennelly about that. 

I did it in another PR like you said and then @tfennelly ask the same question as you but the other way around. I personally do not care, but having this NIT in all PR is getting old. ;)

kzantow: Too funny. :) Not a reason to hold it up at all, obviously.

kzantow: Mozilla examples refer to this method, I suppose I should try to adapt my brain: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf

blueocean-dashboard/src/main/js/redux/actions.js
```diff
@@ -140,12 +142,24 @@ export const actionHandlers = {
         steps[payload.nodesBaseUrl] = payload;
         return state.set('steps', steps);
     },
+    [ACTION_TYPES.REMOVE_STEP](state, { payload }): State {
```
kzantow: :ant: `payload` is entirely nondescript as to what the data contains

michaelneale:
lint failures now...

blueocean-dashboard/src/main/js/redux/actions.js
scherler: Well that is still in my head from my last job. We used it always as "envelope" for the data. This way we had only few attributes in our commands/events, however there is no reason to keep doing it I guess.

I took care of it.

sophistifunk:
Goddammit, is this the same as https://github.com/jenkinsci/blueocean-plugin/pull/521

scherler:
@sophistifunk not really. here we making sure that the steps and logs are getting cleaned when the run has finished

sophistifunk:
I'll see myself out :)

scherler:
@sophistifunk but an excellent way to catch attention to your PR though ;) jeje

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
michaelneale: huh

michaelneale:
Looks like Keith as used the github approval, which is as good as a :bee: ;)

PR #521, opened 2016-09-23
Fix: Unbuilt stages of a run are showing the wrong indicator

sophistifunk:
# Description

Fix to show null node results as "not built" instead of "queued" on completed runs.

See [JENKINS-38247](https://issues.jenkins-ci.org/browse/JENKINS-38247).

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

    josh/JENKINS-38247-notbuilt-indicator * Lint
    josh/JENKINS-38247-notbuilt-indicator * Fix the bug when watching a live build transition to failed, and add
    josh/JENKINS-38247-notbuilt-indicator * Fixed when loading, but not yet when watching a build go from running -> finished

### DIFF
--- a/blueocean-dashboard/src/main/js/components/PipelineRunGraph.jsx
+++ b/blueocean-dashboard/src/main/js/components/PipelineRunGraph.jsx
@@ -10,7 +10,7 @@ function badNode(jenkinsNode) {
     return new Error('convertJenkinsNodeDetails: malformed / missing Jenkins run node.');
 }
 
-function convertJenkinsNodeDetails(jenkinsNode) {
+function convertJenkinsNodeDetails(jenkinsNode, isCompleted) {
     if (!jenkinsNode
         || !jenkinsNode.displayName
         || !jenkinsNode.id) {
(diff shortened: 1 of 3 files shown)


sophistifunk:
screen shot 2016-09-23 at 3 45 02 pm

i386:
@sophistifunk beauty mate 👍

michaelneale:
cc @cliffmeyers

blueocean-dashboard/src/main/js/components/PipelineRunGraph.jsx
```diff
@@ -173,7 +182,7 @@ export default class PipelineRunGraph extends Component {
 PipelineRunGraph.propTypes = {
     pipelineName: string,
     branchName: string,
-    runId: string,
```
scherler: :ant: Then you should remove the attribute in RunDetailsPipeline.jsx as well in the Extensions.Renderer

michaelneale:
:bee: as seems to work well - although @scherler has a comment about removing the runId which may need addressing. Nice.

blueocean-dashboard/src/main/js/components/PipelineRunGraph.jsx
scherler: I take the point which leads to "contract properties" passed to certain ExtensionPoints. ATM we do it on free guess and all properties are passed through anyway. 

However we do not have API style declarations of extension points and the attributes we pass through. Which makes it hard for adopters to create reliable alternatives for our default implementations. 

I say ATM the graph is the only use case that we know of that had used run.id in a certain point of time. For housekeeping I prefer to get rid of it since it is redundant now because we pass full run.

As started I see your point, but see as well the urgent need to declare ExtensionPoints as API where we can deprecate certain attributes to later drop them. ...and to document them to start with,

scherler: I created https://issues.jenkins-ci.org/browse/JENKINS-38482 and assigned to @michaelneale to get the ball rolling. ...„Der Ball muß ins Eckige“. ;)

PR #523, opened 2016-09-23
Dockerfile usable for dogfood

ydubreuil:
This is a follow-up to [JENKINS-37141](https://issues.jenkins-ci.org/browse/JENKINS-37141).

Depends on a fixed Jenkins LTS version instead of latest to ensure reproducible builds.
Don't bypass installation wizard by default as it's here to enforce security.

With this, Docker image built from master branch should be usable for dogfooding, which would make https://github.com/cloudbees/blueocean-docker-image obsolete 

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,3 @@
 *
 !blueocean/target/plugins
-!docker-demo/
+!docker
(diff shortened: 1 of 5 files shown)


michaelneale:
@ydubreuil I wonder why this didn't trigger the normal multibranch build?

michaelneale:
this looks like a tricky one to test - how do we verify it is ok?

ydubreuil:
I did trigger the normal builder, but there was a GitHub request limit issue on CI yesterday, I had to kill the build.

Testing will be possible with the auto-deploy environment, it's nearly ready.

ydubreuil:
Replaced with #527

PR #526, opened 2016-09-26
Update to latest js-extensions, and add an ExtensionPoint for the "New Pipeline" button

sophistifunk:
# Description

See [JENKINS-38321](https://issues.jenkins-ci.org/browse/JENKINS-38321).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

JENKINS-38321-dashboard-new-button * Update dependency on js-extensions to latest, add an ExtensionPoint for the 'new pipeline' action with default of old behaviour

### DIFF
--- a/blueocean-dashboard/src/main/js/components/Pipelines.jsx
+++ b/blueocean-dashboard/src/main/js/components/Pipelines.jsx
@@ -40,9 +40,11 @@ export default class Pipelines extends Component {
                             { organization && ' / ' }
                             { organization && orgLink }
                         </h1>
-                        <a target="_blank" className="btn-secondary inverse" href={newJobUrl}>
-                            New Pipeline
-                        </a>
+                        <Extensions.Renderer extensionPoint="jenkins.pipeline.create.action">
+                            <a target="_blank" className="btn-secondary inverse" href={newJobUrl}>
+                                New Pipeline
+                            </a>
+                        </Extensions.Renderer>
                     </Title>
                 </PageHeader>
                 <main>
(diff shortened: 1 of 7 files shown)


blueocean-dashboard/src/main/js/components/Pipelines.jsx
```diff
@@ -40,9 +40,11 @@ export default class Pipelines extends Component {
                             { organization && ' / ' }
                             { organization && orgLink }
                         </h1>
-                        <a target="_blank" className="btn-secondary inverse" href={newJobUrl}>
-                            New Pipeline
-                        </a>
+                        <Extensions.Renderer extensionPoint="jenkins.pipeline.create.action">
```
michaelneale: neat - what happens when there is an extension - it overrides what is in there?

sophistifunk: Yup. If we need it to be additional rather than replacement we can at some point, just means a bit more complexity keep track of what's a plugin and what's not.

michaelneale: right - seems reasonably. The 'additive' ness would happen if there were multiple things for the one extension point (ie you could put the "built in" behavior as an extension if you wanted it to be additive, right?)

sophistifunk: Zakkary

michaelneale:
:bee: Looks good. Tried it locally (updated with master just to be sure) and seems good to me.

PR #528, opened 2016-09-26
[JENKINS-38329] REGRESSION Errors on activity and branches tabs

scherler:
# Description

See [JENKINS-38329](https://issues.jenkins-ci.org/browse/JENKINS-38329).

Prevent that the error occurs however ATM debugging the issue with keith.

For the AT of this test I need back the folder.js which is atm commented

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

…ting them otherwise we will run into failures the next time. Further in general we should never try to fetch something where we know before hand that it does not exit.

### DIFF
--- a/blueocean-dashboard/src/main/js/components/MultiBranch.jsx
+++ b/blueocean-dashboard/src/main/js/components/MultiBranch.jsx
@@ -59,6 +59,11 @@ export class MultiBranch extends Component {
         }
     }
 
+    componentWillUnmount() {
+       this.props.clearBranchData()
+    }
+
+
     render() {
         const { branches } = this.props;
 
@@ -121,6 +126,7 @@ MultiBranch.contextTypes = {
 MultiBranch.propTypes = {
     branches: array,
     fetchBranches: func,
+    clearBranchData:func,
     children: any,
 };
 
--- a/blueocean-dashboard/src/main/js/components/MultiBranch.jsx
+++ b/blueocean-dashboard/src/main/js/components/MultiBranch.jsx
@@ -60,7 +60,7 @@ export class MultiBranch extends Component {
     }
 
     componentWillUnmount() {
-       this.props.clearBranchData()
+        this.props.clearBranchData();
     }
 
 
@@ -126,7 +126,7 @@ MultiBranch.contextTypes = {
 MultiBranch.propTypes = {
     branches: array,
     fetchBranches: func,
-    clearBranchData:func,
+    clearBranchData: func,
     children: any,
 };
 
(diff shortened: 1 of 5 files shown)


blueocean-dashboard/src/main/js/components/MultiBranch.jsx
```diff
@@ -51,7 +51,7 @@ EmptyState.propTypes = {
 
 export class MultiBranch extends Component {
     componentWillMount() {
-        if (this.context.config && this.context.params) {
+        if (this.context.pipeline && this.context.params && !pipelineBranchesUnsupported(this.context.pipeline)) {
```
kzantow: I don't think this will work

blueocean-dashboard/src/main/js/components/PullRequests.jsx
```diff
@@ -48,7 +48,7 @@ EmptyState.propTypes = {
 
 export class PullRequests extends Component {
     componentWillMount() {
-        if (this.context.config && this.context.params) {
+        if (this.context.pipeline && this.context.params && !pipelineBranchesUnsupported(this.context.pipeline)) {
```
kzantow: ... or this, at *mount* time, data is likely to be fetched and is still `$pending`

blueocean-dashboard/src/main/js/components/MultiBranch.jsx
scherler: @kzantow it works fine, why do you thing it would not?

furthe in PipelinePage the parent we have 

```
        if (pipeline && pipeline.$failed) {
            return <NotFound />;
        }
```

which prevents that from failing

blueocean-dashboard/src/main/js/components/PullRequests.jsx
scherler: we do not fetch this.context.pipeline and willMount is only called once.

blueocean-dashboard/src/main/js/components/MultiBranch.jsx
kzantow: You need `$pending` or the check is useless

blueocean-dashboard/src/main/js/components/PullRequests.jsx
kzantow: ... same as above

blueocean-dashboard/src/main/js/components/MultiBranch.jsx
scherler: not following you

scherler: since in the PP we do 

```
                {isReady && React.cloneElement(this.props.children, { pipeline })}
```

kzantow: ah right 👍

PR #535, opened 2016-09-28
Update Jenkinsfile

michaelneale:
# Description

This should set the build retention to a more reasonable number. 

cc @tfennelly @ydubreuil - look ok? 

@reviewbybees


imeredith:
LGTM if it builds :bee:

michaelneale:
@imeredith yeah it builds - am just running rebuilds to experiment around the npm problem

PR #536, opened 2016-09-28
hotfix/die-shrinkwrap-die

cliffmeyers:
# Description
- DO NOT MERGE
- Removing these files to see if it stabilizes the build in CI


michaelneale:
@cliffmeyers dang it, I can't seem to see it happening on other branches. This one seems good - but perhaps you could try it at different times of day - see if this branch passes where others fail?

michaelneale:
Closing this as we have a better approach and easy enough to bring back

PR #538, opened 2016-09-28
Do check of deps before the main act (faster fail)

michaelneale:
Quick change to do check deps before the build 

cc @scherler @tfennelly @reviewbybees


michaelneale:
that was dumb. Lol

PR #539, opened 2016-09-28
switch around shrinkwrap check

michaelneale:
cc @scherler @reviewbybees

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -10,9 +10,9 @@ node {
   docker.image('cloudbees/java-build-tools').inside {
     withEnv(['GIT_COMMITTER_EMAIL=me@hatescake.com','GIT_COMMITTER_NAME=Hates','GIT_AUTHOR_NAME=Cake','GIT_AUTHOR_EMAIL=hates@cake.com']) {
       try {
+        sh "node ./bin/checkshrinkwrap.js"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -10,9 +10,9 @@ node {
   docker.image('cloudbees/java-build-tools').inside {
     withEnv(['GIT_COMMITTER_EMAIL=me@hatescake.com','GIT_COMMITTER_NAME=Hates','GIT_AUTHOR_NAME=Cake','GIT_AUTHOR_EMAIL=hates@cake.com']) {
       try {
+        sh "node ./bin/checkshrinkwrap.js"
         sh "mvn clean install -B -DcleanNode -Dmaven.test.failure.ignore"
         sh "node ./bin/checkdeps.js"
```
scherler: line 15 should be one line up as well

michaelneale: @scherler no - that actually breaks as there is no node_modules yet - it has to run after the build

michaelneale: ie you will see: 

```
[checkdeps_before_build] Running shell script
+ node ./bin/checkdeps.js
module.js:327
    throw err;
    ^

Error: Cannot find module '../blueocean-dashboard/node_modules/@jenkins-cd/design-language/package.json'
```

michaelneale:
Best to run after it seems

PR #540, opened 2016-09-28
Speed up Maven downloads by using a proxy and increasing DL thread pool size

ydubreuil:
Use a configuration file defined on CI instance to access a cache server to speed up builds.

### DIFF
--- a/blueocean-personalization/package.json
+++ b/blueocean-personalization/package.json
@@ -9,7 +9,8 @@
     "test": "gulp test",
     "test:watch": "gulp test:watch",
     "bundle": "gulp bundle",
-    "bundle:watch": "gulp bundle:watch"
+    "bundle:watch": "gulp bundle:watch",
+    "postinstall": "node ../bin/cleanshrink.js"
   },
   "devDependencies": {
     "@jenkins-cd/eslint-config-jenkins": "0.0.2",
--- a/blueocean-web/package.json
+++ b/blueocean-web/package.json
@@ -8,7 +8,8 @@
     "test": "gulp test",
     "test:watch": "gulp test:watch",
     "bundle": "gulp bundle",
-    "bundle:watch": "gulp bundle:watch"
+    "bundle:watch": "gulp bundle:watch",
+    "postinstall": "node ../bin/cleanshrink.js"
   },
   "devDependencies": {
     "@jenkins-cd/js-builder": "0.0.40",
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,10 +2,13 @@ node {
   deleteDir()
   checkout scm
   configFileProvider([configFile(fileId: 'blueocean-maven-settings', targetLocation: 'settings.xml')]) {
+  configFileProvider([configFile(fileId: 'blueocean-npm-settings', targetLocation: '.npmrc')]) {
 
   docker.image('cloudbees/java-build-tools').inside {
     withEnv(['GIT_COMMITTER_EMAIL=me@hatescake.com','GIT_COMMITTER_NAME=Hates','GIT_AUTHOR_NAME=Cake','GIT_AUTHOR_EMAIL=hates@cake.com']) {
       try {
+        sh 'cp .npmrc $HOME/.npmrc; cat .npmrc'
+        sh '''REGISTRY="$(cat .npmrc | sed -n 's/^registry=\\(.*\\)$/\\1/p')"; sed -i "s|https://registry.npmjs.org|$REGISTRY|" */npm-shrinkwrap.json'''
         sh "mvn clean install -B -DcleanNode -Dmaven.test.failure.ignore -s settings.xml -Dmaven.artifact.threads=30"
         sh "node checkdeps.js"
         step([$class: 'JUnitResultArchiver', testResults: '**/target/surefire-reports/TEST-*.xml'])
@@ -19,7 +22,8 @@ node {
     }
   }
 
-  }
+  } // configFileProvider
+  } // configFileProvider
 }
 
 def sendhipchat() {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -8,7 +8,7 @@ node {
     withEnv(['GIT_COMMITTER_EMAIL=me@hatescake.com','GIT_COMMITTER_NAME=Hates','GIT_AUTHOR_NAME=Cake','GIT_AUTHOR_EMAIL=hates@cake.com']) {
       try {
         sh 'cp .npmrc $HOME/.npmrc; cat .npmrc'
-        sh "mvn clean install -B -DcleanNode -Dmaven.test.failure.ignore -s settings.xml -Dmaven.artifact.threads=30"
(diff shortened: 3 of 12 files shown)


michaelneale:
Nice - needs updating to master. 
Where do the config files come from? (I couldn't work that out). 

Slightly concerned on the dependency on various caches (how reliable will they be?)

Also - does this actually speed things up? I couldn't see that it does (certainly doesn't have the corrupt file problem).

ydubreuil:
> Where do the config files come from? (I couldn't work that out). 

From Config File Provider, so caching settings can be updated instance-wide quickly to disable caching if there's a issue.

> Also - does this actually speed things up? I couldn't see that it does (certainly doesn't have the corrupt file problem). 

It depends on which build agent the build occurs. I agree that on `lxd-docker3`, difference can be small if NPM and repo.jenkins-ci.org are well behaving. When caches are in use, we can be sure that downloads (Maven or NPM) are as fast as possible.

michaelneale:
Nice. I think if this is updated to master, and we are confident in caches, we could move ahead with it. What say you @cliffmeyers ? want to kick it around a bit more? That "sed" but is ugly, but may be livable

michaelneale:
looking good. Not sure why shrinkwrap ones have merge conflicts @cliffmeyers ?

cliffmeyers:
@michaelneale the "resolved" element was removed from all shrinkwraps in this branch. Presumably those files have changed in master.

michaelneale:
ah of course @cliffmeyers ;) 

Very naaaace

cliffmeyers:
I should make one or two more additions to this branch tomorrow before we merge. I wrote a "cleanshrink.js" script but need to check it in and add some hooks into the package.json's for relevant projects.

michaelneale:
nice!

cliffmeyers:
cc @kzantow would love some feedback on this. Short version is that npm-shrinkwrap.json writes these "resolved" properties that hard-code npm registry URL's in them. I added a "postinstall" step to clean them out, and added some validation to the very beginning of Jenkinsfile to fail the build if they are present. I didn't want us to do a pre-process of the build only in the CI environment because it seems risky to have too much divergent behavior between our local environments and CI. Any thoughts on that?

michaelneale:
@cliffmeyers @ydubreuil so is this going to use an npm cache?

ydubreuil:
@michaelneale Yes, it currently uses Nexus as an NPM cache. It can be disabled globally on CI configuration if we want to. Using the cache is a 1 minute win roughly.

michaelneale:
ok good to know!

blueocean-personalization/package.json
```diff
@@ -9,7 +9,8 @@
     "test": "gulp test",
     "test:watch": "gulp test:watch",
     "bundle": "gulp bundle",
-    "bundle:watch": "gulp bundle:watch"
+    "bundle:watch": "gulp bundle:watch",
+    "postinstall": "node ../bin/cleanshrink.js"
```
michaelneale: this introduces a dependency o the local node even when called from mvn front end plugin (not sure if a problem..) doesn't it?

blueocean-web/package.json
```diff
@@ -8,7 +8,8 @@
     "test": "gulp test",
     "test:watch": "gulp test:watch",
     "bundle": "gulp bundle",
-    "bundle:watch": "gulp bundle:watch"
+    "bundle:watch": "gulp bundle:watch",
+    "postinstall": "node ../bin/cleanshrink.js"
```
michaelneale: so I understand the desire to do it here so stuff is same as remove - ie caching is transparent, but it does mean everyone needs to remember to do it everywhere..

ydubreuil: Caching is configured through `Jenkinsfile` which fetches a npmrc file from the CI server. The node part here does not do any caching, it's just to clean up resolved URL of packages from shrinkwrap file. When the resolved URLs are in shrinkwrap file, it's not possible to use a cache server defined in npmrc.

blueocean-personalization/package.json
cliffmeyers: This works fine in the docker environment which actually surprised me. You can actually see this for yourself at the link below, just search for "cleanshrink":

https://ci.blueocean.io/blue/rest/organizations/jenkins/pipelines/blueocean/branches/speed-up-builds/runs/38/log/?start=0

I suspect that `maven-frontend-plugin` is setting up a path that has `node` on it when it executes `npm install`? In which case it should work fine, it might only fail if someone calls `npm install` from their own command line... but if the user has `npm` installed then surely that have `node` available globally too?

Not sure what the proper would fix be unless we changed it to:

`"postinstall": "./node/node ../bin/cleanshrink.js"`

but of course that would fail if a user hadn't run the Maven build at least once to get a local node installed. I guess I'm inclined just to assume that node is available; I'm not really worried about this script running under different versions of node (such as a different version that the user might have installed locally) because the script is pretty vanilla JS.

blueocean-web/package.json
cliffmeyers: Yeah, cleaning the shrink unfortunately is something we need to do to fight against the default behavior of shrinkwrap. I changed the Jenkinsfile so that it validates the shrink before even running the main build, so if a developer commits an untidy shrink it will fail almost immediately.

Unfortunately the hooks we have in npm are limited: postinstall doesn't run if you install a specific dependency, only when doing `npm install` with no args. If anyone else has ideas I'm all ears.

cliffmeyers:
@michaelneale other than the conflict in Jenkinsfile is there anything else you think this PR needs before it's ready to be merged?

michaelneale:
as discussed: HOLDING on this until updating npm. Latest npm fixes some shrinkwrap problems but will stop this way of making npm use a proxy fail.

blueocean-personalization/package.json
michaelneale: yes not sure if front end plugin does that. I think I pinged @tfennelly about that once as being not a bad idea. But in general, post install is not great, if there is any other way to do it.

i386:
@cliffmeyers @michaelneale @ydubreuil can this one be merged now?

ydubreuil:
Maven bits were merged in another PR IIRC. This PR needs a refresh to be merged.

cliffmeyers:
@michaelneale I think originally the intent here was to try to bring some stability to the CI build after addition of shrink started to cause spurious failures. I think that's largely been solved by upgrading the npm version in a past PR.

The next step would be to get npm to actually download from a mirror repo. We could apply a transform to our shrink files so that they'll download from our mirror repo rather than the npm registry. We could use `sed` for that although I'd be inclined to say we should do it with a node script, since we have other code written in JS that deals with shrink validation, etc.

I am not sure if it's a worthwhile investment in time to keep fighting with this in light of the bigger fish we have to fry right now. I also wonder if it's better to not fiddle with npm any longer, and if we want improvements to the build, perhaps consider experimenting with yarn sometime after our 1.0 release.

michaelneale:
@cliffmeyers just in the last day the old .tgz failure with npm has been happening again from time to time, but it seems not common. Should we close this?

PR #541, opened 2016-09-28
Update es6-promise to 3.3.1

paladox:
# Description

Updating es6-promise to 3.3.1

See changelog at https://github.com/stefanpenner/es6-promise/blob/master/CHANGELOG.md#331

Difference between 3.2.1 and 3.3.1.

### DIFF
--- a/blueocean-core-js/package.json
+++ b/blueocean-core-js/package.json
@@ -26,7 +26,7 @@
   "dependencies": {
     "@jenkins-cd/js-modules": "0.0.8",
     "@jenkins-cd/sse-gateway": "0.0.9",
-    "es6-promise": "3.2.1",
+    "es6-promise": "3.3.1",
     "isomorphic-fetch": "2.2.1",
     "jsonwebtoken": "7.1.9",
     "pem-jwk": "1.5.1",
--- a/blueocean-dashboard/npm-shrinkwrap.json
+++ b/blueocean-dashboard/npm-shrinkwrap.json
@@ -1944,9 +1944,9 @@
       "resolved": "https://registry.npmjs.org/es6-map/-/es6-map-0.1.4.tgz"
     },
     "es6-promise": {
-      "version": "3.2.1",
-      "from": "es6-promise@3.2.1",
-      "resolved": "https://registry.npmjs.org/es6-promise/-/es6-promise-3.2.1.tgz"
+      "version": "3.3.1",
+      "from": "es6-promise@3.3.1",
+      "resolved": "https://registry.npmjs.org/es6-promise/-/es6-promise-3.3.1.tgz"
     },
     "es6-set": {
       "version": "0.1.4",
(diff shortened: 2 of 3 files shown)


paladox:
I do not work for CloudBees.

i386:
Hmm the bot seems very confused. 

Would you be able to tell us a little about the motivation for this change?

paladox:
Oh, the motivation is that 3.3.1 includes a couple fixes, like for ie8, and fixes it's globals. I chose not to go with 3.4.0 since I'm unsure if that would break things.

i386:
Cool. If your intention is to get IE 8 support I believe there might be a bit more work to be done to support IE. I have a suspicion that much of our CSS and Server Sent Events usage won't work.

i386:
FWIW we are supporting only the latest versions of all browsers (including IE 11) but contributions are welcome!

paladox:
Oh yep, I found that Object.assign is not supported in any ie release, so we will need to use the es6-shim there, I also found out that sub arrays of array.prototype.map for example array.prototype.map.apply wont work, and it carn't be polyfilled either.

i386:
Ok no worries. @scherler could you take a look when you've got a spare moment?

michaelneale:
yeah the bot is invoked because the PR template wasn't edited to remove that @ mention - we could change it to not include that? 

Anyway - if this can pass the ATH @scherler - why not, seems worthwhile to update 

(shrinkwrap is currently breaking the build too frequently right now, not related to this PR likely)

michaelneale:
This is currently failing due to shrinkwrap json's needing to be updated (./bin/checkshrinkwrap.js will pick that up) - someone can take a look at this.

paladox:
Thanks, fixed now :)

michaelneale:
Passes ATH: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/build_forks/3

And I bench tested locally. This LGTM. 

@tfennelly @cliffmeyers I am not sure of impact of updating a dep and the shrinkwrap for a thing like this, but it seems all good. Would you mind merging it if you agree?

paladox:
Thankyou

tfennelly:
Thanks @paladox. So if you look at the source you should see that we do add some polyfills for IE browsers. The would be added (conditionally) via the main .jelly file in the blueocean-web plugin. I'm 99.99999% sure `Object.assign` was covered by that.

Those changes did get blue ocean running on IE 11. However, that was a few weeks ago now, so I would not be surprised if more work was needed again (knowing IE 😄  ).

@i386 @michaelneale @scherler

tfennelly:
Hmmm ... I wonder who added those in the first place, and why.

paladox:
Hi ie11 support is still broken since I get this error

object doesn't support property or method 'assign'

paladox:
Hi ie11 support is still broken since I get an error about something with 'apply'

paladox:
object doesn't support property or method 'apply'

PR #542, opened 2016-09-28
[binHelper] Implement script to install a dep in various places, incl…

scherler:
# Description

Since we use shrinkwrap you need to install all deps. This scripts provides a handy way to install a lib in different places 

…udes removing the node_modules folder ogf the lib to be installed


tfennelly:
🐝 LGTM from what I can see

scherler:
@kzantow would be more flexible yeah

PR #543, opened 2016-09-29
[FIX JENKINS-38540] - make queue and run consistent with activity

kzantow:
# Description

See [JENKINS-38540](https://issues.jenkins-ci.org/browse/JENKINS-38540).

To test: set queue size to 0, build a freestyle job to get a queued item, go to the BO UI click the queued item, then refresh the browser, you should still have a queued item page rather than a not found.

This also fixes [JENKINS-37843](https://issues.jenkins-ci.org/browse/JENKINS-37843)
# Submitter checklist
- [ ] ~~Ran Acceptance Test Harness against PR changes.~~
# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
@@ -40,7 +40,7 @@ public class QueueContainerImpl extends BlueQueueContainer {
             }
         }
         // JENKINS-38540 - To make this consistent with the activity API, check the runs, too
-        String runId = findQueueInRuns(name);
+        String runId = findRunIdFromQueueId(name);
         if (runId != null) {
             try {
                 StaplerResponse rsp = Stapler.getCurrentResponse();
@@ -56,19 +56,22 @@ public class QueueContainerImpl extends BlueQueueContainer {
 
     /**
      * Finds a queue in the runs based on the queueId
+     * @param queueId incoming queue id, possibly invalid
+     * @return a run id or null if not found
      */
-    private String findQueueInRuns(String name) {
+    private String findRunIdFromQueueId(String queueId) {
         try {
-            long queueId = Long.parseLong(name);
+            long id = Long.parseLong(queueId);
             for (BlueRun run : this.pipeline.getRuns()) {
                 if (run instanceof AbstractRunImpl<?>) {
-                    if (queueId == ((AbstractRunImpl)run).getQueueId()) {
+                    if (id == ((AbstractRunImpl<?>)run).getQueueId()) {
                         return run.getId();
                     }
                 }
             }
         } catch (NumberFormatException e) {
             // not a queue id!
+            return null;
         }
         return null;
     }
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
@@ -9,10 +9,15 @@ import io.jenkins.blueocean.rest.hal.Link;
 import io.jenkins.blueocean.rest.hal.LinkResolver;
 import io.jenkins.blueocean.rest.model.BlueQueueContainer;
 import io.jenkins.blueocean.rest.model.BlueQueueItem;
+import io.jenkins.blueocean.rest.model.BlueRun;
 import jenkins.model.Jenkins;
 
+import java.io.IOException;
 import java.util.Iterator;
 import java.util.List;
+
+import org.kohsuke.stapler.Stapler;
+import org.kohsuke.stapler.StaplerResponse;
 
 /**
  * @author Ivan Meredith
@@ -33,9 +38,39 @@ public class QueueContainerImpl extends BlueQueueContainer {
                 return blueQueueItem;
             }
         }
+        // JENKINS-38540 - To make this consistent with the activity API, check the runs, too
+        String runId = findQueueInRuns(name);
+        if (runId != null) {
+            try {
+                StaplerResponse rsp = Stapler.getCurrentResponse();
+                // Send a redirect, not sure the specific code which would be best. substring to fix double slash
+                rsp.sendRedirect(Jenkins.getInstance().getRootUrl() + pipeline.getLink().toString().substring(1) + "runs/" + runId);
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+            return null;
+        }
         return null;
     }
 
+    /**
+     * Finds a queue in the runs based on the queueId
+     */
+    private String findQueueInRuns(String name) {
+        try {
+            long queueId = Long.parseLong(name);
+            for (BlueRun run : this.pipeline.getRuns()) {
+                if (run instanceof AbstractRunImpl<?>) {
+                    if (queueId == ((AbstractRunImpl)run).getQueueId()) {
+                        return run.getId();
+                    }
+                }
+            }
+        } catch (NumberFormatException e) {
+            // not a queue id!
+        }
+        return null;
+    }
 
     @Override
     public Iterator<BlueQueueItem> iterator() {
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
@@ -7,18 +7,12 @@ import hudson.model.Queue;
 import io.jenkins.blueocean.commons.ServiceException;
 import io.jenkins.blueocean.rest.hal.Link;
 import io.jenkins.blueocean.rest.hal.LinkResolver;
-import io.jenkins.blueocean.rest.hal.Links;
 import io.jenkins.blueocean.rest.model.BlueQueueContainer;
 import io.jenkins.blueocean.rest.model.BlueQueueItem;
-import io.jenkins.blueocean.rest.model.BlueRun;
 import jenkins.model.Jenkins;
 
-import java.io.IOException;
 import java.util.Iterator;
 import java.util.List;
-
-import org.kohsuke.stapler.Stapler;
-import org.kohsuke.stapler.StaplerResponse;
 
 /**
  * @author Ivan Meredith
@@ -39,42 +33,9 @@ public class QueueContainerImpl extends BlueQueueContainer {
                 return blueQueueItem;
             }
         }
-        // JENKINS-38540 - To make this consistent with the activity API, check the runs, too
-        String runId = findRunIdFromQueueId(name);
-        if (runId != null) {
-            try {
-                StaplerResponse rsp = Stapler.getCurrentResponse();
-                // Send a redirect, not sure the specific code which would be best. substring to fix double slash
-                rsp.sendRedirect(Links.ensureTrailingSlash(Jenkins.getInstance().getRootUrl() + pipeline.getLink().toString().substring(1) + "runs/" + runId));
-            } catch (IOException e) {
-                throw new RuntimeException(e);
-            }
-            return null;
-        }
         return null;
     }
 
-    /**
-     * Finds a queue in the runs based on the queueId
-     * @param queueId incoming queue id, possibly invalid
-     * @return a run id or null if not found
-     */
-    private String findRunIdFromQueueId(String queueId) {
-        try {
-            long id = Long.parseLong(queueId);
-            for (BlueRun run : this.pipeline.getRuns()) {
-                if (run instanceof AbstractRunImpl<?>) {
-                    if (id == ((AbstractRunImpl<?>)run).getQueueId()) {
-                        return run.getId();
-                    }
-                }
-            }
-        } catch (NumberFormatException e) {
-            // not a queue id!
-            return null;
-        }
-        return null;
-    }
 
     @Override
     public Iterator<BlueQueueItem> iterator() {
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
@@ -7,6 +7,7 @@ import hudson.model.Queue;
 import io.jenkins.blueocean.commons.ServiceException;
 import io.jenkins.blueocean.rest.hal.Link;
 import io.jenkins.blueocean.rest.hal.LinkResolver;
+import io.jenkins.blueocean.rest.hal.Links;
 import io.jenkins.blueocean.rest.model.BlueQueueContainer;
 import io.jenkins.blueocean.rest.model.BlueQueueItem;
 import io.jenkins.blueocean.rest.model.BlueRun;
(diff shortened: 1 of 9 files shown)


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
```diff
@@ -33,9 +38,39 @@ public BlueQueueItem get(String name) {
                 return blueQueueItem;
             }
         }
+        // JENKINS-38540 - To make this consistent with the activity API, check the runs, too
+        String runId = findQueueInRuns(name);
+        if (runId != null) {
+            try {
+                StaplerResponse rsp = Stapler.getCurrentResponse();
+                // Send a redirect, not sure the specific code which would be best. substring to fix double slash
+                rsp.sendRedirect(Jenkins.getInstance().getRootUrl() + pipeline.getLink().toString().substring(1) + "runs/" + runId);
```
cliffmeyers: This works as expected although the redirect URL doesn't end with a forward slash. Whenever the browser requests a URL without one, Stapler does a redirect to a URL with the forward slash added. So the end result is a little funky in the network debugger:

/queue/5/ -> /runs/25 -> /runs/25/

Could you add a util function that ensures we always get a trailing forward slash on these URL's? It would be easier to debug and should also be more efficient.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunContainerImpl.java
```diff
@@ -49,6 +55,20 @@ public BlueRun get(String name) {
                 }
             }
             if (run == null) {
+                // JENKINS-38540 - To make this consistent with the activity API, check the queue
+                String queuedRun = findRunInQueue(name);
+                if (queuedRun != null) {
+                    try {
+                        StaplerResponse rsp = Stapler.getCurrentResponse();
+                        // Send a 302, temporary redirect. substring to fix double slash
+                        rsp.sendRedirect(Jenkins.getInstance().getRootUrl() + pipeline.getLink().toString().substring(1) + "queue/" + queuedRun);
```
cliffmeyers: Same comment as above with regard to trailing forward slash.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
kzantow: Good catch. Fixed now.

cliffmeyers:
Looks good from here. I'll leave it up to you if you want to wait for @vivek to give a bee. In the mean time I've merged this branch into my own and will push forward. Thanks!!

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
```diff
@@ -33,9 +39,39 @@ public BlueQueueItem get(String name) {
...
+            }
+            return null;
+        }
         return null;
     }
 
+    /**
+     * Finds a queue in the runs based on the queueId
+     */
+    private String findQueueInRuns(String name) {
```
tfennelly: `findQueueInRuns` ? Maybe `findRunIdByQueueId` might more accurately describe what this is doing?

tfennelly: And `name` actually seems to be the `queueId` as a String, no? i.e. not a "name".

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
```diff
@@ -33,9 +39,39 @@ public BlueQueueItem get(String name) {
...
+        try {
+            long queueId = Long.parseLong(name);
+            for (BlueRun run : this.pipeline.getRuns()) {
+                if (run instanceof AbstractRunImpl<?>) {
```
tfennelly: I might be wrong but I think `AbstractRunImpl` doesn't catch all you want to catch.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunContainerImpl.java
```diff
@@ -49,6 +56,20 @@ public BlueRun get(String name) {
                 }
             }
             if (run == null) {
+                // JENKINS-38540 - To make this consistent with the activity API, check the queue
+                String queuedRun = findRunInQueue(name);
+                if (queuedRun != null) {
+                    try {
+                        StaplerResponse rsp = Stapler.getCurrentResponse();
+                        // Send a 302, temporary redirect. substring to fix double slash
+                        rsp.sendRedirect(Links.ensureTrailingSlash(Jenkins.getInstance().getRootUrl() + pipeline.getLink().toString().substring(1) + "queue/" + queuedRun));
```
tfennelly: Seems to duplicate code you also added in `QueueContainerImpl.java` ?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunContainerImpl.java
```diff
@@ -59,6 +80,23 @@ public BlueRun get(String name) {
         return  AbstractRunImpl.getBlueRun(run, pipeline);
     }
 
+    /**
+     * Finds a run in the queue based on the expectedBuildNumber, returns the queue id if found
+     */
+    private String findRunInQueue(String name) {
```
tfennelly: Again ... the name seems confusing to me. "findX" would suggest to me that the function is going to return an X, but the pattern in this PR seems to be the other way around i.e. you are supplying X and returning something else.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
kzantow: well, it's finding the queued item in the runs, name is accurate. name is incoming, might not be a queued id, again, is accurate.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
kzantow: `AbstractRunImpl` is the thing that has the `getQueueId` method

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunContainerImpl.java
kzantow: duplicates a line, but it's different? not sure how that's duplication...

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunContainerImpl.java
kzantow: again, it's finding something in a different place. it's finding a `run` in the list of `queue` items.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunContainerImpl.java
```diff
@@ -59,6 +80,23 @@ public BlueRun get(String name) {
...
+        try {
+            int expectedBuildNumber = Integer.parseInt(name);
+            for (BlueQueueItem i : this.pipeline.getQueue()) {
+                if (expectedBuildNumber == i.getExpectedBuildNumber()) {
+                    return i.getId();
+                }
+            }
+        } catch(NumberFormatException e) {
+            // not an expectedBuildNumber
```
michaelneale: return null from here instead?

michaelneale:
running ATH: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/68/

michaelneale:
ATH says things are kosher. 
I kicked it around locally. Seeing the same 404's as mentioned in another PR, but it doesn't affect the result: 
```
blueocean.js:1505 GET http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/keith/runs/3/ 404 (Not Found)
```
(ie for multibranch, trying a nonsensical URI) - spurious console errors (I don't like unexplained console errors). 

But it appears to work as advertised - nice work. 

I guess we need to await @vivek (who will be around in an Indian timezone from early next week) to review and bless/argue.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
tfennelly: So you are saying that the purpose of this function is to find a queueId (or "queue")? What is the function returning i.e. what service is the function providing? The name suggests to me that it is going to find a "queue" in "runs" and presumably return that queue (as a `String` ?). 😕

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueContainerImpl.java
tfennelly: I'm fairly sure `getQueueId` is defined on `Run` (might be overridden on `AbstractRunImpl`). You might want to double-check that.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunContainerImpl.java
tfennelly: They're a bit different, but very similar other than one of them redirects to `"runs/" + runId` while the other redirects to `"queue/" + queuedRun` and calls `Links.ensureTrailingSlash`. Not a biggie.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunContainerImpl.java
tfennelly: 😕  😄

kzantow:
@cliffmeyers would appreciate if you could revisit this, I've modified things to return `/activities/<run-id-or-expected-build-number>`. This similarly fixes the aforementioned issues. I also significantly optimized the activity fetching. Would love to have @vivek to review these changes.

michaelneale:
@kzantow nice, what sort of optimisations are in there? I can see quite bit of code change in that last commit

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/Containers.java
```diff
@@ -15,7 +15,133 @@
 
     public abstract static class AbstractContainer<T extends Resource> extends Container<T> {
     }
+    
+    /**
+     * Combines multiple {@link Iterable} to treat them as a single one
+     *
+     * @param <T> iterable type
+     */
+    public static class CombinedIterable<T> implements Iterable<T> {
```
michaelneale: surprising, this is the sort of thing I would think already exists... I guess not with the pagination taken into account

michaelneale:
still seems ok, however favourited pipelines don't show it as queued state, and the queued results screen has the correct header state, but nothing in the body (not sure if it always was that way..)

kzantow:
@michaelneale argh, yep this botched favorites quite a bit, checking on that now

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/Containers.java
vivek: @kzantow This change is for very specific use case, that is activity API, where different objects (run and queue) are returned in single iterator and hence need to combine run and queue iterators in to single with intent to enforce pagination parameters across different iterators.

We already have abstraction to enforce pagination criteria over an iterator. I am not convinced this set of changes are needed as such common abstractions would make sense if we have such general use cases.

I have done an alternate implementation that pretty much does same. There is Pageables.combine() that combines iterator with given pagination criteria. There is also ActivityContainer, that uses Pageables.combine(). Its matter of taste but instead of ActivityContainer an anonymous inner class of Container implementation would have worked too.

Another thing, I don't think we can support /activities/:id call reliably so we should not support that. No matter whether you use queue id or run id. So ActivityContainer.get() in my impl always returns null. Instead of /activities/:id, UI should really be using link _links.self.href on each item in activities to fetch individual items.

See my alternate proposal at: https://github.com/jenkinsci/blueocean-plugin/commit/e631fa59144e731925aceecda1e929792921593d.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/Containers.java
```diff
@@ -15,7 +15,133 @@
...
+        }
+    }
+
+    /**
+     * Simple method to provide a combination of {@link Iterable} with the ability to implement a specific
+     * {@link #get(String)} method in a subclass. This will also iterate as efficiently as possible
+     * through the provided iterators to provide paged data - e.g. requesting page 0-9 will only
+     * iterate through the first 10 items, in the order of the provided {@link Iterable}
+     */
+    public abstract static class IterableContainer<T extends Resource> extends Container<T> {
```
cliffmeyers: @vivek there are two scenarios that come to mind:
1. often a user will receive a deep-link to the run details screen for a build that is queued. The URL in this case contains the "expectedBuildNumber" value. With the current setup the UI would need to fetch against /runs/:id and receive a 404. It could then try to fetch the item from /queue but it's very bad to have 404's if we can avoid them.
2. in other scenarios the UI will receive SSE for a job when it becomes queued. We may want to fetch data about the queue item via /queue/:id but frequently the build will transition to a "running" state even in the short amount of time between when the SSE is received and when the fetch is made. In this case the UI would receive a 404 and need to fetch from /runs/:id.

IMO from a cleanliness standpoint and for ease of debugging it's desirable not to have 404's.

vivek: @cliffmeyers Ah ok. Sorry, I remember now, @kzantow mentioned this to me earlier. I guess my objection is only about reliability of this approach. Maybe its fine, given the fact that expectedBuildNumber and run id within a pipeline are always going to be unique.

michaelneale:
should this be updated to master and submitted for re-review? (I didn't see any major objections from @vivek)?

kzantow:
@michaelneale I discussed this with @vivek, I'm going to be updating it, was working on something else last week, though

michaelneale:
ping @kzantow is this still valid? been open for a long time now - can we close this/merge it?

michaelneale:
ping @kzantow this is marked as in review but stil has work-in-progress on it

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/Containers.java
kzantow: @vivek yes, I recall and we discussed this, just pointing out that there is already code to handle this case.

i386:
@cliffmeyers @kzantow is this PR still needed or can it be closed out?

PR #551, opened 2016-10-04
[JENKINS-38013] Set the Cache-Control header for static and adjunct resources

tfennelly:
# Description

See [JENKINS-38013](https://issues.jenkins-ci.org/browse/JENKINS-38013).

Will see about writing a unit test.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

Fyi, before making this change it would always make an `If-Modified-Since` request to the browser, resulting in at least an empty requests/response sequence, which can be costly if you live in the boonies in Australia. E.g. .... see the following wireshark log from before the change (activity page - with "Regular 3G" throttling enabled in the dev tools - rough mimic of slow connection) ....

screenshot 2016-10-04 15 16 38

After making this change, those `If-Modified-Since` requests were eliminated for normal browsing (note these are sill present if you hit the reload button in the browser - in Chrome anyway) 
...

screenshot 2016-10-04 15 28 42

### DIFF
--- a/blueocean-web/src/main/java/io/jenkins/blueocean/ResourceCacheControl.java
+++ b/blueocean-web/src/main/java/io/jenkins/blueocean/ResourceCacheControl.java
@@ -36,6 +36,7 @@ import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpServletResponseWrapper;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
@@ -93,6 +94,8 @@ public final class ResourceCacheControl implements Filter {
         // it matches what we're looking for.
         if (request instanceof HttpServletRequest) {
             if (isCacheableResourceRequest((HttpServletRequest)request)) {
+                HttpServletResponse httpServletResponse = (HttpServletResponse) response;
+
                 //
                 // Set the expiry to one year.
                 //
@@ -104,7 +107,19 @@ public final class ResourceCacheControl implements Filter {
                 // in which case the local cache is fully used (no If-Modified-Since requests for
                 // non-stale resources).
                 //
-                ((HttpServletResponse)response).setHeader("Cache-Control", "public, max-age=31536000");
+                httpServletResponse.setHeader("Cache-Control", "public, max-age=31536000");
+                response = new HttpServletResponseWrapper(httpServletResponse) {
+                    @Override
+                    public void setHeader(String name, String value) {
+                        // Block the setting of the legacy HTTP/1.0 "Expires" header.
+                        // Note that, strictly speaking, this should not be required because
+                        // the HTTP spec dictates that the Cache-Control header takes priority.
+                        // Lets eliminate it anyway in case a browser/intermediary doesn't comply.
+                        if (!name.equalsIgnoreCase("Expires")) {
+                            super.setHeader(name, value);
+                        }
+                    }
+                };
             }
         }
         // continue to execute the filer chain as normal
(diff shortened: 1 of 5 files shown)


blueocean-web/src/main/java/io/jenkins/blueocean/ResourceCacheControl.java
```diff
@@ -0,0 +1,127 @@
...
+                //
+                // Note that this does NOT mean that the browser will never send a request
+                // for these resources. If you click reload in the browser (def in Chrome) it will
+                // send an If-Modified-Since request to the server (at a minimum), which means you at
+                // least have the request overhead even if it results in a 304 response. Setting the
+                // Cache-Control header helps for normal browsing (clicking on links, bookmarks etc),
+                // in which case the local cache is fully used (no If-Modified-Since requests for
+                // non-stale resources).
+                //
+                ((HttpServletResponse)response).setHeader("Cache-Control", "public, max-age=31536000");
```
kzantow: Looks like there might be a conflicting `Expires` header maybe should be filtered out and/or replaced with a far future date?

tfennelly: @kzantow done. Thanks.

michaelneale:
Nice - so this won't affect running things as mvn hpi:run - ie it will reload? (I guess you do a refresh then anyway)

michaelneale:
🐛  when I tried this out, gulp bundle:watch would not trigger it to pick up changes, even on a reload (refresh on browser - command R etc). 

To reproduce: run this, make a change to say Pipelines.jsx and run gulp bundle - it won't show up if you refresh the browser. I compared it with master, and the change did show up. 

I think what is needed is some check to see if it is in dev mode or not

blueocean-web/src/main/java/io/jenkins/blueocean/ResourceCacheControl.java
```diff
@@ -0,0 +1,142 @@
...
+ * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
+ */
+@Restricted(NoExternalUse.class)
+public final class ResourceCacheControl implements Filter {
+
+    private static ResourceCacheControl INSTANCE;
+
+    private final List<String> resourcePrefixes = new ArrayList<>();
+
+    private ResourceCacheControl() {
```
michaelneale: I think we need to have this conditional on if mvn hpi:run is used or not (unfortunately)

michaelneale:
@tfennelly I dabbled on this branch: https://github.com/jenkinsci/blueocean-plugin/tree/fix/nocache - there is a system property set it seems when mvn hpi:run is used - might be worth a look at? (I haven't tested that it actually goes away when run normally, so may want to check that).

tfennelly:
@michaelneale It makes a full request (bypassing the cache) if you do a hard reload of the page. For that reason, I didn't bother switching it off for hpi:run, but we can if ye think it's easier.

michaelneale:
@tfennelly when I tested it, it would not load changes that were bundled (I tried multiple times) until it was turned off, not sure why, but it is a bit of a worry.

tfennelly:
@michaelneale yeah .... if you just do a normal reload ... you need to tell it to bypass the cache.

michaelneale:
@tfennelly yeah I thought I tried hard reload, but changes didn't show up. I wonder if others would be tripped up by this?

tfennelly:
@michaelneale yeah, lets just add the bypass for when running with hpi. No need for the performance benefits there.

tfennelly:
@michaelneale Just added no cache-control for when running with hpi:run.

michaelneale:
ok LGTM - works as advertised. :bee:

PR #554, opened 2016-10-05
Speed up Maven downloads by using a proxy and increasing DL thread po…

ydubreuil:
Split the work done in #540 so we can merge the Maven part that is known to work well. I'll leave the NPM part of #540 to @cliffmeyers 

cc @michaelneale @tfennelly

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,11 +6,12 @@ properties([buildDiscarder(logRotator(artifactNumToKeepStr: '20', numToKeepStr: 
 node {
   deleteDir()
   checkout scm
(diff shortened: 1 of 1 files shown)


tfennelly:
🐝 LGTM

Where is that settings.xml file? I assume it is there or the build should have failed :)

ydubreuil:
> Where is that settings.xml file? I assume it is there or the build should have failed :)

@tfennelly  It's a file managed by Config File provider plugin which contains the Maven mirror settings to use the cache instead of the default jenkins repository

PR #556, opened 2016-10-06
[FIX JENKINS-39225] Internationalisation for Blue Ocean and JDL

scherler:
# Description

See [JENKINS-35845](https://issues.jenkins-ci.org/browse/JENKINS-35845).

https://youtu.be/tV_PhY76uus
# Submitter checklist
# Reviewer checklist
- [ ] Reviewed the code


scherler:
related to this https://github.com/jenkinsci/jenkins/pull/2400

blueocean-web/src/main/js/i18n.js
```diff
@@ -0,0 +1,44 @@
...
+});
+ console.log(lngDetector, 'oooooooo')
+i18n
+  .use(lngDetector)
+  .init({
+    fallbackLng: 'en',
+    // have a common namespace used around the full app
+    ns: ['common'],
+    defaultNS: 'common',
+    preload: ['en', 'de'],
```
kzantow: Assume this would go away?

blueocean-web/src/main/js/main.jsx
```diff
@@ -37,17 +39,18 @@ class App extends Component {
...
         return (
             <div className="Site">
                 <header className="Site-header">
                     <div className="global-header">
                         <Extensions.Renderer extensionPoint="jenkins.logo.top"/>
                         <nav>
-                            <Link to="/pipelines">Pipelines</Link>
-                            <a href="#">Administration</a>
+                            <Link to="/pipelines">{t('pipelines')}</Link>
+                            <a href="#">{t('administration')}</a>
```
kzantow: good syntax, though. i like this better than most of the other i18n providers I've found, which add react components and makes them a non-starter for me

blueocean-web/src/main/js/main.jsx
```diff
@@ -158,9 +161,11 @@ function startApp(routes, stores) {
 
     // Start React
     render(
-        <Provider store={store}>
-            <Router history={history}>{ makeRoutes(routes) }</Router>
-        </Provider>
+          <I18nextProvider i18n={ i18n }>
```
kzantow: why does this have anything to do with react? we should not use react specific stuff here, just make the `t` localization function available globally

tfennelly:
@scherler Hey Thor ... how are the localized language bundles defined with this? How are they loaded etc?

blueocean-web/src/main/js/i18n.js
```diff
@@ -0,0 +1,44 @@
...
+          administration: "Verwaltung",
+        }
+      },
+      en: {
+        common: {
+          login: 'Login',
+          logout: 'Logout',
+          pipelines: "Pipelines",
+          administration: "Administration",
+        }
```
tfennelly: Ok ... now I think I see how the resource bundles are defined. So if I have a plugin, how do I contribute localised strings?

tfennelly:
Hey Thor ... I do think it would be better to make this a normal Javascript component that can be `import`ed (`require`d). In general ... I don't see the point in making something a react component (or any other framework component) unless it absolutely needs to be.

tfennelly:
So to be a bit more specific Thor .... what seems to make sense to me is a simple API that you can call that looks kinda similar to what I think you have here e.g.

```javascript
import LanguageBundle from i18n;

// Get a pack for the "editorcreate" string set.
const editorCreateI8n = new LanguageBundle('editorcreate');

/// blah blah blah

render () {
    return {
        <div>{editorI8n('start')}</div>
    };
}
```

blueocean-web/src/main/js/i18n.js
scherler: in the current form yeah

blueocean-web/src/main/js/main.jsx
```diff
@@ -37,17 +39,18 @@ class App extends Component {
     }
 
     render() {
+        const { t } = this.props;
```
scherler: we are passing functionality to the component as you do in react. Why should we not use a plugin such as react-i18next for the underlying i18next provider. 

http://i18next.com/docs/api/

In case you need it outside the react way see above api, but in React components the above is the way to go IMHO

blueocean-web/src/main/js/main.jsx
scherler: that is the react enabler see above comment for non-react components. They would use i18n service directly.

You asked about blueocean-web/src/main/js/i18n.js that is a simple configure factory for the i18n service this can be used in a non-react context.

scherler:
@tfennelly @kzantow This is a really early snapshot and far from being usable. WIP!!! The integration with extension points and how we will define the languages is still to be determined. I opened the PR so you guys have a preview.

I decided for http://i18next.com/ since it is a pure javascript library which support big language bundles as you both request above. Regarding the integration of that library into our projects (I feel tended to remind you we are using react) I have chosen https://github.com/i18next/react-i18next.

However blueocean-web/src/main/js/i18n.js is that "LanguageBundle" you are talking about which can be imported from any POJSO. Anyway that is as well still in heavy development.

tfennelly:
@scherler yeah I have not forgotten that we are using react as it keeps me awake at night sometimes 😄  . The fact that we are using react does not mean that we have to "reactify" absolutely everything. There's actually a name for that faulty logic / antipattern that I'm sure you're very familiar with .... "[Golden Hammer](https://sourcemaking.com/antipatterns/golden-hammer)".

So ... I'm all for reactifying this and other things if there's a real tangible benefit to it and those benefits out-weight the downsides. I'm not seeing any such benefits here yet. Sorry bud !!

scherler:
@tfennelly The benefit to use a framework like i18next is that it supports big files but delivers them with optimisations. Features like missingKey for translations can be saved to the resources, white listing languages, namespaces, etc.

In react we inject the t function with ```translate(['common'], { wait: true })(App)``` and then use it with ```this.props.t('key')``` however you can do the same in a non react way http://i18next.com/translate/

tfennelly:
@scherler Thor ... I think you are missing my point. I'm totally fine with using the `i18next` package. What I do agree with (atm anyway) is the reactification of that via `react-i18next`. It strikes me as a classic example of over/mis application of a technology (i.e. Golden Hammer).

So it sounds to me like we can use `i18next` directly in react and non-react code. That sounds perfect to me. I would not introduce multiple ways of using this stuff and I would not introduce more than is needed.

I've said this before but I'll just throw it out here again ....  I think we need to be really careful how we use react (and libs in general). Among other things ... we need to minimize the exposure we have to them so that we are not hammered with backward compatibility issues down the line when things change. So, I really am against using every react feature we can uncover, just for the sake of using it, especially when there's a perfectly fine "standard" way of doing things.

> and yes Thor .... that is a c&p of what I said to you on CrapChat 😃  ... I hate retyping.

kzantow:
@scherler @tfennelly agree 100% with Tom on this... same thing I was saying, let's ditch any react specific stuff for now and just make this available as a global library - lots of benefits for doing this, but mostly it makes things consistent and you can always still reference in react dom and properties easily. `i18next` package is fine, but with a wrapper, probably (which might negate using it at all). I'd like to see a slightly different syntax than Tom, maybe. My ideal thing is more like:
```jsx
var t = require('translations')('my-plugin-name'); // getting a reference to the right thing could be different... ask me next week if there's a much easier way (there is, maybe in the works)

// later, as an example in react, in this case:
<table>
...
<th>{t.userName}</th>
<th>{t.userList.birthday}</th>
<th>{t.userList.age}</th>
<th>{t.userList.worth}</th>
<th>{t.userList.successMessageTitle}</th>
...
<td>{user.name}</td>
<td>{t.shortDate(user.birthday)}</td>
<td>{t.number(u.birthday)}</td>
<td style={{background: t.userList.dollarAmountBackgroundImage}}>{t.currency(user.bankRoll)}</td>
<td>{t.userList.successMessage(user.successPhrase)}</td>
...
```

This should be super easy in javascript to accomplish, check this:

```
function blah(arg) {
    return 'blah ' + arg;
}
blah.toString = function() {
    return 'blah';
}

console.log('blah', blah)
console.log('blah(hello)', blah('hello'))
```

Then just make it dependent if there are parameters to replace, and do a recursive lookup for further message resolutions, and you have about the best system you can/need.

Sorry, I'm quite opinionated about this because I've written systems for this in Java before, which worked great. And since this is something fundamental and used by everything, we need to make it as easy as possible to do.

tfennelly:
@kzantow hmmm ... I see what you're doing but it seems a bit funky for my liking for a v1. I'd personally prefer to see it start more simple + use something others have already built Vs reinventing the wheel (wrapping it is fine).

Anyway ... my main concern is that we do not reactify-the-f*ck out of everything 😄 . As for the smaller details of the syntax ... I'm sure we could do it many diff ways and we'd all have slightly different preferences, so I'm happy enough with whatever @scherler comes up with.

kzantow:
@tfennelly here's the problem: we have been prone to getting things wrong the first time. So ok, even if my ideas are wrong, at least they're extremely easy to use. The issue is when we get things wrong that are very verbose, we have more things to change. We have more code that's been written to adapt to these verbose solutions and we have bunches of code to change to adapt to the newly discovered 'right' way of doing things. Let's please stop doing that and get things closer the first time instead of iterating over and over. We're implementing a framework at this point. Frameworks are supposed to be easy to use. Let's be thought leaders, not peons implementing stuff because it needs to be done according to some agile timeframe.

tfennelly:
@kzantow Getting it wrong the first time is not a bad thing and is to be expected. We've gotten things wrong the second time too. Thinking we're going to get things right the first time is not realistic. It's exactly because I'm not expecting us to get it right first time (or maybe even second) that I'm suggesting we should not go ott on it or go reinventing the wheel. Keep it simple if we can is all I'm suggesting + reuse the work and knowledge gained by others - 100K downloads of `i18next` per month ... it's probably not perfect, but must be doing something right and thinking we're going to pull something better out of our cheeks with min effort is crazy talk imo (and would be very arrogant of us).

kzantow:
@tfennelly maybe, but is it arrogant to want something better? Are you saying that it's arrogant to _know_ something is better from experience? I understand your rationale for keeping things simple, but it seems that's been biting us left and right for months now...

tfennelly:
@kzantow you must be trolling me, are you? Go on ... admit it 😉 

I might regret bitting, but 😄  ....

Imo, we collectively as a team "_Now_" jack sh1t on this in the sense that we do not have agreement. To "know" something as an individual is, to me, no different than just having a strong opinion, which is nice, but still just an opinion. We have to get some sort of agreement, which is hard.

The only thing that can possibly be up for debate here in this case when making the argument for reinventing the wheel is the proposed usage model/syntax. I admit I've only looked at i18next and what they have in their docs Vs the samples shown here. Seriously, nobody is going to tell me that we "know" a proposed syntax that we don't really have agreement on is, or ever will be, so radically "better" than their working syntax that it warrants spending X amount of time reinventing the wheel. After that it's a non-contest as far as I'm concerned ... working code with tests, docs, a community of people contributing to it for the last number of years (5 listed on the top level org ... 78 on the core module), 100K downloads a month Vs no working code, tests or anything else.

As for experience ... I and others here have a little bit of that too. I've lost count of how many times I've seen people fall into the trap of "we know better than everyone else", before heading off to waste time solving problems that others have already solved very well. As I said, I have not looked at any other of these OSS libs for doing i18n, but from my sample of 1 that I have seen, I can already see that we need a much stronger case for reinventing the wheel. Personal taste arguments about the positioning of a dot (or whatever) is not good enough imo.

Everything out there done by other people is not sh1t just coz it's not done the way I might have done it. It really infuriates me when people dismiss other peoples work so easily. For me, it's kinda like listening to a climate change or religion Vs science debate. On one side you have the people that have done all the hard work and put it out there (warts and all) for others to pick holes in. On the other side you have ...

tfennelly:
I'm already regretting taking the bate 😉

blueocean-web/src/main/js/i18n.js
michaelneale: Right, ideally it shouldn't matter if they are strings in classic UI, server side messages, or client side js/react - same .properties right?

michaelneale:
@scherler can i18next support at least stuff as simple/complex as: https://wiki.jenkins-ci.org/display/JENKINS/Internationalization#Internationalization-TranslatingmessagereferencesinJelly - as that is where the bar is set - it needs to be able to do that. 

One complication that seems to being ignored is that in a traditional server based web app, there is effectively no cost to look up resources as needed (either lazy loading of .properties and then looking them up in memory from then on). When the app is a SPA, do you deliver all plugins resource bundles on app load (NO, this would add even more load time latency) - therefore there is some "new" complexity around loading and caching these at appropriate times. 

As for how it works, it absolutely needs to cover the hard requirements: 

* https://wiki.jenkins-ci.org/display/JENKINS/Internationalization#Internationalization-TranslatingmessagereferencesinJelly (be able to do that type of thing
* Use properties files and live along side all current translations

scherler:
@cliffmeyers i18next offers: http://i18next.com/translate

- interpolation ```"key": "{{what}} is {{how}}",```
- formatting ```"key": "The current date is {{date, MM/DD/YYYY}}"```
- nesting 

```
"nesting1": "1 $t(nesting2)"
 ....
"nesting2": "2 $t(nesting3)"
```

- singular/plural

```
      "key": "item",
      "key_plural": "items",
```

- Multiple plural 

```
"key_0": "zero"
...
key_3": "few",
"key_4": "many",
```

- interval plural

```
      "key1_plural": "{{count}} items",
      "key1_interval": "(1){one item};(2-7){a few items};(7-inf){a lot of items};",
```

- context

```
      "friend_female": "A girlfriend",
      "friend_male_plural": "{{count}} boyfriends",
....
i18next.t('friend', {context: 'female'});
```

- returnObjects ```i18next.t('tree', {returnObjects: true}); // output: '{"res":"added"}'```
- arrays

```
 "arrayJoin": [
        "line1",
        "line2",
        "line3"
      ],
```
- post processor.

As @tfennelly pointed out I do not think that above feature set is a weekend project and I do not see  any benefit ATM for not using i18next as base.

 I18n is not a simple ```return 'blah ' + arg;``` as @kzantow  want to sell it as can be seen on above features like plural, context, etc. those will need far more logic in translation.

Anyway as discussed with @tfennelly  and @michaelneale I will prepare for dashboard the react wrapper and in personalization the plain i18next and then we can see the differences.

michaelneale:
So on conversations with @kzantow, we have identified the following *hard* requirements for any i18n solution for Jenkins (given it has to work with jenkins): 

1. Must use properties files for resource storage 
2. Must be lazily loaded (on demand) by plugins, as they need them (ie not loaded on app load up front)
3. Must be at least as easy to use in blue ocean plugins as it is in .jelly files in jenkins today. 

To achieve (2) there will need to be an additional endpoint in blue ocean to serve up a plugins resources as json (as needed)
(1) is required as there are existing translations in existing plugins we wish to make use of (need to)
(3) implies that it should be really no extra work to use a string key vs a literal string inside a jsx or js file. For example: 

```
<h1>Output</h1>
<p>...</p>
```
should become: 
```
<h1>${%Output}</h1>
<p>...</p>
```
Or VERY VERY close to that. Perhaps an import, at most (ideally it can be made automatically available). Some function call would also be fine, but the key thing is that the barrier to do it is basically nothing - and in line with https://wiki.jenkins-ci.org/display/JENKINS/Internationalization#Internationalization-TranslatingmessagereferencesinJelly

Interpolation (like MessageFormat) should be possible, eg: 

in `Foo.properties`:
`click.here.link=Click <a href="{0}" >here</a>`

would appear in jsx as something like: 

`<p>${%click.here.link(obj.someMethod(a,b,c))}</p>` (that is how it is in jelly, but in jsx may be more like: 
`<p>{t('click.here.link', obj.someMethod(a,b,c))}</p>`


The import should take care of lazily loading the resources (via some endpoint, ultimately from properties files). 

However it is implemented, the above *must* be met.

cliffmeyers:
One other small request would be that we be able to define date formats like this in the props file:

```
date.short=m/d/yy
date.medium=M d yyyy
date.long=E, d M yyyy
```
And then simply reference them from code like this:

`<p>{t.dateShort(dateValue)}</p>`

or something similarly brief. For two reasons:
- Date formats should always be defined in the props file, not in the code, for ease of translation
- Too many one-off date formats become a chore for contributors to localize.

Here's an example of how the UTF consortium defines short/medum/long/full formats for English:
http://unicode.org/cldr/trac/browser/tags/release-30/common/main/en.xml#L1520

scherler:
one observation which enforces some limitations on properties files. Unfortunately the [MessageFormat](http://docs.oracle.com/javase/6/docs/api/java/text/MessageFormat.html) API does not support named parameters, only argument-index:

```
MessageFormatPattern:
     String
     MessageFormatPattern FormatElement String

FormatElement:
     { ArgumentIndex }
     { ArgumentIndex , FormatType }
     { ArgumentIndex , FormatType , FormatStyle }
```

This has the consequence that we are using a different hook for using named properties for our i18n where this are needed. 

in Messages.properties

```
pipelines=Röhren ##name##
```

and then in jsx

```
<Link to="/pipelines">{t('pipelines', {name: 'of Thor'})}</Link>
```

results in 

screenshot from 2016-10-13 15-23-59

blueocean-web/src/main/js/main.jsx
michaelneale: 👍

michaelneale:
@scherler so the german word for pipeline transliterates as tube ;) 

So you are saying that properties files could be used with named parameters by establishng the convention of ` thing ##paramname##`? 

Is there a reason why positional won't work? (other than not being as nice) - I ask as it is used elsehwere in Jenkins and we are trying to work within its proven i18n work (which is properties files based, with a bunch of existing translations we can re-use)

scherler:
@michaelneale yes default pattern for named parameter is

```
interpolation: {
    prefix: '##',
    suffix: '##',
}
```

However you can as well override that on a per translation basis:

```
{
    t('SSHLauncher.RemoteFSDoesNotExist', {
        0: '000',
        1: '111',
        interpolation: {
            prefix: '{',
            suffix: '}',
        }
    })
}
```

screenshot from 2016-10-14 12-28-15

michaelneale:
@scherler right, so ideally we can keep using positional arguments, for no reason other than to be consistent with existing translations and properties file conventions. If we use named parameters, there is a risk people will get confused and try to use them not in client side js but in classic UIs and get annoyed (I am not sure if this is enough justification, but I wanted to put that out there).

tfennelly:
> I am not sure if this is enough justification, but I wanted to put that out there

@scherler @michaelneale Right ... I really don't think this needs to be backward compatible with the existing/classic translations to this extent. It seems to me like that would be constructing artificial barriers for ourselves 😄 

I think the main use case for this would be porting of existing plugins, right? In that case, I would have thought that the Blue Ocean UI version of the plugin's UI will be different enough to warrant creating of a completely new translation file Vs mixing the two (that would be a mess I would think).

Even if it used exactly the same set of translation strings and there was a requirement to have a Blue Ocean specific copy that did the translations the blue ocean way ... is that really such a big deal? This would be a simple one-off job and be by far the easiest part of porting the plugin to Blue Ocean.

michaelneale:
@tfennelly ok with me, I quite liked the named stuff but I didn't want to create a barrier to re-using existing stuff (but you point out, this is probably not super likely, and doable if it has to be anyway). Cool.

scherler:
@michaelneale @tfennelly I actually implemented it backward compatible now. looks a bit "dumb" but works just fine

```
        const text = i18n.t('bo.core.js.toast.run.stopping', {
            0: name,
            1: runId,
          });
```

Regarding the barrier @tfennelly the base "problem" is that we use properties files and you cannot create deep paths like 

```
{
  parent: {
    key: 'someValue'
  }
}
```

we have some limitation now with the numeric parameters but since the other form does not provide enough benefit I would even stick with this classic form.

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
```diff
@@ -84,9 +85,9 @@ export class PipelinePage extends Component {
                     </Title>
                     }
                     <PageTabs base={baseUrl}>
-                        <TabLink to="/activity">Activity</TabLink>
-                        <TabLink to="/branches">Branches</TabLink>
-                        <TabLink to="/pr">Pull Requests</TabLink>
+                        <TabLink to="/activity">t('bo.dashboard.activity')</TabLink>
+                        <TabLink to="/branches">t('bo.dashboard.branches')</TabLink>
+                        <TabLink to="/pr">t('bo.dashboard.pr')</TabLink>
```
tfennelly: NIT .... maybe we can drop the `bo.` prefix from the keys .... not really needed I think, no?

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
```diff
@@ -116,4 +118,4 @@ PipelinePage.childContextTypes = {
 const selectors = createSelector([pipelineSelector],
     (pipeline) => ({ pipeline }));
 
-export default connect(selectors, actions)(documentTitle(PipelinePage));
+export default translate(['jenkins.plugins.blueocean.dashboard.Messages'], { wait: true })(connect(selectors, actions)(documentTitle(PipelinePage)));
```
tfennelly: It would be nice if we could find a cleaner way of doing this, would it not? Having to add this `translate(blah, blah, blah .....` at the end of all these modules seems a bit tedious.

tfennelly: And does it need to be exported?

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages.properties
```diff
@@ -0,0 +1,12 @@
+bo.dashboard.pipelines.heading=Dashboard
+bo.dashboard.pipelines.new.pipeline=New Pipeline
+bo.dashboard.pipelines.table.health=Health
+bo.dashboard.pipelines.table.branches=Branches
+bo.dashboard.pipelines.table.pr=Pull Requests
+bo.dashboard.loading=Loading...
+bo.dashboard.more=Show more
+bo.dashboard.failing={0} failing
```
tfennelly: Must it be `{0}`, or can it be `{jobName}` (or whatever) ? I really do not think these need to be usable on the server side.

tfennelly:
@scherler Nice work so far Thor. Still marked WiP, so I assume you are planning to do more.

Would be nice if we could find a cleaner/less-verbose way of doing the `translate` provider thing at the end of each module ... people will no doubt bitch about that.

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
scherler: I am in doubt since I saw many examples having a prefix.  see e.g. https://github.com/jenkinsci/ssh-slaves-plugin/blob/master/src/main/resources/hudson/plugins/sshslaves/Messages.properties 

```
SSHLauncher.RemoteFSDoesNotExist
```

If you use multiple bundles you could have the same key, however in i18next you could use ns prefixes to get a specific key.

scherler: ...but yeah looks nicer without

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
scherler: the export is the whole component.

Regarding the react-i18next there are benefits in using it since it will store the keys in a store and takes care of (um)mounting the desired languages and ns bundles.

I left an example in the bo-core-js https://github.com/jenkinsci/blueocean-plugin/pull/556/files#diff-edeae55984cff08bd75ec3c753690ff7R73 I agree that is very clean but those components are not supporting the react cycle as https://github.com/i18next/react-i18next/blob/master/src/translate.js

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages.properties
scherler: see my comments in this PR about the named parameters. in java properties files you only can use numeric values. 

However since i18next is very flexible, you can do ```##name##``` and then in your client side code

```
    t('who.rocks', {
        name: 'Tom the man',
        interpolation: {
            prefix: '##',
            suffix: '##',
        }
    })
}
```

....and what do you mean with server side? Current state of the art in jenkins are those properties file - no ```{name}``` in any of those for sure.

scherler:
@tfennelly  (c'n'p) Regarding the react-i18next there are benefits in using it since it will store the keys in a store and takes care of (um)mounting the desired languages and ns bundles.

I left an example in the bo-core-js https://github.com/jenkinsci/blueocean-plugin/pull/556/files#diff-edeae55984cff08bd75ec3c753690ff7R73 I agree that is very clean but those components are not supporting the react cycle as https://github.com/i18next/react-i18next/blob/master/src/translate.js

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
michaelneale: yes this looks odd - can't actually understand that line of code (I don't think anyone else will either).

tfennelly: @michaelneale I think that's a react "pattern" for kinda mixing functionality i.e. I think  it creates a composite component in the BG... funky stuff.

Maybe we can eliminate this with an annotation type thing e.g. `@Translation('jenkins.plugins.blueocean.dashboard.Messages')` + some build time tooling?

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
+Activity=Aktivit\u00E4ten
```
daniel-beck: Aktivität (Singular)

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Artifacts=Artefakte
+Author=Autor
+Branch=Ableger
```
daniel-beck: Würde ich eher unübersetzt lassen.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Branches=Ableger
+Build=Build
+Changes.by=\u00C4nderungen von {0}
+Changes=\u00C4nderungen
+Changes.condense={0} \u00E4nderungen
+Commit=Commit
+Completed=Komplett
```
daniel-beck: Wie wird das verwendet?

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Dashboard=Instrumententafel
+Date=Datum
+Date.readable.short=MMM DD h:mma Z
```
daniel-beck: Sollte deutsches Datumsformat sein.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Date.readable.long=MMM DD YYYY h:mma Z
```
daniel-beck: Ditto.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Date.duration.format=m[ Minuten] s[ Sekunden]
+Date.duration.hint.format=M [Monate], d [Tage], h[Std.], m[m], s[s]
```
daniel-beck: `M [ Monate] d [ Tage] h[ Stunden] m[ Minuten] s[ Sekunden]`

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Download.artifact=Herunterladen des Artefaktes
```
daniel-beck: *Artefakt herunterladen*

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Duration=Dauer
+Duration.param=Dauer {0}
```
daniel-beck: Was ist der Parameter? Evtl `Dauer: {0}`?

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+# the emptyState keys assume that you will parse them with a markdown parser
+EmptyState.activity=# Auf die Pl\u00E4tze, fertig...\nHmm, es liegen keine Ausf\u00fchrungen f\u00FCr vor.\n\nF\u00FCgen sie ein Commit zum Repository _{0}_ oder starten Sie die Ausf\u00fchrung per Hand.
```
daniel-beck: Weiter oben wurde `Build` englisch gelassen. Wieso ist es hier eine *Ausführung*?

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+EmptyState.artifacts=Es liegen keine Artefakte diese Ausf\u00fchrung vor.\n\n
```
daniel-beck: *… für diese…*

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+EmptyState.branches=# Ableger ziehen\nErstellen Sie einen Ableger (Branch) in Ihrem Repository _{0}_ und Jenkins wird Ihre \u00C4nderungen testen.\n\nProbieren Sie es und werden Sie der/die Held(in) Ihres Teams.
+EmptyState.branches.notSupported=# Ableger nicht unterst\u00FCtzt.\nAbleger (Branch) builds funktionieren nur mit dem _Multibranch Pipeline_ Aufgabentyp. Dies ist nur ein guter Grund Jenkins Pipeline zu nutzen.\n\n[Weitere gute Gr\u00FCnde](https://jenkins.io/doc/book/pipeline-as-code/)
+EmptyState.changes=Es liegen keine \u00C4nderungen f\u00FCr diese Ausf\u00fchrung vor.\n\n
```
daniel-beck: Evtl. eher *Es gibt keine …*? "Es liegen keine … vor" klingt eher nach "Wir wissen es nicht".

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+EmptyState.pr.notSupported=# Integrationsanfrage f\u00FCr \u00C4nderungen (Pull Requests) nicht unterst\u00FCtzt.\nDie Validierung von Integrationsanfragen funktionieren nur mit dem _Multibranch Pipeline_ Aufgabentyp. Dies ist nur ein guter Grund Jenkins Pipeline zu nutzen.\n\n[Weitere gute Gr\u00FCnde](https://jenkins.io/doc/book/pipeline-as-code/)
+EmptyState.pr=# Push me, pull you\nWenn eine Integrationsanfrage in Ihrem Repository _{0}_ erstellt wird, wird Jenkins diese testen und den Status der Integrationsanfrage Github melden.
+EmptyState.tests=Es liegen keine Tests f\u00FCr diese Ausf\u00fchrung vor.\n\n
```
daniel-beck: Siehe `EmptyState.changes`

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Enable=Aktivieren
+Failed=Fehlgeschlagen {0}
```
daniel-beck: Kontext ist hier relevant. Eher *{0} Fehlschläge*? Evtl. sogar *{0} Testfehler*?

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Failing={0} fehlgeschlagen
+Failures=Fehler
+Fixed=Fixiert
```
daniel-beck: Kommt das aus dem Google-Übersetzer?

*behoben*, *korrigiert*. Was verwendet Jenkins (core) für sowas?

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Health=Gesundheit
+Latest.build=Letzte Ausf\u00fchrung
```
daniel-beck: Build lassen oder nicht? Inkonsistent.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Latest.message=Letzte Nachricht
```
daniel-beck: Evtl. Meldung?

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Loading=Laden...
+Message=Nachricht
+More=Mehr anzeigen
+Name=Name
+New.failures=Neue Fehler
+New.Pipeline=Neue Pipeline
+No.changes=Keine \u00C4nderungen
+noSteps=Es gibt keine Schritte.
```
daniel-beck: Liest sich seltsam. *Diese Pipeline definiert keine Build-Schritte* oder sowas evtl.?

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Not.found.heading=Seite nicht gefunden (404)
+Not.found.message=Jenkins konnte die Seite die Sie suchen nicht finden. Bitte kontrollieren Sie die URL oder dr\u00FCcken Sie den zur\u00FCck Knopf.
```
daniel-beck: *… die gesuchte Seite…*

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Author=Autor
+Branch=Ableger
+Branches=Ableger
+Build=Build
+Changes.by=\u00C4nderungen von {0}
+Changes=\u00C4nderungen
+Changes.condense={0} \u00E4nderungen
+Commit=Commit
+Completed=Komplett
+Dashboard=Instrumententafel
```
daniel-beck: Würde es als Dashboard lassen, oder es einfach *Übersichtsseite* (oder nur *Übersicht*) nennen.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+No.changes=Keine \u00C4nderungen
+noSteps=Es gibt keine Schritte.
+Not.found.heading=Seite nicht gefunden (404)
+Not.found.message=Jenkins konnte die Seite die Sie suchen nicht finden. Bitte kontrollieren Sie die URL oder dr\u00FCcken Sie den zur\u00FCck Knopf.
+Open.dashboard=Instrumententafel anzeigen
+Passed=Erfolgreich {0}
+Passing={0} erfolgreich
+Passing.singular=Erfolgreich
+Pipeline=Pipeline
+PR=Integrationsanfrage
```
daniel-beck: Wirklich?

Würde es als PR als feststehenden Begriff/Fremdwort lassen. Integrationsanfrage verwirrt nur.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+More=Mehr anzeigen
+Name=Name
+New.failures=Neue Fehler
+New.Pipeline=Neue Pipeline
+No.changes=Keine \u00C4nderungen
+noSteps=Es gibt keine Schritte.
+Not.found.heading=Seite nicht gefunden (404)
+Not.found.message=Jenkins konnte die Seite die Sie suchen nicht finden. Bitte kontrollieren Sie die URL oder dr\u00FCcken Sie den zur\u00FCck Knopf.
+Open.dashboard=Instrumententafel anzeigen
+Passed=Erfolgreich {0}
```
daniel-beck: Sieht mir eher nach `{0} erfolgreich` aus, aber es fehlt Kontext.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Passing={0} erfolgreich
+Passing.singular=Erfolgreich
+Pipeline=Pipeline
+PR=Integrationsanfrage
+Run.now=Ausf\u00fchrung starten
+Serenity=Wie schaut es?
+Show.complete.logs=Ganze Protokoll
```
daniel-beck: Ganzes Protokoll anzeigen

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Not.found.heading=Seite nicht gefunden (404)
+Not.found.message=Jenkins konnte die Seite die Sie suchen nicht finden. Bitte kontrollieren Sie die URL oder dr\u00FCcken Sie den zur\u00FCck Knopf.
+Open.dashboard=Instrumententafel anzeigen
+Passed=Erfolgreich {0}
+Passing={0} erfolgreich
+Passing.singular=Erfolgreich
+Pipeline=Pipeline
+PR=Integrationsanfrage
+Run.now=Ausf\u00fchrung starten
+Serenity=Wie schaut es?
```
daniel-beck: ???

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+noSteps=Es gibt keine Schritte.
+Not.found.heading=Seite nicht gefunden (404)
+Not.found.message=Jenkins konnte die Seite die Sie suchen nicht finden. Bitte kontrollieren Sie die URL oder dr\u00FCcken Sie den zur\u00FCck Knopf.
+Open.dashboard=Instrumententafel anzeigen
+Passed=Erfolgreich {0}
+Passing={0} erfolgreich
+Passing.singular=Erfolgreich
+Pipeline=Pipeline
+PR=Integrationsanfrage
+Run.now=Ausf\u00fchrung starten
```
daniel-beck: *Jetzt starten*?

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Serenity=Wie schaut es?
+Show.complete.logs=Ganze Protokoll
+Size=Gr\u00F6sse
+Skipped.singular=\u00DCbersprungen
+Skipped=\u00DCbersprungen {0}
```
daniel-beck: `{0} skipped`

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+Status=Status
+Summary=Zusammenfassung
+Tests.passing=Alle Tests erfolgreich
+Tests.passing.count=Sehr gut! Alle {0} Tests f\u00FCr sind erfolgreich.
```
daniel-beck: "diese Pipeline" fehlt zwischen "für" und "sind".

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
daniel-beck: *starten Sie die Pipeline per Hand*? Im Englischen ist es auch *Pipeline*.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
```diff
@@ -0,0 +1,64 @@
...
+EmptyState.branches=# Ableger ziehen\nErstellen Sie einen Ableger (Branch) in Ihrem Repository _{0}_ und Jenkins wird Ihre \u00C4nderungen testen.\n\nProbieren Sie es und werden Sie der/die Held(in) Ihres Teams.
+EmptyState.branches.notSupported=# Ableger nicht unterst\u00FCtzt.\nAbleger (Branch) builds funktionieren nur mit dem _Multibranch Pipeline_ Aufgabentyp. Dies ist nur ein guter Grund Jenkins Pipeline zu nutzen.\n\n[Weitere gute Gr\u00FCnde](https://jenkins.io/doc/book/pipeline-as-code/)
+EmptyState.changes=Es liegen keine \u00C4nderungen f\u00FCr diese Ausf\u00fchrung vor.\n\n
+EmptyState.pr.notSupported=# Integrationsanfrage f\u00FCr \u00C4nderungen (Pull Requests) nicht unterst\u00FCtzt.\nDie Validierung von Integrationsanfragen funktionieren nur mit dem _Multibranch Pipeline_ Aufgabentyp. Dies ist nur ein guter Grund Jenkins Pipeline zu nutzen.\n\n[Weitere gute Gr\u00FCnde](https://jenkins.io/doc/book/pipeline-as-code/)
+EmptyState.pr=# Push me, pull you\nWenn eine Integrationsanfrage in Ihrem Repository _{0}_ erstellt wird, wird Jenkins diese testen und den Status der Integrationsanfrage Github melden.
```
daniel-beck: *GitHub* (großes H).

Ansonsten wieder die furchtbare Integrationsanfrage.

daniel-beck:
I didn't comment on `EmptyState.branches`, `EmptyState.branches.notSupported`, and `EmptyState.pr.notSupported` (in German) but they also need some work. They are very obviously directly translated from English.

scherler:
@daniel-beck thank you for your feedback I will now update the translations and yes I tried to stay as close to the original as possible.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: screenshot from 2016-10-25 11-49-05

IMHO that needs to stay in Plural

daniel-beck:
> I tried to stay as close to the original as possible. 

… which makes for clumsy translations.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: not sure how serious https://github.com/danielauener/git-auf-deutsch is but he is using Zweig instead. ```Ableger``` is more Fork I guess.

scherler: However we translated it in spanish I would think we shoud do the same in german

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: screenshot from 2016-10-25 11-49-05

letzte spalte, besser ```Beendet```?

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: ok

daniel-beck:
Maybe @fredg02 @albers @wolfs are interested in reviewing the German translations here?

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: screenshot from 2016-10-25 11-49-05

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: Build werde ich einheitlich mit Ausführung übersetzen

scherler: The problem is that the english one is IMHO as well wrong since that message is the same for a freestyle projects

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: +1

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: War so stolz auf ```Integrationsanfrage``` auf spanish ```Petición de cambio```

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: ist der test view, werde ein paar tests erstellen und ein paar bilder ziehen

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: Korrigiert

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: +1

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: screenshot from 2016-10-25 12-37-04

Aber das ist ATM sowieso ```display: 'none'``` nicht sicher warum es drin ist.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: siehe oben ist ATM in einem hidden block
screenshot from 2016-10-25 12-37-04

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
scherler: Sehr gut! Alle {0} Tests waren erfolgreich.

ist IMO besser

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages.properties
```diff
@@ -0,0 +1,64 @@
...
+Serenity=How's the serenity?
+Show.complete.logs=Show complete log
+Size=Size
+Skipped.singular=Skipped
+Skipped=Skipped {0}
+Status=Status
+Summary=Summary
+Tests.passing=All tests are passing
+Tests.passing.count=Nice one! All {0} tests for this pipeline are passing.
+Tests=Tests
```
cliffmeyers: I have two suggestions here:
- I think we should avoid mixed case for the key names and stick with all lowercase and using dot as a delimiter when needed.
- I think this file would also benefit from namespacing. It's hard to see where a key like "Summary" actually corresponds to in the UI. I would suggest keys named in this fashion:

pipelinelist.header.failing=Failing
pipelinelist.row.failing={0} failing
rundetails.artifacts.download=Download Artifact
(or run.details... if you prefer)

I think generally the localization key should identify the place on the screen where the label exists, not the label itself. On a past project we had many different screens with a "Search" button and defined the "search" label in one place, then discovered later that in Spanish sometimes Search is "busqueda" and other times it's "buscar".

Not trying to be critical here, just trying to point out things I've seen be a problem on past projects.

blueocean-core-js/src/js/i18n/i18n.js
```diff
@@ -0,0 +1,45 @@
...
+    // order and from where user language should be detected
+    order: ['querystring', 'navigator'],
+    // keys or params to lookup language from
+    lookupQuerystring: 'language',
+});
+/**
+ * configure the backend for our locale
+ */
+export const xhr = new XHR(null, {
+    loadPath: '/jenkins/i18n/resourceBundle?language={lng}&baseName={ns}',
```
cliffmeyers: Does this path need to integrate with `UrlConfig` to get the proper "jenkins path"? Seems like this could break for deployments where standard "jenkins" context isn't used.

blueocean-dashboard/src/main/js/components/MultiBranch.jsx
```diff
@@ -13,40 +14,34 @@ import { pipelineBranchesUnsupported } from './PipelinePage';
...
-            </p>
-
-            <p>
-                Give it a try and become a hero to your team.
-            </p>
-
-            <button>Enable</button>
+            <Markdown>
+                {t('EmptyState.branches', { 0: repoName })}
+            </Markdown>
```
cliffmeyers: Lately some members on the team have expressed concerns about depending on too many React libs since it may limit our ability to upgrade React down the road. Do we really need markdown here or could we just have plug a few translation keys into a few HTML elements?

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
cliffmeyers: The syntax used in core JS seems very simple and easy to understand. Is there a reason we can't just use that in the rest of the app, or am I misunderstanding something?

blueocean-web/src/main/js/main.jsx
cliffmeyers: I like the brevity of this syntax too, although I think we should just get a reference to "t" via import / require.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages.properties
scherler: If you see the history of this PR you find @tfennelly requesting the opposite. Having it changed it once already I am not really keen to do it again (at least not in this PR)

blueocean-core-js/src/js/i18n/i18n.js
scherler: Yeah good catch

blueocean-dashboard/src/main/js/components/MultiBranch.jsx
scherler: I think markdown is needed here since the alternative is creating 4/5 keys and wire them in a really ugly way. I am using markdown and not dangerouslySetInnerHTML to prevent xss attacks via i18n keys. 

I further think that we may want to use markdown in general much more since this would make it easy to create a markdown2pdf parser in the future.

scherler:
@cliffmeyers https://github.com/i18next/react-i18next/blob/master/src/translate.js the component is designed to support the react life cycle. In not using it we would need to support the same logic somehow. 

I further do not see any benefit that ```t``` is coming from an import rather then injected via properties as it is custom in react.

scherler:
BTW @cliffmeyers @tfennelly 

```
export default translate(['jenkins.plugins.blueocean.dashboard.Messages'], { wait: true })(connect(selectors, actions)(documentTitle(PipelinePage)));
```

we can use https://lodash.com/docs/4.16.4#flow and this would become

```
const enhanced = _.flow(
  translate(['jenkins.plugins.blueocean.dashboard.Messages'], { wait: true }),
  connect(selectors, actions),
  documentTitle
);

export default enhanced(PipelinePage);
```

However that would create a dep on lodash.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
daniel-beck: > However we translated it in spanish I would think we shoud do the same in german

This is absolutely meaningless, it depends on what terms are used in each language.

Compare 'Computer' in German (which is both more common and less colloquial than 'Rechner') to 'ordinateur' in French.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
daniel-beck: Fertiggestellt?

scherler:
@tfennelly @cliffmeyers I just created a small ```compose``` function to wire different functions so we do not need to use lodash. e.g. for the PipelinePage

```
const composed  = compose(
  translate(['jenkins.plugins.blueocean.dashboard.Messages'], { wait: true }),
  connect(selectors, actions),
  documentTitle
);

export default composed(PipelinePage);
```

blueocean-dashboard/src/main/js/components/RunDetails.jsx
```diff
@@ -215,4 +217,9 @@ const selectors = createSelector(
     [runSelector, isMultiBranchSelector, previousSelector],
     (run, isMultiBranch, previous) => ({ run, isMultiBranch, previous }));
 
-export default translate(['jenkins.plugins.blueocean.dashboard.Messages'], { wait: true })(connect(selectors, actions)(RunDetails));
+const composed  = compose(
```
michaelneale: even less clear what this does now cc @tfennelly> Went from boiler plate to magical more verbose boilerplate.

michaelneale:
Getting there - that latest change for `compose` looks even more strange (perhaps @tfennelly can find a way to make all that stuff go away)

blueocean-dashboard/src/main/js/components/RunDetails.jsx
michaelneale: even less clear what this does now cc @tfennelly> Went from boiler plate to magical more verbose boilerplate. 👎🏻

blueocean-dashboard/src/main/js/util/compose.js
```diff
@@ -0,0 +1,15 @@
...
+ * with the `this` binding of the created function, where each successive
+ * invocation is supplied the return value of the previous.
+ *
+ * Basically, we’re just nesting functions inside each other so that when they’re called with a final value,
+ * the result explodes outwards through each layer.
+ *
+ * @param {...(Function|Function[])} [funcs] The functions to invoke.
+ *
+ */
+export default compose  = (fn, ...rest) =>
```
michaelneale: yeah not sure this adds anything really (maybe I am missing the point)

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages.properties
cliffmeyers: Sorry, I missed that, but I have to say that I disagree with @tfennelly on that. You can't ask people to translate one word to another without understanding the context. Using the longer key names makes it obvious where the label is being used and will make life much easier for anyone submitting a translation. If we can reach consensus on the key naming convention I would be happy to contribute some commits, as I feel quite strongly about this. Curious on thoughts from @kzantow @michaelneale on this.

blueocean-core-js/src/js/i18n/i18n.js
cliffmeyers: I don't follow. In your `i18n.js` you should just be able to do this:

`import urlConfig from '../UrlConfig';`

then

`urlConfig.getJenkinsRootUrl()`

and in most cases that should just return `/jenkins` I believe.

cliffmeyers:
@scherler I looked at the i18next React component and it looks like the main functionality it provides is to refresh components if the selected language is changed by the user. Is that the functionality you're referring to, or is there something else it's doing too that is more subtle?

If it's just the refresh, then I think a locale change should just trigger a full-page browser refresh that will load the right bundles and render a new UI. It's not like users are going to change their locales over and over so the in-place refresh seems like an unnecessary optimization to me. I also feel like a developer should just be able to add a resource bundle, import the `t` function and use it. I don't think a developer should have to do component composition for something like this that seems on the surface pretty simple to me.

scherler:
@cliffmeyers a dev does not ```add``` a resource bundle since that is creating a Message.properties in the resource dir, which would be done once.

If you refer to using a specific bundle, or as we name it namespace, than we doing exactly that. I limited the usage of the ```translate```  now to 3 components. 

Regarding the react-i18next it further stores the keys in a store, so it will be quicker in retrieval. Further I do not see any benefit in doing a hard reload of a whole page when there are actually only two strings on that page to be refreshed.

....and I totally miss the point about the problem of composing function since that is something really common in react.

scherler:
@cliffmeyers thanks for pointing me to the '../UrlConfig' !!!

cliffmeyers:
@scherler Can you explain how putting the keys in the store leads to quicker retrieval? Do you mean from a performance standpoint?

I guess my point is that most Jenkins plugins authors are Java developers. They are probably at most moderately proficient with JS and very likely have never used React before. If our I18N API was "import this function from a library and invoke it, and it just works" then there is no learning curve. Doing things "The React Way" (component composition, setting stuff on props instead of using an import) means you need to have a good grasp on React to understand what's going on. That's fine for our team but I don't think most people will get it immediately.

And I still stand by the concern that the more React-specific libraries we depend on, the more likely we are to face upgrade challenges down the road.

I respect the work you've done in this PR, I just think we have some considerations (developer learning curve, keeping our React dependencies light) that are important to hash out.

scherler:
@cliffmeyers I still need to i18nify personalization and there I will show the way you want to see (actually it is shown in the core-js as well). I only using the react-i18next in dashboard and in web (there more or less as enabler).  Regarding the upgrade challenge on that specific plugin is close to null since the author is the same as the underlying i18next component.


Benefits for me why I using it in dashboard

**only get ns needed**

```
      componentWillMount() {
        this[translateFuncName] = this.i18n.getFixedT(null, namespaces);
      }
``` 

**wait for namespaces loaded**

```
this.i18n.loadNamespaces(namespaces, () => {
          if (this.mounted) this.setState({ ready: true });
          if (wait && this.mounted) bind();
        });
        if (!wait) bind();
```

Regarding the learning curve about higher order components (HOC) in react, I think is not that high. They original appeared when mixins got deprecated.

https://medium.com/@franleplant/react-higher-order-components-in-depth-cf9032ee6c3e#.5wb1b9nht 

> A Higher Order Component is just a React Component that wraps another one.

So the nature of the beast is that a react component is a function, so the above is results in a function that wraps a function.

```
const wrapped = outerFunction(innerFunction);
wrapped(props); // => is equal to outerFunction(innerFunction)(props)
```

michaelneale:
There was a conversation (hangout) as the discussion here was getting "sprawling" @cliffmeyers did you take any notes you can share?

pom.xml
```diff
@@ -18,7 +18,7 @@
   <properties>
     <java.level>7</java.level>
     <jackson.version>2.2.3</jackson.version>
-    <jenkins.version>2.7.1</jenkins.version> <!-- Should be kept in sync with Dockerfile FROM statement-->
+    <jenkins.version>2.26-SNAPSHOT</jenkins.version> <!-- Should be kept in sync with Dockerfile FROM statement-->
```
michaelneale: We actually won't be able to depend on this it looks like, we need to find a way to make it work with 2.7 lineage as that is the most common one out there and people take a while to upgrade

cliffmeyers:
Outcomes from yesterday's meeting:
- Java properties files will be used to store localization data.
   - [ ] @scherler will confirm that there are no major limitations using i18next with properties files (as .json is the default)
- Localization files will use "dot delimited keys" (e.g. home.pipelineslist.header.summary=Summary) for each label location. This will make it easier for translators to understand where a label is used. It will result in some duplicated labels but will give maximum flexibility for various languages.
   - [ ] @cliffmeyers will update the PR to use keys according to this convention.
- We will use raw i18next JS APIs for localization and avoid taking on i18next-react dependency. We want to limit React libs where possible as they may be incompatible with future React versions and complicate upgrades. We also want to avoid l10n boilerplate where possible.
   - Long term we will load localization data via js-modules similar to how JS and CSS are loaded. This will ensure that l10n data is loaded before a module initializes and eliminate need to make React components "i18next aware" via i18next-react.
      - [ ] @cliffmeyers will create JIRA ticket to track this
   - [ ] @scherler will remove i18next-react and implement a workaround in this PR.
   - If the user changes the locale, it will require a browser refresh to display. This is a trade-off to simplify downstream UI / React code so it does not need boilerplate to handle on-the-fly locale changes as this is a rare use case.
- Only unresolved issue is whether to continue using react-markdown. There is a similar concern here regarding React upgrade challenges.

cc @michaelneale @kzantow @sophistifunk

tfennelly:
@cliffmeyers I added a ticket for the js-modules/js-builder changes - [JENKINS-39322](https://issues.jenkins-ci.org/browse/JENKINS-39322).

michaelneale:
I am -1 on using markdown (those libraries have a way of creeping in dependencies). 

one other requirement (new) is that blue ocean works (even if the translations don't) on 2.7.x LTS base for mass adoption

scherler:
@michaelneale https://github.com/i18next/react-i18next/issues/209 

> but be aware if you put user content there not escaped you open your site to xss attacks.

To overcome that problem markdown is beautiful. If you look on our use cases (all emptyState boxes) are using markdown shows the power of the approach. You can use one key for the whole box, allowing the translator freely place links and use all markdown syntax to structure the outcome. 

Looking into the alternatives
- dangerouslySetInnerHTML - the cross-site scripting (XSS) attack danger is to high IMHO. it is to easy to inject a xss attack in e.g. bo-web via key and gates of hell are open
- https://github.com/i18next/react-i18next#interpolate-component it is out since we decided, we do not use that lib anymore
- splitt keys and surround them with markup - IMO not feasible  due to diversity of language and structure

Thereby I came to the conclusion that markdown is the only **real** solution to give translator the possibility to use a certain subset of markup via markdown without exposing ourselfs to xss hacks.

@michaelneale is there another alternative?

scherler:
BTW we are using [react-remarkable]( https://www.npmjs.com/package/react-remarkable) which has **one** dep to [remarkable](https://www.npmjs.com/package/remarkable) which has **two** [autolinker](https://www.npmjs.com/package/autolinker) and [argparse](https://www.npmjs.com/package/argparse)

The later has [sprintf-js] (https://www.npmjs.com/package/sprintf-js) as only dep.

Trying to that the fingerprint of the lib is minimal but the benefit using it for **our** usecase is **BIG**

cliffmeyers:
Brody is redesigning the empty states here, so perhaps we should hold off on this for now. Maybe the new design will be simpler and more amenable to L10N w/o the need for Markdown?  https://issues.jenkins-ci.org/browse/JENKINS-37295

michaelneale:
I am even more -1 on markdown for that - if it is just empty states. There doesn't need to be near turing complete content in the localised empty state messages, even html markup should not be used. 

I am not sure if the XSS applies here when it is content from the server. If a user can compromise that, why not just change the javascript itself.

scherler:
@michaelneale for now that is just the emptyState boxes however that problem is a general problem. 

As soon we have the slightest of markup we will run into that problem. Due to the nature of language it is close to imposible to cover all cases where this may happen.

e.g. in german the keys maybe [into, htmlTags, innerText, outro] but in spanish that may have to be [htmlTag, innerText, intro, outro] trying to say that if we want markup in translations (in my experience there is no way around it in the long run) we need something like markdown. ...and I do not really understand the problem behind it to use it, I have to admit (see above close to no footprint)

michaelneale:
@scherler if it has to happen, then markup is ok (ie HTML) but sparingly. The XSS is not applicable in this case any more than if you say editing a .jsx file is a XSS risk (as I understand it). Markup is terrible.

cliffmeyers:
@scherler @michaelneale in talking to Brody yesterday, he plans to redesign the "empty states" to be much simpler: image, single blurb of text, action button (optional). I don't think markdown will be necessary for those, so I'd still advocate holding off on changing anything in this PR. That's the last I will speak of it :)

michaelneale:
@scherler "Well since we getting the i18n resources from REST it is an easy exercise to do a man-in-the-middle attack and deliver keys that result in a xss attack. It it not the same as jsx since you would need to edit it, in i18n you just put a tier in between and boom goes the dynamite."

We get everything from rest - you put ANYTHING in the middle and you can do that. How is the js served? how are the other requests served? I still don't get what is different about this resource, it is a server side resource like anything else. XSS is for user contributed content in a webapp, which this is not (this is code). Imagine the translations were in a .js file - would you say the same thing? Its just not a helpful or valid argument. 

We are battling dependencies all the time, unless there is another reason for markdown (or asciidoc) then we shouldn't use it (and as cliff says, likely not needed). Are there existing translations in Jenkins that use html or markdown in the contents of the strings?

sophistifunk:
I've been trying to just let this float by rather than jump in and make it even noisier, but are you guys all sure you're talking about the same thing, and that thing is actually XSS? 

If we're fetching message raw resources via REST they should be returned *un-processed* with the substitutions happening client-side. If we're fetching localised messages, then they should *only* have values mapped into the templates from server-side data. The only thing coming from the client should be entity ids.

michaelneale:
@scherler are you concerned if that some unsafe HTML method is used, then the *values* from elsewhere (not the Messages.properties) may create some danger? 

I think from what cliff is saying, we shouldn't have markup in the translations (as often it is very implementation specific), but I am curious if this is something that is occuring elsewhere in jenkins

michaelneale:
@scherler I see this is marked for review, but has some failing tests still (you can see on CI link above).

scherler:
@michaelneale yeah, the problem with windows not defined :( I just tried to fix that but now personalization tests are not working at all. :(

michaelneale:
@scherler ok, thanks.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages_de.properties
daniel-beck: This makes no sense at all.

PR #558, opened 2016-10-07
Update to use the new plugin parent pom (2.17)

tfennelly:
# Description

Update the build to use the new plugin parent pom (2.17). This involved a few small tweaks to some of the `package.json` files, adding `mvnbuild` and `mvntest` scripts (executed by the maven profiles).

### DIFF
--- a/blueocean-dashboard/.mvn_exec_node
+++ b/blueocean-dashboard/.mvn_exec_node
@@ -1,0 +1,1 @@
+Simple marker file to tell maven to execute node build profiles.
--- a/blueocean-dashboard/package.json
+++ b/blueocean-dashboard/package.json
@@ -9,7 +9,9 @@
     "test": "gulp test",
     "test:watch": "gulp test:watch",
     "bundle": "gulp bundle",
-    "bundle:watch": "gulp bundle:watch"
+    "bundle:watch": "gulp bundle:watch",
+    "mvnbuild": "gulp bundle",
+    "mvntest": "gulp test lint"
   },
   "devDependencies": {
     "@jenkins-cd/eslint-config-jenkins": "0.0.2",
--- a/blueocean-personalization/.mvn_exec_node
+++ b/blueocean-personalization/.mvn_exec_node
@@ -1,0 +1,1 @@
+Simple marker file to tell maven to execute node build profiles.
--- a/blueocean-personalization/package.json
+++ b/blueocean-personalization/package.json
@@ -9,7 +9,9 @@
     "test": "gulp test",
     "test:watch": "gulp test:watch",
     "bundle": "gulp bundle",
-    "bundle:watch": "gulp bundle:watch"
+    "bundle:watch": "gulp bundle:watch",
+    "mvnbuild": "gulp bundle",
+    "mvntest": "gulp test lint"
   },
   "devDependencies": {
     "@jenkins-cd/eslint-config-jenkins": "0.0.2",
--- a/blueocean-web/.mvn_exec_node
+++ b/blueocean-web/.mvn_exec_node
@@ -1,0 +1,1 @@
+Simple marker file to tell maven to execute node build profiles.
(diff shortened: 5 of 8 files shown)


tfennelly:
@vivek @imeredith hey guys ..... this branch is failing because of the error below ... what can we do?

```
[INFO] Analyzed bundle 'Module :: BlueOcean :: REST API implementation' with 45 classes
[WARNING] Rule violated for bundle Module :: BlueOcean :: REST API implementation: complexity covered ratio is 0.56, but expected minimum is 0.60
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] Blue Ocean UI Parent ............................... SUCCESS [ 12.162 s]
[INFO] Module :: BlueOcean :: Commons API ................. SUCCESS [  7.028 s]
[INFO] Module :: BlueOcean :: Web module .................. SUCCESS [01:41 min]
[INFO] Module :: BlueOcean :: Rest module ................. SUCCESS [  5.592 s]
[INFO] Module :: BlueOcean :: JWT module .................. SUCCESS [ 31.088 s]
[INFO] Module :: BlueOcean :: REST API implementation ..... FAILURE [01:52 min]
[INFO] Module :: BlueOcean :: Pipeline REST API implementation SKIPPED
[INFO] Module :: BlueOcean :: Events ...................... SKIPPED
[INFO] Module :: BlueOcean :: Dashboard ................... SKIPPED
[INFO] Module :: BlueOcean :: Personalization ............. SKIPPED
[INFO] Module :: BlueOcean :: Config ...................... SKIPPED
[INFO] BlueOcean beta ..................................... SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 04:36 min
[INFO] Finished at: 2016-10-07T10:35:06+00:00
[INFO] Final Memory: 95M/1556M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.jacoco:jacoco-maven-plugin:0.7.6.201602180812:check (default-check) on project blueocean-rest-impl: Coverage checks have not been met. See log for details.
```

michaelneale:
@tfennelly is this a routine bump of the parent pom (I assume it has some goodness). that is an odd error, it implies an optional check for coverage is now on by default cc @vivek

tfennelly:
I just updated master and set it to use the previous version of the plugin parent pom (2.16) and it is failing the same, so this must be something that has changed in the parent pom between versions 2.11 and 2.16 (inclusive).

tfennelly:
`2.13` is the one that breaks it. https://github.com/jenkinsci/plugin-pom/commit/8f3f36d9d1c8a29d1f822bbcf7a0a8fe328ef332 seems to be the change that causes it to fail.

tfennelly:
Yeah ... it's the presence (or lack of) the `<argLine>-Djava.awt.headless=true</argLine>` config on the `maven-surefire-plugin` plugin that's causing the error. No idea how that is effecting `jacoco-maven-plugin` complexity evaluations.

[See these commits](https://github.com/jenkinsci/blueocean-plugin/commits/parent-pom-2.16) for what works and then what doesn't work.

michaelneale:
@tfennelly the `-Djava.awt.headless=true` param is just for convenience - if you remove it does it work? 

Unfortunately if you don't have that then the maven build keeps stealing focus (perhaps could remove it but people can have it in their .profile - although not sure if that fails it too). @tfennelly maybe drop a ticket in viveks queue to investigate?

tfennelly:
@michaelneale I'm afraid it is more than convenience in this case ... the build fails without it. As I said, if this `argLine` is present the build passes, if not present it does not pass.

Easiest solution would be if there was a config for the jacoco plugin that lowers the coverage requirements. But it is hard to see how that setting is causing a change to the coverage levels. Weird as f***.

michaelneale:
@tfennelly ah ok - well it is a good setting to have anyway. Can it be put in the mvn config somewhere? (I have mine set in .profile: `export JAVA_TOOL_OPTIONS='-Djava.awt.headless=true'`)

Although it is somewhat alarming that setting trips that up...

tfennelly:
> Although it is somewhat alarming that setting trips that up...

Yes ... totally weird actually, but I'm sure there's a locally explanation once you get to the bottom of it; there always is :)

PR #559, opened 2016-10-08
Prefer the displayName rather than the fullName for prettier display

alecharp:
# Description

See [JENKINS-38042](https://issues.jenkins-ci.org/browse/JENKINS-38042).

There is no unit test as the displayed value of the `Link` in `PipelineRowItem` is not tested at this point. 

I "manually" validated the change: I created a job which I named `my-ugly-project` and set its __Display name__ to `My Beautiful project`. I also reacted another project without changing the __Display name__ to validate it didn't affect the current behavior. But as the `displayName` is filled with the `fullName` when it's not specified, then the front part of this plugin has no impact on this.

I couldn't run a complete build of the project as the current `master` branch has code inspections issues (lines being too longs) even before I coded my patch.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/PipelineRowItem.jsx
+++ b/blueocean-dashboard/src/main/js/components/PipelineRowItem.jsx
@@ -45,7 +45,7 @@ export default class PipelineRowItem extends Component {
         const activitiesURL = `${baseUrl}/activity`;
 
         const pathInJob = fullName.split('/').slice(0, -1).join(' / ');
-        const formattedName = `${pathInJob? `${pathInJob} / ` : ''}${displayName ? displayName.split('/').join(' / ') : ''}`;
+        const formattedName = `${pathInJob ? `${pathInJob} / ` : ''}${displayName}`;
         const nameLink = (
             <Link to={activitiesURL}>
                 { showOrganization ?
--- a/blueocean-dashboard/src/main/js/components/PipelineRowItem.jsx
+++ b/blueocean-dashboard/src/main/js/components/PipelineRowItem.jsx
@@ -33,6 +33,7 @@ export default class PipelineRowItem extends Component {
             numberOfFailingBranches,
             numberOfSuccessfulPullRequests,
             numberOfFailingPullRequests,
+            displayName,
             } = pipeline;
 
         const hasPullRequests = !simple && (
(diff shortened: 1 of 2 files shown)


michaelneale:
thanks @alecharp - there are test(s) failing - which do cover this code. If you adjust them to pass then this might be good to go!

alecharp:
oh crap. I only looked at `src/test/js/pipelines-spec.js` which doesn't test the content of the link. I'm working on it.

scherler:
@alecharp there had been https://issues.jenkins-ci.org/browse/JENKINS-35901 where the fullName had been preferred of displayName. @i386 or @cliffmeyers would know the specifics but I guess it has something to do with multibranch and/or folder projects. Did you try with multibranch projects?

alecharp:
@scherler the "normal" pipeline job doesn't have any "Display" field in it's configuration. Only the MultiBranch Pipeline has. By the way, the `fullName` and the `displayName` have the same value if no specific `displayName` is configured. 

If the job creator did specified a `displayName` we should use it. It meant to have a "nicer" human readable value. If the blueocean plugin is about to have a better / nicer UX, this should be part of the plugin, don't you think? Maybe the [JENKINS-35901](https://issues.jenkins-ci.org/browse/JENKINS-35901) was created when only `name` was provided in the REST impl.

scherler:
@alecharp I took screenshots of what we have now

screenshot from 2016-10-10 15-29-08

you can see the "folder" structure and the project name.

screenshot from 2016-10-10 15-30-04

Applying your patch you will loose the path and AFIAK @i386 wants that path. However maybe you can rewrite the patch to include the path AND the display name.

alecharp:
yes I confirm this. Working on it.

michaelneale:
ok thanks @alecharp and thanks @scherler for clarifying this

blueocean-dashboard/src/main/js/components/PipelineRowItem.jsx
```diff
@@ -44,7 +44,8 @@ export default class PipelineRowItem extends Component {
         const pullRequestsURL = `${baseUrl}/pr`;
         const activitiesURL = `${baseUrl}/activity`;
 
-        const formattedName = displayName ? displayName.split('/').join(' / ') : '';
+        const pathInJob = fullName.split('/').slice(0, -1).join(' / ');
+        const formattedName = `${pathInJob? `${pathInJob} / ` : ''}${displayName ? displayName.split('/').join(' / ') : ''}`;
```
scherler: ``` ${displayName ? displayName.split('/').join(' / ') : ''}``` should be ``` ${displayName}```

scherler:
:bee: when CI passes

PR #560, opened 2016-10-10
Move to pipeline-model-definition-plugin 0.3

abayer:
# Description

Update to the latest release (0.3) of pipeline-model-definition-plugin.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/pom.xml
+++ b/blueocean-pipeline-api-impl/pom.xml
@@ -81,7 +81,7 @@
         <dependency>
             <groupId>org.jenkinsci.plugins</groupId>
             <artifactId>pipeline-model-definition</artifactId>
-            <version>0.2</version>
+            <version>0.3</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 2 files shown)


abayer:
Whoops, somehow borked the `pipeline-stage-step` dependency! Fixed and force-pushed a new commit.

abayer:
Ok, the test failures are because `pipeline-model-definition` now executes empty stages for stages that were planned to run after a failed stage. That's fine and dandy, but it looks like, despite what I thought the result would be, the `FlowNode` that ends up getting the failure on it is the *last* one, not the one that actually failed. I'm trying to see if there's some way to set a `FlowNode` to failed while still continuing, but that logic is, so far as I can tell, very very very wrapped up in execution proper.

@vivek - any thoughts?

michaelneale:
@abayer is this a bug with Kyoto or just a test assertion issue? I can see it would make more sense to have the failure on the flownode that was running... (I know vivek was reworking things with bismuth so that may help here possibly)

abayer:
@michaelneale So it's at the very least a test assertion issue as a result of the changes to "execute" the empty post-failure stages. Then there's the question of where the failure shows up...

michaelneale:
@abayer so - did you want to look at the issue from the pipeline model side first? We can then update this later.

PR #561, opened 2016-10-11
Don't run hpi:assemble-dependencies again

ydubreuil:
It's not necessary to run hpi:assemble-dependencies to prepare the ATH plugin archive. It's already done in https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean/pom.xml#L132 to make plugins available for building the development Docker image.

This should reduce build time by 1 minute.


michaelneale:
@tfennelly can you review and approve this?

PR #568, opened 2016-10-18
[FIX JENKINS-38811] Fix IE adjunt loading

tfennelly:
# Description

See [JENKINS-38811](https://issues.jenkins-ci.org/browse/JENKINS-38811).

This broke because of a change we made in js-builder a few weeks ago wrt how it generates the paths to bundle assets.


michaelneale:
:bee: nice catch. How did that ever work?

tfennelly:
@michaelneale It worked before I broke it with a change to `js-builder`. See description.

sophistifunk:
Awesome, that `s/-/_/` dealie annoyed me when I first set pulled the editor out of a temporary JDL fork :)

bbm-bueno

PR #569, opened 2016-10-22
[FIX JENKINS-38013] Fix bloated JavaScript bundles

tfennelly:
# Description

See [JENKINS-38013](https://issues.jenkins-ci.org/browse/JENKINS-38013).

The blueocean web bundle is still huge (nearly 3 Mb) but we need to tackle that some other way (see [JENKINS-39191](https://issues.jenkins-ci.org/browse/JENKINS-39191)). The other bundles are considerably smaller now. It was a bug I had left in `js-builder`.

This PR also removes unneeded `gulpfile.js` files from builds and tweaks them to [use the `js-builder` CLI instead](https://github.com/jenkinsci/js-builder#cli).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


pom.xml
```diff
@@ -20,8 +20,8 @@
     <jackson.version>2.2.3</jackson.version>
     <jenkins.version>2.7.1</jenkins.version> <!-- Should be kept in sync with Dockerfile FROM statement-->
     <javadoc.exec.goal>javadoc-no-fork</javadoc.exec.goal> <!-- stop initialize phase plugins executing twice -->
-    <node.version>5.8.0</node.version>
-    <npm.version>3.7.3</npm.version>
+    <node.version>6.4.0</node.version>
+    <npm.version>3.10.3</npm.version>
```
tfennelly: Tried 6.9.1 and 3.10.9 but there doesn't seem to be installers for the build slaves.

michaelneale: @tfennelly ?? they are pulled down as binaries by the front end plugin. The slave environment shouldn't matter right?

blueocean-personalization/src/main/js/model/FavoritesSseListener.js
```diff
@@ -19,10 +19,21 @@ class FavoritesSseListener {
...
-            jobListener,
-            (event) => this._filterJobs(event)
-        );
+        try {
+            this.sseBus.subscribeToJob(
+                jobListener,
+                (event) => this._filterJobs(event)
+            );
+        } catch (e) {
+            if (!this.sseBus.connection && !global.window) {
```
michaelneale: Is this to fix the failing mocha tests that you uncovered?

blueocean-personalization/package.json
```diff
@@ -3,19 +3,19 @@
...
-    "lint": "gulp lint",
-    "lint:fix": "gulp lint --fixLint",
-    "lint:watch": "gulp lint:watch --continueOnLint",
-    "test": "gulp test",
-    "test:watch": "gulp test:watch",
-    "bundle": "gulp bundle",
-    "bundle:watch": "gulp bundle:watch",
-    "mvnbuild": "gulp bundle",
-    "mvntest": "gulp test lint"
+    "lint": "jjsbuilder --tasks lint",
```
michaelneale: will this trouble people who were used to calling gulp directly?

blueocean-config/gulpfile.js
```diff
@@ -1,7 +0,0 @@
-var builder = require('@jenkins-cd/js-builder');
```
michaelneale: The READMEs refer to gulp all over... need some work done to replace this. Assumig people won't just be able ot run gulp anymore... directly (as no gulpfile) :ant:

michaelneale:
So when running kzantow/failure-project, I am seeing failures re-building the master branch (karoke): 


```
blueocean.js:1505 GET http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/keith/branches/master/runs/9/steps/ 500 (Server Error)rawFetchJSON @ blueocean.js:1505fetchJSON @ blueocean.js:1565(anonymous function) @ jenkins-js-extension.js:14633(anonymous function) @ blueocean.js:80862dispatch @ blueocean.js:80920getNodeAndSteps @ jenkins-js-extension.js:14578(anonymous function) @ jenkins-js-extension.js:14591
blueocean.js:1462 Error: Server Error(…)
```

Also if I run the "michaelneale" branch, I see the status of steps as "unknown" purple, which is wrong... 

still looking, but this seems to be specific to this branch

michaelneale:
this LGTM - can't really punch holes in it. 

Main :ant: is that docs are now wrong, and need to "socialise" amongst people that they don't run "gulp bundle:watch" anymore but run npm commands... this may take some fingers getting used to so that is my only :ant: - but if docs updated, I am probably ok with it.

michaelneale:
cc @scherler @imeredith @kzantow @sophistifunk this change means you don't invoke gulp directly - are you ok with this? (there are no gulp files in the blueocean code) - which means uniform builds (good) but a change in muscle memeory - are you ok with this?

sophistifunk:
As long as it works *without* a globally installed js-builder, ie I can `npm run jjsbuilder` I've been bitten too many times by global tools in JS-land.

michaelneale:
When I run "npm run test" I get this: 

```
[16:35:53] Will use babel config from .babelrc
[16:35:53] 'js_bundle_jenkins-js-extension_bundle_1' errored after 524 ms
[16:35:53] SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:387:25)
    at Object.Module._extensions..js (module.js:422:10)
    at Module.load (module.js:357:32)
    at Function.Module._load (module.js:314:12)
    at Module.require (module.js:367:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (/Users/michaelneale/projects/paas/blueocean-plugin/blueocean-dashboard/node_modules/@jenkins-cd/js-builder/internal/pipeline-transforms/require-stub-transform.js:9:22)
    at Module._compile (module.js:413:34)
```

sophistifunk:
Shit, I've seen that but I can't remember how I fixed it.

tfennelly:
@michaelneale Oh yeah ... I forgot about the readme docs. I'll update. Thanks for reminding me.

tfennelly:
@michaelneale as for the error you got .... Update to a more recent node and it should go away. If it proves to be a problem for lots then I'll make a chance in `js-builder` to get rid of it.

tfennelly:
@michaelneale Updated any docs I could find that refer to gulp. There were only a small few.

tfennelly:
@michaelneale BTW ... there's still a `gulpfile.js` .... in `blueocean-web` ... it's still needed in "special" cases were extra stuff needs to be done.

blueocean-personalization/src/main/js/model/FavoritesSseListener.js
tfennelly: Yes

pom.xml
tfennelly: No, we have maven profiles that use the maven download plugin to download the node and npm tars (the frontend plugin just executes node/npm commands, assuming they are already in place). The slave env has a particular arch that these profiles look at before deciding which tars to download for that arch. That part was failing for the newer versions, but was fine when I dropped back to these versions (which are also "fairly new"). We can try again but I think these are fine for now.

blueocean-personalization/package.json
tfennelly: I don't think so. Anyone that has worked in node land for more than a day knows they need to check the `scripts` in the `package.json`.

blueocean-config/gulpfile.js
tfennelly: I only found a small few refs. I fixed the ones I could find.

michaelneale:
@tfennelly ok will try updating node. I had to downgrade it as JDL wouldn't build with the newer version (fun)

tfennelly:
@michaelneale the clue for you was ... `npm run XXX` failed but `mvn XXX` would pass .... the main diff there is the fact that the first uses the node/npm installed on your machine while the other uses the node/npm installed in the build itself (version set in the `pom.xml`).

pom.xml
scherler: current stable versions are:

```
node -v 
v6.8.1
npm -v 
3.10.8
```

We should use them.

michaelneale:
:bee: but please email the team that gulp command probably wont work and to use npm equivalents. THanks.

scherler:
:ant: the core-js is not included

tfennelly:
@scherler core-js doesn't use js-builder, so not relevant.

scherler:
@tfennelly right

tfennelly:
@scherler so how about a bee then ?? 😄

tfennelly:
I have a bee from @michaelneale anyway, so will go with that if I don't get another in the next little while.

scherler:
@tfennelly I am starting to worry about ya, bro, see above _I_ approved the PR and gave you the small insect.

michaelneale:
@tfennelly he gave a bee above ;)

michaelneale:
@scherler I think he needs something more obvious ...

scherler:
http://images.allposters.com/images/ATA/24808BP.jpg @michaelneale hopefully that is enough for @tfennelly  ;)

tfennelly:
@scherler ooops ... sorry man ... I didn't notice that :) Thanks.

PR #567, opened 2016-10-17
Feature/jenkins 38594 create pipeline begin

cliffmeyers:
# Description
- This one still needs a proper Dialog component from the JDL and some security applied. I opened this PR for now and am forking a new branch for JENKINS-38595 so I can make further progress.
- See [JENKINS-38594](https://issues.jenkins-ci.org/browse/JENKINS-38594).

# Submitter checklist
- [x] Link to JIRA ticket in description, if appropriate.
- [x] Change is code complete and matches issue description
- [x] Appropriate unit or acceptance tests or explanation to why this change has no tests
- [ ] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below.
- [x] Ran Acceptance Test Harness against PR changes.

# Reviewer checklist
- [ ] Run the changes and verified the change matches the issue description
  - Load up app, press "Create Pipeline": confirm classic job creation screen is launched in new tab.
  - Add ?blueCreate to the Blue Ocean URL, then click "Create Pipeline": confirm that a very crummy looking temp dialog overlays the main screen. Test that "CLOSE" and "back button" both navigate back to prior screen
  - Test that a "deep link" to the create pipeline route displays the dialog. It will overlay over a mostly empty screen, since there is no prior route.
- [x] Reviewed the code
- [ ] Verified that the appropriate tests have been written or valid explanation given

@jenkinsci/code-reviewers @reviewbybees

PR #572, opened 2016-10-25
Feature/jenkins 38595 create pipeline scm list

cliffmeyers:
# Description
- This stacks on top of #567 so the changes to the routing can be ignored.
- This is a work-in-progress to get some early feedback on a proposed API for "Create Pipeline" flows. I have implemented a basic flow for Git with most of the forms but nothing has been skinned or componentized yet. There is also a fake flow for Github just for the purposes of showing how multiple plugins would contribute their flows. Some explanation of the code:
  - `CreatePipeline` is the main file that renders the dialog and content. It renders the buttons via `CreatePipelineScmListRenderer` and the content below via `CreatePipelineStepsRenderer`. Plugins create a subclass of `ScmProvider` and register to the extension point.
  - Check out `GitScmProvider` and `GitCreationFlow`. This shows how the flow for Git is put together. The "steps" are in `CompletedStep` and `ConnectStep`. My goal here was to allow someone full freedom to do whatever they want in a step but still have an easy way to render the steps, mark them as complete, and complete the flow. The Manager, Api and Status classes are really just the internal plumbing.
  - The inner works of flow management is done in `MultiStepFlow` and `FlowStep`.
  - I added some things in js-extensions, mainly a `SandboxedComponent` that renders any children passed to it via a separate `ReactDOM.render` call. I wanted to avoid creating multi extension points for the list, creation flow (and re-entrant flow, which will come later) because I felt it would be confusing for the plugin author and make communication between the separate parts of code more difficult. So there's a single extension point that returns different React elements that are rendered arbitrarily in their own sandboxes.
- See [JENKINS-38595](https://issues.jenkins-ci.org/browse/JENKINS-38595).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-dashboard/src/main/js/creation/CreatePipelineStepsRenderer.jsx
```diff
@@ -0,0 +1,52 @@
+/**
+ * Created by cmeyers on 10/17/16.
+ */
+import React, { PropTypes } from 'react';
+import Extensions from '@jenkins-cd/js-extensions';
+import VerticalStep from './VerticalStep';
+
+const Sandbox = Extensions.SandboxedComponent;
```
michaelneale: What is a sandboxed component?

blueocean-dashboard/src/main/js/creation/git/ConnectStep.jsx
```diff
@@ -0,0 +1,190 @@
...
+    SSH: 'SSH',
+    SYSTEM_SSH: 'SYSTEM_SSH',
+    USER_PASS: 'USER_PASS',
+};
+
+/**
+ * Component that accepts repository URL and credentials to initiate
+ * creation of a new pipeline.
+ */
+export default class ConnectStep extends React.Component {
```
michaelneale: Should this be called "GitConnectStep"? 

This is in dashboard know, but I think @imeredith has proposed before it could one day migrate to another existing plugin - still the intention?

michaelneale:
Very nice @cliffmeyers I like what I see (few minor comments/suggestions). 

I can never quite work out with github, but does this PR stack on top of https://github.com/jenkinsci/blueocean-plugin/pull/567 ?

blueocean-dashboard/src/main/js/PipelineRoutes.jsx
```diff
@@ -82,9 +83,16 @@ function persistModalBackground() {
  * the background context.
  */
 function handleNavigationChangeToFromModal(prevState, nextState, replace, callback) {
-    if (nextState.params.runId && (prevState == null || !prevState.params.runId)) {
+    const enterRunDetails = nextState.params.runId && (prevState == null || !prevState.params.runId);
+    const leaveRunDetails = (prevState !== null && prevState.params.runId) && !nextState.params.runId;
+    const enterCreatePipeline = nextState.location.pathname.indexOf('/create-pipeline') !== -1 &&
```
cliffmeyers: This branch was forked off the one from #567 so this code has been cleaned up there (and your comments on that PR will be addressed)

blueocean-dashboard/src/main/js/creation/CreatePipelineStepsRenderer.jsx
cliffmeyers: From jsdoc:  "Component that renders any passed-in children via a separate ReactDOM.render call. Useful for rendering untrusted React elements: errors will be trapped and displayed so the main UI isn't broken."

blueocean-dashboard/src/main/js/creation/git/ConnectStep.jsx
cliffmeyers: Renamed in bd2f370 - good suggestion.

Right, the intent is that everything in the "git" subfolder could be picked up and dropped into its own plugin. The code does use the `MultiStepFlow` and `FlowStep` so we either need to make that available via DI (Keith's work) or put in blueocean-core-js. Right now the `GitScmProvider` is registering itself via the extension point, so this is all built in a way that plugins can implement their own creation flows.

cliffmeyers:
@michaelneale that's right, this is stacked on #567. I'll update the description to reflect that.

michaelneale:
Looking good @cliffmeyers - keep it coming. Code looks neat and it seems to work nicely thus far. Looks like some bits of code could be unit tested so that may be worth while.

cliffmeyers:
@michaelneale @imeredith made some minor changes to this PR. I think now is a good time to merge so I can get rolling on the creation flow w/ git.

michaelneale:
@cliffmeyers ok - anythign new you think I should review? 
If ATH passes and you are comfortable, previous approvals apply

cliffmeyers:
@michaelneale you can have a look at the fancy new Dialog, that's really the only thing that has changed.

michaelneale:
@cliffmeyers yeah I had a lot in the PR sandbox and 🐝

PR #579, opened 2016-10-28
Task/jenkins 35845 i18n key rename

cliffmeyers:
# Description
- See [JENKINS-35845](https://issues.jenkins-ci.org/browse/JENKINS-35845).
- Merging this to @scherler branch so he can see more easily what I changed. Figured this was easier than just committing directly to his branch.


cliffmeyers:
@scherler we should get this merged to your branch sooner rather than later before there is drift in the code.

blueocean-dashboard/src/main/js/components/Activity.jsx
```diff
@@ -88,23 +88,24 @@ export class Activity extends Component {
         }
 
         const latestRun = runs[0];
+        const head = 'pipelinedetail.activity.header';
```
scherler: why not move this to top level and use

```
t(`${head}.xxx`)
```

all the time (especially the one above)

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages.properties
```diff
@@ -1,23 +1,81 @@
-Activity=Activity
```
scherler: IMO it makes sense to have a strict alphabetic order for the keys of our properties. Meaning we should do something like

```
sort -t\  -n Messages.properties > Messages.properties_sort
#review
mv Messages.properties_sort Messages.properties
```

on them before commit.

cliffmeyers: Sorry, my unix-fu isn't up to snuff. Can you describe the desired end state?

blueocean-dashboard/src/main/js/components/Activity.jsx
cliffmeyers: Yeah, I'll admit I am not sure which approach is the best. Part of me felt that using the full keys in all places would make it easier to search across the code base and see where certain keys were used without having to search for substrings of the keys. On the other hand, it is a little verbose and repetitive in the code. Curious what the the rest of the team feels about this, cc @imeredith @kzantow @tfennelly @sophistifunk @michaelneale

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages.properties
scherler: just sorting the file using the key column

scherler: @cliffmeyers do not worry will do it in JENKINS-35845

PR #587, opened 2016-11-06
Initial ResourceManager impl

i386:
Instead of working within the constraints of the Adjunct system we should just copy its code from stapler and make our own resource manager that plays by our rules.

### DIFF
--- a/blueocean-web/src/main/java/io/jenkins/blueocean/ResourceManager.java
+++ b/blueocean-web/src/main/java/io/jenkins/blueocean/ResourceManager.java
@@ -1,0 +1,135 @@
+package io.jenkins.blueocean;
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import hudson.Plugin;
+import hudson.PluginManager;
+import hudson.PluginWrapper;
+import hudson.Util;
+import jenkins.model.Jenkins;
+import jnr.ffi.annotations.In;
+import org.apache.commons.lang.StringUtils;
+import org.kohsuke.stapler.HttpResponses;
+import org.kohsuke.stapler.MetaClass;
+import org.kohsuke.stapler.StaplerRequest;
+import org.kohsuke.stapler.StaplerResponse;
+
+import javax.servlet.ServletContext;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+import java.net.URL;
+import java.util.Arrays;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+
+import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
+import static javax.servlet.http.HttpServletResponse.SC_FORBIDDEN;
+import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
+
+/**
+ * Plugin-aware resource management system
+ */
+@Singleton
+public class ResourceManager {
+
+    /**
+     * Map used as a set to remember which resources can be served.
+     */
+    private final ConcurrentHashMap<String,String> allowedResources = new ConcurrentHashMap<>();
+
+    @Inject
+    private PluginManager pluginManager;
+
+    @Inject
+    private ServletContext context;
+
+    private static final String KEY = ResourceManager.class.getName();
+
+    public ResourceManager() {
+        install();
+    }
+
+    /**
+     * Handle URLs like /resources/my-plugin/c82be88/org/jenkins-ci/plugins/bundle.js
+     * @param req request
+     * @param rsp response
+     * @throws IOException on error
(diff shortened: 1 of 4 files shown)


blueocean-web/src/main/java/io/jenkins/blueocean/ResourceManager.java
```diff
@@ -0,0 +1,135 @@
...
+
+import static javax.servlet.http.HttpServletResponse.SC_BAD_REQUEST;
+import static javax.servlet.http.HttpServletResponse.SC_FORBIDDEN;
+import static javax.servlet.http.HttpServletResponse.SC_NOT_FOUND;
+
+/**
+ * Plugin-aware resource management system
+ */
+@Singleton
+public class ResourceManager {
```
i386: @tfennelly this is a bit of an experiment I did over the weekend that implements a new resource loading system that replaces adjunct in Blue Ocean. It works much the same as Adjunct but has a few key differences.

blueocean-web/src/main/java/io/jenkins/blueocean/ResourceManager.java
```diff
@@ -0,0 +1,135 @@
...
+    private ServletContext context;
+
+    private static final String KEY = ResourceManager.class.getName();
+
+    public ResourceManager() {
+        install();
+    }
+
+    /**
+     * Handle URLs like /resources/my-plugin/c82be88/org/jenkins-ci/plugins/bundle.js
```
i386: @tfennelly the URL format is as follows

`/resources/my-plugin/c82be88/org/jenkins-ci/plugins/bundle.js`

# `/resources/` gets served off of our root action
# `my-plugin` is the name of the plugin the resource belongs to
# `c82be88` is a hash of the plugin version and the jenkins version. That means if either of those change new resources are downloaded.
# `org/jenkins-ci/plugins/bundle.js` is the resource path within the plugin JAR

blueocean-web/src/main/java/io/jenkins/blueocean/ResourceTag.java
```diff
@@ -0,0 +1,68 @@
...
+import org.xml.sax.SAXException;
+
+import java.io.IOException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import javax.servlet.ServletContext;
+
+@NoContent
+public class ResourceTag extends TagSupport {
```
i386: @tfennelly started replicating all the stapler adjunct tags here but ran out of time. Looks reasonably straight forward to copy this jelly tag lib into BO and migrate them to use `Resource Manager`.

blueocean-web/src/main/java/io/jenkins/blueocean/ResourceManager.java
```diff
@@ -0,0 +1,135 @@
...
+            throw HttpResponses.error(SC_NOT_FOUND, new IllegalArgumentException("Plugin " + pluginName + " is disabled"));
+        }
+
+        String resourcePath = StringUtils.join(Arrays.copyOfRange(atoms, 2, atoms.length), '/');
+        URL resource = plugin.classLoader.getResource(resourcePath);
+
+        if(resource == null) {
+            throw HttpResponses.error(SC_NOT_FOUND,new IllegalArgumentException("No such adjunct found: "+path));
+        } else {
+            long expires = MetaClass.NO_CACHE ? 0 : TimeUnit.DAYS.toMillis(90);
```
i386: I think this is where you'd want to set all the headers you put in `ResourceCacheControl`?

i386:
I don't think we want a new stapler version for this - all the code can happily be contained in Blue Ocean.

blueocean-web/src/main/java/io/jenkins/blueocean/ResourceManager.java
```diff
@@ -0,0 +1,135 @@
...
+@Singleton
+public class ResourceManager {
+
+    /**
+     * Map used as a set to remember which resources can be served.
+     */
+    private final ConcurrentHashMap<String,String> allowedResources = new ConcurrentHashMap<>();
+
+    @Inject
+    private PluginManager pluginManager;
```
vivek: 🐛 I have been trying to keep blueocean-web free of jenkins dependency. Idea really is for Jenkins 3/4 time this remains pretty abstract module free of hard dependency on Jenkins core packages. Could this not be moved to blueocean-dashboard? Hard to enforce at maven level as we pick stapler (version) from jenkins core.

i386: @vivek actually the resource manager should be split out and moved to its own plugin. This was to just demonstrate an approach.

vivek: @i386 👍

i386:
@tfennelly Ill close this one now

PR #590, opened 2016-11-08
[FIX JENKINS-39587] refactor to not use context but use standard properties

scherler:
# Description

In some circumstances, that are not reproducible, the listing of pipelines has shown a blank list.
I do not why I had this circumstances, but the usage of context had been wrong anyhow in this place.

https://issues.jenkins-ci.org/browse/JENKINS-39587

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
+++ b/blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
@@ -13,17 +13,6 @@ import * as pushEventUtil from './util/push-event-util';
 const { object, array, func, node, string } = PropTypes;
 
 class OrganizationPipelines extends Component {
-    // FIXME: get rid of context use
-    getChildContext() {
-        if (this._getOrganizationName()) {
-            return {
-                pipelines: this.props.organizationPipelines,
-            };
-        }
-        return {
-            pipelines: this.props.allPipelines,
-        };
-    }
 
     componentWillMount() {
         const config = this.context.config;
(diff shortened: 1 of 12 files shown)


alecharp:
:bee: once the build is validated

alecharp:
sorry but :bug: for the moment. see http://pr.sandbox.blueocean.io/pr-590/blue/organizations/jenkins/multibranch/activity/ which is not working and goes into an infinite loop

blueocean-dashboard/src/main/js/OrganizationPipelines.jsx
```diff
@@ -132,10 +121,17 @@ class OrganizationPipelines extends Component {
...
+        let pipelines = null;
+        if (!allPipelines && !organizationPipelines) {
             return null;
         }
-        return this.props.children;
+        if (allPipelines && allPipelines.$success) {
+            pipelines = allPipelines;
+        } else if (organizationPipelines && organizationPipelines.$success) {
+            pipelines = organizationPipelines;
+        }
```
tfennelly: So reading the above code .... both `allPipelines` and `organizationPipelines` cannot both be non-null at the same time, or does `allPipelines` always take precedence when they're both non-null?

tfennelly: Maybe a comment here for other readers would be a good idea.

scherler:
@alecharp that is fixed now

scherler:
@alecharp that is fixed now

alecharp:
:bee: nicely done!

i386:
@scherler good to see a bug fix like this get done 💯

PR #591, opened 2016-11-09
Fix possible NPE

i386:
# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
+++ b/blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
@@ -59,22 +59,27 @@ public class BlueMessageEnricher extends MessageEnricher {
         message.set(EventProps.Jenkins.jenkins_org, OrganizationImpl.INSTANCE.getName());
 
         String channelName = message.getChannelName();
-        if (channelName.equals(Events.JobChannel.NAME)) {
-            JobChannelMessage jobChannelMessage = (JobChannelMessage) message;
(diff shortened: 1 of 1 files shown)


blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
```diff
@@ -59,22 +59,27 @@ public void enrich(@Nonnull Message message) {
...
-                }
-            }
+        if (!channelName.equals(Events.JobChannel.NAME)) {
+            return;
+        }
+
+        JobChannelMessage jobChannelMessage = (JobChannelMessage) message;
+        ParameterizedJobMixIn.ParameterizedJob job = jobChannelMessage.getJob();
+        Link jobUrl = LinkResolver.resolveLink(job);
+        if (jobUrl == null) {
```
i386: Missing this null check

Do we even have findbugs running?

imeredith: @i386 I'm not sure if this is the fix we want. For sure the null check is good, but it could be hiding issues. If a job exists we should be able to resolve the think afaik. The question I have is what job cant resulve a link? And how can we report issues like this in the future as it will affect the frontend sse handlers

tfennelly:
🐝 if this is needed in a hurry

Agree with @imeredith that there's another issue here too wrt why `LinkResolver.resolveLink(job)` would return null for a job instance that was just received from an Jenkins core event.

tfennelly:
Also agree on Findbugs .... I campaigned for that months ago, but failed.

PR #593, opened 2016-11-09
Feature/jenkins 38848: Credential enumeration API

vivek:
# Description

See [jenkins 38848](https://issues.jenkins-ci.org/browse/jenkins 38848).

## Credential API to get all credential in a domain:
```
GET /jenkins/blue/rest/organizations/jenkins/credentials/:storeId/domains/:domainId/credentials/:credentialId
```

Example:

```
GET /jenkins/blue/rest/organizations/jenkins/credentials/system/domains/_/credentials/

[ {
  "_class" : "io.jenkins.blueocean.rest.impl.pipeline.CredentialApi$Credential",
  "_links" : {
    "self" : {
      "_class" : "io.jenkins.blueocean.rest.hal.Link",
      "href" : "/organizations/jenkins/credentials/system/domain1/e27dc632-d497-4021-9190-bd7e4296f791/"
    }
  },
  "description" : "admin/******:_:Username with password",
  "displayName" : "admin/******",
  "fingerprint" : null,
  "fullName" : "system/domain1/e27dc632-d497-4021-9190-bd7e4296f791",
  "id" : "e27dc632-d497-4021-9190-bd7e4296f791",
  "typeName" : "Username with password"
} ]
```

Current implementation introduces new method in: ApiRoutable.isChildOf(Routable parent). This will enable other plugins to add their URL path to blueocean API roots. We already allow it at root, see 'ApiHead.java' . This change allows plugins to add to organizations path dynamically, '/organizations/:id/xyz'.

## Credential Search API

```
GET /jenkins/blue/rest/search?q=type=credential
```

Example:

```
GET /jenkins/blue/rest/search?q=type:credential

[ {
  "_class" : "io.jenkins.blueocean.rest.impl.pipeline.CredentialApi$Credential",
  "_links" : {
    "self" : {
      "_class" : "io.jenkins.blueocean.rest.hal.Link",
      "href" : "/organizations/jenkins/credentials/system/domain1/26a39941-b08c-4406-a763-620ed05c01d0/"
    }
  },
  "displayName" : "admin/******",
  "fingerprint" : null,
  "fullName" : "system/domain1/26a39941-b08c-4406-a763-620ed05c01d0",
  "id" : "26a39941-b08c-4406-a763-620ed05c01d0",
  "typeName" : "Username with password",
  "description" : "admin/******:domain1:Username with password"
}, {
  "_class" : "io.jenkins.blueocean.rest.impl.pipeline.CredentialApi$Credential",
  "_links" : {
    "self" : {
      "_class" : "io.jenkins.blueocean.rest.hal.Link",
      "href" : "/organizations/jenkins/credentials/system/domain2/18266caf-f546-4411-9cf0-f12ff2d9880a/"
    }
  },
  "displayName" : "joe/******",
  "fingerprint" : null,
  "fullName" : "system/domain2/18266caf-f546-4411-9cf0-f12ff2d9880a",
  "id" : "18266caf-f546-4411-9cf0-f12ff2d9880a",
  "typeName" : "Username with password",
  "description" : "joe/******:domain2:Username with password"
} ]
```

### Create credential using SSH file on master
````
{
  "credentials" : {
    "privateKeySource" : {
      "privateKeyFile" : "~/.ssh/blah",
      "stapler-class" : "com.cloudbees.jenkins.plugins.sshcredentials.impl.BasicSSHUserPrivateKey$FileOnMasterPrivateKeySource"
    },
    "passphrase" : "ssh2",
    "scope" : "GLOBAL",
    "description" : "ssh2 desc",
    "$class" : "com.cloudbees.jenkins.plugins.sshcredentials.impl.BasicSSHUserPrivateKey",
    "username" : "ssh2"
  }
}
````
### Create credential using default ssh file on master (~/.ssh)

````
{
  "credentials" : {
    "privateKeySource" : {
      "stapler-class" : "com.cloudbees.jenkins.plugins.sshcredentials.impl.BasicSSHUserPrivateKey$UsersPrivateKeySource"
    },
    "passphrase" : "ssh2",
    "scope" : "GLOBAL",
    "description" : "ssh2 desc",
    "$class" : "com.cloudbees.jenkins.plugins.sshcredentials.impl.BasicSSHUserPrivateKey",
    "username" : "ssh2"
  }
}
````
### Create credential using ssh private key directly entered

````
{
  "credentials" : {
    "privateKeySource" : {
      "privateKey" : "abcabc1212",
      "stapler-class" : "com.cloudbees.jenkins.plugins.sshcredentials.impl.BasicSSHUserPrivateKey$DirectEntryPrivateKeySource"
    },
    "passphrase" : "ssh2",
    "scope" : "GLOBAL",
    "description" : "ssh2 desc",
    "$class" : "com.cloudbees.jenkins.plugins.sshcredentials.impl.BasicSSHUserPrivateKey",
    "username" : "ssh2"
  }
}
````

### Create credential using username and password
````
{
  "credentials" : {
    "password" : "abcd",
    "stapler-class" : "com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl",
    "scope" : "GLOBAL",
    "description" : "joe desc",
    "$class" : "com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl",
    "username" : "joe"
  }
}
````

### Create Credential Response

````
{
  "_class" : "io.jenkins.blueocean.rest.impl.pipeline.credential.CredentialApi$Credential",
  "_links" : {
    "self" : {
      "_class" : "io.jenkins.blueocean.rest.hal.Link",
      "href" : "/organizations/jenkins/credentials/system/domains/domain1/credentials/b12ddc62-9f82-401e-b417-fe6dfad961dc/"
    }
  },
  "displayName" : "ssh2 (ssh2 desc)",
  "fingerprint" : null,
  "fullName" : "system/domain1/b12ddc62-9f82-401e-b417-fe6dfad961dc",
  "id" : "b12ddc62-9f82-401e-b417-fe6dfad961dc",
  "typeName" : "SSH Username with private key",
  "description" : "ssh2 desc",
  "domain" : "domain1"
}
````

## TODOs
- [x] Api to serve all credentials for a given user.
- [x] credential.description should be set to 'domain:type' if description is empty. Ideally credentials should have domain and store to it's payload for completeness, but this is something that needs to be done in credentials plugin. 
- [x] Create credential API

# Submitter checklist

Started : https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/312/

# Reviewer checklist
- [ ] Reviewed the code

### CODE REVIEW

**Comment by vivek** [comment]
@cliffmeyers Please take a look, try it out and let me know what you think.

**Comment by imeredith** [comment]
@vivek LGTM, only question I have is, is there a short type id we can use? .as in.. usernamePasword or something.

**Comment by i386** [comment]
@imeredith like a capability? ;)

**Comment by vivek** [comment]
@imeredith it comes directly from credentials plugin, not something we control.

**Comment by vivek** [comment]
ATH passes, https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/312/

**Comment by imeredith** [comment]
@i386 no, like an id for the credential type. Because we will need something like it for the credential creation api.

**Comment by imeredith** [comment]
:bee: from me then

**Comment by vivek** [comment]
@imeredith we are not going to use the type in creation API. Credential API needs clients to send $class to send which credential creator needs to be created - that relates to type.

**Comment by vivek** [comment]
@imeredith I will wait tomorrow for @cliffmeyers bee before merging.

**Comment by i386** [comment]
@vivek fantastic to see this in PR 💯

**Comment by cliffmeyers** [comment]
@vivek this looks pretty great to me. I do have a few questions that relate a little more to UI integration and I will cc @i386 on this for feedback:

1. mostly for @i386: thinking about multi-tenancy: when a user initiates create flow, how does the system decide which organization the pipeline will be created in? This is important because the REST URLs in play here are all hang off an organization: for creating the pipeline, for querying the credential list, for creating a new credential. If the answer is "don't worry about that yet", that's fine, I can use "jenkins" without painting us into a corner now.
2. @vivek just wanted to confirm my understanding of the model: there are _many organizations_ which have _one-to-many stores_ which have _1-M domains_ which have _1-M credentials_.
3. If that's correct, putting aside organizations, it would take the UI quite a few REST calls to gather all of the credentials in all of the domains in all of the stores. Would it be possible to enhance the `/search` API to query for a flat list of credentials?
4. If we do that, perhaps the `CredentialApi$Credential` could be enhanced to include `domain` and `store` names so that each object has some context? If the master had credentials spread out across multiple stores and domains and we used a single `/search` API call to fetch them, presumably we need some way to provide the store/domain context in the UI.

**Comment by i386** [comment]
@cliffmeyers there would be a selector on the flow at some point but don't worry about that for the time being.

**Comment by michaelneale** [comment]
ping @vivek looks like this is OK if you can just ACK cliffs question 2 and so on above.

**Comment by vivek** [comment]
@michaelneale Already discussed these among other things with @cliffmeyers. These resulted in to couple of work to be done, search API for credentials, and add domain:type to red description if description is empty. I am going to add them as task items in this PR.

**Comment by vivek** [comment]
@cliffmeyers both additional features credential search API and description default is implemented. Please take a look.

**Comment by michaelneale** [comment]
I wasn't able to get this to work for search - it returned an empty list. I have a global credential for github that I use in multibranch things. 

`http://localhost:8080/jenkins/blue/rest/organizations/jenkins/credentials/system/domains/_/credentials/` works and returns what I expect, ie: 

```
[
{
"_class": "io.jenkins.blueocean.rest.impl.pipeline.CredentialApi$Credential",
"_links": {
"self": {
"href": "/organizations/jenkins/credentials/system/_/github/"
}
},
"description": "",
"displayName": "michaelneale/******",
"fingerprint": null,
"fullName": "system/_/github",
"id": "github",
"typeName": "Username with password"
}
]
```
However:
`http://localhost:8080/jenkins/blue/rest/search?q=type:credential`
returns `[]` - so this seems to be a :bug:

screen shot 2016-11-15 at 4 24 26 pm

**Comment by vivek** [comment]
@michaelneale I suspect you have not pulled latest changes related to search API? I ran the same on `90652f` and it works fine:

```
GET http://localhost:8080/jenkins/blue/rest/search?q=type:credential

[
  {
    "_class": "io.jenkins.blueocean.rest.impl.pipeline.CredentialApi$Credential",
    "_links": {
      "self": {
        "href": "/organizations/jenkins/credentials/system/_/fde4d2bd-64d3-4813-af9b-67b716463a7d/"
      }
    },
  "displayName": "admin/******",
  "fingerprint": null,
  "fullName": "system/_/fde4d2bd-64d3-4813-af9b-67b716463a7d",
  "id": "fde4d2bd-64d3-4813-af9b-67b716463a7d",
  "typeName": "Username with password",
  "description": "admin/******:_:Username with password"
  }
]
```

michaelneale:
@vivek yep looks grand now :bee:

cliffmeyers:
Tested search API and examined fields, they LGTM 🐝

michaelneale:
I guess can go in any time assuming ATH etc happy (not sure if this is needed yet, but up to vivek)

michaelneale:
as discussed, hopefully these api changes go into credenitals plugin itself (which they should very shortly) so may not be needed.

vivek:
@cliffmeyers @michaelneale create credential API implemented and ready for review.

michaelneale:
yep looks good and ATH seems happy

michaelneale:
:bee:

cliffmeyers:
@vivek a few questions about the APIs:

- Can I assume that the UI always needs to provide the `$class`, `stapler-class` and `scope`properties in the request? I assume that `description` is optional?
- For "private key directly entered" - the UI spec shows just a text area for the key; there was no option to enter a `username` or `passphrase`. Can these be safely omitted in creation?
- For "using default ssh file on master", the flow in the UI might not be what you were expecting. The user simply selects the radio button for this option and then proceeds with creation. Would the UI call into the create credential API and then proceed? I think that would have the effect of creating a new credential each time a user proceeds through the flow - unless there is some way to handle this creation in a special way to avoid duplicates?
- For "Create credential using SSH file on master" - we don't actually have that option in the UI but I suspect you were just adding it for completeness.

Will go ahead and test some of these APIs now and provide additional feedback. Thanks!

vivek:
@cliffmeyers 

> Can I assume that the UI always needs to provide the $class, stapler-class and scopeproperties in the request? I assume that description is optional? Update: from my testing this all appears to be the case; let me know if there are any gotchas.

Yes. $class, stapler-class and scopes mandatory and description is optional.

> For "private key directly entered" - the UI spec shows just a text area for the key; there was no option to enter a username or passphrase. Can these be safely omitted in creation?
Update: looks like they can, although it appears that username will be defaulted to the Jenkins username if omitted. Is that the correct behavior?

passphrase is optional and username is not needed for all types of ssh credentials.

> For "using default ssh file on master", the flow in the UI might not be what you were expecting. The user simply selects the radio button for this option and then proceeds with creation. Would the UI call into the create credential API and then proceed? I think that would have the effect of creating a new credential each time a user proceeds through the flow - unless there is some way to handle this creation in a special way to avoid duplicates? Update: I tested this and made the following request twice (see below) and I do see that two different credentials will be created. Not sure what's the best way to approach this:

UI can fetch credentials and if there is ssh credential type then use it or have user select. If you call creation of such key multiple times they get created multiple times. 

> For "Create credential using SSH file on master" - we don't actually have that option in the UI but I suspect you were just adding it for completeness in the REST API.

Thats right:)

cliffmeyers:
@vivek is there some way we could differentiate between the "default .ssh key on master" and other types of ssh keys? I noticed that the default uses a stapler-class value of `BasicSSHUserPrivateKey$UsersPrivateKeySource` - is there some way we could abstract that through a capability to differentiate it? If so, then when the UI fetches the credential list it would know that's the correct credential to re-use, and the user would not be forced to make a selection. It would probably be difficult for the user to differentiate between some of the credentials anyways - if you recall that discussion we had a month or so back.

vivek:
> @vivek is there some way we could differentiate between the "default .ssh key on master" and other types of ssh keys? 

We can't because its not persisted as part of credentials model. default .ssh or file based or direct ssh key entry are different way to create but internally they are simply ssh credential type. If at all this needs such 'creation mechanism persisted as part of model' that would be RFE on credentials plugin. This is out of scope for this PR.

cliffmeyers:
@vivek I agree; we can handle it later on if we need to. This PR is ready to go from my perspective! 🐝

PR #594, opened 2016-11-10
JENKINS-39463# Integrate pipeline-model-definition 0.6

vivek:
# Description

See [JENKINS-39463](https://issues.jenkins-ci.org/browse/JENKINS-39463).

It's still a work in progress. However declarative plugin 0.5 brings some in compatible changes, needs change in script in certain instance, also adds unexecuted nodes. Which is great, it may result in we getting rid of union eventually (we still need it for legacy non-blocking stages). 

@i386 @michaelneale still a WIP but please try this branch out as I suspect it might create certain ux issues. I found one issue related to failed node not reported correctly, see https://issues.jenkins-ci.org/browse/JENKINS-39631.

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code


i386:
@vivek Ill start testing this.

i386:
@vivek the only problem I can see apart from the duplicate steps we have previously discussed, is that the Pipeline Visualisation now goes to the last step in the pipeline which hasn't been executed.

jenkins-39463 3 2016-11-11 09-26-00

i386:
LGTM - no need to raise that bug for @scherler either

jenkins-39463 6 2016-11-11 11-01-21

abayer:
yay. =)

vivek:
@michaelneale @i386 its ready to go. Feel free to test it.

vivek:
@tfennelly @scherler @michaelneale ATH test script using declarative need to change. I have opened a PR for that, https://github.com/jenkinsci/blueocean-acceptance-test/pull/66.

michaelneale:
:bee:

Not 100% sure why I follow the null result/state though, but I figure it makes sense to someone

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/NodeRunStatus.java
```diff
@@ -72,8 +73,12 @@ public NodeRunStatus(GenericStatus status){
                 this.state =  BlueRun.BlueRunState.FINISHED;
                 break;
             case NOT_EXECUTED:
-                this.result = BlueRun.BlueRunResult.NOT_BUILT;
-                this.state = BlueRun.BlueRunState.QUEUED;
+                // It happens when declarative plugin strting from 0.5
+                // sends NOT_EXECUTED for nodes thats not run in case earlier node failed.
+                // we want to represent them as totally unknown in blueocean so that UI can
+                // represent them accordingly
+                this.result = null;
```
i386: I don't have enough context to answer this. What's the situation @vivek ?

michaelneale: how will the UI know to show them appropriately if its all just null?

michaelneale:
Running ATH with the ATH branch here: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/declarative-0.6-upgrade/

michaelneale:
yes passes the ATH so yay!

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/NodeRunStatus.java
vivek: @michaelneale @i386 @scherler  I think Result can still be NOT_EXECUTED but not sure about state. Available states are, QUEUED,RUNNING,PAUSED,FINISHED. So either we keep state as null, or state same as result that is NOT_EXECUTED?

vivek: Btw, null was returned so that it doesn't break UI, that is it shows up as a gray circle.

michaelneale: well I am ok with it as it is - but I expect that GUI will need to know that it *was* skipped vs some other reason right @i386 (at least eventually?)

michaelneale: basically, we can leave this as is if @i386 is happy. We may need ot revisit it when the GUI work is done to show the fancy skip graphic

i386: if thats what this means then the result should be `NOT_EXECUTED` and the state should be `SKIPPED`?

vivek:
@michaelneale Added explicit result and status for stage/steps not build yet. After jenkins report success and if I have your bee, we can merge it.

vivek:
@michaelneale ATH passes https://ci.blueocean.io/job/ATH-Jenkinsfile/job/declarative-0.6-upgrade/5/console.

michaelneale:
:bee: this seems to work for now.

PR #600, opened 2016-11-15
[FIX JENKINS-39624] Pre-load user data on page config object

tfennelly:
# Description

See [JENKINS-39624](https://issues.jenkins-ci.org/browse/JENKINS-39624).

Pre-load user data on page config object + change personalization to not fetch the user info from the rest API.

# Submitter checklist
- [x] __READD__ shrinkcrap where removed - 99fb1e0 and 27f3ff9.

# Reviewer checklist
- [ ] Reviewed the code


blueocean-core-js/src/js/security.js
```diff
@@ -30,7 +30,7 @@ function isSecurityEnabled() {
 }
 
 function isAnonymousUser() {
-    return config.getInitialUser() === 'anonymous';
+    return config.getInitialUser().id === 'anonymous';
```
cliffmeyers: If this could ever be null/undefined perhaps there should be a guard there.

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
```diff
@@ -36,7 +36,7 @@ export function pipelineBranchesUnsupported(pipeline) {
 
 const classicConfigLink = (pipeline) => {
     let link = null;
-    if (AppConfig.getInitialUser() !== 'anonymous') {
+    if (AppConfig.getInitialUser().id !== 'anonymous') {
```
cliffmeyers: Perhaps we could standardize on using the Security.isAnonymousUser() here? Realize you didn't write this code, but having "anonymous" magic string isn't good IMO.

blueocean-dashboard/src/main/js/components/RunDetails.jsx
```diff
@@ -36,7 +36,7 @@ const { func, object, any, string } = PropTypes;
 
 const classicConfigLink = (pipeline) => {
     let link = null;
-    if (AppConfig.getInitialUser() !== 'anonymous') {
+    if (AppConfig.getInitialUser().id !== 'anonymous') {
```
cliffmeyers: see above

blueocean-personalization/src/main/js/components/FavoritesProvider.js
```diff
@@ -6,7 +6,8 @@ import { connect } from 'react-redux';
 import { createSelector } from 'reselect';
 import { List } from 'immutable';
 
-import { userSelector, favoritesSelector } from '../redux/FavoritesStore';
+import { AnonUser, User } from '../model/User';
```
cliffmeyers: Looks like `AnonUser` might be unused in this file.

blueocean-personalization/src/main/js/model/User.js
```diff
@@ -18,6 +19,10 @@ export class User extends Record(
     isAnonymous() {
         return false;
     }
+
+    static current() {
+        return new User(AppConfig.getInitialUser());
```
cliffmeyers: Sorry to be annoying about this, I just wish we could get any code dealing with auth or user into the same lib (`Security`). Maybe this `getInitialUser` should be pseudo-private, or not exist at all (let `Security` reach in and grab the object)

blueocean-config/pom.xml
```diff
@@ -19,9 +19,13 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>blueocean-web</artifactId>
     </dependency>
-      <dependency>
-          <groupId>${project.groupId}</groupId>
-          <artifactId>blueocean-commons</artifactId>
-      </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>blueocean-rest-impl</artifactId>
```
michaelneale: I don't think this is ideal right @i386 ?

michaelneale:
@tfennelly @i386 is there any way to measure if this is an improvement, or what is expected (as it is all about improving subjective or actual performance)?

blueocean-core-js/src/js/security.js
tfennelly: Yeah, was just getting it working first. Ideally, that `User` class that's in personalization should really be in core-js and then we could use that with guarded accessor methods etc

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
tfennelly: Yeah, will change to use that. Was going to to earlier actually and then just decided to leave it for now but since you say it too then ...

blueocean-personalization/src/main/js/model/User.js
tfennelly: I think `getInitialUser` is fine (actually .... I don't see the point in the "Initial" part of the name and why it's not just called `getUser`, but anyway) because getting user details is not just a security thing imo ... it's primarily an identity thing. Testing if a user has auth to do certain things would (imo) be more of a security thing.

blueocean-config/pom.xml
tfennelly: @michaelneale Where is the circular dep? I'm quite certain that's not the case.

The only other way of doing this would be to start implementing some extension points or something, which seems to me like over engineering/thinking.

One thing that I was thinking though was that the likes of this module needing to depend on `blueocean-rest-impl` does seem wrong to me. I'd have thought it should depend on `blueocean-rest` and that it should provide some factory methods for accessing the active user etc, using the write `@Extension` to create the `BlueUser` instance etc. As far as I can see, that's not possible with the current structure i.e. you need to use `blueocean-rest-impl` directly. Seems wrong to me.

tfennelly:
> @tfennelly @i386 is there any way to measure if this is an improvement, or what is expected (as it is all about improving subjective or actual performance)?

@michaelneale Well it def makes a difference because there's one less fetch happening on page load. Whether it's noticeable is a different issue :)

In general ... I would expect there to be a cumulative effect here i.e. you're not going to notice a difference by just eliminating one fetch ... you'll see it after eliminating 2 or 3 of them (or whatever).

blueocean-config/pom.xml
tfennelly: Chatted with @michaelneale and his comment was wrt to my last comment about the seeming need to use `blueocean-rest-impl` instead of just `blueocean-rest`. I'll touch base with @vivek in case there's some CDI magic that needs to be used.

scherler:
As soon Jenkins like the pr

tfennelly:
Build failure is because of shrinkcrap. Will sort that out now.

tfennelly:
@michaelneale Going to merge this and we can address the issue of which rest plugin to use in a separate PR because I'm very likely going to be having a follow-up PR to this wrt runs etc

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/ModelObjectSerializer.java
```diff
@@ -0,0 +1,73 @@
...
+import java.io.IOException;
+import java.io.StringWriter;
+import java.io.Writer;
+
+/**
+ * Simple Jenkins Model Object serializer.
+ *
+ * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
+ */
+public class ModelObjectSerializer {
```
vivek: This is great, thanks @tfennelly:) 🐝

i386:
Nice 👍

PR #604, opened 2016-11-17
JENKINS-39661# /runs and /activity performance fix

vivek:
# Description

See [JENKINS-39661](https://issues.jenkins-ci.org/browse/JENKINS-39661).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultiBranchPipelineImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultiBranchPipelineImpl.java
@@ -1,6 +1,5 @@
 package io.jenkins.blueocean.rest.impl.pipeline;
 
-import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import hudson.Extension;
 import hudson.model.Item;
@@ -24,7 +23,6 @@ import io.jenkins.blueocean.rest.model.BlueQueueItem;
 import io.jenkins.blueocean.rest.model.BlueRun;
 import io.jenkins.blueocean.rest.model.BlueRunContainer;
 import io.jenkins.blueocean.rest.model.Container;
-import io.jenkins.blueocean.rest.model.Containers;
 import io.jenkins.blueocean.rest.model.Resource;
 import io.jenkins.blueocean.service.embedded.rest.AbstractPipelineImpl;
 import io.jenkins.blueocean.service.embedded.rest.BlueFavoriteResolver;
@@ -288,12 +286,36 @@ public class MultiBranchPipelineImpl extends BlueMultiBranchPipeline {
 
             @Override
             public Iterator<BlueRun> iterator() {
+                throw new ServiceException.NotImplementedException("Not implemented");
+            }
+
+            @Override
+            public Iterator<BlueRun> iterator(int start, int limit) {
                 List<BlueRun> c = new ArrayList<>();
-                for(final BluePipeline b: getBranches()) {
-                    for(final BlueRun r: b.getRuns()) {
-                        c.add(r);
-                    }
+
+                List<BluePipeline> branches = Lists.newArrayList(getBranches().list());
+                sortBranchesByLatestRun(branches);
+
+                int l = limit/branches.size() > 0 ? limit/branches.size() : 1;
+
+                int s=0;
+                if(start > 0) {
+                    s = start - l > 0 ? start - l : 0;
                 }
+
+                int count = 0;
+                int retry = 0;
+                while(retry < 5 && count < limit) {
+                    for (BluePipeline b : branches) {
+                        Iterator<BlueRun> it = b.getRuns().iterator(s, l);
+                        while (it.hasNext()) {
+                            count++;
+                            c.add(it.next());
+                        }
+                    }
+                    retry++;
+                }
+
                 Collections.sort(c, new Comparator<BlueRun>() {
                     @Override
                     public int compare(BlueRun o1, BlueRun o2) {
@@ -302,6 +324,7 @@ public class MultiBranchPipelineImpl extends BlueMultiBranchPipeline {
                 });
 
                 return c.iterator();
+
             }
 
             @Override
@@ -309,6 +332,18 @@ public class MultiBranchPipelineImpl extends BlueMultiBranchPipeline {
                 throw new ServiceException.NotImplementedException("This action is not supported");
             }
         };
+    }
+
+    static void sortBranchesByLatestRun(List<BluePipeline> branches){
+        Collections.sort(branches, new Comparator<BluePipeline>() {
+            @Override
+            public int compare(BluePipeline o1, BluePipeline o2) {
+                Long t1 = o1.getLatestRun() != null ? o1.getLatestRun().getStartTime().getTime() : 0;
+                Long t2 = o2.getLatestRun() != null ? o2.getLatestRun().getStartTime().getTime() : 0;
+
+                return t2.compareTo(t1);
+            }
+        });
     }
 
     @Override
@@ -373,6 +408,27 @@ public class MultiBranchPipelineImpl extends BlueMultiBranchPipeline {
 
     @Navigable
     public Container<Resource> getActivities() {
-        return Containers.fromResource(getLink(), Lists.newArrayList(Iterators.concat(getQueue().iterator(), getRuns().iterator())));
+        return new Container<Resource>() {
+            @Override
+            public Resource get(String name) {
+                return null;
+            }
+
+            @Override
+            public Link getLink() {
+                return MultiBranchPipelineImpl.this.getLink().rel("activities");
+            }
+
+            @Override
+            public Iterator<Resource> iterator() {
+                throw new ServiceException.NotImplementedException("Not implemented");
+            }
+
+            @Override
+            public Iterator<Resource> iterator(int start, int limit) {
+                return AbstractPipelineImpl.activityIterator(getQueue(), getRuns(), start, limit);
+            }
+        };
+
     }
 }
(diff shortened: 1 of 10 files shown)


vivek:
@i386 @tfennelly @michaelneale besides code review it will be great if you guys can try it to see the perf improvement. Thanks.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractPipelineImpl.java
```diff
@@ -210,7 +209,58 @@ public static String getRecursivePathFromFullName(BluePipeline pipeline){
...
+                if(queueIterator.hasNext()){
+                    return queueIterator.next();
+                }else{
+                    return runIterator.next();
+                }
+            }
+
+            @Override
+            public void remove() {
+                //noop
```
kzantow: Probably throw an exception?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractPipelineImpl.java
```diff
@@ -210,7 +209,58 @@ public static String getRecursivePathFromFullName(BluePipeline pipeline){
 
     @Override
     public Container<Resource> getActivities() {
-        return Containers.fromResource(getLink(), Lists.newArrayList(Iterators.concat(getQueue().iterator(), getRuns().iterator())));
+
+        return new Container<Resource>(){
+            @Override
+            public Iterator<Resource> iterator() {
+                return Lists.newArrayList(Iterators.concat(getQueue().iterator(), getRuns().iterator())).iterator();
```
kzantow: I also had a solution for this we talked about before... was just getting  back to that PR today, actually. My solution avoided reading all runs if the iterator wasn't read fully.

vivek: @kzantow Since we have pagination enabled by default, iterator(start, limit) is what gets called.  So iterator() gets called only when iterator(start, limit) is not overridden. See PagedResponse, Container classes. For this reason I did not try to optimize it here, maybe I could throw exception from this iterator or use pagination default and delegate to iterator(start, limit)?

i386: @vivek won't creating a new `ArrayList` like this read all the items from the iterator and add them to the list? `Iterators.concat()` would be more appropriate?

kzantow: @i386  The thing is it's not really actually being called right now, but yes, it would.

michaelneale: From what @vivek is saying this shouldn't be called, perhaps it should be an UnsupportedOperationException ?

michaelneale:
test failures: 

```
getMultiBranchPipelineActivityRuns - io.jenkins.blueocean.rest.impl.pipeline.MultiBranchTest
a few seconds
Error
expected:<3> but was:<15>
Stacktrace
java.lang.AssertionError: expected:<3> but was:<15>
	at org.junit.Assert.fail(Assert.java:88)
```

vivek:
@michaelneale @tfennelly @i386 @imeredith Implemented optimization of multi-branch pipelines /activities and /runs API. @tfennelly it will be great if you could test it.

My testing shows:

For 6000 pipeline jobs:
- First time load 7.1 sec to 212 ms
- reloads 625ms to 59ms

For multi-branch pipeline jobs (5 branches) with total about 1000+ jobs (most on master):
- First time load 2.3 sec to 437 ms
- reloads 86ms to 80 ms

i386:
@vivek nice thats a great improvement!

michaelneale:
Looks reasonable to me - quick tyre kick and nothing is on fire so... 

@tfennelly if this looks ok to you, and ATH passes etc - maybe worth a bee?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunSearch.java
```diff
@@ -83,6 +84,40 @@ public Link getLink() {
...
+        if(job != null){
+            pipelines = ImmutableList.of(job);
+        }else{
+            pipelines = Jenkins.getInstance().getItems(Job.class);
+        }
+
+        int skipCount=0;
+        int limitCount=0;
+        for (Job p : pipelines) {
+            RunList<? extends Run> runList = p.getBuilds();
```
tfennelly: @vivek I still think we have a problem here. I created a job, did a few hundred runs on it and then tried loading the activity for that job. The limit was 26, but `getBuilds` is returning a `List` with all the jobs in it. If there are 1000s of jobs and Jenkins has just been restarted, this will result in every one of those Runs being loaded from disk, which seems wrong to me when you're only interested in the first 26.

Did you look at the [comment I left on the JIRA ticket](https://issues.jenkins-ci.org/browse/JENKINS-39661?focusedCommentId=277090&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-277090)? I did an experiment using `LazyBuildMixIn` and `RunMap`. See https://github.com/tfennelly/blueocean-plugin/tree/JENKINS-39661

I'll do another test here with a few 1000 runs and report back.

tfennelly: Okay ... tried it with 3000+ runs of a very simple job similar to the last time and it does seem a lot better, but it is still loading all `Run`s for the job, which I do think would end up being a problem for a more complex build with more stuff loaded up on the `Run` instance (slower to deserialize the persisted form).

I assume the improvement comes from the fact that we are now only creating the `BlueRun` instances for the number of runs needed (`limit`), which is great, but I would still consider this to be a potential issue under more real-life situations. After all, `LazyBuildMixIn` etc would appear to have been created exactly for the purpose of avoiding this "loading everything" situation.

kzantow: @tfennelly getBuilds() returns a lazy list, though. We just can't enumerate it all, or it _will_ load them up.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultiBranchPipelineQueueContainer.java
```diff
@@ -57,4 +63,41 @@ public Link getLink() {
         }
         return queueItems.iterator();
     }
+
+    @Override
+    public Iterator<BlueQueueItem> iterator(int start, int limit) {
+        List<BluePipeline> branches = Lists.newArrayList(multiBranchPipeline.getBranches().list());
+        sortBranchesByLatestRun(branches);
+
+        int l = limit/branches.size() > 0 ? limit/branches.size() : 1;
```
kzantow: 🐛  what if 0 branches?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunSearch.java
tfennelly: @kzantow Not from what I can see both looking at the code, and looking in a debugger. See [`Job.getBuilds`](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/Job.java#L701) ... calls `_getRuns().values()`, where `_getRuns()` returns a `SortedMap` ... that's not a lazy loading of them afaics.

kzantow: When I debugged this before, getRuns() for a job (at least a pipeline job) was lazy loaded. If it's changed, this is definitely a problem.

vivek: @tfennelly I looked at your usage of LazyBuildMixIn and thought its pretty same, what I missed was .values(), I will try it out.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultiBranchPipelineQueueContainer.java
vivek: Thanks, good catch. I will be fixed.

vivek:
@tfennelly @kzantow pushed commit with your issues fixed. Give it a try when you get a chance.

michaelneale:
Looks like good progress so far, and like we discussed last week in terms of the algorithm. 
@kzantow spotted a bug, and @tfennelly was going to re-test it, so still a bit of a WIP I gather but we probably could get this in ASAP

tfennelly:
LGTM 👍

michaelneale:
had a discussion on this: this will change to calculate some MAX_ROWS set, which is then sorted and paginated server side (with a tuneable value). This means that pagination still makes sense and the page will work mostly as people expect but with a smaller amount of data (until we have a REAL DB we can query...)

vivek:
@michaelneale @i386 code pushed. Please review.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultiBranchPipelineImpl.java
```diff
@@ -55,6 +54,8 @@
 public class MultiBranchPipelineImpl extends BlueMultiBranchPipeline {
     /*package*/ final MultiBranchProject mbp;
 
+    private static final int MAX_MBP_RUNS_ROWS = Integer.getInteger("MAX_MBP_RUNS_ROWS", 250);
```
i386: Should this be centralized in `io.jenkins.blueocean.commons.BlueOceanConfigProperties` ?

i386:
Thanks for getting this done :)

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultiBranchPipelineImpl.java
vivek: I don't think so. Common properties are the ones that are used by UI, sent to UI using JSON. This property is more of something privately used by pipeline-api-impl module.

michaelneale:
LGTM :bee:

PR #609, opened 2016-11-19
Lucene PoC

i386:
Uses Lucene on run data to improve searchability of runs. Big performance upsides!

Things to investigate:
* How to index existing runs on the fly
* How we recover from corrupt indexes
* Single index per multi-branch pipeline
** Querying index for specific branch
* Refactor `blueocean-indexing` so it does not have a hard dependency on pipeline
* General operability

### DIFF
--- a/blueocean-indexing/src/main/java/io/jenkins/blueocean/indexing/Index.java
+++ b/blueocean-indexing/src/main/java/io/jenkins/blueocean/indexing/Index.java
@@ -1,0 +1,132 @@
+package io.jenkins.blueocean.indexing;
+
+
+import com.google.common.base.Function;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+import hudson.model.Item;
+import hudson.model.Run;
+import io.jenkins.blueocean.rest.model.BluePipeline;
+import io.jenkins.blueocean.rest.model.BlueRun;
+import jenkins.model.Jenkins;
+import org.apache.commons.lang.SystemUtils;
+import org.apache.lucene.analysis.standard.StandardAnalyzer;
+import org.apache.lucene.document.Document;
+import org.apache.lucene.index.DirectoryReader;
+import org.apache.lucene.index.IndexReader;
+import org.apache.lucene.index.IndexWriter;
+import org.apache.lucene.index.IndexWriterConfig;
+import org.apache.lucene.index.Term;
+import org.apache.lucene.search.IndexSearcher;
+import org.apache.lucene.search.Query;
+import org.apache.lucene.search.ScoreDoc;
+import org.apache.lucene.search.Sort;
+import org.apache.lucene.search.TopDocs;
+import org.apache.lucene.store.Directory;
+import org.apache.lucene.store.FSDirectory;
+import org.apache.lucene.store.NIOFSDirectory;
+import org.apache.lucene.store.SimpleFSDirectory;
+
+import javax.annotation.Nullable;
+import java.io.Closeable;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+
+class Index<T> implements Closeable {
+
+    @SuppressWarnings("unchecked")
+    public static Index<BluePipeline> openItems() throws IOException {
+        return openInternal(Jenkins.getInstance().getRootDir());
+    }
+
+    @SuppressWarnings("unchecked")
+    public static Index<BlueRun> openRuns(Item item) throws IOException {
+        return openInternal(item.getRootDir());
+    }
+
+    private static Index openInternal(File root) throws IOException {
+        File file = new File(root, "index");
+        if (!file.exists()) {
+            file.mkdirs();
+        }
+        // On Linux we can safely use NIOFSDirectory
+        return new Index(SystemUtils.IS_OS_WINDOWS ? new SimpleFSDirectory(file.toPath()) : new NIOFSDirectory(file.toPath()));
+    }
+
+    private final FSDirectory index;
+
+    public Index(FSDirectory index) {
+        this.index = index;
+    }
+
+    /**
+     * Add documents to index
+     * @param items the items to add to the index
+     * @param transformer to transform the item to a document
+     * @throws IOException
+     */
+    public void addDocuments(Iterable<T> items, Function<T, Document> transformer) throws IOException {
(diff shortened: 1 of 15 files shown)


blueocean-indexing/src/main/java/io/jenkins/blueocean/indexing/Transformers.java
```diff
@@ -0,0 +1,170 @@
...
+                    }
+
+                    @Override
+                    public BluePipelineNodeContainer getNodes() {
+                        return null;
+                    }
+
+                    @Override
+                    public Collection<BlueActionProxy> getActions() {
+                        return null;
```
i386: @vivek not sure if this is going to be a problem?

blueocean-indexing/src/main/java/io/jenkins/blueocean/indexing/RunLoaderImpl.java
```diff
@@ -0,0 +1,38 @@
...
+    @Inject
+    IndexService index;
+
+    @Override
+    public Iterable<BlueRun> getRuns(Job job, Link parent) {
+        Index<BlueRun> runIndex = index.getIndex(job);
+        try {
+            return runIndex.query(
+                new MatchAllDocsQuery(),
+                100, new Sort(new SortedNumericSortField(sortField(BlueRun.START_TIME), Type.LONG, true)),
```
i386: @vivek need to figure out how we can make this iterable page through hits returned via the query so we can remove the limit here. I am sure there is something todo this in Lucene but haven't had time to investigate.

blueocean-indexing/src/main/java/io/jenkins/blueocean/indexing/Index.java
```diff
@@ -0,0 +1,126 @@
...
+            }
+        }
+    }
+
+    public Iterable<T> query(Query query, int limit, @Nullable Sort sort, final Function<Document, T> transformer) throws IOException {
+        try (IndexReader reader = DirectoryReader.open(index)) {
+            final IndexSearcher searcher = new IndexSearcher(reader);
+            final TopDocs docs = searcher.search(query, limit, sort);
+            final List<ScoreDoc> scoreDocs = Arrays.asList(docs.scoreDocs);
+            return ImmutableList.copyOf(Iterables.transform(scoreDocs, new Function<ScoreDoc, T>() {
```
i386: TODO: don't read all the docs into memory

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunContainerImpl.java
```diff
@@ -38,30 +39,12 @@ public Link getLink() {
...
-            run = runList.getLastBuild();
-        }
-        return  AbstractRunImpl.getBlueRun(run, pipeline);
+        return RunLoader.get().getRun(name, job, pipeline);
     }
 
     @Override
     public Iterator<BlueRun> iterator() {
-        return RunSearch.findRuns(job, pipeline.getLink()).iterator();
+        return RunLoader.get().getRuns(job, pipeline.getLink()).iterator();
```
i386: @vivek  One thing this broke is the executing runs now don't show on /activity. Since they are continuously updating we can't put them in the index.

blueocean-indexing/src/main/java/io/jenkins/blueocean/indexing/Transformers.java
```diff
@@ -0,0 +1,170 @@
...
+            document.add(new StringField(BlueRun.TYPE, input.getType(), Store.YES));
+            document.add(new StringField(BlueRun.RUN_SUMMARY, input.getRunSummary(), Store.YES));
+            document.add(new LongField(BlueRun.START_TIME, input.getStartTime().getTime(), Store.YES));
+            document.add(new LongField(BlueRun.END_TIME, input.getEndTime().getTime(), Store.YES));
+            document.add(new LongField(BlueRun.ENQUEUE_TIME, input.getEnQueueTime().getTime(), Store.YES));
+            document.add(new LongField(BlueRun.DURATION_IN_MILLIS, input.getDurationInMillis(), Store.YES));
+            document.add(new LongField(BlueRun.ESTIMATED_DURATION_IN_MILLIS, input.getEstimatedDurtionInMillis(), Store.YES));
+            document.add(new NumericDocValuesField(sortField(BlueRun.START_TIME), input.getStartTime().getTime()));
+            document.add(new NumericDocValuesField(sortField(BlueRun.END_TIME), input.getEndTime().getTime()));
+            document.add(new NumericDocValuesField(sortField(BlueRun.ENQUEUE_TIME) + "_sort", input.getEnQueueTime().getTime()));
```
i386: 🐛  Need to remove ` + "_sort"` on the next few lines as `sortField` does the prefixing

tfennelly:
@i386 ahhh I love it .... and what a coincidence .... I was just talking to @michaelneale about this very thing last week. I was thinking of using something like CouchDB (would handle indexing etc and provides view features etc for dicing and slicing existing data in different ways).

Maybe we can make this an `@Extension` point so that we can leave the door open to implementing different "providers" ala how analytics does it.

Anyway ... I love it. I was thinking of doing this as my xmas hols scunkworks project (on the sly when the wife is out), but I'm delighted to see you doing it because when a PM does something like this then it has some chance of actually going somewhere.

i386:
> Maybe we can make this an @Extension point so that we can leave the door open to implementing different "providers" ala how analytics does it.

@tfennelly that abstraction is already there - Its called [`RunLoader`] :) (https://github.com/jenkinsci/blueocean-plugin/blob/aa18684ae04ecb007b6ced401d22a39201ebbfde/blueocean-rest/src/main/java/io/jenkins/blueocean/rest/RunLoader.java)

> I'm delighted to see you doing it because when a PM does something like this then it has some chance of actually going somewhere.

LOL well it might just at that ;)

blueocean-indexing/src/main/java/io/jenkins/blueocean/indexing/Index.java
```diff
@@ -0,0 +1,223 @@
...
+            }
+
+            @Override
+            public void close() throws IOException {
+                queryIterator.close();
+            }
+        };
+    }
+
+    public interface ClosableIterator<I> extends Iterator<I>, Closeable {}
```
i386: BOs paging code would have to call `close` somewhere up the stack of the iterator supports closing

PR #614, opened 2016-11-21
Bug/jenkins 39890 tables no ellipses

cliffmeyers:
# Description
- Work in progress; wanted to solicit some feedback before porting Branches and PR's to use similar. I'm not in love with using React components for all this
- See [JENKINS-39890](https://issues.jenkins-ci.org/browse/JENKINS-39890).
- See [JENKINS-39888](https://issues.jenkins-ci.org/browse/JENKINS-39888).
- TODO
   - [ ] Port CellRow back to a stateless function component

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


cliffmeyers:
cc @sophistifunk

blueocean-dashboard/src/main/js/components/Runs.jsx
```diff
@@ -61,61 +62,53 @@ export default class Runs extends Component {
...
-        const RunCol = (props) => <td className="tableRowLink">
-            <a onClick={open} href={`${UrlConfig.getJenkinsRootURL()}/blue${runDetailsUrl}`}>{props.children}</a>
-        </td>;
-        
+
         const openRunDetails = (newUrl) => {
             location.pathname = newUrl;
             router.push(location);
         };
-        return (<tr key={id} onClick={open} id={`${pipeline}-${id}`} >
```
cliffmeyers: This click handler was causing a double `router.push` that was breaking the "back" functionality in Run Details (JENKINS-39888)

blueocean-dashboard/src/main/less/core.less
```diff
@@ -390,9 +390,11 @@ table.highlight {
     width: 100%;
     height: 100%;
     display: block;
-    padding: 0.5em 0;
+    padding: 0.5em 0.8em 0.5em 0;
     text-decoration: none;
     color: inherit;
+    line-height: 22px; // force height to fill space to ensure clicks are "captured"
```
cliffmeyers: I don't really like this approach but the TR elements were introducing minor space above/below each TD. I tried positioning them absolutely but then there were vertical alignment issues. Maybe `flexbox` and `alignItems` would handle that better.

blueocean-dashboard/src/main/js/util/smart-fetch.js
```diff
@@ -140,25 +140,27 @@ class Pager {
...
-                    startIndex: existingData.length > 0 ? this.startIndex : first,
-                });
-                Object.freeze(outData); // children are already frozen, only shallow freeze here
-                onData(outData);
-            })
-            .catch(err => {
-                debugLog(' -- error: ', url, err);
-                onData(assignObj(concatenator(this, existingData), { $failed: err }));
-            });
+            .then(
```
sophistifunk: Could we do this in a separate PR to get it in ASAP?

blueocean-dashboard/src/main/less/core.less
sophistifunk: Agreed

blueocean-dashboard/src/main/less/core.less
```diff
@@ -390,9 +390,11 @@ table.highlight {
...
+    .text-overflow;
```
sophistifunk: TBH I'd rather the overflow stuff be attached to another class that is added by default by our `<Table>` component, with a switch to exclude that class. That way when somebody wants wrapping on one column, they're not going to ditch the entire thing. We also might want to separate the no-wrap from the text-overflow, because if I have one column that's tall because of icons or something, you might as well wrap a long line once or twice before cutting it off with ellipses.

blueocean-dashboard/src/main/js/util/smart-fetch.js
cliffmeyers: Opened #615 for this.

cliffmeyers:
@sophistifunk perhaps once jenkinsci/jenkins-design-language#118 is ready you can merge to master and publish a new JDL, or publish a beta JDL so I can incorporate those changes here?

blueocean-dashboard/src/main/less/core.less
cliffmeyers: Addressed in a9e244f - `line-height` hack no longer necessary.

blueocean-dashboard/src/main/js/components/CellLink.js
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * A table cell with an embedded react-router link.
+ */
+export function CellLink(props) {
+    const url = props.linkUrl || '';
+
+    return (
+        <td className="tableRowLink">
+            <Link to={url}>{props.children}</Link>
```
scherler: that will blow up i18n or any feature that depends on ```query``` parameters I am afraid.

Have a look in src/main/js/components/PipelineRowItem.jsx

```
                    <Link to={activitiesURL} query={location.query}>
                        <ExpandablePath path={fullDisplayPath} />
                    </Link>
```

blueocean-dashboard/src/main/less/core.less
cliffmeyers: Is that something you wanted to incorporate into a future change to the table? In jenkinsci/design-language#118 the `text-overflow` mixin is still being applied to all `td` and `td` elements nested in a `.jdl-table`

blueocean-dashboard/src/main/js/components/CellLink.js
cliffmeyers: Can you clarify? Does `i18n` require a query param to be added to every page in order to set a non-default locale?

blueocean-dashboard/src/main/js/util/smart-fetch.js
kzantow: 👍  nice catch on this

blueocean-dashboard/src/main/js/components/CellLink.js
sophistifunk: Ew, I certainly hope not. Reinventing the wheel is pants. Especially when it's square.

sophistifunk:
Regarding the text-overflow extraction into a separate style, I just wanted to get some feedback before I jumped in and changed it but I'll do that this morning if nobody objects :)

blueocean-dashboard/src/main/js/util/smart-fetch.js
michaelneale: yass - good oen - separate micro PR if this one is still baking? 

(while cliff is baking)

michaelneale:
@cliffmeyers is still still a WIP? as this has a few things in it

blueocean-dashboard/src/main/js/util/smart-fetch.js
sophistifunk: @michaelneale it's already done in another PR.

cliffmeyers:
@michaelneale yes, although I think it just needs porting to Branches and PR's which won't take long. Still waiting on clarification from @scherler and blessing/bee from @sophistifunk . Would like to get this ready for merge tomorrow.

sophistifunk:
It's all good with me 🐝

blueocean-dashboard/src/main/js/components/CellLink.js
michaelneale: yes I would hope i18n could use a cookie optionally or some other flag.

michaelneale:
One :bug: I found (not sure if specific to here) is if I open a multibranch project, and click on somewhere on the branch column it takes me back to the dashboard, which doesn't seem right. 

The reloading of the activity list also looks odd as well (when I close the modal, but not sure if related).

blueocean-dashboard/src/main/js/components/CellLink.js
scherler: well we have a couple of resolver in place ATM browser and query (that can be extended to cookies, localStore, ...). However since we do not use react-i18next anymore we would need to implement a reload on language changes (e.g. browser preferred language depends on that).

Anyway it is a general problem in our code base. Any link that is not passing the current query is removing it from the url.

For example if we implement filters or search they would use query parameter to instruct underlying code to filter results. ```?filter=type&search=multibranch``` if we do not pass this query to the links we remove this filters/search context. 

Meaning we need to be very careful when we remove the query from the url.

sophistifunk:
I've published JDL 0.0.90 with the table changes.

A couple of additional BO changes will be needed in this PR (besides updating the JDL dependency version of course):
- [ ] Remove "fixed" className from `<Table/>` use sites
- [ ] Add "u-highlight-rows" className to `<Table/>` use sites that need it. I *think* that's all of them that currently exist, but you might want to double-check.

cliffmeyers:
@michaelneale yes, good catch, that was a bug I introduced since that column is wrapped in the `IfCapability` "conditional" component. Fixed it.

cliffmeyers:
@michaelneale I think this is ready for a final round of testing. The Branches and PR's tabs have been ported over to the new logic. The only thing I haven't addressed are some alignment issues that I believe are filed in separate JIRA issues. I'd like to handle those in a follow-on PR so we can get these other bugs resolved.

cliffmeyers:
ATH passed locally.

```
2016-11-29 11:20:40,791 [Thread-3] INFO   (InputStreamHandler.java:47) com.github.eirslett.maven.plugins.frontend.lib.DefaultGulpRunner - OK. 673  total assertions passed. (6m 0s)
2016-11-29 11:20:40,831 [Thread-3] INFO   (InputStreamHandler.java:47) com.github.eirslett.maven.plugins.frontend.lib.DefaultGulpRunner - [11:20:40] Finished 'default' after 6 min
```

michaelneale:
:bee:

PR #615, opened 2016-11-22
bug / smart fetch fault handler

cliffmeyers:
# Description
- While working on an unrelated bug, I noticed that React errors were being silently swallowed on some screens. On the Activity tab, if an error occurred during render after data fetch, the screen would break with no error being printed in console. I tracked this down to the use of `catch()` in the `smart-fetch` library. I switched it to use a fault handler instead and now errors display as expected.
- Relevant discussion: https://github.com/facebook/react/issues/4199

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

…t React errors aren't swallowed

### DIFF
--- a/blueocean-dashboard/src/main/js/util/smart-fetch.js
+++ b/blueocean-dashboard/src/main/js/util/smart-fetch.js
@@ -140,25 +140,27 @@ class Pager {
             Fetch.fetchJSON(url) // Fetch data
             .then(data => augmenter.augmentCapabilities(data))
             .then(successAndFreeze)) // add success field & freeze graph
-            .then((data) => {
-                debugLog(' -- success: ', url, data);
-                // fetched an extra to test if more
-                const hasMore = data.length > limit;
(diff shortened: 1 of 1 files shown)


cliffmeyers:
Created separate PR for this at the request of @sophistifunk .  Also cc @kzantow

cliffmeyers:
@kzantow yes. I suppose another way to handle it would just be to rethrow the error from within the `catch` function. Maybe we should revisit this when we have a discussion about handling and logging errors.

scherler:
I was really hard to find the "real" change due to the unrelated whitespace changes. :(

kzantow:
@scherler you can always append [?w=1](https://github.com/jenkinsci/blueocean-plugin/pull/615/files?w=1) to the end of the `/files` URL. Unfortunately, can't comment that way...

scherler:
@kzantow right, ...or not introduce it to begin with. Anyway all good.

PR #598, opened 2016-11-14
Use global style colors for try ocean button

kxepal:
When custom Jenkins styling is used, like [jenkins-contrib-themes][1], 
the "Try" button breaks the style and looks [alien][2]. This patch fixes 
[this][3] by removing blue background, allowing to reuse global one.

[1]: https://jenkins-contrib-themes.github.io/jenkins-material-theme
[2]: http://i.imgur.com/R8B6sKS.png
[3]: http://i.imgur.com/hwxJwBj.png

PR #617, opened 2016-11-22
Fix border color for try blue ocean button

kxepal:
In the #598 PR I did a big mistake and forgot to fix border color of the button as well while I did everything right locally as screenshot shows. I just notices that issue after plugin update, when it overrided mine patched one. Sorry for that! ):


i386:
No problems and thanks for fixing it :)

michaelneale:
🚢

PR #619, opened 2016-11-23
JENKINS-39965 Add extension point for top of runs changes tab

i386:
# Description

Add an extension point at the top of the runs changes tab called jenkins.pipeline.run.changes.top.

PM by day 🎩 , plugin developer by night 🌔 

https://github.com/i386/blueocean-jira-plugin

See [JENKINS-39965](https://issues.jenkins-ci.org/browse/JENKINS-39965).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/records.jsx
+++ b/blueocean-dashboard/src/main/js/components/records.jsx
@@ -65,6 +65,7 @@ export class RunRecord extends Record({
     state: null,
     type: null,
     commitId: null,
+    actions: null,
 }) {
     isQueued() {
         return this.state === 'QUEUED';
--- a/blueocean-dashboard/src/main/js/components/RunDetailsChanges.jsx
+++ b/blueocean-dashboard/src/main/js/components/RunDetailsChanges.jsx
@@ -1,4 +1,5 @@
 import React, { Component, PropTypes } from 'react';
+import Extensions from '@jenkins-cd/js-extensions';
 import { CommitHash, EmptyStateView, ReadableDate, Table } from '@jenkins-cd/design-language';
 import Markdown from 'react-remarkable';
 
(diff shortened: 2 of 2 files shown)


i386:
PTAL @cliffmeyers @kzantow

scherler:
@i386 would it not make sense to extract the "Table" code into the default implementation of that extension point? Do we have an implementer for that extension point?

i386:
For https://github.com/i386/blueocean-jira-plugin

i386:
@scherler yes, I think we want some high level `PanelExtensionPoint` here that requires the implementor to provide an optional title and required content body.

In the screenshot below, everything within the red box is defined by my plugin. For the commits table, we only should be showing this header when there are > 1 extensions (we dont want to see the Changes header if its the only thing there).

jira

Something like:
```
class JiraPanel extends Panel {
   panelTitle() { return "JIRA Issues"; }
   renderContent() {
      return (<IssuesTable ..../>);
   }
}
```

WDYT @tfennelly @scherler @kzantow ?

blueocean-dashboard/src/main/js/components/records.jsx
```diff
@@ -65,6 +65,7 @@ export class RunRecord extends Record({
     state: null,
     type: null,
     commitId: null,
+    actions: null,
 }) {
     isQueued() {
```
i386: Perhaps we should have a helper method here like:
```
getActions(class) {
  return this.actions.filter(action => action._class == class);
}
```

i386:
@michaelneale breaks nothing but I really believe we need a better extension point here. It's given me some ideas of what I'd like to see from the API.

blueocean-dashboard/src/main/js/components/records.jsx
```diff
@@ -65,6 +65,7 @@ export class RunRecord extends Record({
     state: null,
     type: null,
     commitId: null,
+    actions: null,
```
scherler: that is fine

blueocean-dashboard/src/main/js/components/records.jsx
scherler: IMO ```getActions``` should return the object array and not only the ```_class``` 

If you want such a method then it should be something like ```getActionClasses```

michaelneale:
there are test failures here

blueocean-dashboard/src/main/js/components/records.jsx
imeredith: @scherler filter just removes objects from array, .map would return just the _clsss.

```
getActions(class) {
  if(!class) {
    return this.actions;
  }
  return this.actions.filter(action => action._class == class);
}
```

scherler: Yeah right @imeredith so better ```filterActions```

PR #626, opened 2016-11-25
[Fix_version_numbers_and_move_testing_to_prepublish_and_drop_preinsta…

scherler:
# Description

emergency version sync

…ll] Fix version numbers and move testing to prepublish and drop preinstall

### DIFF
--- a/bin/crap
+++ b/bin/crap
@@ -1,0 +1,7 @@
+#!/bin/bash
+home=`pwd`
+modules=('blueocean-dashboard' 'blueocean-personalization' 'blueocean-web')
+echo $home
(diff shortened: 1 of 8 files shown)


bin/crap
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
michaelneale: :bug: what is this file hahahaha

scherler: It saves you lots of hour of frustration with shringcrap. We can argue the name but imo that should be a script. However the important is the sync of version.

michaelneale: ok - ifyou can rename it - why not

PR #627, opened 2016-11-25
[FIX JENKINS-40027_REGRESSION_cant_open_branches_that_have_/_in_the_name]…

scherler:
# Description

See [JENKINS-40027](https://issues.jenkins-ci.org/browse/JENKINS-40027).
https://github.com/jenkinsci/blueocean-acceptance-test/pull/72

need to run above against this branch

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

… fix path


blueocean-core-js/src/js/paths/rest.js
```diff
@@ -23,7 +27,7 @@ export default {
 
     run({ organization, pipeline, branch, runId }) {
         if (branch) {
-            return `${this.pipeline(organization, pipeline)}branches/${branch}/runs/${runId}/`;
+            return `${this.pipeline(organization, pipeline)}branches/${this.encodeSlashes(branch)}/runs/${runId}/`;
```
michaelneale: -1 - this wasn't used before and a different method was used. It should encode things the way it did previously (I think there was a more general encoding funciton)>

michaelneale: also this encode function must be unit tested as well, and not be specific to just slashes necessarily. What other branch name things may need to be encoded, what was in use before etc.

michaelneale:
thanks @imeredith if you are happy with this, and it passes the ATH with thors test case to cover it, then perhaps we should merge this so we can cut a release tomorrow?

PR #628, opened 2016-11-25
[JENKINS-39804] Script to build the official BlueOcean image

ydubreuil:
# Description

See [JENKINS-39804](https://issues.jenkins-ci.org/browse/JENKINS-39804).

This is my first take at providing a set of scripts to build an official BlueOcean image. Code is based off the script used to build Jenkins weekly images (see https://github.com/jenkinsci/docker/blob/master/weekly.sh). Like for Jenkins weekly images, the idea would be to trigger `build.sh` after every release.

The image is based on top of the latest Jenkins LTS. It bundles all BO plugins and their dependencies.

### DIFF
--- a/docker/official/Dockerfile
+++ b/docker/official/Dockerfile
@@ -1,0 +1,11 @@
+# This image is based off the latest Jenkins LTS
+FROM jenkins:latest
+
+USER root
+
+COPY blueocean/target/plugins /usr/share/jenkins/ref/plugins/
+
+RUN for f in /usr/share/jenkins/ref/plugins/*.hpi; do mv "$f" "${f%%hpi}jpi"; done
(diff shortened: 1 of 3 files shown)


michaelneale:
:bee:

ydubreuil:
@michaelneale is it okay to merge? I see `ready-for-merge` tag but also  `needs-review`

michaelneale:
@ydubreuil yes - "needs-review" simply is a call for review. When ready-for-merge is added, then it means all good. I don't think this one hurts anyone so if it works for you, why not ;)

We leave the "needs-review" tag on just to make it clear it is still an active PR for people to look at.

ydubreuil:
Thanks for the explanation @michaelneale  I'm happy with it as a first step, I'm gonna merge it.

PR #629, opened 2016-11-27
Adds all the capabilities of a Action to a _capabilities collection o…

i386:
Adds all the capabilities of a Action to a _capabilities collection on the response and adds RunRecord.getActions(classOrcapability).

I was playing over the weekend with supporting other testing plugins other than JUnit (e.g. [robotframework](https://github.com/i386/robot-plugin/tree/blueocean))

The capability on the RobotBuildAction tells me I can find a similar JUnit style structure at this action (robot framework seems to have copy/pasted half of the Junit code and depended on another half - I had to return a similar data structure to JUnit here).

Responses for `Action` now include a `_capabilities` array. This contains ONLY the capabilities of the type and its super classes.

```
{
  "_class": "hudson.plugins.robot.RobotBuildAction",
  "_links": {
    "self": {
      "href": "/blue/rest/organizations/jenkins/pipelines/robotframeworktest/runs/1/robot/"
    }
  },
  "_capabilities": [
    "io.jenkins.actions.testResult"
  ],
  "failCount": 8,
  "skipCount": 0,
  "totalCount": 16,
  "urlName": "robot"
}
```

**I strongly believe we should be removing the /classes endpoint and inlining the capabilities into the response**

The hypothesis that was used to support returning `Capabilities.allSuperClassesAndCapabilities`  on `/classes` was based on the idea that plugins would not have to be modified if the plugin developer could walk its classes and detect that an action was either an instance of or subclass of a type they were interested in. This is faulty thinking - many plugins will required modification for the `@ExportedBean` or `@Exported` annotations for them to even be exposed by Blue Ocean REST. Plugins have to be modified to work with Blue Ocean.

I propose a system where any `Resource` or `Container` has a map of the capabilities that its objects support (the `RobotBuildAction` being an example for the `Resource`)

For example, a response for a container would look like:
```
{
  "_capabilities": {
    "io.example.exampleResourceA": ["io.example.capability.A", "io.example.capability.SomethingElse"],
    "io.example.exampleResourceB": ["io.jenkins.stats.coverage"]
  },
  "collection": [
    {
      "_class": "io.example.exampleResourceA"
      ...
    },
    {
      "_class": "io.example.exampleResourceB"
      ...
    },
    {
      "_class": "io.example.exampleResourceB"
      ...
    }
  ]
}
```

A resource:
```
{
  "_class": "io.example.my.resource",
  "_capabilities": [
    "io.jenkins.blueocean.pipeline"
  ]
}
```

Which would remove the need entirely for `/classes`, the extra requests but keep the responses for containers small. When a `Container` is rendered, individual `Resource`s should not return their own `_capabilities` array.

@vivek @kzantow @michaelneale

### DIFF
--- a/blueocean-dashboard/src/main/js/components/records.jsx
+++ b/blueocean-dashboard/src/main/js/components/records.jsx
@@ -65,6 +65,7 @@ export class RunRecord extends Record({
     state: null,
     type: null,
     commitId: null,
+    actions: null,
 }) {
     isQueued() {
         return this.state === 'QUEUED';
@@ -81,6 +82,17 @@ export class RunRecord extends Record({
 
     getComputedResult() {
         return this.isCompleted() ? this.result : this.state;
+    }
+
+    /** Get action by class or capability */
+    getActions(class) {
+        return run.actions.filter(action => {
+            if (action._class == class) {
+                return true;
+            }
+            var testCapability = action._capabilities.filter(capability => capability == class);
+            return testCapability.length > 0;
+        });
     }
 }
 
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ActionProxiesImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ActionProxiesImpl.java
@@ -1,12 +1,16 @@
 package io.jenkins.blueocean.service.embedded.rest;
 
 import hudson.model.Action;
+import io.jenkins.blueocean.rest.Capabilities;
 import io.jenkins.blueocean.rest.Reachable;
 import io.jenkins.blueocean.rest.hal.Link;
 import io.jenkins.blueocean.rest.model.BlueActionProxy;
 import org.kohsuke.stapler.export.ExportedBean;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+import java.util.Set;
 
 /**
  * @author Vivek Pandey
@@ -45,6 +49,12 @@ public class ActionProxiesImpl extends BlueActionProxy {
     @Override
     public String get_Class() {
         return action.getClass().getName();
+    }
+
+    @Override
+    public Collection<String> getCapabilities() {
+        Set<String> capabilities = Capabilities.allCapabilities(action.getClass());
+        return capabilities.isEmpty() ? null : capabilities;
     }
 
     @Override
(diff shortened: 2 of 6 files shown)


imeredith:
We already add _capabilities automatically to all response objects in Fetch.fetchJSON. It only triggers a call to /classes if its a capability it has not seen. We have ideas to make this use localStorage to keep the requests down even further.

This new method would work, although we would still need to transform those capabilities into _capabilities for each object we store the collections as individual objects in stores.

vivek:
>The hypothesis that was used to support returning Capabilities.allSuperClassesAndCapabilities on /classes was based on the idea that plugins would not have to be modified if the plugin developer could walk its classes and detect that an action was either an instance of or subclass of a type they were interested in. This is faulty thinking - many plugins will required modification for the @ExportedBean or @Exported annotations for them to even be exposed by Blue Ocean REST. Plugins have to be modified to work with Blue Ocean.

In some cases yes but in general we expect most plugin actions to expose it's data as it is. For example, JUnit plugin didn't need to be modified to expose all the test results. This is the main reason we are keeping maps of such API where a plugin action tells what class it is (_class element) and UI can check this _class supported capabilities using /classes API. 

Real problem is introducing _capabilities in every JSON serialized bean's (Action, ExtensionPoint, Extension...) subgraphs, that would need such changes everywhere - unless just like _class stapler or some interceptor somehow automatically adds this _capabilities element on every serialized bean.  For example, if a plugin's action exposes an object, you can capture it's capabilities but it's subgraph exposed via sub-url path relative to that plugin might expose a different set of JSON object which might serve different capabilities but they are not going to be serialized unless modified. With /classes it can work just fine as long as that bean is JSON serializable. 

This sort of proposal was discussed in early design of capabilities and was not considered in favor of doing generic lookup of capabilities using /classes API. @kohsuke feel free to comment here otherwise we can discuss it in our next design discussion this week.

i386:
@vivek Ill close this out but I'd like to chat with you further about this.

michaelneale:
would this make the response in lists even heaver as the plugin installations increase? we have an explosion of _class data already in actions for the dashboard listing, for example. Adding more to that would be a disaster (but that may not be related, if this is for not loading lists of things)

i386:
@michaelneale it is data that is fetched in a separate request. The suggestion was just to inline all the type info in every response and save on the extra fetching.

Fetching is supposed to happen once per page load but thats not the case today.

michaelneale:
@i386 ACK

PR #643, opened 2016-12-01
[JENKINS-39761] Artifact perfomance changes

imeredith:
# Description

Adds a resource for artifacts. Lists only 100 of them.

@vivek please review backend changes. 

See [JENKINS-39761](https://issues.jenkins-ci.org/browse/JENKINS-39761).

JDL PR https://github.com/jenkinsci/jenkins-design-language/pull/120
ATH PR https://github.com/jenkinsci/blueocean-acceptance-test/pull/82

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ArtifactContainerImpl.java
```diff
@@ -0,0 +1,91 @@
...
+
+            @Override
+            public Link getLink() {
+                return new Link(getUrl());
+            }
+        };
+    }
+
+    @Override
+    public Iterator<BlueArtifact> iterator() {
```
vivek: Default implementation of iterator() should not hardcode to 100.  Instead implement iterator(start, limit), this way pagination will be enforced early on for clients calling with start and limit parameters.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ArtifactContainerImpl.java
```diff
@@ -0,0 +1,91 @@
...
+    public BlueArtifact get(final String name) {
+        final VirtualFile file = run.getArtifactManager().root().child(name);
+        try {
+            if(file == null || !file.exists() || !file.isFile()) {
+                return null;
+            }
+        } catch (IOException e) {
+            throw new ServiceException.UnexpectedErrorException("Something is wrong with an artifact",e);
+        }
+        return new BlueArtifact() {
```
vivek: Why not call ArtifactImpl() instead of anonymous instance?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ArtifactImpl.java
```diff
@@ -0,0 +1,44 @@
...
+
+import hudson.model.Run;
+import io.jenkins.blueocean.rest.hal.Link;
+import io.jenkins.blueocean.rest.model.BlueArtifact;
+import org.kohsuke.stapler.Stapler;
+
+public class ArtifactImpl extends BlueArtifact {
+    final private Run run;
+    final private Run.Artifact artifact;
+    public ArtifactImpl(Run run, Run.Artifact artifact) {
```
vivek: Pass 'Reachable parent' parameter so that you can compute it's reference like you do in ArtifactContainerImpl.java.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ArtifactImpl.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @Override
+    public String getPath() {
+        return artifact.relativePath;
+    }
+
+    @Override
+    public String getUrl() {
+        return Stapler.getCurrentRequest().getContextPath() +
```
vivek: It should be returning getLink()?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ArtifactImpl.java
```diff
@@ -0,0 +1,44 @@
...
+        try {
+            return artifact.getFileSize();
+        } catch (NumberFormatException e) {
+            return 0;
+        }
+    }
+
+    @Override
+    public Link getLink() {
+        return new Link(getUrl());
```
vivek: See my above comment on using Reachable from constructor to construct self Link of Artifactimpl and then return it here.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
```diff
@@ -9,6 +9,7 @@
 import org.kohsuke.stapler.export.ExportedBean;
 import org.kohsuke.stapler.verb.POST;
 import org.kohsuke.stapler.verb.PUT;
+import org.omg.PortableInterceptor.DISCARDING;
```
vivek: Not used import?

vivek:
@imeredith Left some comments. Implementation looks fine in general, I think there is redundancy and it can be simplified easily. It will also help if you can provide some details in description of this PR, like what artifact rest URLs this implementation is going to support. Maybe it's easier to chat if not clear from my comments.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ArtifactContainerImpl.java
imeredith: Main reason is there is no way to get the single instance of the artifact,

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/ArtifactContainerImplTest.java
```diff
@@ -0,0 +1,49 @@
...
+    @Test
+    public void testArtifact() throws Exception {
+        FreeStyleProject p = j.createFreeStyleProject(JOB_NAME);
+        p.getBuildersList().add(new Shell("mkdir -p test/me/out; touch test/me/out/{{a..z},{A..Z},{0..99}}.txt"));
+        p.getPublishersList().add(new ArtifactArchiver("**/*"));
+        Run r = p.scheduleBuild2(0).waitForStart();
+
+        r = j.waitForCompletion(r);
+
+        Map artifact = request().get("/organizations/jenkins/pipelines/"+JOB_NAME+"/runs/"+r.getId()+"/artifacts/test%252Fme%252Fout%252F0.txt").build(Map.class);
```
imeredith: @vivek I can't get the get() method on ArtifactContainerImpl.java to route match, what am I doing wrong/ x.x

blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
```diff
@@ -2,21 +2,77 @@ import React, { Component, PropTypes } from 'react';
...
+
+const ZipFileDownload = (props) => {
+    const { zipFile, t } = props;
+    if (!zipFile) {
+        return null;
+    }
+
+    return (<div className="downloadAllArtifactsButton"><a className="btn-secondary" target="_blank" title={t('rundetail.artifacts.button.downloadAll', { defaultValue: 'Download all artifact as zip' })} href={`${UrlConfig.getJenkinsRootURL()}${zipFile}`}>
+                Download All
+            </a></div>
```
cliffmeyers: I don't usually nit about formatting but the line up above is quite long :)

blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
```diff
@@ -2,21 +2,77 @@ import React, { Component, PropTypes } from 'react';
...
+};
 /**
  * Displays a list of artifacts from the supplied build run property.
  */
+@observer
 export default class RunDetailsArtifacts extends Component {
+    componentWillMount() {
+        const { result } = this.props;
+        if (result) {
+            this.artifacts = this.context.activityService.fetchArtifacts(result._links.self.href);
```
cliffmeyers: Tiny nit, I wonder if it makes more sense to pass in "result" here and let `ActivityService` do the fetching.

blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
```diff
@@ -2,21 +2,77 @@ import React, { Component, PropTypes } from 'react';
...
+
+const ArtifactListingLimited = (props) => {
+    const { artifacts, t } = props;
+
+    if (!artifacts || artifacts.length < 100) {
+        return null;
+    }
+
+    return (<div className="artifactListingLimited">
+        <EmptyStateView tightSpacing><Markdown>{t('rundetail.artifacts.limit', { defaultValue: 'Only showing the first 100 artifacts' })}</Markdown></EmptyStateView>
```
cliffmeyers: Does this actually need to use Markdown? Also I don't see any changes to the resource files that add this key. Is that a 🐛  ?

blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
```diff
@@ -27,33 +83,41 @@ export default class RunDetailsArtifacts extends Component {
...
+                <ArtifactListingLimited artifacts={artifacts} t={t} />
+                <Table headers={headers} className="artifacts-table">
+                    { artifacts.map(artifact => (
+                        <tr key={artifact.url}>
+                            <td>{artifact.path}</td>
+                            <td>
+                                <FileSize bytes={artifact.size} />
+                            </td>
+                            <td className="download">
+                                <a target="_blank" title={t('rundetail.artifacts.button.download', { defaultValue: 'Download the artifact' })} href={`${UrlConfig.getJenkinsRootURL()}${artifact.url}`}>
```
cliffmeyers: Another long line :)  (sorry!)  Might be worth just building the href before actually returning the JSX.

blueocean-dashboard/src/main/less/run-details-artifact.less
```diff
@@ -0,0 +1,8 @@
+.downloadAllArtifactsButton {
+    text-align: center
+}
```
cliffmeyers: Just wanted to double check this is needed - I had thought that `.btn-secondary` would apply `text-align: center` to the `a` tag's text.

cliffmeyers (commented):
Nice job especially with finding `mobx-utils`: I hadn't ever seen that before.

Just a few small nits and one possible 🐛  with the missing i18n change? Otherwise looks good to me.

cliffmeyers:
@imeredith not anything to worry about in this PR but just thinking out loud: seems like eventually the Pager API could get tweaked to align a little more closely with the `mobx-utils` pending stuff? Consistency there would be nice.

michaelneale:
@imeredith Cliff had some comments (eg markdown) - assuming ATH on that branch passes (can you past a link?)

blueocean-dashboard/src/main/less/run-details-artifact.less
imeredith: No that centers the button in the middle of the page.

blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
imeredith: We use markdown in a lot of similar situations. The reason I added markdown is that the default look isnt right without the markdown element.

kzantow: seems weird to not just use a `markdown` function...

blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
```diff
@@ -2,21 +2,77 @@ import React, { Component, PropTypes } from 'react';
...
+ZipFileDownload.propTypes = {
+    zipFile: string,
+    t: func,
+};
+
+
+const ArtifactListingLimited = (props) => {
+    const { artifacts, t } = props;
+
+    if (!artifacts || artifacts.length < 100) {
```
kzantow: Is the limit `100`? Should this actually be `101` or `<=`?

blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
kzantow: agree, with mobx, could this just call: `activityService.getArtifacts(activity)` during `render` where that method returns an empty list, and subsequently calls to fetch it, resulting in a re-render once it's no longer pending, so we don't need to do all this ridiculousness with `componentWillMount`?

blueocean-dashboard/src/main/less/run-details-artifact.less
kzantow: maybe change the classname?

kzantow (approved):
Only minor issue is if there were exactly 100 artifacts, maybe you should return an extra 1, and only display 100 then test for size > 100?

blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
imeredith: Good catch.

blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
michaelneale: seems weird that it needs to be markdown. I can't think of a case where a translation needs rich elements in it, it should be plain text only.

imeredith: I agree. For this PR i'm leaving it how it is though.

blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
imeredith: Yeah I think you are right. Not in scope of this PR though I think.

blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
imeredith: Actaully thats correct. If its <100 we don't render it. If its >=100 we do.

imeredith: Oh actaully I get what you mean.Honestly I think we need to have a nextPage link or something that tells us if there are more.

kzantow: still needs to be < 101?

michaelneale:
I am having troubles with this passing ATH locally, even master branch, something is not right. The infamous log-karaoke/stages.js test (which is really really hard to read) chokes in some way... 

this also needs updating to master (which may be part of the problem) - although unrelated changes.

michaelneale:
via some miracle, got this to pass on CI with the master branch of the ATH: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/573/ - it seems to have upset some "timing sensitive" tests - @imeredith may need a little more time to sort out ATH issues (perhaps relaxing some timing, which I am playing with) as well as the new tests. Close...

michaelneale:
:bee: nice work, and thanks for persisting with ATH. 

I think this can go in now (don't forget ATH branch too).

PR #646, opened 2016-12-03
JENKINS-40135# NPE fix with non-block stages without steps in between

vivek:
# Description

See [JENKINS-40135](https://issues.jenkins-ci.org/browse/JENKINS-40135).

Following script cause NPE, this PR fixes that. Similar script in ATH (src/test/resources/multibranch_1 ) fails with same NPE. I think this used to pass on master, perhaps underlying workflow-api was upgraded. In any case MemoryFlowChunk.getLastNode() is returning null that its not supposed to and that resulted in NPE. A non-null check has been added and https://issues.jenkins-ci.org/browse/JENKINS-40200 is raised.

```
node {
    stage 'Stage 1'
    stage 'Stage 2'
       echo 'hello'
}
```

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees 

Root cause is MemoryFlowChunk.getLastNode() returns null, its supposed to be @Nonnull.
JENKINS-40200 raised, as work around null check is added.

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
@@ -226,7 +226,9 @@ public class PipelineNodeGraphVisitor extends StandardChunkVisitor implements No
         }
 
         TimingInfo times = null;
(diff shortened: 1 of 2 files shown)


michaelneale:
nice catch - checking with ATH but nice...

michaelneale:
:bee: I wonder how this ever worked in the ATH

vivek:
@michaelneale Probably underlying workflow-api was upgraded and it brought in this regression.

PR #638, opened 2016-12-01
FIX Jenkins 39229 regression initial stage run does not show graph

scherler:
# Description

See [JENKINS-39229](https://issues.jenkins-ci.org/browse/JENKINS-39229).
ATH test https://github.com/jenkinsci/blueocean-acceptance-test/pull/79

# Submitter checklist
- [x] Link to JIRA ticket in description, if appropriate.
- [x] Change is code complete and matches issue description
- [x] Appropriate unit or acceptance tests or explanation to why this change has no tests
- [x] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below.
- [x] Ran Acceptance Test Harness against PR changes. https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/510

# Reviewer checklist
- [ ] Run the changes and verified the change matches the issue description
- [x] Reviewed the code
- [x] Verified that the appropriate tests have been written or valid explanation given

@reviewbybees

PR #648, opened 2016-12-06
[FIX JENKINS-39955] re-run broken due to queue/executing timing/disconnect

kzantow:
# Description

Re-run button was broken in a number of cases. Added code to retry in the case of a queued item and failed node/step fetches.

Can test this with the pipeline provided in the ticket, it runs or 15 seconds with a number of stages and logs.

See [JENKINS-39955](https://issues.jenkins-ci.org/browse/JENKINS-39955).

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
+++ b/blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
@@ -80,12 +80,34 @@ export class RunDetailsPipeline extends Component {
     }
 
     componentWillReceiveProps(nextProps) {
-        if (this.props.result.isQueued()) {
+        if (nextProps.result.isQueued()) {
             return;
         }
         const followAlong = this.state.followAlong;
         this.mergedConfig = this.generateConfig({ ...nextProps, followAlong });
-
+        
+        if (this.mergedConfig.supportsNode && (!nextProps.nodes || !nextProps.nodes[this.mergedConfig.nodeKey])) {
+            const tryNodes = () => {
+                // in order to handle queued items during re-run, we need to retry fetching nodes periodically
+                if (!this.props.nodes || !this.props.nodes[this.mergedConfig.nodeKey]) {
+                    this.props.fetchNodes(this.mergedConfig);
+                }
+            };
+            setTimeout(tryNodes, 500);
+        }
+        
+        if (!nextProps.steps || !nextProps.steps[this.mergedConfig.stepKey]) {
+            const trySteps = () => {
+                this.props.fetchSteps({ ...this.mergedConfig, refetch: true });
+                if (!this.props.steps || !this.props.steps[this.mergedConfig.stepKey]) {
+                    setTimeout(trySteps, 500);
+                } else {
+                    this.setState({ followAlong: true });
+                }
+            };
+            setTimeout(trySteps, 500);
+        }
+        
         // we do not want any timeouts if we are not doing karaoke
         if (!this.state.followAlong && this.timeout) {
             clearTimeout(this.timeout);
--- a/blueocean-dashboard/src/main/js/redux/actions.js
+++ b/blueocean-dashboard/src/main/js/redux/actions.js
@@ -774,6 +774,7 @@ export const actions = {
     fetchLog(cfg) {
         return (dispatch, getState) => {
             const data = getState().adminStore.logs;
+            
             let config = cfg;
             if (!config.nodesBaseUrl) {
                 config = { ...config, nodesBaseUrl: calculateNodeBaseUrl(config) };
--- a/blueocean-dashboard/src/main/js/redux/actions.js
+++ b/blueocean-dashboard/src/main/js/redux/actions.js
@@ -774,7 +774,6 @@ export const actions = {
     fetchLog(cfg) {
         return (dispatch, getState) => {
             const data = getState().adminStore.logs;
-            
             let config = cfg;
             if (!config.nodesBaseUrl) {
                 config = { ...config, nodesBaseUrl: calculateNodeBaseUrl(config) };
--- a/blueocean-dashboard/src/main/js/redux/actions.js
+++ b/blueocean-dashboard/src/main/js/redux/actions.js
@@ -771,11 +771,12 @@ export const actions = {
      Get a specific log for a node, fetch it only if needed.
      key for cache: logUrl = calculateLogUrl
      */
-    fetchLog(config) {
+    fetchLog(cfg) {
         return (dispatch, getState) => {
             const data = getState().adminStore.logs;
+            let config = cfg;
             if (!config.nodesBaseUrl) {
-                config.nodesBaseUrl = calculateNodeBaseUrl(config);
+                config = { ...config, nodesBaseUrl: calculateNodeBaseUrl(config) };
             }
             const logUrl = calculateLogUrl(config);
             if (
--- a/blueocean-dashboard/src/main/js/redux/actions.js
+++ b/blueocean-dashboard/src/main/js/redux/actions.js
@@ -774,6 +774,9 @@ export const actions = {
     fetchLog(config) {
         return (dispatch, getState) => {
             const data = getState().adminStore.logs;
+            if (!config.nodesBaseUrl) {
+                config.nodesBaseUrl = calculateNodeBaseUrl(config);
+            }
             const logUrl = calculateLogUrl(config);
             if (
                 config.fetchAll ||


michaelneale:
gulp errors

imeredith:
:bee: To how it works now. One small issue is that it prints 404s for /steps to console. I assume that is expected since the karoke mode seems to work. If its expected it probably shouldnt be in console. 

Is there an ATH that covers this?

michaelneale:
@imeredith @kzantow no we need ATH coverage for this. 

Something like: https://github.com/jenkinsci/blueocean-acceptance-test/blob/master/src/test/js/multibranch/multibranchOpening.js#L39 - there is a test that checks that re-run button works at least - but it doesn't look for stage graph showing.

michaelneale:
Would be nice to test with ATH like: https://github.com/jenkinsci/blueocean-acceptance-test/blob/master/src/test/js/failing.js#L83

however, we need a failing job that has stages, so we can press rerun, and check stages show up again (so maybe a new test case)

michaelneale:
Passes ATH: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/560/ - @kzantow nice ot update to master just to be sure. I'll look at an ATH test if I can

kzantow:
@michaelneale I'll look at adding an acceptance test, was fighting with a github failure for a while

michaelneale:
@kzantow https://github.com/jenkinsci/blueocean-acceptance-test/pull/83 - is my suggested test

kzantow:
@imeredith yes, the problem is pipeline queue vs. running inconsistency, a running event is sent, tries to fetch the nodes & stages, but they don't always exist hence a 404. This just retries until it has data. Not the best solution... this problem with pipeline status necessitates these hacks.

michaelneale:
Trying to have a test that captures the graph updating, or perhaps the log showing up?

michaelneale:
@kzantow actually my PR for test does cover this, BUT, and this is the funny bit, the re-run isn't actually a problem unless you are in mvn hpi:run land - ie I tried on dogfood, re-run works as expected - GAH

michaelneale:
ok this gets a :bee: from me. ATH confirms it: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/bug%252FJENKINS-39955-re-run-broken/4/

Of course it always worked in non hpi:run mode - but now it is covered both on master and this branch: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/bug%252FJENKINS-39955-re-run-broken/5/

Go for it.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
```diff
@@ -80,12 +80,34 @@ export class RunDetailsPipeline extends Component {
...
             return;
         }
         const followAlong = this.state.followAlong;
         this.mergedConfig = this.generateConfig({ ...nextProps, followAlong });
-
+        
+        if (this.mergedConfig.supportsNode && (!nextProps.nodes || !nextProps.nodes[this.mergedConfig.nodeKey])) {
+            const tryNodes = () => {
+                // in order to handle queued items during re-run, we need to retry fetching nodes periodically
+                if (!this.props.nodes || !this.props.nodes[this.mergedConfig.nodeKey]) {
```
tfennelly: `this.props` or `nextProps` ?

tfennelly: Guessing `this.props` is right ... maybe a comment wrt the difference/relationship between `this.props` and `nextProps` and what's being waited for here. Waiting for a node (karaoke?) associated with the rerun to appear or something?

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
```diff
@@ -80,12 +80,34 @@ export class RunDetailsPipeline extends Component {
...
+                    this.props.fetchNodes(this.mergedConfig);
+                }
+            };
+            setTimeout(tryNodes, 500);
```
tfennelly: And what happens if it's still not there after 500ms?

tfennelly: I see below that you set another timeout to call `trySteps` again. Maybe you mean to do the same here?

michaelneale:
This doesn't fix https://issues.jenkins-ci.org/browse/JENKINS-40369 @kzantow  - I can still reproduce things even with this, so likely another issue. 

@tfennelly had some other comments. So we can either address JENKINS-40369 here or elsewhere

michaelneale:
@kzantow I think if @tfennelly comments are addressed and ATH happy - can we merge this - I just checked and this also resolves: https://issues.jenkins-ci.org/browse/JENKINS-40372 which is good. 

Can we get this in for next release please?

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
scherler: Hmm could we not relay on SSE events to fetch when needed? 

Using from  https://github.com/jenkinsci/blueocean-plugin/pull/638/files#diff-67a79c3214f72b25c36381b96fdbebc0 e.g.  `pipeline_start` event and then in the case add

```
            case 'pipeline_start':
                {
                    delete this.mergedConfig.node;
                    fetchNodes({ ...this.mergedConfig, refetch });
                }
```

That would handle start events.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
```diff
@@ -80,12 +80,34 @@ export class RunDetailsPipeline extends Component {
...
+        }
+        
+        if (!nextProps.steps || !nextProps.steps[this.mergedConfig.stepKey]) {
+            const trySteps = () => {
+                this.props.fetchSteps({ ...this.mergedConfig, refetch: true });
+                if (!this.props.steps || !this.props.steps[this.mergedConfig.stepKey]) {
```
scherler:
```
if (!this.mergedConfig.currentSteps) 
```

is the same.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
```diff
@@ -80,12 +80,34 @@ export class RunDetailsPipeline extends Component {
...
+            const tryNodes = () => {
+                // in order to handle queued items during re-run, we need to retry fetching nodes periodically
+                if (!this.props.nodes || !this.props.nodes[this.mergedConfig.nodeKey]) {
+                    this.props.fetchNodes(this.mergedConfig);
+                }
+            };
+            setTimeout(tryNodes, 500);
+        }
+        
+        if (!nextProps.steps || !nextProps.steps[this.mergedConfig.stepKey]) {
```
scherler:
```
if (!this.mergedConfig.currentSteps) 
```

is the same.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
michaelneale: perhaps worth a try @kzantow ?

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
michaelneale: @scherler you knowledge of the Deep Magic of the karaoke mode is showing :)

michaelneale:
@kzantow @scherler do you want to keep this one? likely still valid problem but only seems apparent in hpi:run mode

scherler:
@michaelneale I am looking at this again, since I am looking to solve https://issues.jenkins-ci.org/browse/JENKINS-40648 

I must admit I do not fancy the re-fetch too much and I am looking ATM do understand why the correct SSE events are not going through and do the work of refetching.

scherler:
I used two changes only of this PR and it seems that is enough to fix the re-run button

```
diff --git a/blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx b/blueocean-dashboard/src/main/js/compo
nents/RunDetailsPipeline.jsx
index e06db72a..27065eea 100644
--- a/blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
+++ b/blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
@@ -80,7 +80,7 @@ export class RunDetailsPipeline extends Component {
     }
 
-        if (this.props.result.isQueued()) {
+        if (nextProps.result.isQueued()) {
             return;
         }

```

```
diff --git a/blueocean-dashboard/src/main/js/redux/actions.js b/blueocean-dashboard/src/main/js/redux/actions.js
index f631818f..43705709 100644
--- a/blueocean-dashboard/src/main/js/redux/actions.js
+++ b/blueocean-dashboard/src/main/js/redux/actions.js
@@ -771,9 +771,13 @@ export const actions = {
      Get a specific log for a node, fetch it only if needed.
      key for cache: logUrl = calculateLogUrl
      */
-    fetchLog(config) {
+    fetchLog(cfg) {
             const data = getState().adminStore.logs;
+            let config = cfg;
+            if (!config.nodesBaseUrl) {
+                config = { ...config, nodeBaseUrl: calculateNodeBaseUrl(config) };
+            }
             const logUrl = calculateLogUrl(config);
             if (

```

kzantow:
This one turned out to be a rat hole. Closing in favor of https://github.com/jenkinsci/blueocean-plugin/pull/801 unless someone actually needs this for anything.

PR #652, opened 2016-12-08
[INFRA-987] Jenkinsfile to build the official docker image

ydubreuil:
# Description

See [INFRA-987](https://issues.jenkins-ci.org/browse/INFRA-987).

Provides a Jenkinsfile to build on Jenkins infra.

@rtyler

### DIFF
--- a/docker/official/Jenkinsfile
+++ b/docker/official/Jenkinsfile
@@ -1,0 +1,15 @@
+#!groovy
+
+/* Only keep the 10 most recent builds. */
+properties([[$class: 'BuildDiscarderProperty', strategy: [$class: 'LogRotator', numToKeepStr: '10']]])
+
+node('docker') {
+  stage('Checkout') {
+    deleteDir()
+    checkout scm
+  }
+
+  stage('Build and publish container') {
+      sh 'docker/official/build.sh'
+  }
+}


rtyler:
@ydubreuil the only thing really missing here is any sort of polling functionality either on SCM or peridic building.

How often is this supposed to be running? Depending on the answer to that, the `properties` step should be updated

michaelneale:
@rtyler @ydubreuil ideally this would be triggered of tags if possible. This means there is a 1:1 correspondence with a blue ocean release and official image - that would be ideal, right?

i386:
@michaelneale very ideal

michaelneale:
so basically `jenkinsci/blueocean:b13-1.0.0` == `jenkinsci/blueocean:latest`

updated when a new beta is published (ie a tag appears in this github repo) would be the dream.

ydubreuil:
The script is part of the blueocean repo only to be able to list all existing tags and sort them by creation date (there's no API to do that on GitHub). Blueocean code from this repo is not build at all. The script is written in a way that the latest created tag updates `jenkinsci/blueocean:latest`.

Image build is cancelled if the image corresponding to the latest tag already exists.

Triggering when a tag is created should be good as plugins are not downloaded from UC but from Maven, so when the tag for the new release is pushed, artifact are available from Maven repo (UC has a 6 to 12 hours delay IIRC).

@rtyler So the question is, is it possible to trigger a build when a tag is pushed?

rtyler:
@ydubreuil insofar that any Pipeilne can set up SCM Polling options, yes. It's possible, I haven't done it with tags specifically, but you might have to spend some time playing with the snippet generator to make something with `properties` work :frowning:

ydubreuil:
Played a bit with the snippet generator, there's nothing to trigger only when a tag is pushed. I could write some code to only build when the current HEAD is actually a tag, but I'm afraid it will be triggered before the build for that tag completes.

In a ideal world, this pipeline would be triggered when the Maven release completes.

@michaelneale can you tell how releases are done? Is the tag pushed after Maven binaries are uploaded? Is it built on the same Jenkins instance this pipeline will run?

rtyler:
> @michaelneale can you tell how releases are done? Is the tag pushed after Maven binaries are uploaded? Is it built on the same Jenkins instance this pipeline will run?

It will not, the trusted.ci instance that this Jenkinsfile will execute on is only for internal infrastructure releases and other sensitive builds/pipelines

rtyler:
I'm going to merge this, and if I can find some time this week @ydubreuil I will try to implement a Pipeline which polls tags

i386:
@ydubreuil what is the status of this one?

ydubreuil:
@rtyler did you have time to schedule this pipeline?

PR #654, opened 2016-12-10
[JENKINS-39225] Improved (reverted) German translation of Branch(es) …

svijee:
# Description

See [JENKINS-39225](https://issues.jenkins-ci.org/browse/JENKINS-39225).

I've reverted the German translations of "Branch(es)" and "Pull-Request(s)". No one uses the words "Ableger" and/or "Integrationsanfrage" for those words. Both are highly confusing. And IMHO one should not simply translate common technical english words into German. (e.g. "Pipeline" isn't translated and that's good!)

Translations seems to be included in #556 

# Submitter checklist
- [X] Link to JIRA ticket in description, if appropriate.
- [X] Change is code complete and matches issue description
- [ ] Appropriate unit or acceptance tests or explanation to why this change has no tests
- [ ] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below.
- [ ] Ran Acceptance Test Harness against PR changes.

# Reviewer checklist
- [ ] Run the changes and verified the change matches the issue description
- [ ] Reviewed the code
- [ ] Verified that the appropriate tests have been written or valid explanation given

…and Pull-Request(s)

PR #655, opened 2016-12-10
Small fixes for i18n documentation

svijee:
# Description

I've made some small fixes to the I18N documentation which I saw, when I was reading it for my other PR (#654). Nothing big actually. ;)


michaelneale:
thanks @svijee I will let @scherler approve this as he is the poly lingual expert on the team for i18n ...

PR #665, opened 2016-12-21
Bug/fix core js tests

cliffmeyers:
# Description
- Fix tests for blueocean-core-js that were broken and had been disabled (!) in gulpfile.
- core-js is now built and tested from CI via Jenkinsfile.
- We now have a "test-entrypoint.js" file that runs before the tests, useful for mocking or other config setup.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


cliffmeyers:
@michaelneale when you are online today perhaps we can touch base on this. I burned a lot of time trying to get the tests to run as part of CI but the image has a very old node and npm installed that fails to install. Hopefully your docker-fu can help me sort it out.

michaelneale:
@cliffmeyers I have a fully sick dope branch with an alt Dockerfile and Jenkinsfile that will have modern versions (based on ubuntu 16) - will send you an email with that, you can incorporate that into this branch (is a smaller change)

cliffmeyers:
@michaelneale that's awesome, excited to take a look!

michaelneale:
@cliffmeyers https://github.com/jenkinsci/blueocean-plugin/pull/667 has my changes

Perhpas if you merge in: https://github.com/jenkinsci/blueocean-plugin/pull/667/files

There is a new Dockerfile.build which contains the CI environment, and then a tiny change to the Jenkinsfile to have a step that builds the environment, and then uses it in docker pipieline (really should just be 2 line change). 

That will then give you a new nodejs version, and if you need to add more tools - can modify the Dockerfile.build for it. This was borrowed from the ATH project - which Tom and I did - and seems to work nicely.

blueocean-core-js/package.json
```diff
@@ -30,9 +29,8 @@
     "@jenkins-cd/sse-gateway": "0.0.10",
     "es6-promise": "4.0.5",
     "eslint": "2.13.1",
-    "eslint-plugin-react": "4.3.0",
```
cliffmeyers: This should not have been added to "dependencies"

blueocean-core-js/package.json
```diff
@@ -59,8 +57,8 @@
     "chai": "3.5.0",
     "del": "2.2.2",
     "enzyme": "2.4.1",
-    "eslint-plugin-react": "6.3.0",
```
cliffmeyers: This was being ignored because of the entry in "dependencies"; so bringing it to 4.3.0 so it's in alignment.

blueocean-core-js/package.json
```diff
@@ -5,8 +5,7 @@
   "main": "dist/js/index.js",
   "scripts": {
     "gulp": "gulp",
-    "test": "gulp test",
-    "prepublish": "gulp test; gulp"
```
cliffmeyers: I removed prepublish because anytime you do `npm install` it runs this, and if you are sorting out shrink issues this tends to fail all the time.

kzantow: anytime you do `npm install` _locally_, no? If you're actually doing `npm install` from npm is it doesn't actually run this, right? Seems like we probably would still want it...

blueocean-core-js/src/js/i18n/i18n.js
```diff
@@ -130,3 +134,16 @@ export default function i18nTranslator(pluginName, namespace = toDefaultNamespac
...
+
+/**
+ * Create i18n resources bound to a plugin name.
+ * @param pluginName
+ * @param i18nKeys
+ */
+export function mockTranslator(pluginName, i18nKeys = {}) {
+    const cacheKey = buildCacheKey(pluginName);
+    translatorCache[cacheKey] = function mockTranslate(key, opts = {}) {
+        const resolved = i18nKeys[key];
```
kzantow: Maybe for tests would it be even better to just return the translation key, maybe prefixed with something to indicate it was properly run through the translate function? Like... `return 'T/' + key`

blueocean-core-js/package.json
cliffmeyers: It runs before a `npm publish`, and also any time you run `npm install` without any parameters within the project folder. Running before a publish is fine; but in the other case, I've found it to be a real hassle.

blueocean-core-js/src/js/i18n/i18n.js
cliffmeyers: Used a different approach in commit b4a99ec based on your suggestion: just return the localization key. I think all of our tests want to avoid caring about i18n at all; testing against just the original key value is probably the simplest way to do check that things exist in the UI without having to take into account that a translator is involved.

cliffmeyers:
I'll put it back in, and just live with disabling it myself locally ;) Hopefully we can automate some of this stuff soon or obviate it via fancy new extensions API.

cliffmeyers:
ATH passed locally.

```
016-12-22 15:46:32,911 [Thread-3] INFO   (InputStreamHandler.java:47) com.github.eirslett.maven.plugins.frontend.lib.DefaultGulpRunner - OK. 940  total assertions passed. (9m 24s)
2016-12-22 15:46:32,983 [Thread-3] INFO   (InputStreamHandler.java:47) com.github.eirslett.maven.plugins.frontend.lib.DefaultGulpRunner - [15:46:32] Finished 'default' after 9.42 min

```

PR #666, opened 2016-12-21
Task/jenkins 40256 integrate new form controls

cliffmeyers:
# Description
- See [JENKINS-40256](https://issues.jenkins-ci.org/browse/JENKINS-40256).
- Mostly this involved tweaking button and link styles

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


blueocean-core-js/package.json
```diff
@@ -59,8 +55,8 @@
     "chai": "3.5.0",
     "del": "2.2.2",
     "enzyme": "2.4.1",
-    "eslint-plugin-react": "6.3.0",
-    "flow-bin": "^0.34.0",
+    "eslint-plugin-react": "4.3.0",
```
scherler: downgrade?

blueocean-core-js/package.json
```diff
@@ -1,12 +1,11 @@
 {
   "name": "@jenkins-cd/blueocean-core-js",
-  "version": "0.0.35",
+  "version": "0.0.37-nojdl2",
   "description": "Shared JavaScript libraries for use with Jenkins Blue Ocean",
   "main": "dist/js/index.js",
   "scripts": {
     "gulp": "gulp",
-    "test": "gulp test",
-    "prepublish": "gulp test; gulp"
```
scherler: why loose prepublish?

blueocean-core-js/package.json
cliffmeyers: The same dep was defined in both "dependencies" (as 4.3.0) and "devDependencies" (as 6.3.0). So 4.3.0 was the effective version used by the project. I removed it from "dependencies" (should not have been added there in the first place) so this change here is just adjusting the version number to what it really was before. We are using 4.3.0 in all other projects, and if I recall there are some peerDeps conflicts with js-builder when trying to use the newer version anyhow, so I think 4.3.0 is the right choice here.

cliffmeyers:
@scherler or @kzantow or anyone else, this one should be ready to go now.

michaelneale:
nice - however I think to finish this off the ATH will need to be adjusted. As the ATH isn't fully self aware yet, it uses various styles and xpaths to look for the appearance of elements. This branch seems to make it fail, eg: 

```
Timed out while waiting for element <nav.page-tabs a>
```
in the smoke tests (I think). 

So unfortunately :bug: until it is adjusted to match (I suspected a few of these changes may affect it). There is a chance they are spurious failures but they look legit to me.

cliffmeyers:
@michaelneale thanks for heads up, will look into this shortly. The class name changes were small but perhaps they've tripped something up. Don't believe I changed anything with respect to `nav.page-tabs a' but will double-check that too.

michaelneale:
@cliffmeyers yeah it seems related (that error could be a bum steer) - but worth looking into.

michaelneale:
Something looks like it has gone wrong with this branch: 

screen shot 2016-12-29 at 11 32 44 am
screen shot 2016-12-29 at 11 31 45 am

Also there are many many ATH tests failing (not just one) - so not sure if something incorrect was committed, but its definitely not ready :bug:

blueocean-core-js/src/js/NotFound.jsx
```diff
@@ -15,7 +15,7 @@ export default () => (
                 defaultValue: 'Page not found (404)',
             })}</h3>
             <div className="message">{translate('Not.found.message', { defaultValue: 'Jenkins could not find the page you were looking for. Check the URL for errors or press the back button.' })}</div>
-            <div className="actions"><Link to="/" className="btn btn-primary inverse">{translate('Open.dashboard', { defaultValue: 'Open Dashboard' })}</Link></div>
```
michaelneale: I expect this would upset the ATH, as one example

michaelneale: will need to run ATH locally and check all the breakages (likely small changes to a few strings of what to look for). 

More seriously it just looks quite wrong when I tried it

cliffmeyers:
I'll give this a serious once-over in the morning. Perhaps there was an issue during merge that I didn't catch somehow. I remember having some serious pain with shrinkwrap after the merge on Fri.

michaelneale:
@cliffmeyers right - yeah the ATH is screaming for a reason. The Run button on freestyle/non multibranch seemed to be missing too (which the ATH picked up). Shrinkwrap having a "moment". I think it has had enough of 2016!

cliffmeyers:
@michaelneale figured this one out. There were two issues, both resolved:
- While dealing with shrink merges on Fri, I published a prod version of the JDL, not realizing that the tab changes from #668 had already been merged to master in JDL. I did not rebuild and re-test the app after this (shame, shame) so I missed the funked up styling and regression. This was fixed easily in commit a51505d
- The i18n test in the ATH broke when the button classes changed slightly. I added a new "semantic" class name that will be less brittle in 6c3f889 and updated the ATH to use it in jenkinsci/blueocean-acceptance-test#91

cliffmeyers:
I ran the latest of this branch against the updated ATH code (locally) and it passed:

```
2016-12-29 10:18:47,438 [Thread-3] INFO   (InputStreamHandler.java:47) com.github.eirslett.maven.plugins.frontend.lib.DefaultGulpRunner - OK. 964  total assertions passed. (9m 28s)
```

blueocean-core-js/src/js/NotFound.jsx
cliffmeyers: This one did not break actually. The only test I saw that was selecting against a class such as `.btn  .btn-primary  .btn-secondary` was in the  "loadGerman" test. I fixed that issue in 6c3f889

cliffmeyers:
@michaelneale I think this is ready for merge but will defer to you in case you'd like to do one more round of testing on your end.

michaelneale:
PASSES ATH and looks good. https://ci.blueocean.io/job/ATH-Jenkinsfile/job/task%252FJENKINS-40256-integrate-new-form-controls/4/

:bee: nice work @cliffmeyers !

michaelneale:
BTW please merge the ATH branch before merging this @cliffmeyers

PR #667, opened 2016-12-22
Mic/new pipeline

michaelneale:
Just for reference...


cliffmeyers:
@michaelneale this worked great, and I have integrated it into #665. Thanks for the help.

michaelneale:
changes are now in thanks to cliff

PR #674, opened 2016-12-31
Just return what the artifact zip file path would be to avoid adding …

i386:
…~6ms lookup of artifacts for each run

PTAL @imeredith 

# Description

See [JENKINS-40738](https://issues.jenkins-ci.org/browse/JENKINS-40738).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


imeredith:
@i386 this saves 5ms? damn.  I think for now this is fine, in the future it would be good to. Download button in RunDetails will only be shown if there are artifacts anyway so I think its fine. :bee:

i386:
Yeah it does. Thanks @imeredith :)

PR #675, opened 2017-01-01
Exclude poorly performing FlowGraphAction from action proxy results

i386:
… FlowGraphAction) from REST responses.

I was able to get a 53x improvement on dashboard load times. Longer term we should be fetching only the actions we need.

*Before*
before

*After*
after

PTAL @vivek 

# Description

See [JENKINS-40741](https://issues.jenkins-ci.org/browse/JENKINS-40741).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ActionProxiesImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ActionProxiesImpl.java
@@ -1,5 +1,8 @@
 package io.jenkins.blueocean.service.embedded.rest;
 
+import com.google.common.base.Predicate;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Iterables;
 import hudson.model.Action;
 import io.jenkins.blueocean.rest.Reachable;
 import io.jenkins.blueocean.rest.hal.Link;
@@ -8,14 +11,18 @@ import org.kohsuke.stapler.export.ExportedBean;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
 /**
  * @author Vivek Pandey
  */
 public class ActionProxiesImpl extends BlueActionProxy {
-
     private final Action action;
     private final Reachable parent;
     private static final Logger logger = LoggerFactory.getLogger(ActionProxiesImpl.class);
+    private static final ImmutableSet<String> BANNED_ACTIONS = ImmutableSet.of("org.jenkinsci.plugins.workflow.job.views.FlowGraphAction");
 
     public ActionProxiesImpl(Action action, Reachable parent) {
         this.action = action;
@@ -53,5 +60,22 @@ public class ActionProxiesImpl extends BlueActionProxy {
             return parent.getLink().rel(getUrlName());
         }
         return null;
+    }
+
+    public static Collection<BlueActionProxy> getActionProxies(List<? extends Action> actions, Reachable parent){
+        List<BlueActionProxy> actionProxies = new ArrayList<>();
+        for(Action action : Iterables.filter(actions, new Predicate<Action>() {
+            @Override
+            public boolean apply(Action action) {
+                return !BANNED_ACTIONS.contains(action.getClass().getName());
+            }
+        })){
+            if(action == null){
+                continue;
+            }
+            actionProxies.add(new ActionProxiesImpl(action, parent));
+        }
+        return actionProxies;
+
     }
 }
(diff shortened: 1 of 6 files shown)


tfennelly:
Seeing as this is a temp workaround anyway, is there any reason why it not just added as part of `PipelineImpl.getActionProxies` for now, with the list of "banned" action types maintained in a hardcoded list?

i386:
@tfennelly either/or :)

vivek:
@i386 

We are inlining run details (latestRun) in every pipeline. UI doesn't really need all elements of it, why not UI fetch what it needs from the latestRun. For example:

```
http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/?tree=latestRun[id,result,state,startTime,estimatedDurationInMillis],*
```

We should try this instead of putting a hack to exclude certain actions (globally) instead of fixing the perf issue that is pretty localized to certain APIs. 

This also goes along with extensibility mechanism that @kzantow  discussed a month back or so. Also hard blocking actions in backend is not good idea in general, as certain UI extensions might ask for it. In fact we have been discussing using tree api for quite sometime and its getting pushed off. 

Unless we determine its hard to do tree API way, to me its 🐛

i386:
@vivek can you cook up some tickets demonstrating how this should be done? What pages should this be applied to - Dashboard, Activity, Branches, etc?

vivek:
@i386 lets start with Dashboard first. There are two API calls we need to change:

1. GET /organizations/jenkins/pipelines/

Here append with: `?tree=latestRun[id,result,state,startTime,estimatedDurationInMillis],*`

2. GET /search/?q=type:pipeline;excludedFromFlattening:jenkins.branch.MultiBranchProject,hudson.matrix.MatrixProject&filter=no-folders&start=0&limit=26

Here append:
`&tree=latestRun[id,result,state,startTime,estimatedDurationInMillis],*`

Of course tree parameters I came up with is example, someone from UI team should even trim further or add more elements as needed by the page.

For pages fetching activities and branches API, it could be on similar lines, just that the pages need to tweak elements in the tree parameters based on whats needed from response graph.

tfennelly:
@i386 my question was because it seems to me that this PR is more complicated than it needs to be. I was hoping you'd explain why you went for the more complicated option.

i386:
To make it easier to ban other options (just edit the file)

tfennelly:
@i386 tbh, I don't see how that's easier. The process would be the same either way i.e. editing a single file to add to a list. In fact, having it in a Java file would give you typing e.g. `BANNED.add(FlowGraphAction.class);` (which would also handle refactoring etc). Even if it did make adding other classes easier ... this isn't a "feature", so the added complexity doesn't warrant it imo.

Note I'm not adding a bug since I don't agree with doing that kinda thing on other people's PRs. That said, I think if I were operating in the same mode as others then I probably would add one, since I do think over-engineering is a bad idea and we should try help each other avoid it where we can.

i386:
@vivek is that actually going to work though? The `FlowGraphAction` is brought in on the `result`s actions list.

i386:
@tfennelly I don't mind how its done so long as this performance issue goes away and goes away fast

michaelneale:
this seems to be failing the ATH though: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/787/console

michaelneale:
ah, @i386 could you update this branch to master - then I think ATH may be happier. 
@vivek I think tree api and other things are worth exploring, but a smaller change like this is worthwhile considering IMO for maximum impact.

i386:
@michaelneale done.

michaelneale:
passes ATH https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/792/ now 

will kick some tyres locally...

michaelneale:
I actually like this, and kind of agree with @tfennelly that ideally it could have been just a class/constants vs a configurable file (but I guess there could be some benefit to overridding things dynamically as we are just discovering these as time goes on, and it was suggested). I think @i386 will consider changing it to be a class vs a text file. 

@vivek I don't think we have scope to change things across the dashboard to use the tree api just yet (and when we do, good...). This is overall a pretty small change.

vivek:
@michaelneale I am pretty sure there would be other actions that might cause perf impacts and I don't want it to keep adding stuff in it. Thats my main concern. I guess for this lets keep it small PR with what @tfennelly proposes and get don't with it. If we realize we need to keep adding more stuff to it, we want to reevaluate it.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ActionProxiesImpl.java
```diff
@@ -8,14 +11,18 @@
...
+
 /**
  * @author Vivek Pandey
  */
 public class ActionProxiesImpl extends BlueActionProxy {
-
     private final Action action;
     private final Reachable parent;
     private static final Logger logger = LoggerFactory.getLogger(ActionProxiesImpl.class);
+    private static final ImmutableSet<String> BANNED_ACTIONS = ImmutableSet.of("org.jenkinsci.plugins.workflow.job.views.FlowGraphAction");
```
i386: @vivek @michaelneale @tfennelly moved banned actions to a immutable set in this class.

vivek: 👍 🐝

PR #678, opened 2017-01-04
[FIX-JENKINS-40648_Input_state_isnt_updated_in_activity_branch_listing_fr…

scherler:
# Description

See [JENKINS-40648](https://issues.jenkins-ci.org/browse/JENKINS-40648).

Uses two new job events from the related PR's to fire when pipeline is paused and when it gets unpaused again. 

# Related PRs
- https://github.com/jenkinsci/sse-gateway-plugin/pull/14
- https://github.com/jenkinsci/pubsub-light-module/pull/4

[FIX-Video](https://www.youtube.com/watch?v=Dn22CTWp4_0&feature=youtu.be)
jenkins_40648_input_state_isn_t_updated_in_activity_branch_listing_from_sse

[ZEN-Video](https://youtu.be/XYw-uwmRLhU)
jenkins_40648

Contains https://github.com/jenkinsci/blueocean-plugin/pull/679 from @vivek  and parts of https://github.com/jenkinsci/blueocean-plugin/pull/648 to fix the re-run button from @kzantow 

# ath

https://github.com/jenkinsci/blueocean-acceptance-test/pull/95

# Submitter checklist
https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/761/
https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/793/

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


scherler:
I am unsure on how I can publish the related PR libs, since we need the SNAPSHOT version for it to make Jenkins pass

vivek:
@scherler https://github.com/jenkinsci/blueocean-plugin/pull/679 for your review.

scherler:
@vivek I merged your PR in this one since it works like a charm

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
```diff
@@ -160,6 +168,22 @@ private Message newMessage(PipelineEventChannel.Event event, FlowNode flowNode,
...
+                    // We will publish on the job channel
+                    try {
+                        PubsubBus.getBus().publish(new RunMessage(this.run)
+                            .setEventName(Events.JobChannel.job_run_paused)
+                        );
+                    } catch (MessageException e) {
+                        LOGGER.log(Level.WARNING, "Error publishing Run pause event.", e);
+                    }
+                }
+                message.set(PipelineEventChannel.EventProps.pipeline_is_paused, String.valueOf(pausedForInputStep));
```
tfennelly: If you look at the existing property names in `PipelineEventChannel.EventProps` you should see that they are nothing like `pipeline_is_paused`. Something like `pipeline_run_state` with values of `PAUSED` etc seem like they'd be more consistent, and will allow for other states in time Vs an explosion of property names, one foe each state.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/BlueOceanGraphListener.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+public interface BlueOceanGraphListener extends ExtensionPoint{
+    /**
+     * This event is sent when an input step moves from paused to continue state. That is when an input form is submitted.
+     *
+      * @param inputStep {@link InputStep} which got executed by sumitting parameters
+     *  @param run {@link WorkflowRun} associated with this input step
+     */
+    void onStepContinue(InputStep inputStep, WorkflowRun run);
+}
```
tfennelly: This interface smells wrong to me ... it has a very generic name i.e. `BlueOceanGraphListener`, yet seems to be tied to a very specific step i.e. `InputStep`. I'm sure I'll be told this is ok/right, but just flagging that it smells wrong to me. Prob just a naming thing.

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
tfennelly: Hmmm `pipeline_step_is_paused` is, imo, as inconsistent as `pipeline_is_paused`, but not to worry 😺

kzantow: Agree with @tfennelly, we should transition to a single property with different values for the current run state, though I agree with @scherler `pipeline_step_is_paused` is more accurate, since only one branch could be waiting on input while others are executing.

scherler: @kzantow exactly and further there can be more then one step in a pipeline paused for more input e.g. in parallel each branch can have an input step that is paused. 

...so I rename it and we are fine?

scherler: On a side note this prop is not yet used but I think it will come in handy for the karaoke mode. If you prefer we can drop it now and add it when really put to use.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/BlueOceanGraphListener.java
vivek: @kzantow @tfennelly agreed. I started at making a proper graph listener and it ended up being very specific to input step. 

@scherler do you want to rename this interface to. PipelineInputListener or BlueOceanPipelineInputListener?

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
tfennelly: Ah ok ... hadn't thought about multiple branches .... one of them paused and others running.

So @scherler, does that not mean that the pubsub changes we made are not quite right either?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/BlueOceanGraphListener.java
scherler: @vivek go nuts on it. Checkout the branch and do what you think is best. I am done now (just fixed some glitches in the matrix, ...karaoke I meant. ;)

vivek: @scherler sure, will send my changes across.

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
scherler: well @tfennelly they are good as gold. a job could sent x paused events for x parallel stages nothing in our changes would prevent that, right? the only thing we may need to implement the "matching logic" to validate the unpaused, to match back to specific step in run, but my current testing did not reveal any problems on that front.

tfennelly: I suppose we just need to be clear what "PAUSED" means i.e. it might not mean that nothing is happening i.e. it really just means "some part of it" is paused.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
```diff
@@ -144,7 +145,11 @@ public HttpResponse submitInputStep(StaplerRequest request) {
 
             Object o = parseValue(execution, JSONArray.fromObject(body.get(PARAMETERS_ELEMENT)), request);
 
-            return execution.proceed(o);
+            HttpResponse response =  execution.proceed(o);
```
kzantow: returns an `HttpResponse` what the heck?

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
```diff
@@ -172,6 +196,22 @@ private void publishEvent(Message message) {
...
 
+    @Extension
+    public static class InputStepPublisher implements PipelineInputStepListener {
+
+        @Override
+        public void onStepContinue(InputStep inputStep, WorkflowRun run) {
+            // fire an unpaused event in case the input step has received its input
+            try {
+                PubsubBus.getBus().publish(new RunMessage(run)
+                        .setEventName(Events.JobChannel.job_run_unpaused)
```
kzantow: might be more complicated than this, what if multiple stages are waiting for input?

kzantow:
Looking over this, I don't really see a differentiation between a paused job and a job waiting for input, could there be a difference? E.g. one parallel stage is waiting for input but the job is still executing? @vivek do you know?

kzantow:
@scherler @vivek I got some clarification from @jglick about this, there could be multiple input steps which each independently prompt for input, while the pipeline continues execution. We probably need to differentiate between an actual 'paused' pipeline and one that has a stage 'waiting for input'

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
vivek: Is that a problem? We want to notify listener only after calling proceed().

vivek:
@kzantow what is 'actual 'paused' pipeline'? When does that happen and how does underlying model reflect it? 

In general sense pipeline Run state will be paused even if there are multiple kind of pauses. Each node metadata could provide more info in terms of whats causing this pause. For example in case its waiting for input then there is going to be 'input' element in that node describing input step details.

kzantow:
@vivek you could also press the pause button, which doesn't rely on input, in that case the pipeline would finish each executing step and pause, awaiting re-start. But in the case of waiting for input, the entire pipeline wouldn't be paused, and in the case of receiving input, it may STILL be waiting for input...

vivek:
@kzantow Thanks. 

@michaelneale we discussed it last time, not sure what our resolution was. API is not handling pipeline 'pause'. In fact showing pipeline 'paused' if one of its input is waiting for input is confusing as there could be parallels still running. Underlying model shows it as running in this case. We should discuss Monday and resolve how we want to tackle.

kzantow:
@michaelneale @vivek from the UI perspective, it's fine to show as paused or whatever we want. My main concern is a) receiving input doesn't mean that the pipeline is continuing execution, it could STILL have other pending inputs and b) that we can differentiate between a pipeline that is actually executing irrespective of input and one that is not executing at all

vivek:
@kzantow Right. To address this, either we can use different state at run state level to distinguish the two or show run state as paused only for actual pipeline 'pause', that is in case of pipeline waiting for input is still shown as running.

michaelneale:
sounds like more discussion needed. I am not sure of overall state, but if a pipeline is waiting for input "at all" the user will likely want to know about it if fav'd or elsewhere, so in that sense it doesn't matter so much if there are still parallel things actually running, it is the pause for input that matters most to the user (if you had to have one state represent things).

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
michaelneale: this is just to prompt the front end to go fetch its new state, not the actual final word on it, so this should be ok. Its just a notification really (like "you've got mail").

scherler:
regarding the discussion on pause of a job, that is ATM not "supported" as a quick test revealed. The REST backend does not report a run as "paused". In case we want to support that we can create a new ticket. IMO for now the PR should be merged and then we do any changes in a later iteration.

michaelneale:
We should exclude "paused" from "waiting for input" for the purposes of this feature. Blue Ocean doesn't yet have a design to pause or resume a whole job without using an input step, so we have to put that to the side (even though it may be a same/similar to input, although the first time I tried to pause a job it blew up badly so I am not entirely sure it is a widely used feature compared to input).

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
```diff
@@ -165,17 +188,16 @@ export class RunDetailsPipeline extends Component {
...
-                            // console.log('nodes fetching via sse triggered');
-                            delete this.mergedConfig.node;
-                            fetchNodes({ ...this.mergedConfig, refetch });
+                            // console.log('nodes fetching via ssePipeline triggered');
+                            refetchNodes();
                         } else {
-                            // console.log('only steps fetching via sse triggered');
-                            fetchSteps({ ...this.mergedConfig, refetch });
+                            // console.log('only steps fetching via ssePipeline triggered');
+                            setTimeout(fetchSteps({ ...this.mergedConfig, refetch }), 500);
```
michaelneale: any reason for this specific delay?

pom.xml
```diff
@@ -218,7 +218,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>sse-gateway</artifactId>
-            <version>1.10</version>
+            <version>1.11-SNAPSHOT</version>
```
michaelneale: this will need to change to released version obviously...

michaelneale:
@scherler seems to work as advertised and ATH is ok - unless @kzantow objects (given that paused job is a slightly difference scope) then I think this is :bee: from me. This also needs to be udpated to used published versions of the sse plugin(s) etc etc...

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
scherler: I was trying/adopted it from the PR of the rerun button, but it did not really fixed the glitches that I can sometime see.

scherler:
@tfennelly could you assist on publishing released versions of the sse-gateway and related pubsub modules? TIA!

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
```diff
@@ -105,9 +106,13 @@ export class RunDetailsPipeline extends Component {
...
-        if (this.listener.sse) {
-            sseConnection.unsubscribe(this.listener.sse);
-            delete this.listener.sse;
+        if (this.listener.ssePipeline) {
+            sseConnection.unsubscribe(this.listener.ssePipeline);
+            delete this.listener.ssePipeline;
+        }
+        if (this.listener.sseJob) {
+            sseConnection.unsubscribe(this.listener.sseJob);
+            delete this.listener.sseJob;
```
tfennelly: In general, I'd wrap cleanup code like this in try/finally blocks so as to ensure that if something fails in an early phase of the cleanup, that that doesn't cause an exception that causes the other phases of cleanup to be skipped. Prob not an issue here as I'm sure `sseConnection.unsubscribe` catches exceptions.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
```diff
@@ -62,7 +62,8 @@ export class RunDetailsPipeline extends Component {
             }
         }
 
-        this.listener.sse = sseConnection.subscribe('pipeline', this._onSseEvent);
+        this.listener.ssePipeline = sseConnection.subscribe('pipeline', this._onSseEvent);
+        this.listener.sseJob = sseConnection.subscribe('job', this._onSseEvent);
```
tfennelly: Hmmm ... one listener receiving events from different channels? Yes it can work that way, but not sure I like that idea tbh ... seems like a recipe for confusing other people that will be reading the code => possible source of those people making bad changes to the code ... bugs etc

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
tfennelly: In fact, considering that the code is using the same listener instance in both cases (`this._onSseEvent`), then I'm wondering what's going to happen here.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
```diff
@@ -144,13 +149,31 @@ export class RunDetailsPipeline extends Component {
...
+            // we get events from the pipeline and the job channel, they have different naming for the id
+            if (event.pipeline_run_id !== runId && event.jenkins_object_id !== runId) {
                 // console.log('early out');
                 throw new Error('exit');
             }
             // we turn on refetch so we always fetch a new Node result
             const refetch = true;
+            const refetchNodes = () => {
+                delete this.mergedConfig.node;
+                setTimeout(fetchNodes({ ...this.mergedConfig, refetch }), 50);
```
tfennelly: Why's the timeout needed here? Maybe a comment would be good.

tfennelly: And what's magic about 50 ms?

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
scherler: hmm, the listener is an Object `{}` that subscribe to two different channels and then stores them. Not sure how we could do it differently.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
tfennelly: These magic numbers make me nervous and wonder are we papering over some other issue that should be resolved so as to make it more deterministic etc.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
tfennelly: It's probably ok, but I just don't like the idea of registering a single listener function instance to multiple things. It's being a bit too clever for my liking and a recipe for future bugs.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
scherler: I answered that one a bit above

> I was trying/adopted it from the PR of the rerun button, but it did not really fixed the glitches that I can sometime see

The theory is that given a certain timeout our store would fetch something more reliable. However the bigger problem is that if have a sequence like:

- job_run_queue_enter
- job_run_queue_left
- pipeline_start

which will fire three requests but since they are async we can not be sure that the will return in that order.

tfennelly:
One or two minor questions there bug overall LGTM so a 🐝 .

As for the "paused" Vs "waiting for input" ... I don't think we should be overloading the user with too many subtle "concepts". To me, "paused" can be simply defined as meaning "this run needs some user interaction in order to complete". The reason why can be multiple things and I don't really see how it helps the user to break things down (needs input or simply needs to be "unpaused") ... all they want is a cue to draw their attention to the fact that they need to get involved "here". In my view of things ... pressing the pause button is effectively the same thing as dynamically inserting an input step into the pipeline asking for user input of the form "press ok when ok to continue the run" (like a manual approval input step).

tfennelly:
@scherler `pubsub-light:1.4` and `sse-gateway:1.11` are now released.

PR #680, opened 2017-01-04
[JENKINS-37244] Consistent run terminology

imeredith:
# Description

'Build' changed to 'Run' in a few places. I did not touch any translations.

See [JENKINS-37244](https://issues.jenkins-ci.org/browse/JENKINS-37244).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


blueocean-dashboard/src/main/js/components/Activity.jsx
```diff
@@ -75,15 +75,15 @@ export class Activity extends Component {
         const head = 'pipelinedetail.activity.header';
 
         const status = t(`${head}.status`, { defaultValue: 'Status' });
-        const build = t(`${head}.build`, { defaultValue: 'Build' });
+        const run = t(`${head}.run`, { defaultValue: 'Run' });
```
i386: why do we specify the "default value"? Shouldn't this be in i18n properties?

imeredith: I assumed it was for missing translations in other languages, but maybe there is a default propsl. Maybe @scherler knows ?

scherler: Yes that is for translation only and it is the last fallback in case all i18n  services are down. You would see the `defaultValue` which by definition should be the same as in the english i18n locale.

i386:
Though it would be reasonable to expect that i18n is always working now that we bundle it as a plugin right? What say you @kzantow?

kzantow:
@i386 irrespective of bundling at as a plugin, it is implemented and functional as far as I know, and we should maintain at least the English default translations, so there is no valid reason to specify a default. I need to find the details, but it should behave like Java resource bundles, falling back to the English (default) translations if none is found for the selected locale

blueocean-dashboard/src/main/js/components/Activity.jsx
kzantow: @scherler this sounds perfect. But if "i18n services are down" that means "Jenkins is down" and we have bigger problems, I don't think it makes any sense to have defaults like this.

@imeredith 🐜  despite the shorter bits of typing, I really don't like the `const head = ...`, worse for maintainability, can't search for strings in the codebase and doing actual javascript work for no real reason...

i386:
@kzantow perhaps its is worth removing these as we go? If you agree then @imeredith should ditch the defaults for the code he's touched in this PR?

michaelneale:
:bee:

PR #682, opened 2017-01-05
JENKINS-39658# Include failed step's error to it's log

vivek:
# Description

See [JENKINS-39658](https://issues.jenkins-ci.org/browse/JENKINS-39658).

Try following pipeline script:

```
def foo = null

node {
    stage('blah') {
        sh "echo 42"
        foo.bar = 42
    }
}
```

It should appear in the log of last step. @michaelneale @i386 if you have more examples go ahead and test with them.

What I noticed is that step specific errors are present in that step's logs. Whats missing is that if there are error that are outside of this step but at the block level they will now be captured in the last step's log.

cc: @svanoort PTAL

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


vivek:
@i386 Depends, the failure might get captured in blown up step's log, if that step got recorded. If this failed step did not get recorded - no corresponding FlowNode, then previous step's log should show it's failure. Point is enclosing block's end node is marked with last failure and I am appending this log to the last step's log.

I would like to test it though, can you give me example of such archive step that fails? I tried 'step([$class: 'ArtifactArchiver', artifacts: '/target1/foo.jar', fingerprint: true])' but there is no log corresponding to it, the build is marked failure though. All I see is 'Archiving artifacts'.

i386:
@vivek did you get something in the full console log for that?

vivek:
@i386 No, nothing in console log.

i386:
@vivek I wrote a very quick step that throws an exception on a fork of this branch (see https://github.com/jenkinsci/blueocean-plugin/commit/213854399c924919b7afb7160819a7f0b892dd19).

I ran this Pipeline called `Exploding Snap`
```
pipeline {
    agent docker:'java'
    stages {
        stage("first") {
            steps {
                sh 'ping -c 5 google.com'
            }
        }
        stage("second") {
            steps {
                sh "ping -c 5 google.com"
                explode "Michael Neale broke this Pipeline. He is the worst."
            }
        }
    }
}
```

This is great as the exception message `Michael Neale broke this Pipeline. He is the worst.` is appended to the step. However, no stack trace is appended but I suspect we want stack traces in these error cases? WDYT @michaelneale ?

exploding snap 4 2017-01-06 15-15-21

🐛  now if I restart my Jenkins server the step is successful and the log can no longer be expanded. Not sure whats going on there?

exploding snap 4 2017-01-06 15-23-02

vivek:
>  I'd like to see some more tests that exercise this PR a bit more. Perhaps we could use ExplodingStep in our test code? EDIT: @michaelneale told me the error step throws an exception. Maybe use that?

I tried both. error step doesn't write log file, no LogAction so there was no log returned for error step. I added support for steps which do not have logs but there is error, such errors are going to be returned by that step's log API.

For exploding step, this step has  error (ErrorAction) as well as has log (LogAction). Its already handled by steps log API.

In both error and exploding step case, after jenkins restart ErrorAction and LogAction are missing from corresponding FlowNodes. Its reproducible via stage view plugin as well. I have open ticket on workflow-api and workflow-basic-step: https://issues.jenkins-ci.org/browse/JENKINS-40874, if its fixed there, jenkins restart issue should go away.

i386:
@vivek thanks I will let Ryan and Sam know about that ticket and we will chase it up.

michaelneale:
error step works great. 

one nit is that in the example, the step still shows as success even though it has appended the log and failure. *ideally* that would change the step to be a failure, as at first it wasn't clear to me where to look (and if it was failure, it would autoexpand). 

This is one small nit, and I am ok to :bee: this as long as we create a follow on feature to fix that if it is too hard to do now. 

This passes the ATH it looks like, so over to you @vivek

PR #683, opened 2017-01-06
[FIX JENKINS-40080] Cache class info in browser storage

tfennelly:
# Description

.. and the infra for caching other data we might want and eliminate more backend calls.

See [JENKINS-40080](https://issues.jenkins-ci.org/browse/JENKINS-40080). This JIRA was a higher priority a few weeks ago but, for some reason, has dropped out completely now.

These changes eliminate all but the min calls to get class info from the backend by storing the class info in the browser `localStorage` (via a new [@jenkins-cd/storage](https://www.npmjs.com/package/@jenkins-cd/storage) package).

The main part of the changes though (and most interesting from a wider perspective) are the parts around the `localStorage` (see 2392247), how it's invalidated etc. These are bits of infra that we can use for caching other such data on the client and so eliminate more backend calls (e.g. localization data). The actual changes for caching the classes info were minimal (see 90397e2).

# Submitter checklist
- [x] Ran Acceptance Test Harness against PR changes. [ATH: master/781](https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/781/).

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-core-js/src/js/i18n/i18n.js
+++ b/blueocean-core-js/src/js/i18n/i18n.js
@@ -15,6 +15,7 @@ export const defaultLngDetector = new LngDetector(null, {
     lookupQuerystring: 'language',
 });
 const prefix = urlConfig.getJenkinsRootURL() || '';
+const FALLBACK_LANG = '';
 
 function newPluginXHR(pluginName) {
     let pluginVersion = store.getPluginVersion(pluginName);
@@ -91,7 +92,7 @@ const pluginI18next = (pluginName, namespace = toDefaultNamespace(pluginName)) =
         defaultNS: namespace,
         keySeparator: false, // we do not have any nested keys in properties files
         debug: false,
-        fallbackLng: '',
+        fallbackLng: FALLBACK_LANG,
         load: 'currentOnly',
         interpolation: {
             prefix: '{',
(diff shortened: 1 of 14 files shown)


michaelneale:
has conflicts but seems reasonable to me. Any compatibility impacts on local storage? If caching was working "as expected" (ie not doing requests it didn't need to do because headers told browser not to) would we still want this? 

This seems to work nicely, and ATH passes so, unless this causes hassles... why not LGTM

blueocean-core-js/src/js/i18n/i18n.js
```diff
@@ -136,7 +137,13 @@ export default function i18nTranslator(pluginName, namespace) {
     const I18n = pluginI18next(pluginName, namespace);
 
     // Create and cache the translator instance.
-    translator = I18n.getFixedT(defaultLngDetector.detect(), namespace);
+    let detectedLang;
```
michaelneale: so this isn't a functional change so much as making sure it falls back correctly?

tfennelly: Right ... see the comment on a666296.

tfennelly:
@scherler Thanks man

> you are the sub-project man

Only when it makes sense, and it does here :)

PR #687, opened 2017-01-08
Do not return action proxies that are not exported beans

i386:
On my test data it took a total of 245 ms (15.75KB) to load 48 runs. After this change it took 174ms (13.5KB) to load the same data **resulting in a 71% improvement**.

OOTB it reduces the amount of actions returned for a WorkflowRun BlueRun from 12 to 5.

# Description

See [JENKINS-40880](https://issues.jenkins-ci.org/browse/JENKINS-40880).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
@@ -1,8 +1,10 @@
 package io.jenkins.blueocean.rest.impl.pipeline;
 
+import com.google.common.base.Predicate;
 import hudson.ExtensionList;
 import hudson.FilePath;
 import hudson.console.AnnotatedLargeText;
+import hudson.model.Action;
 import hudson.model.FileParameterValue;
 import hudson.model.ParameterDefinition;
 import hudson.model.ParameterValue;
@@ -31,6 +33,7 @@ import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.framework.io.ByteBuffer;
 
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import javax.servlet.ServletException;
 import java.io.IOException;
 import java.io.Reader;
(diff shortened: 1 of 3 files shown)


i386:
PTAL @vivek - not sure if you want a test for this or not considering the size.

michaelneale:
This looks like it needs a test change: 

```
java.lang.AssertionError
	at org.junit.Assert.fail(Assert.java:86)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.junit.Assert.assertNotNull(Assert.java:712)
	at org.junit.Assert.assertNotNull(Assert.java:722)
	at io.jenkins.blueocean.rest.impl.pipeline.PipelineNodeTest.getPipelineJobRunStepLogTest(PipelineNodeTest.java:1513)
```

can you check that test locally?

i386:
@michaelneale looks like `LogActionImpl` is used by BO but isn't a serializable action. WDYT @vivek?

michaelneale:
there is no risk there are other things in that boat other than LogActionImpl?

vivek:
@i386 Actual action serialization already happens if its ExportedBean. So the only saving your change is bringing is serializing wrapper object. I wonder why is there is such an expense serializing ActionProxy minus action if its not exported bean?

ActionProxiesImpl.java:

```
@Override
    public Object getAction() {
        if(action.getClass().isAnnotationPresent(ExportedBean.class)){
            return action;
        }else{
            return null;
        }
    }
```


@i386 @michaelneale LogActionImpl action is used to detect if a a step could have log or not. See logDisplayHelper.js:60. So that needs to change if we do this.

i386:
@vivek I think serialization is attempted and throws an exception in Stapler (later caught) which is the cause of the slow down.

vivek:
@i386 Maybe there is something going on, ActionProxy.getAction() is @Exported. Stapler would do nothing if ActionProxy.getAction() returns null. Thats how its supposed to work. do you see exception being thrown in that case?

vivek:
LGTM 🐝 pending ATH and CI run

michaelneale:
we don't have a functioning ATH so confidence - I am not sure about.

michaelneale:
I'll give it a good run - this is nice.

michaelneale:
:bee: passed ATH and CI for me - so go for it.

PR #690, opened 2017-01-10
[FIX JENKINS-40487] REST data page preload optimizations

tfennelly:
# Description

See [JENKINS-40487](https://issues.jenkins-ci.org/browse/JENKINS-40487).

* `PipelineStatePreloader`: Should eliminate all cold load pipeline details REST API calls e.g. `/jenkins/blue/organizations/jenkins/ATH`.
* `PipelineBranchRunStatePreloader`: Should eliminate all cold load MBP branch run details REST API calls e.g. `/jenkins/blue/organizations/jenkins/ATH/detail/activateFolder/1/pipeline/`.
* `FavoritesStatePreloader`: Should eliminate all cold load favorites REST API calls e.g. `/jenkins/blue/rest/users/tfennelly/favorites/`.

Kinda hard to manually test other than to compare the network tab in the browser from before and after. There should be fewer REST calls on a cold page load.

# Submitter checklist
- [x] Ran Acceptance Test Harness against PR changes. [ATH-Jenkinsfile/master #813](https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/813/).

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-core-js/src/js/fetch.js
+++ b/blueocean-core-js/src/js/fetch.js
@@ -268,7 +268,7 @@ export const Fetch = {
 };
 
 function trimRestUrl(url) {
-    const REST_PREFIX = 'blue/rest/organizations';
+    const REST_PREFIX = 'blue/rest/';
     const prefixOffset = url.indexOf(REST_PREFIX);
 
     if (prefixOffset !== -1) {
--- a/blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
+++ b/blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2016, CloudBees, Inc.
+ * Copyright (c) 2017, CloudBees, Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
--- a/blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
+++ b/blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
@@ -1,0 +1,83 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package io.jenkins.blueocean.preload;
+
+import hudson.Extension;
+import hudson.model.User;
+import io.jenkins.blueocean.commons.BlueUrlTokenizer;
+import io.jenkins.blueocean.commons.RESTFetchPreloader;
+import io.jenkins.blueocean.commons.stapler.ModelObjectSerializer;
+import io.jenkins.blueocean.rest.Reachable;
+import io.jenkins.blueocean.rest.model.BlueFavorite;
+import io.jenkins.blueocean.rest.model.BlueFavoriteContainer;
+import io.jenkins.blueocean.service.embedded.rest.UserImpl;
+import net.sf.json.JSONArray;
+import net.sf.json.JSONObject;
+
+import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.util.Iterator;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Preload user favorites.
+ *
+ * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
+ */
+@Extension
+public class FavoritesStatePreloader extends RESTFetchPreloader {
+
+    private static final Logger LOGGER = Logger.getLogger(FavoritesStatePreloader.class.getName());
+
+    @Override
+    protected FetchData getFetchData(@Nonnull BlueUrlTokenizer blueUrl) {
+        User jenkinsUser = User.current();
+
+        if (jenkinsUser != null) {
+            UserImpl blueUser = new UserImpl(jenkinsUser);
+            BlueFavoriteContainer favoritesContainer = blueUser.getFavorites();
+
+            if (favoritesContainer != null) {
+                JSONArray favorites = new JSONArray();
+                Iterator<BlueFavorite> favoritesIterator = favoritesContainer.iterator();
+
+                while(favoritesIterator.hasNext()) {
+                    Reachable favorite = favoritesIterator.next();
+                    try {
+                        favorites.add(JSONObject.fromObject(ModelObjectSerializer.toJson(favorite)));
+                    } catch (IOException e) {
+                        LOGGER.log(Level.FINE, String.format("Unable to preload favorites for User '%s'. Serialization error.", jenkinsUser.getFullName()), e);
+                        return null;
+                    }
+                }
+
+                return new FetchData(favoritesContainer.getLink().getHref(), favorites.toString());
+            }
+        }
+
+        // Don't preload any data on the page.
+        return null;
+    }
+}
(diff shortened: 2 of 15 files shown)


tfennelly:
There are other optimizations that could be made but not going to do those on this iteration. Will add a JIRA for those

blueocean-core-js/src/js/fetch.js
```diff
@@ -268,7 +268,7 @@ export const Fetch = {
 };
 
 function trimRestUrl(url) {
-    const REST_PREFIX = 'blue/rest/organizations';
+    const REST_PREFIX = 'blue/rest/';
```
michaelneale: I assume this is deliberate?

michaelneale:
nice, thanks for openening  https://issues.jenkins-ci.org/browse/JENKINS-40941 

Once again, needs updating due to some stuff that went in today (nothing much). wil try this out shortly!

michaelneale:
seems to not set anthing on fire, nice. 

@tfennelly if you update to master at leasure, perhaps if we can get the PR previewer back on line then we can compare this with gtmetrics and have a more numbers based comparison to see if this helps with that page load, and decide with that?

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
```diff
@@ -0,0 +1,83 @@
...
+                    }
+                }
+
+                return new FetchData(favoritesContainer.getLink().getHref(), favorites.toString());
+            }
+        }
+
+        // Don't preload any data on the page.
+        return null;
+    }
```
cliffmeyers: One thing I noticed in the different preloaders is that they have a similar code structure in `getFetchData` where they check if they should be active, build up data if they should, otherwise return null. From API standpoint, I wonder if `RESTFetchPreloader` should have an additional method like:

`boolean shouldPreload(BlueUrlTokenizer blueUrl)`

Which determines whether the extension actually does anything. It would give a little structure to that logic check, and maybe help remind people that they should be making an explicit check in that method to avoid someone accidentally writing a preloader that fires for all URL's.

If you decided that was worthwhile, then you'd just need a tweak in `RESTFetchPreloader.getStateJson()` which calls that method first and only proceeds if it returns true. 

I don't think it's a big deal, but it might result in a small improvement in the API (and slightly less nesting of if statements in `getFetchData`)

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/PipelineRunsStatePreloader.java
```diff
@@ -0,0 +1,116 @@
...
+            // saving the frontend the overhead of requesting them.
+
+            BlueRunContainer runsContainer = pipeline.getRuns();
+            Iterator<BlueRun> runsIterator = runsContainer.iterator(0, DEFAULT_LIMIT);
+            JSONArray runs = new JSONArray();
+
+            while(runsIterator.hasNext()) {
+                BlueRun blueRun = runsIterator.next();
+                try {
+                    runs.add(JSONObject.fromObject(ModelObjectSerializer.toJson(blueRun)));
```
cliffmeyers: I see this LOC pop up a few times:  `JSONObject.fromObject(ModelObjectSerializer.toJson(thing))` - would there be any benefit to adding a convenience method like `RESTFetchPreloader.toJSON` to make it a little more concise?

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/PipelineRunsStatePreloader.java
```diff
@@ -0,0 +1,116 @@
...
+        if (blueUrl.lastPartIs(BlueUrlTokenizer.UrlPart.PIPELINE)) {
+            // e.g. /blue/organizations/jenkins/f1%2Ff3%20with%20spaces%2Ff3%20pipeline/
+            return true;
+        } else if (blueUrl.lastPartIs(BlueUrlTokenizer.UrlPart.PIPELINE_TAB, "activity")) {
+            // e.g. /blue/organizations/jenkins/f1%2Ff3%20with%20spaces%2Ff3%20pipeline/activity/
+            return true;
+        }
+
+        return false;
+    }
```
cliffmeyers: You are already doing something similar to the `shouldPreload` method here, so maybe that brings some weight to the idea.

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/PipelineRunsStatePreloader.java
```diff
@@ -0,0 +1,116 @@
...
+
+    @Override
+    protected FetchData getFetchData(@Nonnull BlueUrlTokenizer blueUrl) {
+        BluePipeline pipeline = getPipeline(blueUrl);
+
+        if (pipeline != null) {
+            // It's a pipeline page. Let's prefetch the pipeline runs and add them to the page,
+            // saving the frontend the overhead of requesting them.
+
+            BlueRunContainer runsContainer = pipeline.getRuns();
```
kzantow: :bug: This would cause another case of: https://issues.jenkins-ci.org/browse/JENKINS-40924 it just needs to be switched to `.getActivities()`

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
kzantow: I'd prefer to keep it this way, keeps the preloaders simpler instead of maintaining a separate method

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
```diff
@@ -0,0 +1,83 @@
...
+                    Reachable favorite = favoritesIterator.next();
+                    try {
+                        favorites.add(JSONObject.fromObject(ModelObjectSerializer.toJson(favorite)));
+                    } catch (IOException e) {
+                        LOGGER.log(Level.FINE, String.format("Unable to preload favorites for User '%s'. Serialization error.", jenkinsUser.getFullName()), e);
+                        return null;
+                    }
+                }
+
+                return new FetchData(favoritesContainer.getLink().getHref(), favorites.toString());
```
kzantow: :ant: I would rather not do all the serialization in the preloaders, though... just return the serializable objects and have the method take care of serializing them however makes the most sense

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
```diff
@@ -0,0 +1,83 @@
...
+            BlueFavoriteContainer favoritesContainer = blueUser.getFavorites();
+
+            if (favoritesContainer != null) {
+                JSONArray favorites = new JSONArray();
+                Iterator<BlueFavorite> favoritesIterator = favoritesContainer.iterator();
+
+                while(favoritesIterator.hasNext()) {
+                    Reachable favorite = favoritesIterator.next();
+                    try {
+                        favorites.add(JSONObject.fromObject(ModelObjectSerializer.toJson(favorite)));
```
kzantow: :ant: I would rather not do all the serialization in the preloaders, though... just return the serializable objects and have the method take care of serializing them however makes the most sense

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
cliffmeyers: That's a good suggestion.

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
cliffmeyers: Personally, I'd prefer a class with a few smaller methods than a class with one large one ;)  But to each his own.

tfennelly: Yeah, this is one of those subjective/personal taste things. I prefer it as it is because, if split up, you'd then find that you'd have some code duplication between the `shouldPreload` and the `getFetchData` because the `getFetchData` would need to get many of the same things as the `shouldPreload` in order to get at the data to be preloaded (e.g. looking up runs on jobs etc). Of course you'd then say that you could split out helper methods etc, but that's just adding more conceptual weight to the whole thing, not to mention the fact that it would result in the same code searches happening multiple times per request, unless you start making these preloaders stateful etc, and that's def something we do not need to be doing.

So @cliffmeyers, if it's all the same with you, I'd prefer to leave that as is.

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/PipelineRunsStatePreloader.java
tfennelly: `shouldPreload` ?

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/PipelineRunsStatePreloader.java
tfennelly: @kzantow fair enough ... will have a look at that. Thanks Keith !

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
cliffmeyers: Yep, doesn't matter.

tfennelly:
@cliffmeyers @kzantow Thanks for the review guys. Wrt some of the comments relating to the API structure (whether `getFetchData` should return preserialized or not), they're not really relating to this PR (we're already there and are something I'd debate anyway), so I'm not planning to do anything with them here. Hope ye're ok with that.

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
```diff
@@ -0,0 +1,83 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2016, CloudBees, Inc.
```
kzantow: :ant: it's 2017 btw ;)

tfennelly: Ooops ... c&p :) Thanks.

blueocean-dashboard/src/main/java/io/jenkins/blueocean/preload/PipelineRunsStatePreloader.java
tfennelly: @kzantow Actually this wasn't a change in this PR ... just showed up because I renamed the class. Just as well though, otherwise you wouldn't have spotted it. Thanks !!!

kzantow: Ah, well I fixed it locally in the other class...

tfennelly:
@kzantow addressed the bug you spotted. See 0dbcd23.

Thanks for catching that !!

blueocean-core-js/src/js/fetch.js
tfennelly: It was indeed.

PR #692, opened 2017-01-10
[FIX-JENKINS-38805_User_can_run_parametrised_pipeline]

scherler:
# Description

See [JENKINS-38805](https://issues.jenkins-ci.org/browse/JENKINS-38805).

- extending the RunButton to be able to start parametrised jobs where ever the "old" RunButton has placed
- Refactor parameter implementation to use mobx 
- move parameter into core

# state of the nation

The feature is ready to review and playing around. 

#  ATH coverage

https://github.com/jenkinsci/blueocean-acceptance-test/pull/100

Current master ATH run on the branch https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/894/
Current branch ATH https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/919/

# Next steps
- [ ] create ATH to cover the basic steps with input parameter 

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


blueocean-dashboard/src/main/js/components/InputParameters.jsx
```diff
@@ -0,0 +1,214 @@
...
+
+    /**
+     * change a specific parameter value and update the state.
+     * @param index - which parameter we need to change
+     * @param event - the event leading to the change
+     */
+    changeParameter(index, event) {
+        // console.log('onChange', index, event);
+        const originalParameters = this.state.parameters;
+        originalParameters[index].defaultParameterValue.value = event;
```
cliffmeyers: Per discussion in Gitter, it looks like the `value` property is observable here, so modifying it outside of `@action` decorator of `action()` function will throw this error when mobx `useStrict(true)` has been set. I believe that is set globally in core-js right now.

This may seem annoying but in a way I think it's good because it makes it very explicit when state change happens. Observable / data binding can be a slippery slope whenever any code can modify the data when there are shared references.

If you refer to core-js `ToastService` that's a very simple example where modification of an array is managed by two different `@action` methods. I think this is a good practice so that we're explicit and consistent of where we do state mutations. 

On the other hand, if I understand the code correctly, you are pulling metadata about the input parameters from a job, then rendering them dynamically for user modification when launching a run. To me, you don't want to modify the underlying data in the data model from this user interaction. So I think what you probably want to do is clone the data here that drives the input parameters. If you still need advantages of observability and UI data binding, I write a quick class which encapsulates the form state and is build from a clone of the pipeline data. That way this data can be modified freely without changing the underlying data model.

michaelneale:
great to see this stuff kick off !

blueocean-dashboard/src/main/js/components/InputParameters.jsx
```diff
@@ -0,0 +1,214 @@
...
+        const ok = t('parameterized.pipeline.submit', { defaultValue: 'Build' });
+        const parametersArray = Object.keys(parameters).map(key => parameters[key]);
+
+        // console.log('state', this.state);
+        // console.log('stateToFormSubmit', this.stateParametersToArray());
+        const cancelCaption = t('rundetail.input.cancel');
+        const cancelButton = (<a title={cancelCaption} onClick={() => this.cancelForm()} className="btn inputStepCancel run-button btn-secondary" >
+            <span className="button-label">{cancelCaption}</span>
+        </a>);
+        return (<ModalView
```
i386: @scherler is this the dialog component that @sophistifunk built or a new impl of the Modal? I believe we should be using the dialog not a new component (cc @brody)

blueocean-dashboard/src/main/js/components/InputParameters.jsx
```diff
@@ -0,0 +1,214 @@
...
+ * Creating a "<form/>"less form to submit the input parameters requested by the user in pipeline.
+ *
+ * We keep all form data in state and change them onChange and onToggle (depending of the parameter
+ * type). We match the different supported inputTypes with a mapping functions
+ * @see supportedInputTypesMapping
+ * That mapping delegates to the specific implementation where we further delegate to JDL components.
+ * In case you want to register a new mapping you need to edit './parameter/index' to add a new mapping
+ * and further in './parameter/commonProptypes' you need to include the new type in the oneOf array.
+ */
+export default class InputParameters extends Component {
```
michaelneale: tiny thing, but I worry that the name "Input" may confuse people looking at code for input vs params (they are very similar). It says parameters so this is ok... perhaps "ParameterForm" or something? just a suggestion. Nice commentary ;)

scherler: noted, the whole thing needs a second iteration to remove some code redundancies. I create the PR to get feedback from cliff.

blueocean-dashboard/src/main/js/components/InputParameters.jsx
scherler: @i386 that is Modal view from jdl. I was not aware about the dialog component. will use that instead.

i386:
Thanks Thor :)

blueocean-dashboard/src/main/js/components/Activity.jsx
```diff
@@ -103,45 +108,44 @@ export class Activity extends Component {
...
             <article className="activity">
                 { showRunButton &&
-                <RunButton
-                  runnable={pipeline}
-                  latestRun={latestRun}
-                  buttonType="run-only"
-                  onNavigation={onNavigation}
-                  innerButtonClasses="btn-secondary"
-                />
+                    <ParametersRunButton
```
michaelneale: So this contains the regular run button if I read it correctly. 
This is maybe a little bit confusing name wise - it probably still should be "RunButton" but then it has components in it for normal and parametrised run button I think.

blueocean-dashboard/src/main/js/components/InputStep.jsx
```diff
@@ -114,7 +85,7 @@ export default class InputStep extends Component {
                         error.response = response;
                         throw error;
                     }
-                    return response;
+                    return response.json();
```
michaelneale: curious about this change - how did it work before?

michaelneale:
good progress, did a quick look: 

screen shot 2017-01-12 at 8 31 45 pm


Also, in that case, @brody @i386 - should it not be "abort" but "cancel" or "close" as at that point it isn't actually running, so abort doesn't make as much sense (I don't know what the design showed).

i386:
Few things here @scherler:
* That has to be `Cancel` instead of `Abort`
* The dialog is too thin. @brody can you please provide guidance here?
* I realise that there is a z-index problem that is creating the line across the dialog. This needs to be fixed in this PR or at least immediately after.
* The dialog type you are using will overflow the content area of the dialog and always show the buttons. Is that how you are using it? I believe @sophistifunk allows the dialog to show buttons in a standard way and we should be using that.

brody:
@scherler the design in zeplin is 464px wide. At least make it 450px wide and it'll be beautiful.

blueocean-dashboard/src/main/js/components/Activity.jsx
scherler: @michaelneale yes it contains the regular `RunButton` and shows it when no `parameters` are detected. 

Regarding naming `ParametersAwareRunButton` had been to long but that would be the most specific name.

scherler:
> The dialog type you are using will overflow the content area of the dialog and always show the buttons. Is that how you are using it? I believe @sophistifunk allows the dialog to show buttons in a standard way and we should be using that.

I am using it as shown in the related story: https://github.com/jenkinsci/jenkins-design-language/blob/master/src/js/stories/DialogStories.jsx#L58

https://github.com/jenkinsci/blueocean-plugin/pull/692/files#diff-e999489b5e8809d41a4a6ba443345688R134 meaning as I understand I am using it in the standard way.

scherler:
I am starting now to make it more appealing.

scherler:
screenshot from 2017-01-13 13-53-23

i386:
@scherler the "Build" button should be "Run"

michaelneale:
Getting some test failures @scherler - worth looking into

blueocean-dashboard/src/main/js/components/Activity.jsx
```diff
@@ -103,45 +106,45 @@ export class Activity extends Component {
...
-                                    pipeline,
-                                    key: index,
-                                    changeset: latestRecord,
-                                }}
-                                />
-                            );
-                        })
-                    }
-                </Table>
+                { runs.length > 0 &&
```
michaelneale: Is there any real diff here? It looks like formatting only, but its a dozen odd lines so hard to see - is there some tool doing that as it is a bit annoying as git can't tell it apart from real refactoring.

blueocean-dashboard/src/main/js/components/Activity.jsx
```diff
@@ -103,45 +106,45 @@ export class Activity extends Component {
...
+                            })
+                        }
+                    </Table>
                 }
 
-                {runs && runs.length > 0 &&
-                <button disabled={this.pager.pending || !this.pager.hasMore} className="btn-show-more btn-secondary" onClick={() => this.pager.fetchNextPage()}>
-                    {this.pager.pending ? t('common.pager.loading', { defaultValue: 'Loading...' }) : t('common.pager.more', { defaultValue: 'Show more' })}
-                </button>
+                { runs && runs.length > 0 &&
```
michaelneale: ditto with whitespace only changes???

michaelneale:
very nice - working great so far @scherler!

blueocean-dashboard/src/main/js/components/Activity.jsx
scherler: yeah sorry about that.

blueocean-web/src/main/resources/jenkins/plugins/blueocean/web/Messages_fr.properties
```diff
@@ -3,6 +3,9 @@ login=Connexion
 logout=D\u00e9connexion
 Not.found.heading=Page non trouv\u00e9e (404)
 Not.found.message=Jenkins n'a pas pu trouver la page que vous recherchiez. V\u00e9rifiez l'URL pour des erreurs ou appuyez sur le bouton de retour. 
+parametrised.pipeline.cancel=Annuler
+parametrised.pipeline.header=Input indispensable
+parametrised.pipeline.submit=Lancer
```
alecharp: I would say "Démarrer"

blueocean-web/src/main/resources/jenkins/plugins/blueocean/web/Messages_fr.properties
```diff
@@ -3,6 +3,9 @@ login=Connexion
 logout=D\u00e9connexion
 Not.found.heading=Page non trouv\u00e9e (404)
 Not.found.message=Jenkins n'a pas pu trouver la page que vous recherchiez. V\u00e9rifiez l'URL pour des erreurs ou appuyez sur le bouton de retour. 
+parametrised.pipeline.cancel=Annuler
+parametrised.pipeline.header=Input indispensable
```
alecharp: I would say "Entrée(s) indispensable" as "Input" is not a true French word. But the job configuration use the terms "parameters" so we could say "Paramètre(s) indispensable", which would be better for the configuration -> usage understanding.

scherler:
@michaelneale IMO ready for prime time. see https://github.com/jenkinsci/blueocean-acceptance-test/pull/100 and passing https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/919/

scherler:
https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/JENKINS-3.pkuu4orao08q.-pipeline/25/pipeline passed

scherler:
https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/944  passed

PR #698, opened 2017-01-11
JENKINS-40466# Show cause of blockage for a stage in DAG

vivek:
# Description

See [JENKINS-40466](https://issues.jenkins-ci.org/browse/JENKINS-40466).

@michaelneale @i386 PTAL. For now I think its fine as stages are all sequential and at present declarative only allows agent inside stages. When declarative allows parallel execution of stage then I might need some ways to associate blocking flow node in queue with the flow node while walking the graph. I have asked Jesse if we can access underlying context from queued item.task.

I tried this script:

```
pipeline {
  agent none
  stages {
    stage ('first') {
     agent { 
      label 'first' 
     }
     steps{
          sh 'echo "from first"'
      }
    }
  }
}
```

GET http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/:pipelineId/runs/:runId/nodes/

```
[
   {
      "state" : "RUNNING",
      "result" : "UNKNOWN",
      "startTime" : "2017-01-10T13:35:30.688-0800",
      "id" : "4",
      "actions" : [],
      "_links" : {},
      "durationInMillis" : 8690535,
      "causeOfBlockage" : "Waiting for next available executor",
      "_class" : "io.jenkins.blueocean.rest.impl.pipeline.PipelineNodeImpl",
      "displayName" : "first",
      "edges" : [],
      "input" : null
   }
]
```
# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


i386:
@vivek saw the comment on the JIRA ticket. Maybe easier to send Jesse a to make that field public?

michaelneale:
@vivek I think this is breaking karaoke in its current form: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/837/console - I have tried it a few times with no luck.

michaelneale:
Yes there are a few things I found: 

* Watch: 
* Note that karaoke is broken in the build stage
* Still shows up the generic waiting message (I didn't see any steps about causes of blockage)
 (this was using the https://github.com/kzantow/failure-project `michaelneale-blockscoped` branch in a multibranch project

Also: 
* ATH picks this failure up as it should
* When I tried it with the example in the description of this PR, I can see it flashes up the message quickly but then it disappears - is this expected? (I guess it is synthetic). 

So still some :bug:'s but I don't think this would be easy

vivek:
@michaelneale This PR adds new element to node element, that is causeofBlockage. I do not expect any regression except for the fact that this PR also upgrades declarative to 0.8.1 and there are breaking syntax changes. So ATH is expected to fail. Try declarative-0.8.1-fix ATH branch, I have put a fix in. For some reason when I try to run ATH locally it hangs forever but this should be it.

Upgrade to declarative is needed so that agent element is allowed inside stage block.

> Note that karaoke is broken in the build stage

But this is how it looks exactly on master. I just tried that.

> Still shows up the generic waiting message (I didn't see any steps about causes of blockage)
(this was using the https://github.com/kzantow/failure-project michaelneale-blockscoped branch in a multibranch project

Not sure what you are expecting but the changes are all at API level, for UI to show the messages it has to use newly introduced field on node 'causeOfBlockage' and display it at node level. I looked at Jenkinsfile in michaelneale-blockscoped branch, don't see any code to indicate blocking.

vivek:
@i386 Opened ticket https://issues.jenkins-ci.org/browse/JENKINS-41010

michaelneale:
ok, passes ATH - so :bee: from me (I had some misunderstandings above). Do we need anyone like @scherler to check the api will be right for the UI?

vivek:
@michaelneale Will be good to point UI mockup to @scherler. From API standpoint its matter of picking causeOfBlock from node and if its present simply display it based on the UI mockup. We can do that as part of separate ticket or as part of this. Up to you.

michaelneale:
@scherler how does this look to you? 

@vivek merge this at will if you don't hear back in a day or so.

i386:
@vivek how do you check if the whole Pipeline is blocked from executing in the case of

```
node("does-not-exist") {
  ...
}
```

?

i386:
@vivek WDYT if we showed the `node` steps up as normal steps and the cause of blockage hung off of those? That way they could be `running` and their display name could be the `causeOfBlockage` or a descrption of that `node` step (we will get this when Sam implements step names) - then there would be nothing for us to implement on the UI and the user would get more info on how nodes were allocated as part of their run?

vivek:
@i386 this is doable. API can always return this step with name 'CauseOfBlockage: ....' of a blocked stage, no log message to keep things simple. Needs no support from step names as such as its a fabricated step return by /nodes/:id/steps/ API.

i386:
@vivek in this case would the `node` step show up after the fact? e.g. when the run is successful

i386:
@vivek to be clear what I am after would look like this with no frontend changes

artboard

If that is doable could we reopen JENKINS-40466?

i386:
Where ever we could not tie the blockage to a step (freestyle or pipeline level `agent`) we would show something like

blueocean sketch 2017-01-14 10-34-08

vivek:
@i386 I repoed JENKINS-40466. Will do a separate PR though.

i386:
@vivek thank you 👍

PR #700, opened 2017-01-11
[FIX JENKINS-39794] SSE event for queue task complete - for MBP indexing

tfennelly:
# Description

Depends on:
* https://github.com/jenkinsci/sse-gateway-plugin/pull/16
* https://github.com/jenkinsci/pubsub-light-module/pull/5

See [JENKINS-39794](https://issues.jenkins-ci.org/browse/JENKINS-39794).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


cliffmeyers:
@tfennelly @michaelneale @vivek here's my findings with the first POC of this:

If I create a project pointing to a Git repo I don't have permissions to, OR a Git repo without a Jenkins file, the result is basically the same:

No Permissions Scenario
git-create-1-no-permissions

No Jenkinsfile Scenario
git-create-2-no-jenkinsfile

If successful, I see a bunch of events, as I would expect:
git-create-3-successful

So as it stands now, the UI logic could do the following:

1. After creation REST API call completes, begin to listen for SSE with job URL that starts with the HAL href that was returned by the REST API
2. Once we receive a "job_crud_created" for a branch with a HAL href that starts with the above href, we know that Jenkins has at least scanned the repo and found something useful.

We have some limitations though:
1. We don't really know if there was an error. After we receive the "job_run_queue_left" event, we could start a timer and hope that new events flow in soon, but we can't really distinguish between "this isn't a valid job" and the executors just being busy.
2. We can't provide any useful feedback about the error.

Question for @michaelneale or @i386: should we let the user sit for 30s or so, then show a message like "the system is still working, taking you to job now..." and then perhaps direct them to the activity tab after that... which would at least allow them an easy way to get to classic jenkins and dig into the details.  And if we receive the "job_crud_created" event we show more of a success message but then navigate them to /activity right after?

tfennelly:
> We don't really know if there was an error. After we receive the "job_run_queue_left" event, we could start a timer and hope that new events flow in soon, but we can't really distinguish between "this isn't a valid job" and the executors just being busy.

When you get this event I wonder is there an API call you ca make to get the status?

> could start a timer and hope

And monkeys might fly out of my butt 🐒  ... you know you're in trouble when you're resorting to magic timers in the hope that something might have happened in the timer window 😨

cliffmeyers:
@tfennelly yeah, it's absolutely not a long-term solution. I don't know options exist within these plugins to propagate new events out, or if things need to be added whether they can even be released and made available within our release window. So, I'm trying to be creative so we can at least get _something_ :)

The best solution would be if underlying plugin actually fired out an error event when something goes wrong with some information about what happened... but we still need to wait for it. I'd rather add a timer with an acceptable threshold that tells the user "sorry, something bad happened, maybe go here instead?" than leave them on a page that says "Waiting..." indefinitely.

michaelneale:
@cliffmeyers 

> Question for @michaelneale or @i386: should we let the user sit for 30s or so, then show a message like "the system is still working, taking you to job now..." and then perhaps direct them to the activity tab after that... which would at least allow them an easy way to get to classic jenkins and dig into the details. And if we receive the "job_crud_created" event we show more of a success message but then navigate them to /activity right after?

Hrm... yeah I think some timeout with a message is a good start, but we will need a way at some point to pull the log/error messages in tht case. We can't rely on going back to classic to diganose as it is really hard to. We may need a new api to yield this data though. 

Are there no events when things do go wrong? There is a log for org and branch scanning, which should be exposed via an api (and some status) so even outside of SSE we should be able to pull out the fact it is good or bad, and a log of why it went bad and display it. 

Aside: should we move discussion to email list?

tfennelly:
Where are we on this? Is there an API that can be called after the event to get some detail? Maybe there's some message enrichment we can do too if there's some aditional error/state data on the queue item used to create the event.

tfennelly:
Ok, some changes that I think give @cliffmeyers what he needs.

There's a new event now, named `job_run_queue_task_complete`. Look for those events on the frontend where `job_ismultibranch: "true"`.

screenshot 2017-01-18 11 26 51

cliffmeyers:
Would be good to see this one land sometime this week. Perhaps @michaelneale or @vivek could give some input as to the Java code? Functionally this is great from my standpoint.

tfennelly:
Okay, I'm going to spin news versions of `pubsub-light` and `sse-gateway`.

I'm going to convert `pubsub-light` from a `jenkins-module` and into a `hpi`. I was originally thinking it might make its way into jenkins core at some point (and so the `jenkins-module` made sense) but I don't see that ever happening now. Making it a `hpi` would be a lot less trouble in future because it would mean we could make changes to `pubsub-light` (e.g. new events like for this) without having to also do a release of the `sse-gateway`.

tfennelly:
@cliffmeyers made changes to use new plugins + rebased etc.

blueocean-events/pom.xml
```diff
@@ -16,6 +16,10 @@
     <dependencies>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>pubsub-light</artifactId>
```
michaelneale: gosh git is so terrible at XML diffs. Still. yet maven keeps using xml

michaelneale:
code looks OK to me - new dependency - pubsub lite? is that a new module? 

Would this be somewhat covered or testable via the ATH?

cliffmeyers:
@tfennelly was having a chat related to this with @i386 and he was keen on us digging a little deeper to see if there's more information we could put into the error event itself. I'm trying to recall if the outcome from the `Future` had an exception as to why it failed? I seem to remember a "result" field that we'd have to cast to get at, but perhaps we could do that in the Enricher? I will touch base with you on this again tomorrow when I'm awake and online.

cliffmeyers:
cc @vivek (see above) as we had discussed getting at similar data through the API.

michaelneale:
:bee: if cliff is happy!

tfennelly:
@cliffmeyers I'll have a look to see if there's an error we can attach to the event.

tfennelly:
@cliffmeyers @i386 @michaelneale After checking in a debugger and then asking @jglick, I can confirm that there's no programmatic way of attaching the cause of the indexing failure to the generated event. The raw SUCCESS/FAILED result is all that's available atm. Seems like we'll need a REST API for getting at the cause of the failure via parsing of the indexing task log.

michaelneale:
:bee:

tfennelly:
Btw ... blocked on this waiting for https://github.com/jenkins-infra/repository-permissions-updater/pull/198 to be resolved.

PR #706, opened 2017-01-12
[hotfix/syntaxError] JENKINS-40977 

scherler:
# Description

See [JENKINS-40977](https://issues.jenkins-ci.org/browse/JENKINS-40977).

With the help of FF I could limit the problem to some JSON parse method. Surrounding all with try catch gets rid of the error in the console. 

screen shot 2017-01-12 at 10 52 05 am

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


cliffmeyers:
So this fixes the errors in the console but have we not found the actual underlying error causing it? I assume maybe it's malformed data being created by prefetch?

blueocean-core-js/src/js/fetch.js
```diff
@@ -290,7 +290,10 @@ function getPrefetchedDataFuture(url) {
                 // the prefetched data as the value.
                 if (trimRestUrl(preFetchEntry.restUrl) === trimmedUrl) {
                     try {
-                        return Promise.resolve(JSON.parse(preFetchEntry.data));
+                        const parse = JSON.parse(preFetchEntry.data);
+                        return Promise.resolve(parse);
+                    } catch (e) {
+                        console.error('preFetch Error', e);
```
tfennelly: How about adding `preFetchEntry.data` to the error logging so we can see what's wrong there?

tfennelly: And the same in the other places where parse errors are being caught.

cliffmeyers: Perhaps we should actually rethrow the error with prefetch.data attached, so we can see it in rollbar? My understanding is that we haven't been able to repro this locally, but if it shows up in rollbar then probably easier for us to pin down what's going on.

scherler: @tfennelly you mean to the `catch`? I can do that but the problem the error is logged into the nirvana

tfennelly: Whatever ... I think it would be useful to somehow see that data and get an idea why it's not parseable.

kzantow: +1, let's solve this at the root of the issue

scherler: if we would know the root cause, we could do that.

cliffmeyers: Could we rethrow a new error in the catch block which has the prefetch.data attached to it? So the details show up in rollbar?  (unless console.errors show up automatically, but I didn't think they do)

```
} catch (e) {
    console.error('preFetch Error', e);
    const err = new Error();
    error.prefetchData = preFetchEntry.data;
    throw err;
}
```

scherler: @cliffmeyers I will have a look. However the prefetch seems not the root cause. Still trying to pin down where the problems happens. FF were throwing problems in parsing the sse-event so maybe it is some faulty events. Still debugging some more.

scherler: hmm interesting, I just reverted the core-js changes and only left the dashboard one and now I cannot see the problem anymore. It may be really a problem of faulty sse-event markup.

blueocean-core-js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@jenkins-cd/blueocean-core-js",
-  "version": "0.0.45",
+  "version": "0.0.46-unpublished",
```
kzantow: why this? 👎  agree with @tfennelly we should just not keep adding random suffixes to the version

scherler: that is new. Until now we used the "maven" way. Things in master are "snapshots" if that has changed then we need to make this public knowledge.

scherler: Anyway fixed that now

cliffmeyers:
I'm a little confused - thought this PR was fixing the syntax error thing? Now all I see in the diff is removal of some `setTimeout`s wrapping `fetchNodes` ?

scherler:
@cliffmeyers and moving the parse of the sse-event into the try-catch block

scherler:
@kzantow done

michaelneale:
nice one @scherler!

I'll raise a glass of german beer tonight, this one was annoying. I thought it may be bad json data from the event but couldnt' see it.

i386:
@scherler fixing this is so awesome

scherler:
@i386 @michaelneale danke  und Prost ;)

PR #710, opened 2017-01-14
JENKINS-41047# ErrorMessage improvement

vivek:
# Description

See [JENKINS-41047](https://issues.jenkins-ci.org/browse/JENKINS-41047).

@cliffmeyers Its implemented as we discussed. I have fixed error messages for git and GitHub creation APIs. Please review.

```
POST http://localhost:8080/blue/rest/organizations/jenkins/pipelines/
{
  "scmConfig" : { }
}
```

Should result in this error message:

```
{
  "message" : "Failed to create Git pipeline",
  "code" : 400,
  "errors" : [ {
    "message" : "name is required",
    "code" : "MISSING",
    "field" : "name"
  }, {
    "message" : "$class is required",
    "code" : "MISSING",
    "field" : "$class"
  } ]
}
```

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/ErrorMessage.java
+++ b/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/ErrorMessage.java
@@ -1,0 +1,80 @@
+package io.jenkins.blueocean.commons;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Describes JSON based error message.
+ *
+ * {
+ *     "code": 401,
+ *     "message": "login failed",
+ *
+ *     "errors" : [ {
+ *          "message" : "Invalid password",
+ *          "code" : "INVALID",
+ *          "field" : "password"
+ *     } ]
+ * }
+ *
+ * @author Vivek Pandey
+ **/
+public class ErrorMessage {
+
+    public final String message;
+
+    public final int code;
+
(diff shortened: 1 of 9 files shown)


michaelneale:
Seems good to me - :bee: as long as ATH and @cliffmeyers happy (mostly @cliffmeyers, as he knows what to expect)

vivek:
@cliffmeyers At present these are well known codes, I have added them in README,https://github.com/jenkinsci/blueocean-plugin/tree/improvement/JENKINS-41047/blueocean-rest#error-message. I can further improve the doc I have if it helps?

cliffmeyers:
@vivek I was thinking along the lines of specific error codes for specific fields. For example, would a developer integrating with the Git creation REST API know that unique name is a requirement, and that ALREADY_EXISTS is an error they might receive? Perhaps we don't need to document to that extent, but if we don't, it would be hard for the client (such as the UI) to know which errors to try to account for.

This is not a blocker for merge of this PR just something I figured was worth discussing / thinking about.

vivek:
@cliffmeyers I see. I was thinking documenting these errors at one place, like what I have now. These error codes meaning is generic and won't change from APIs to apis. I guess you are saying mention these codes on each APIs that report them so that clients know they need to look at them, right?

cliffmeyers:
@vivek I think for the top-level error codes (400, 401) that is fine. I was talking more about the specific field-level errors in the `errors` array. If we have some very specific validation or business logic error that only makes sense in the context of a single API endpoint, putting that in a general list might be confusing.

I'm really fine with it either way, and just trying to think ahead a bit. Putting them in one place is fine, and if for some reason it becomes a problem later, we can always change it then.

PR #711, opened 2017-01-14
Internalize Stapler so we can move away from core releases

i386:
# Description

PTAL @vivek 

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/Export.java
+++ b/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/Export.java
@@ -1,0 +1,94 @@
+package io.jenkins.blueocean.commons.stapler.export;
+
+import hudson.ExtensionList;
+import io.jenkins.blueocean.commons.stapler.export.TreePruner.ByDepth;
+import jenkins.model.Jenkins;
+import jenkins.security.SecureRequester;
+import org.kohsuke.stapler.StaplerRequest;
+import org.kohsuke.stapler.StaplerResponse;
+
+import javax.servlet.ServletException;
+import java.io.IOException;
+import java.io.Writer;
+import java.net.HttpURLConnection;
+
+public class Export {
+
+    static ModelBuilder MODEL_BUILDER = new ModelBuilder();
+
+    public static void doJson(StaplerRequest req, StaplerResponse rsp, Object bean) throws IOException, ServletException {
+        if (req.getParameter("jsonp") == null || permit(req, bean)) {
+            rsp.setHeader("X-Jenkins", Jenkins.VERSION);
+            rsp.setHeader("X-Jenkins-Session", Jenkins.SESSION_HASH);
+            serveExposedBean(req, rsp, bean, req.getParameter("jsonp") == null ? Flavor.JSON : Flavor.JSONP);
+        } else {
+            rsp.sendError(HttpURLConnection.HTTP_FORBIDDEN, "jsonp forbidden; implement jenkins.security.SecureRequester");
+        }
+    }
+
+    private static boolean permit(StaplerRequest req, Object bean) {
+        for (SecureRequester r : ExtensionList.lookup(SecureRequester.class)) {
+            if (r.permit(req, bean)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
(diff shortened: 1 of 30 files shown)


blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/Export.java
```diff
@@ -0,0 +1,94 @@
+package io.jenkins.blueocean.commons.stapler.export;
```
vivek: I guess Export is our own code, not moved in from Stapler. We should keep it separate from rest of Stapler classes. For example once core allows plugin specific stapler jar and we remove exported package and change Export.doJson() without affecting other parts of the code. So maybe move it to `io.jenkins.blueocean.commons.stapler`?

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/pageable/PagedResponse.java
```diff
@@ -21,7 +21,7 @@
 /**
  * Renders {@link Pageable} to HTTP by honoring the current page, links to next page, etc.
  *
- * @author Kohsuke Kawaguchi
+ * @author Kohsuke KawaguchiF
```
vivek: typo:)

vivek:
@i386 code wise looks good to me, will do some testing 👍 

I think we should have a script to copy code from stapler exported package in future, for example any bug fixes in stapler exported package, we want to bring in. Such as https://github.com/stapler/stapler/pull/89.

Some concerns/observation:

- As it sounds from the PR title, it will fix issue of upgrading core for stapler issues is not completely true. Yes it will avoid upgrade for json serialization functionality but not for the rest of stapler functionalities. So any fixes in stapler web framework would still require upgrade. 
- As discussed, its temporary measure - that is till core provides a way to use plugin mandated stapler version without upgrade of core, we keep this solution around.

vivek:
@i386 `mvn clean install` fails due to javadoc generation error. Following patch in blueocean-commons/pom.xml fixes it (exclude exported package from generation):

```
diff --git a/blueocean-commons/pom.xml b/blueocean-commons/pom.xml
index 833331d..10df76d 100644
--- a/blueocean-commons/pom.xml
+++ b/blueocean-commons/pom.xml
@@ -21,4 +21,17 @@
     </dependency>
   </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <version>2.10.4</version>
+        <configuration>
+          <excludePackageNames>io.jenkins.blueocean.commons.stapler.export</excludePackageNames>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
 </project>
```

Overall looks good! I completed with my testing. Tree api works fine for both pipelines and runs API. ATH passes. Pending this pom changes and some comments/concerns posted in earlier comment, LGTM 🐝

i386:
@vivek awesome - glad you approve. Doing it with a script won't add much time to upgrading export as there are a few APIs I've had to hand delete/change for compatibility. I've also ditched all the python, ruby and xml variants to slim this down

i386:
@vivek I also copied over the `MethodHandle` optimisations so we will have those in the next beta too

vivek:
@i386 I left other comment on the code, looks like you missed them:) also, I guess you are going to pull in Jesse fix for WSOD improvement, once it's merged, in later PR?

i386:
@vivek oops I'll come back to it this afternoon when I get off my flight.

I'll move over the WSOD changes when they land.

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/Export.java
i386: Can do

i386:
@vivek Fixed in https://github.com/jenkinsci/blueocean-plugin/commit/f04880d6018e07197b34c8194b75ff6563683703 and https://github.com/jenkinsci/blueocean-plugin/commit/9fc12c4fd58f170ef9aa17f137c74c4ffc870980

vivek:
@i386 thanks 👍

PR #712, opened 2017-01-16
[JENKINS-36975] Show admin link at the correct times

imeredith:
# Description

'Administration' Link now links to /manage. IT only shows if user is logged in OR if security is disabled (and therefore anyone can access it).

No ATH changes as we cant log in with ath at this time.

See [JENKINS-36975](https://issues.jenkins-ci.org/browse/JENKINS-36975).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-web/src/main/js/main.jsx
+++ b/blueocean-web/src/main/js/main.jsx
@@ -33,6 +33,27 @@ function loginOrLogout(t) {
         }
     }
 }
+
+// Show link when only when someone is logged in...unless security is not configured,
+// then show it anyway.
+const AdminLink = (props) => {
+    const { t } = props;
+    const showLink = !Security.isSecurityEnabled() || !Security.isAnonymousUser();
+
+    if (showLink) {
+        var adminCaption = t('administration', {
+            defaultValue: 'Administation',
+        });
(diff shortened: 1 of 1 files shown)


i386:
@imeredith one thing here we are missing is a role check for if you can manage the system. e.g. If `Jenkins.getInstance().hasPermission(Jenkins.ADMINISTER)` is false then the user shouldn't see the admin link

michaelneale:
yeah - as long as it is in line with classic, and if it doesn't show the configure link, neither should blue ocean cc @imeredith @i386

michaelneale:
@i386 we can't see any existing api to return that information, the permissions that exist are per pipeline. It would need to be added to user etc, whereas the scope of this ticket wasn't quite this specific. I propose this goes in as is and then we follow up with more user permission info at some point

i386:
One field to put on the global window object "isAdminstrator" that is the result of the snippet above.

i386:
https://issues.jenkins-ci.org/browse/JENKINS-41096

michaelneale:
:bee: from me as it does what it says. unless @i386 wants to make it include specific permissions check (in which case fine to hold it open until that is in)

PR #722, opened 2017-01-17
FIX Jenkins 41016 move input buttons to left

scherler:
# Description

See [JENKINS-41016](https://issues.jenkins-ci.org/browse/JENKINS-41016) and  [JENKINS-41030](https://issues.jenkins-ci.org/browse/JENKINS-41030) 

screenshot from 2017-01-17 18-17-49

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


blueocean-dashboard/src/main/js/components/InputStep.jsx
```diff
@@ -69,17 +70,17 @@ export default class InputStep extends Component {
      * Submit the form as "cancel" out of the state data id.
      */
     cancelForm() {
-        const { id } = this.state;
-        parameterApi.cancelInputParameter(this.state.href, id);
+        const { href, id } = this.state;
+        parameterApi.cancelInputParameter(href, id);
```
tfennelly: An example of how I don't get some of these new ES6 "enhancements".

It's a puzzle to me how ...

```javascript
const { href, id } = this.state;
parameterApi.cancelInputParameter(href, id);
```

is an improvement on ....

```javascript
parameterApi.cancelInputParameter(this.state.href, this.state.id);
```

Just say'n 😈 and I know @scherler will call me an old fashioned fart and to 🔥 in hell 😄

blueocean-dashboard/src/main/less/forms/index.less
```diff
@@ -5,41 +5,45 @@
...
-        align-items: center;
-        max-width: 490px;
-    }
-
-    .inputBody .form-element {
-        padding-top: 5px;
-        padding-bottom: 5px;
+    .inputBody {
+        max-width: 400px;
+        min-width: 400px;
```
tfennelly: Will some people have an issue with this fixed width?

blueocean-dashboard/src/main/js/components/InputStep.jsx
michaelneale: Get off my lawn!

i386:
Looking good @scherler

michaelneale:
:bee: as long as ATH happy etc etc etc. 

Nice one!

blueocean-dashboard/src/main/js/components/InputStep.jsx
scherler: @tfennelly for me it reduces redundant information while reading the code. 

```
parameterApi.cancelInputParameter(this.state.href, this.state.id);
```

I consider the information `this.state` as disturbing, since I consider `href` and `id` as `const` which I do not care where they come from. It is more I can now do:

```
const { href } = this.props;
const { id } = this.state;
```
 
and 

```
parameterApi.cancelInputParameter(href, id);
```

stays the same.

When reading the diff like this for me it is cristal clear that I have change the `href ` source from `state` to `props`, this is not that obvious when the whole `cancelInputParameter` line will be changed. In that case I need to really compare whether the call has changed or only source .

scherler: BTW from now on I would use the logging that @tfennelly introduced before the api call for debugging reasons.

```
            if (logger.isDebugEnabled()) {
                logger.debug(`Will send cancel input call to ${href} for input step id: ${id}`);
            }
```

...and that looks so much cleaner then prefixing everything with `this.state`

blueocean-dashboard/src/main/less/forms/index.less
scherler: @tfennelly good question. ...but not sure how else to force the inputBody width to be 400px

PR #726, opened 2017-01-18
don't try to run npm, it's not available in official jenkins image

ydubreuil:
This should fix the PR deployer which is failing to build Docker images because `npm` is not available in standard Jenkins image.

Regression introduced in #665

@reviewbybees @michaelneale @cliffmeyers


ydubreuil:
PR deployer worked on this branch, Docker image successfully built.

cliffmeyers:
This was my mistake: apologies. When I was trying to get the core-js tests executing as part of blueocean CI, I was trying to install a later version of `npm` without realizing that this wouldn't work... and it slipped through review.

PR #650, opened 2016-12-07
[JENKINS-40299] Adapt to the new SCM API for detecting change requests

stephenc:
# Description

Adapt to the correct API for detecting change requests, see [JENKINS-40138](https://issues.jenkins-ci.org/browse/JENKINS-40138).

Blue Ocean ticket [JENKINS-40299](https://issues.jenkins-ci.org/browse/JENKINS-40299).

No tests because your existing tests should cover it and if they don't that's your problem, the API has moved and you are broken without this ;-)

# Submitter checklist
- [x] Link to JIRA ticket in description, if appropriate.
- [x] Change is code complete and matches issue description
- [x] Appropriate unit or acceptance tests or explanation to why this change has no tests
- [ ] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below.
- [ ] Ran Acceptance Test Harness against PR changes.

# Reviewer checklist
- [ ] Run the changes and verified the change matches the issue description
- [ ] Reviewed the code
- [ ] Verified that the appropriate tests have been written or valid explanation given

@reviewbybees 

…quests

PR #730, opened 2017-01-18
For testing of proposed SCM api changes

michaelneale:
This is to test SCM proposed changes

Versions to test from snapshot: 

cloudbees-folder is 5.17-20170118.220629-5
branch-api is 2.0.2-20170118.220727-1

Also, building from source would help out as well: 

```
$ mkdir ~/tmp/scm-api-fun
$ cd ~/tmp/scm-api-fun
$ git clone https://github.com/stephenc/cloudbees-folder-plugin.git
$ cd cloudbees-folder-plugin
$ git checkout jenkins-41124
$ mvn clean install
$ cd ..
$ git clone https://github.com/stephenc/branch-api-plugin.git
$ cd branch-api-plugin
$ git checkout jenkins-41124
$ mvn clean install
$ cd ..
$ git clone https://github.com/jenkinsci/blueocean-plugin.git
$ cd blueocean-plugin
$ vi # and modify the poms so that cloudbees-folder is 5.17-SNAPSHOT and branch-api is 2.0.2-SNAPSHOT
```

The production code to change would be from this older PR: https://github.com/jenkinsci/blueocean-plugin/pull/715/files (we may have to revert the test cases around URI changes). Or https://github.com/jenkinsci/blueocean-plugin/pull/650

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Blue Ocean test
+# Blue Ocean SCM api testing
 Blue Ocean is the next generation user experience for Jenkins. You can learn more about its features and roadmap on the [Blue Ocean project website](https://jenkins.io/projects/blueocean/).
 
 Join the community and [![Gitter](https://badges.gitter.im/jenkinsci/blueocean-plugin.svg)](https://gitter.im/jenkinsci/blueocean-plugin?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge)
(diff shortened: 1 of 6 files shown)


michaelneale:
cc @vivek work in progress

michaelneale:
The remaining failing test is due to it detecting that a snapshot is in use

PR #740, opened 2017-01-22
[FIX JENKINS-40326] server connection failure handling

tfennelly:
# Description

See [JENKINS-40326](https://issues.jenkins-ci.org/browse/JENKINS-40326).

Depends on https://github.com/jenkinsci/sse-gateway-plugin/pull/15

@brody and @i386 will probably want some visual/styling tweaks.

To test ... start jenkins, connect to BO interface with a browser ... kill BO .... error screen should appear .... restart jenkins ... error screens should change from gray to blue once the connection is ok again and then the page should reload.

Not sure how we can automate the testing of this, but will have a think about that.

15s demo video ...

[demo video](https://youtu.be/3IR0mckBDgo)

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


blueocean-web/src/main/js/components/BackendConnectFailure.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+        return (
+            <Fullscreen className={`errorscreen ${cssClass}`}>
+                <div className="message-box">
+                    <h3>{title}</h3>
+                    <div className="message">
+                        {message}
+                    </div>
+                </div>
+            </Fullscreen>
```
tfennelly: Didn't create a `<ErrorScreen>` component for this yet.

Not sure it's worth creating tbh. How many full screen error pages do we think we're going to need? We have 404 for when the page doesn't exist and then we have this full page error for then we can't connect to the backend. What else? These were really easy to create with the `<Fullscreen>` component and I'm not sure it's worth generalising it any more than that for a small few use cases.

michaelneale: yeah I expect things may end up with little header like popovers, or toasts etc for more subtle cases. so I agree. We don't really want that many. +1

michaelneale:
:bee: and passes ATH. 

I am not sure how to describe this other than "beautiful". Truly magnificent. Solves that ticket and the general case.

i386:
Might be best to meet with @brody sometime later this week to walk through it

michaelneale:
This is good to go now - we can leave it and make more busy work merging. This does what it says, re-uses the same design for full screen errors

tfennelly:
@i386 whatever it is you don't like should be very easy to resolve, assuming it's just some css/text tweaks that are needed. I was expecting that anyway.

blueocean-dashboard/src/main/less/connection-lost.less
```diff
@@ -0,0 +1,30 @@
...
+        /*
+         * For some reason, we needed to set the font-weight on the
+         * h3 used in the errorscreen when disconnected. If not, it
+         * appears more bold than it does if the same screen is
+         * shown when connected, plus when compared to the 404 screen.
+         * Weird :(
+         */
+        h3 {
+            font-weight: 400;
+        }
```
tfennelly: This was one of the weird changes I had to make in order for the header font weight to appear same as it does when we have a connection i.e. if we display this component when the connection is ok then we don't need this setting. There are no failed requests for fonts etc in the network tab ... weird.

scherler: Hmm could be a weird css path issue somewhere. To make sure they are the same, move it to line 2 

```
* > h3 {...}
```

tfennelly: But, I mean, it should not be needed at all !! It's not needed if I display this component when the connection is ok ... why is it needed when the connection is not ok? There are no resource loading errors ... the computed CSS rules are exactly the same (in the browser dev tools). It's weird.

scherler: maybe something in the background gets activated like some error listener that sets some css switched in some parent. I find it is hard to compare the exact result tree from css.

tfennelly: right ... must be something like that but I can't see anything yet.

tfennelly: So if I remove this `h3` rule and unconditionally show the error connection screen (i.e. when the connection is actually fine) I see the font-weight as expected (i.e. the same as for the 404 screen) ...

screenshot 2017-01-23 12 14 18

If I remove the unconditional behaviour and only show that same screen when there is no connection, notice how the font-weight is heavier ....

screenshot 2017-01-23 12 30 43

Only difference is that there's a connection for the first, while there's no connection for the second. Wtf ??

kzantow: Maybe something to do with loaded CSS from dashboard or elsewhere?

tfennelly: They look like different fonts, don't they? No errors in the network tab though and the dev tools state the same thing for the `font-family` in use.

screenshot 2017-01-23 12 45 44

kzantow: That's a weird one, I'd ignore it for now and revisit it if rears it's head again.

tfennelly: That's the problem ... for whatever reason, the browser is deciding to render one of the fallback fonts when there's no connection ....

screenshot 2017-01-23 13 01 30

kzantow: I can take a quick look and see if I notice anything, if you want

tfennelly: @kzantow if it's something that tickles your fancy, then thanks and by all means ... I was going to fire a question to a mailing list of some sort and see if someone can tell me the reason why :)

tfennelly: I failed to find anything on SO or google.

kzantow: The demo looks fine, which includes this oddity, right?

tfennelly: @kzantow you mean the video? If so, that's because I have a specific `font-weight` tweak in there to make it look "kinda" similar ... they're not the same though ... different fonts.

I suppose this is really a small issue that we can come back to if we find an answer to it.

kzantow: I think this should just be using `LatoLatinWeb` there isn't a `LatoLatinWebLight` loaded in the browser (there are 3: regular, bold, semibold - none of which seem like they'd provide 'light')... that's why it's falling back to `Helvetica Neue`

tfennelly: We could do that, but I'd prefer to know why it's falling back and what will happen on a browser that can't render `LatoLatinWebLight`, assuming that's a possibility (otherwise why do we have fallbacks at all).

kzantow: That LatoLatin is a web font we're providing. The issue is the `LatoLatinWebLight` is an unknown font family, at least while that dialog is shown... maybe we have something that loads it up later, but I think maybe it was just something that was mistyped at some point? It works to use the right font when using the known `LatoLatinWeb`.

tfennelly: There you go ... it works fine if I first browse to another page that's using that "light" version. I thought we were using that all over, but no - I missed the "light" part of it 😄 

So I need to make sure we get that loaded :)

tfennelly: And the fact that it never attempts to load it threw me too (no failed network requests).

kzantow: Ah yes, seems like it should just load at first with the other Lato fonts if we're ever using it.

cliffmeyers: Yeah IMO we should just drop the "WebLight" thing altogether, as all of @brody 's stuff that I've seen is just using Lato. Probably makes sense for us to use Lato over LatoLatin so we support non-latin glyphs too. (I'm not a font expert, but for something as widely-used as Jenkins restricting to Latin char set seems inappropriate).

tfennelly: I've sorted this for this PR. The debate on whether or not to keep or ditch specific fonts can be had at a different time imo.

blueocean-web/src/main/js/components/BackendConnectFailure.jsx
```diff
@@ -0,0 +1,113 @@
...
+            });
+        });
+    }
+
+    render() {
+        if (this.state.connectionOkay) {
+            // Force loading of fonts etc. We need this because we need to make
+            // sure all required resources are available when there's no connection.
+            return (
+                <Fullscreen className="hidden">
```
kzantow: oh, i get it now. browser optimizes resource loading until the first css rule references a particular font. in this case browser already knows there's no connection, so it doesn't load the font... fun problems to solve.

kzantow: wonder if the browser would be smart enough to not load a font since the element is hidden... hah

tfennelly: Yeah, I was wondering that too but seems to be fine. Not the end of the world anyway if it does ... it'll just fallback.

tfennelly: Note i had to inline the background SVG in the CSS too hahaha

brody:
@tfennelly you'll have to let me know if the reconnect action is possible, if not, having the toast appear to let the user know the connection is lost & have the toast dismiss itself when the connection is regained.

connection lost

michaelneale:
@brody the reconnect is not necessary (nor possible) as it is continuously trying to reconnect - I guess at most it would be a reload? 

Also I am not sure if the toast is clear enough for this. It doesn't stop them interacting which may do bad things.. maybe... will see what @tfennelly says.

i386:
@michaelneale you could put a div over the page that darkens the screen slightly in that case so you can't click anywhere but the toast.

brody:
@i386 @michaelneale white at 50% opacity or black at 25% opacity. The white looks more like the disabled state of the forms.
image

michaelneale:
yeah I like the darkening. 

I am still sad as I love the big waves :(

i386:
IMO darkening is consistent with other overlays (popovers, dialogs) where you can't interact with the page. I could go either way.

brody:
`rgba(0,0,0,0.25)` it is

kzantow:
👍  on the darkening. still a big 👎  on not having something that looks like an error to display an error.

michaelneale:
@tfennelly so making it something *like* toast is the go. Last I heard @kzantow and @brody were jelly wrestling to work out what form that should take exactly. So if you want to convert it to the suggested toast like thing, go for it, otherwise park it for 1 day while they sort it out.

tfennelly:
@brody No action is possible, so the little x on the toast is not really meaningful either. You ok with not having the little x ?

michaelneale:
@tfennelly yeah - it can't be actioned or dismissed, so why not. if you check gitter there are a few sketches he proposed.

tfennelly:
@brody see https://youtu.be/z7X1KmL_0Cs and let me know if it's ok wrt what you want

tfennelly:
@scherler looks like toast, but really is turkey ... nothing in common with the toast service aside from the styling ... trying to shoehorn in the toast service would do nothing other than make it more complicate .. it's currently very simple.

scherler:
@tfennelly :rofl: happy thanksgiving then ;)

brody:
@tfennelly I'll repost my comments from Gitter here so you dont need to wade through the comments.

Looks pretty good. 
- We’ll need a fade in & out for the background because it flashing on & off is jarring. 
- The position of the fail & success toasts are a couple of pixels off & i’ll need to add a transition for those too because we need some movement to draw the eye. 

I’ll work on both of those today and flick something through.

PR #743, opened 2017-01-23
[JENKINS-39857_Display_steps_durations_in_Blue_Ocean] Add duration in…

scherler:
…formation to the result item

# Description

See [JENKINS-39857](https://issues.jenkins-ci.org/browse/JENKINS-39857) and [JENKINS-41282](https://issues.jenkins-ci.org/browse/JENKINS-41282).

# Depends on

https://github.com/jenkinsci/jenkins-design-language/pull/135

# showview

https://youtu.be/nao7UX0v8r0

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


michaelneale:
Unfortunately I think this one needs to be done the same time as: https://issues.jenkins-ci.org/browse/JENKINS-41282 so that the humanize function is the same in both places.

scherler:
@michaelneale we use the humanize function from jdl so when we change it we change it over there. Speaking of which https://github.com/moment/moment/issues/348 seems that the "normal" one is not given more precised output and it is recommended to use either https://github.com/EvanHahn/HumanizeDuration.js or write our own http://stackoverflow.com/questions/14157341/how-can-i-humanize-this-complete-duration-in-moment-js-javascript

michaelneale:
IMO makes sense to have our own one as it will likely change.

scherler:
@michaelneale agree, further it could be done with `format` which we already use in the rest.

scherler:
@michaelneale  fixes now as well https://issues.jenkins-ci.org/browse/JENKINS-41282

PR #744, opened 2017-01-23
Fix spelling of Administration

rhutchison:
Correct the default spelling of Administration navigation menu.


i386:
@rhutchison oops - good catch! Thank you.

michaelneale:
Gosh, how embarrassing... good catch

PR #752, opened 2017-01-26
[JENKINS-39345] Async bundle startup scripts

tfennelly:
# Description

So we can attach async "startup scripts" to bundles, allowing us to ensure that e.g. i18n resources are loaded before the bundle executes ( [JENKINS-39345](https://issues.jenkins-ci.org/browse/JENKINS-39345) ).

Depends on https://github.com/jenkinsci/js-builder/pull/11

Other probably uses of this feature would be js extension metadata, allowing us to ensure that it's loaded before the extensions are executed etc. CC @kzantow.

Not really possible to write meaningful tests for this imo, unless someone has a good idea that "m missing.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-dashboard/src/main/js/components/PipelinePage.jsx
+++ b/blueocean-dashboard/src/main/js/components/PipelinePage.jsx
@@ -10,13 +10,15 @@ import {
     TabLink,
     WeatherIcon,
 } from '@jenkins-cd/design-language';
-import { i18nTranslator, NotFound, User, Paths } from '@jenkins-cd/blueocean-core-js';
+import { i18nTranslator, NotFound, User, Paths, logging } from '@jenkins-cd/blueocean-core-js';
 import { Icon } from '@jenkins-cd/react-material-icons';
 import PageLoading from './PageLoading';
 import { buildOrganizationUrl, buildPipelineUrl, buildClassicConfigUrl } from '../util/UrlUtils';
 import { documentTitle } from './DocumentTitle';
 import { observer } from 'mobx-react';
 import { observable, action } from 'mobx';
+
+const logger = logging.logger('io.jenkins.blueocean.dashboard.PipelinePage');
 
 const RestPaths = Paths.rest;
 /**
@@ -68,6 +70,7 @@ export class PipelinePage extends Component {
         const isReady = !!pipeline;
 
         if (!pipeline && this.error) {
+            logger.log(`Error finding pipeline page for ${fullName}.`, this.error);
             return <NotFound />;
         }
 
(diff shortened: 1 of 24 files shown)


blueocean-dashboard/src/main/js/components/PipelinePage.jsx
```diff
@@ -68,6 +70,7 @@ export class PipelinePage extends Component {
         const isReady = !!pipeline;
 
         if (!pipeline && this.error) {
+            logger.log(`Error finding pipeline page for ${fullName}.`, this.error);
```
scherler: should that not be `logger.error` or `warn`

js-extensions/@jenkins-cd/subs/extensions-bundle.js
```diff
@@ -162,3 +184,47 @@ function assertHasJenkinsJsExtensionsDependency(message) {
...
+            for (var i = 0; i < files.length; i++) {
+                var file = files[i];
+                if (propertiesFileMatcher.test(file)) {
+                    // Strip off the different parts we don't want. It might be that
+                    // resource bundle basenames can have underscores in them according
+                    // to specs (I didn't find a ref one way or the other), but trying to match
+                    // language/country/variant codes exactly and strip them off is going to
+                    // make the code a pita to impl/understand, so lets just say we only detect
+                    // basenames that do NOT contain underscores. If someone insists on using
+                    // underscores they can manually define the bundles in the yaml file.
```
scherler: Do I understood the comment correctly that we only need the basename. Maybe add to the above a small example:

```
cd blueocean-plugin/blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard
tree 
.
├── Messages_de.properties
├── Messages_es.properties
├── Messages_fr.properties
└── Messages.properties
```

In case of the dashboard plugin we have above resource bundles. The only file we are interested in in `Messages.properties` - the base name of the bundle - all other files are variants/translations

tfennelly:
Moved to https://github.com/jenkinsci/blueocean-plugin/pull/769

blueocean-dashboard/src/main/js/components/PipelinePage.jsx
tfennelly: Def not `error` and I don't think `warn` either ... it's not an "error" in the sense that the app is not working correctly. It's just a user doing something strange pasting in a wrong url or something.

js-extensions/@jenkins-cd/subs/extensions-bundle.js
tfennelly: Ok, will add that. Thanks

PR #756, opened 2017-01-27
JENKINS-41462# branch name encoding fix

vivek:
# Description

See [JENKINS-41462](https://issues.jenkins-ci.org/browse/JENKINS-41462).

Reopened previously approved but reverted PR https://github.com/jenkinsci/blueocean-plugin/pull/751 due to failing ATH. Reopening.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-dashboard/src/main/js/util/UrlUtils.js
+++ b/blueocean-dashboard/src/main/js/util/UrlUtils.js
@@ -46,8 +46,16 @@ export const buildRunDetailsUrl = (organization, pipeline, branch, runId, tabNam
 /*
  * helper to clean the path replace(/%2F/g, '%252F')
  * @param input
+ *
+ * XXX: This was used where encodeURIComponent() should have been used. If it has not other uses,
+ *      it should perhaps be removed.
  */
 export const uriString = (input) => encodeURIComponent(input).replace(/%2F/g, '%252F');
+
+/**
+ * Double encode name, feature/test#1 is encoded as feature%252Ftest%25231
+ */
+export const doubleUriEncode = (input) => escape(encodeURIComponent(input));
 
 // general fetchAllTrigger
 export const fetchAllSuffix = '?start=0';
@@ -112,7 +120,7 @@ export function calculateNodeBaseUrl(config) {
         `${_appURLBase}/rest/organizations/jenkins/` +
         `pipelines/${name}`;
     if (isMultiBranch) {
-        return `${baseUrl}/branches/${uriString(branch)}/runs/${runId}/nodes/`;
+        return `${baseUrl}/branches/${doubleUriEncode(branch)}/runs/${runId}/nodes/`;
     }
     return `${baseUrl}/runs/${runId}/nodes/`;
 }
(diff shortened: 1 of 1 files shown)


michaelneale:
This passes the ATH now - was a temporary failure. I would like this to be blessed by @cliffmeyers however. I will try this out by hand next week to sniff around otherwise.

vivek:
@michaelneale no hurry and perhaps if @scherler PTAL as well.

michaelneale:
yeap :bee: from me, seems sound on all fronts. Hopefully @cliffmeyers can ACK this as a good idea.

scherler:
@vivek as I understood we doing double encode

` Double encode name, feature/test#1 is encoded as feature%252Ftest%25231`

because of fails in ATH?

The case that I found is in folder.js that uses that enconding, but not sure whether I misunderstood the context

blueocean-dashboard/src/main/js/util/UrlUtils.js
```diff
@@ -46,9 +46,17 @@ export const buildRunDetailsUrl = (organization, pipeline, branch, runId, tabNam
 /*
  * helper to clean the path replace(/%2F/g, '%252F')
  * @param input
+ *
+ * XXX: This was used where encodeURIComponent() should have been used. If it has not other uses,
+ *      it should perhaps be removed.
  */
 export const uriString = (input) => encodeURIComponent(input).replace(/%2F/g, '%252F');
```
cliffmeyers: Yes, let's remove this as part of the PR. Its only usages are in this file.

blueocean-dashboard/src/main/js/util/UrlUtils.js
```diff
@@ -46,9 +46,17 @@ export const buildRunDetailsUrl = (organization, pipeline, branch, runId, tabNam
...
 
+/**
+ * Double encode name, feature/test#1 is encoded as feature%252Ftest%25231
+ */
+export const doubleUriEncode = (input) => escape(encodeURIComponent(input));
```
cliffmeyers: `escape` is deprecated. This should use `encodeURIComponent` twice.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape

vivek:
@scherler 

>because of fails in ATH?

Nothing to do with ATH. User reported bug where branch names with special chars (other than '/') results in 404. For example branch 'feature/test#1' results in 404. This is fix for that.

blueocean-dashboard/src/main/js/util/UrlUtils.js
vivek: Ok.

blueocean-dashboard/src/main/js/util/UrlUtils.js
vivek: Ack.

vivek:
@cliffmeyers included your comments in this commit, PTAL.

vivek:
ATH passes https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/1184/.

PR #768, opened 2017-01-31
JENKINS-41348# BlueOcean specific CredentialsProvider

vivek:
Provides User store/scoped credentials by using appropriate properties
on OrganizationFolder.

# Description

BlueOcean credentials provider is a proxy to credentials stored in user store. 

This is how it works:

- When an OrganizationFolder is created, we attach authenticated user's user Id and credential Id of the credential in authenticated user's store with USER scope.
- Later, when job runs, it asks BlueOceanCredentialsProvider to give credentialId for the OrganizationFolder created in previous step, it looks for these properties and returns right Credentials object by looking in to user's store.
- A github OrganizationFolder configuration page on classic lists the BlueOcean Provider credential

image

### Changes to credential creation for git pipeline flow (for @cliffmeyers)
curl -v -XPOST -u admin:admin -H ‘content-type: application/json’ http://localhost:8080/jenkins/blue/rest/organizations/jenkins/credentials/user/ 

```
{
  "credentials" : {
    "privateKeySource" : {
      "privateKey" : "abcabc1212",
      "stapler-class" : "com.cloudbees.jenkins.plugins.sshcredentials.impl.BasicSSHUserPrivateKey$DirectEntryPrivateKeySource"
    },
    "passphrase" : "ssh2",
    "scope" : "USER",
    "domain" : "blueocean-git-domain",
    "description" : "ssh2 desc",
    "$class" : "com.cloudbees.jenkins.plugins.sshcredentials.impl.BasicSSHUserPrivateKey",
    "username" : "ssh2"
  }
}
```

There are three changes:

- POST URL is changed to /jenkins/blue/rest/organizations/jenkins/credentials/user/ 
- domain element in request body:  for git use case its recommended UI passes domain name, for example 'blueocean-git-domain'. If its not supplied this credential will be created in default domain 'blueocean-domain'. 
See [JENKINS-41348](https://issues.jenkins-ci.org/browse/JENKINS-41348).
- scope element in request body should be set to USER. 

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


i386:
Gosh you're fast :)

vivek:
@stephenc PTAL, specially at [BlueOceanCredentialsProvider.java](https://github.com/jenkinsci/blueocean-plugin/blob/bug/JENKINS-41348-userscope/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java.)

vivek:
@i386 its been bothering me, better to get it out of the way sooner than later:) We also need similar approach for git ssh keys - these are stored in system store. Will chat with @cliffmeyers tomorrow to get it wrapped with this PR.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
```diff
@@ -0,0 +1,262 @@
...
+ * {@link CredentialsProvider} to serve credentials stored in user store.
+ *
+ * It works by looking for {@link FolderPropertyImpl} on a folder. This should allow
+ * using such user scoped properties with {@link jenkins.branch.OrganizationFolder} or {@link jenkins.branch.MultiBranchProject}
+ * or even a Folder.
+ *
+ * @author Stephen Connoly
+ * @author Vivek Pandey
+ */
+@Extension(optional = true)
```
stephenc: Not an optional extension.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
```diff
@@ -0,0 +1,262 @@
...
+public class BlueOceanCredentialsProvider extends CredentialsProvider {
+    private static final String DISPLAY_NAME = "BlueOcean Folder Credentials";
```
stephenc: :ant: I18N this in a `Messages` resource bundle

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
```diff
@@ -0,0 +1,262 @@
...
+                return store;
+            }
+
+            /**
+             * {@inheritDoc}
+             */
+            @Override
+            public String getIconFileName() {
+                return isVisible()
+                        ? "/plugin/credentials/images/48x48/folder-store.png"
```
stephenc: :ant: please use a Blue Ocean specific icon

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
```diff
@@ -0,0 +1,262 @@
...
+                super(BlueOceanCredentialsProvider.class);
+                this.abstractFolder = abstractFolder;
+                this.storeAction = new CredentialsStoreActionImpl(this);
+            }
+
+            @Nonnull
+            @Override
+            public List<Domain> getDomains() {
+                //XXX: how to get domain from given Credentials object? Maybe attach domain name to AbstractFolder?
+                return Collections.emptyList();
```
stephenc: Collections.singletonList(Domain.global())

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
```diff
@@ -0,0 +1,262 @@
...
+
+            @Nonnull
+            @Override
+            public ModelObject getContext() {
+                return abstractFolder;
+            }
+
+            @Override
+            public boolean hasPermission(@Nonnull Authentication a, @Nonnull Permission permission) {
+                return abstractFolder.getACL().hasPermission(a,permission);
```
stephenc: may need to tweak this, but ok for now

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
```diff
@@ -0,0 +1,262 @@
...
+        if(isApplicable(object)) {
+            return new FolderPropertyImpl.StoreImpl((AbstractFolder) object);
+        }
+        return null;
+    }
+
+    @Override
+    public Set<CredentialsScope> getScopes(ModelObject object) {
+        if(isApplicable(object)){
+            return ImmutableSet.of(CredentialsScope.USER);
```
stephenc: Actually you are `CredentailsScope.GLOBAL` if we are to be strictly correct... this may require changes in the credentials API to allow you to "clone" a credential while changing it's scope

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
```diff
@@ -0,0 +1,262 @@
...
+                Authentication auth = user.impersonate();
+                return CredentialsMatchers.filter(
+                        CredentialsProvider.lookupCredentials(
+                                type,
+                                Jenkins.getInstance(),
+                                auth,
+                                Collections.<DomainRequirement>emptyList()
+                        ),
+                        CredentialsMatchers.allOf(CredentialsMatchers.withId(prop.getId()),
+                                CredentialsMatchers.withScope(CredentialsScope.USER))
```
stephenc: I am now wondering if perhaps we shouldn't be more explicit and just target the credential stores scoped to the user and bypass the general `lookupCredentials`, e.g.

```
for (CredentialStore store: CredentialsProvider.lookupStores(user)) {
  // if store has credentials with id, return credential
}
// return empty list;
```

as matching based on scope here is ringing some alarm bells

blueocean-github-pipeline/src/test/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubOrgFolderTest.java
```diff
@@ -2,47 +2,84 @@
...
 import org.junit.Assert;
 import org.junit.Test;
 
+import java.io.IOException;
 import java.util.Map;
 
 /**
  * @author Vivek Pandey
  */
 public class GithubOrgFolderTest extends PipelineBaseTest {
```
stephenc: Missing a test that the credential gets resolved, see [this](https://github.com/jenkinsci/cloudbees-folder-plugin/blob/ff0c53014c346dac3d891b150b0234c9c4c29c53/src/test/java/com/cloudbees/hudson/plugins/folder/properties/FolderCredentialsProviderTest.java#L140-L287) for inspiration but I do not thing you need to go quite so far as to actually trigger a build. I think you could get away with seeing if the credential can be looked up as SYSTEM using the standard credentials API when you add your property and if you remove the property the credential goes missing again, even though the credential is in "bob"'s user store

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
```diff
@@ -343,11 +343,23 @@ private static StandardUsernamePasswordCredentials findUsernamePasswordCredentia
...
-    private static @CheckForNull StandardUsernamePasswordCredentials findUsernamePasswordCredential(String id){
+    static StandardUsernamePasswordCredentials findUsernamePasswordCredential(String uri, String credentialId){
+        return CredentialsMatchers.firstOrNull(
+                CredentialsProvider.lookupCredentials(
+                        StandardUsernamePasswordCredentials.class,
+                        Jenkins.getInstance(),
+                        Jenkins.getAuthentication(),
+                        URIRequirementBuilder.fromUri(uri).build()),
+                CredentialsMatchers.allOf(CredentialsMatchers.withId(credentialId),
+                        CredentialsMatchers.withScope(CredentialsScope.USER))
```
stephenc: This is a bad smell, should not need to match based on scope

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
stephenc: You may need to specify an ordinal to ensure your priority is ahead of the standard stores

stephenc:
@vivek said:

> We also need similar approach for git ssh keys - these are stored in system store. Will chat with @cliffmeyers tomorrow to get it wrapped with this PR.

If you need to handle ssh keys also then actually the `BlueOceanCredentialsProvider` will have to expose up to two credential ids rather than the single id I initially pseudo-coded

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    private boolean isApplicable(ModelObject object){
+        return object instanceof AbstractFolder &&
+                ((AbstractFolder)object).getProperties().get(FolderPropertyImpl.class) != null;
+    }
+
+    public static class FolderPropertyImpl extends AbstractFolderProperty<AbstractFolder<TopLevelItem>> {
+        private final String user;
+        private final String id;
```
stephenc: Sounds like you now need this to be a `Collection<String>` (probably a `Set`) rather than a single value

stephenc:
It's looking good and definitely a better approach than stuffing the credentials into the global store, just some polish recommended before merging

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: Hmm, I thought this is stock folder credential one and can be used for other CredentialsProvider extension. This is anyway read-only credential provider and is a proxy really. Not sure we really need an icon created for it. @i386?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: @stephenc Why global? The credentials it exposes are in user scope.

blueocean-github-pipeline/src/test/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubOrgFolderTest.java
vivek: Ok, will add a test.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
vivek: Why is that? I want to keep it strict and if there is GitHub access key based credentials it better come from USER scope or we reject it.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: @stephenc jenkins organization folder is either going to be setup by blueocean either using ssh credential(BasicSSHUserPrivateKey) or GitHub accesskey based credential (StandardUsernamePassword), not both. So why a collection?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: This provider is proxy exposing credentials from USER scope, for this reason I exposed it with USER scope. Exposing them with GLOBAL scope sounds incorrect to me.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: I can give it try, but I saw other places your concern about using USER scope as not a good idea. I guess I need to understand why this is not a good idea - this provider is really a locked down provider very specific to keys generated by blueocean in user store with USER scope.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: yeah APIs have i18N missing, maybe good time to get started on it.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: ack.

vivek:
@stephenc applied your suggested changes, tests, i18n etc. except the one related to avoiding USER scope as I think this provider should continue doing it - more strict version of checks. If its a bad idea I want to know why.

vivek:
@cliffmeyers git specific credential creation in user store and git pipeline creation with user store keys implementation is completed. PTAL. In general for this ticket, all cases have been covered.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
stephenc: Because you are not using the authorise project plugin, the authentication will be ACL.SYSTEM so you want the scope GLOBAL as that allows the credential to be used by jobs running as SYSTEM or jobs running as an authentication with the USE_ITEM permission (limited to jobs in the org folder)

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
stephenc: Well if doing checkout with ssh keys you will need *both* the api credentials for scan/index *and* the ssh key for checkout

blueocean-git-pipeline/src/main/java/io/jenkins/blueocean/blueocean_git_pipeline/GitPipelineCreateRequest.java
```diff
@@ -40,6 +42,10 @@ public GitPipelineCreateRequest(String name, BlueScmConfig scmConfig) {
 
     @Override
     public BluePipeline create(Reachable parent) throws IOException {
+        User authenticatedUser =  User.current();
+        if(authenticatedUser == null){
+            throw new ServiceException.UnauthorizedException("Must login to create a pipeline");
+        }
```
stephenc: You probably will need some additional checks here, such as:

* Did the system admin disable the per-user credentials store?
turning off user credentials store

* Did the system admin disable this Blue Ocean User Credentials Proxy/Gateway credential store

* Did the system admin block storage of the credential type we need to store in the user credentials store
screen shot 2017-02-01 at 09 08 01

Probably not a blocker for merging this, but rather something you will need to be aware of and decide what you want to do about it and how to behave if the system admin has made configuration choices about credentials storage that prevent the blue ocean pipeline functionality from working

blueocean-git-pipeline/src/main/java/io/jenkins/blueocean/blueocean_git_pipeline/GitPipelineCreateRequest.java
```diff
@@ -48,13 +54,20 @@ public BluePipeline create(Reachable parent) throws IOException {
...
 
-        //XXX: set credentialId to empty string if null or we get NPE later on
-        String credentialId = scmConfig.getCredentialId() == null ? "" : scmConfig.getCredentialId();
-
         TopLevelItem item = create(Jenkins.getInstance(), getName(), MODE, MultiBranchProjectDescriptor.class);
 
         if (item instanceof WorkflowMultiBranchProject) {
             WorkflowMultiBranchProject project = (WorkflowMultiBranchProject) item;
+
+            if(scmConfig.getCredentialId() != null) {
```
stephenc: Safer to use `StringUtils.isNotBlank(scmConfig.getCredentialId())`

blueocean-git-pipeline/src/main/java/io/jenkins/blueocean/blueocean_git_pipeline/GitPipelineCreateRequest.java
```diff
@@ -48,13 +54,20 @@ public BluePipeline create(Reachable parent) throws IOException {
...
+                project.addProperty(
+                        new BlueOceanCredentialsProvider.FolderPropertyImpl(authenticatedUser.getId(),
+                                scmConfig.getCredentialId()));
+            }
+
+            //XXX: set credentialId to empty string if null or we get NPE later on
+            String credentialId = scmConfig.getCredentialId() == null ? "" : scmConfig.getCredentialId();
```
stephenc: Or `String credentialId = StringUtils.defaultString(scmConfig.getCredentialId())`

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
```diff
@@ -55,27 +60,57 @@ public BluePipeline create(Reachable parent) throws IOException {
...
-        TopLevelItem item = create(Jenkins.getInstance(), getName(), DESCRIPTOR, CustomOrganizationFolderDescriptor.class);
 
-        if (item instanceof OrganizationFolder) {
-            GitHubSCMNavigator gitHubSCMNavigator = new GitHubSCMNavigator(apiUrl, orgName, credentialId, credentialId);
-            if (sb.length() > 0) {
-                gitHubSCMNavigator.setPattern(sb.toString());
+        User authenticatedUser =  User.current();
+        if(authenticatedUser == null){
+            throw new ServiceException.UnauthorizedException("Must login to create a pipeline");
+        }
```
stephenc: repeat of my comment w.r.t. the GitPipelineCreateRequest authorization check

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
```diff
@@ -235,57 +231,14 @@ public HttpResponse validateAndCreate(@JsonBody JSONObject request) {
...
-            }
-
-            if(githubCredential == null){
-                if(!store.addCredentials(domain, credential)){
-                    throw new ServiceException.UnexpectedErrorException("Failed to add credential to domain");
-                }
-
+            if(githubCredential == null) {
+                CredentialsUtils.createCredentialsInUserStore(
+                        credential, authenticatedUser, DOMAIN_NAME, new URI(getUri()));
```
stephenc: Will the `getUri()` return the GitHub Enterprise URL if the user is creating a pipeline against GHE?

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
stephenc: Still think this is a bad smell. Given that this credentials provider has very specific needs, I suspect we actually want to be directly checking the credential stores for the `User` and not bothering with the scope check at all

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
```diff
@@ -358,7 +323,7 @@ private static StandardUsernamePasswordCredentials findUsernamePasswordCredentia
                         Jenkins.getAuthentication(),
                         Collections.<DomainRequirement>emptyList()),
                 CredentialsMatchers.allOf(CredentialsMatchers.withId(id),
-                        CredentialsMatchers.anyOf(CredentialsMatchers.instanceOf(StandardUsernamePasswordCredentials.class)))
+                        CredentialsMatchers.withScope(CredentialsScope.USER))
```
stephenc: Bad smell

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
```diff
@@ -70,7 +66,7 @@
     private static final String USER_EMAIL_SCOPE = "user:email";
     private static final String USER_SCOPE = "user";
     private static final String REPO_SCOPE = "repo";
-    private static final String DOMAIN_NAME="github-domain";
+    private static final String DOMAIN_NAME="blueocean-github-domain";
```
stephenc: So, you will need to have a fall-back strategy to cover the case where the admin has disabled the standard per-user store and replaced it with an alternative implementation (think Hashicorp Vault, CyberArk, etc). The alternative implementation is not required to support multiple domains and may indeed decide to only expose all stored credentials in the `global` domain.

Your code should therefore be able to fall back and handle the situation where the add domain fails and either attempt to add the credentials into the global domain or declare that you cannot proceed.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
```diff
@@ -0,0 +1,261 @@
...
+        }
+        return Collections.emptyList();
+    }
+
+    @Nonnull
+    @Override
+    public <C extends IdCredentials> ListBoxModel getCredentialIds(@Nonnull Class<C> type,
+                                                                   @Nullable ItemGroup itemGroup,
+                                                                   @Nullable Authentication authentication,
+                                                                   @Nonnull List<DomainRequirement> domainRequirements,
```
stephenc: Given that you are creating the credentials in a domain in the user store, we should probably be applying the same domain validation in the `getCredentials` and `getCredentialIds` methods. This would mean storing either the `URL` or the `DomainSpecification` as a field in the `FolderPropertyImpl`

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
stephenc: The stock icon is for the "stock" FolderCredentialsProvider from the folders plugin. You need your own so that a user/admin sees that the credentials are coming from this proxy/gateway provider

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
stephenc: actually what I see now is that you want something different, you want

`Collections.singletonList(/*domain for "blueocean-github-domain"*/)`

The domain is about matching the URL the credentials are requested for against the URLs that the credentials are allowed to be used against.

Scope is a different cross-cutting requirement

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
stephenc: Do need to tweak this, as this is a read only domain we want to always return false for the `Credentials/CREATE`, `Credentials/DELETE`, `Credentials/MANAGE_DOMAIN` and `Credentials/UPDATE` permissions

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
```diff
@@ -0,0 +1,261 @@
...
+
+            @Override
+            public boolean hasPermission(@Nonnull Authentication a, @Nonnull Permission permission) {
+                return abstractFolder.getACL().hasPermission(a,permission);
+            }
+
+            @Nonnull
+            @Override
+            public List<Credentials> getCredentials(@Nonnull Domain domain) {
+                return BlueOceanCredentialsProvider.getCredentials(Credentials.class, abstractFolder);
```
stephenc:
```
if (domain == /*domain for "blueocean-github-domain"*/) {
    return BlueOceanCredentialsProvider.getCredentials(Credentials.class, abstractFolder);
}
return Collections.emptyList();
```

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/CredentialApi.java
```diff
@@ -32,6 +33,8 @@
 
     private  final CredentialsStoreAction credentialStoreAction;
     private  final Reachable parent;
+    public static final String DOMAIN_NAME = "blueocean-domain";
```
stephenc: this is a different domain name from that used elsewhere

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
```diff
@@ -0,0 +1,261 @@
...
+        return Collections.emptySet();
+    }
+
+    private boolean isApplicable(ModelObject object){
+        return object instanceof AbstractFolder &&
+                ((AbstractFolder)object).getProperties().get(FolderPropertyImpl.class) != null;
+    }
+
+    public static class FolderPropertyImpl extends AbstractFolderProperty<AbstractFolder<TopLevelItem>> {
+        private final String user;
```
stephenc: - [ ] Need a field to store the domain name(s) that this / these credentials are exposed through
- [ ] Need a field to store either the `DomainSpecification` or the URI (and instantiate the domain specification on demand)... my preference is to store the `DomainSpecification`

You also need to solve for the case of GitHub / BitBucket where there is the possibility of both scan credentials and checkout credentials... where you need the two different credentials, to further complicate matters the domain specifications for these will be different, e.g. the SSH key will need to match both the `ssh` and `git` protocols while the API token used for scanning will need to match `http` and `https`

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: I see. So it's more for routing to this proxy. I will give it a try.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: Agreed.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/CredentialApi.java
vivek: @stephenc thats right, the way its is currently, GitHub credentials are stored in 'github-domain' got for git keys, UI/client can create credentials in domain of their choice. 'blueocean-domain' is default domain where they get saved. Is this problem?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: @stephenc  blueocean allowes credentials to be stored in different domain. For example GitHub credentials could be stored in different domain than git credentials.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: @stephenc Yes I realize storing domain name is better along with domain specification. Is there Credentials API or somewhere I could use code that creates domain spec. I do it for github/https, bit tedious to take in to account different URI types.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: Ack.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
vivek: Yes.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
vivek: >Your code should therefore be able to fall back and handle the situation where the add domain fails and either attempt to add the credentials into the global domain or declare that you cannot proceed.

I think we would rather fail than creating in global scope. So add credential is then going to fail and will be reported as error to UI.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
vivek: @stephenc Given I am going to be scanning in context of domain in user store, that is blueocean generated credentials are going to be created in user store in specific domain and searched accordingly, dropping USER scope is fine.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: You don't necessarily need both. I can do MBP job creation only using my ssh key. https://gist.github.com/vivek/7a200b9cd6d645a78402b0ce5b1e56a8.

Our UI flow, 
- For Git case, we let user create either api key or ssh keys. In case of git, its ssh, or access key (username/password) but not both. 
- For GitHub, no ssh key, only GitHub accessKey.

blueocean-git-pipeline/src/main/java/io/jenkins/blueocean/blueocean_git_pipeline/GitPipelineCreateRequest.java
vivek: Good point. I will open another ticket to get it addressed properly as it may need some UX change.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
vivek: @stephenc It seems to be working how it is, but I will change it to GLOBAL. All I am making sure is that its not less secure and these credentials do not get outside of the scope of the org folder they are attached to. Which I think is not the case.

vivek:
@stephenc I think I have got everything landed here. Last items DomainRequirement/DomainSpecification has been added. It also goes over user credential store with correct impersonation. PTAL so that we can proceed further.

BlueOcean store specific icon is another item but it should not block this PR.

stephenc:
> BlueOcean store specific icon is another item but it should not block this PR.

That's why I gave it an :ant: (which is more a niggle than a bug) :-D

Will take a look later today

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
```diff
@@ -55,35 +67,95 @@ public BluePipeline create(Reachable parent) throws IOException {
...
-            return new GithubOrganizationFolder(organizationFolder, parent.getLink());
+                }
+            }
+            throw new ServiceException.UnexpectedErrorException(msg, e);
         }
         return null;
     }
 
-     static void validateCredentialId(String credentialId, OrganizationFolder item, GitHubSCMNavigator navigator) throws IOException {
+    private String GithubCredentialsDomain(String apiUri) {
```
stephenc: ugh... methods start with lowercase character

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
```diff
@@ -55,35 +67,95 @@ public BluePipeline create(Reachable parent) throws IOException {
...
+        URI uri;
+        try {
+            uri = new URI(apiUri);
+        } catch (URISyntaxException e) {
+            throw new ServiceException.BadRequestExpception(new ErrorMessage(400, "Failed to create pipeline")
+                    .add(new ErrorMessage.Error("scmConfig.uri",
+                            ErrorMessage.Error.ErrorCodes.INVALID.toString(), "Invalid github API URI: "+e.getMessage())), e);
+        }
+        String path = StringUtils.defaultIfBlank(uri.getPath(), "/");
+        boolean githubEnterprise = false;
```
stephenc: unnecessary boolean

blueocean-github-pipeline/src/test/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubOrgFolderTest.java
```diff
@@ -1,48 +1,126 @@
...
+        organizationFolder.addProperty(prop);
+
+        // lookup for created credential id in system store, it should resolve to previously created user store credential
+        StandardCredentials c = Connector.lookupScanCredentials(organizationFolder, null, credential.getId());
+        assertNotNull(c);
+        assertTrue(c instanceof StandardUsernamePasswordCredentials);
+        StandardUsernamePasswordCredentials usernamePasswordCredentials = (StandardUsernamePasswordCredentials) c;
+        assertEquals(credential.getId(), usernamePasswordCredentials.getId());
+        assertEquals(credential.getPassword().getPlainText(),usernamePasswordCredentials.getPassword().getPlainText());
+        assertEquals(credential.getUsername(),usernamePasswordCredentials.getUsername());
```
stephenc: Missing a test where there is a credential with the same ID in the system store, verifying that your proxy store is a higher ordinal and hence will get asked first

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
stephenc: you ACKed but did nothing in this method

stephenc: and in fact you kind of got the idea backwards where you applied it in `getCredentials` ....

The DomainSpecification should be in the property so that you can only return the credential for usage against the GitHub server.

1. User enters `https://github.com/jenkinsci/blueocean-plugin` as SCM URL
2. doFillScanCredentialsItems creates a domain requirements from that URL and asks for all appropriate credentials
3. Your provider gets asked for credentials... you need to apply your property's `DomainSpecification` to see if this matches the domain that you are exposing the user's credentials as
4. Ok, so that matches, now we go and check the user store for credentials with ID (from property) and in a domain matching the `DomainRequirement` of the `BlueOceanProxyDomainRequirement`

You have done 4, you still need 3 or the users credentials will leak further than intended

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
```diff
@@ -0,0 +1,307 @@
...
+            StoreImpl(AbstractFolder abstractFolder) {
+                super(BlueOceanCredentialsProvider.class);
+                this.abstractFolder = abstractFolder;
+                this.storeAction = new CredentialsStoreActionImpl(this);
+            }
+
+            @Nonnull
+            @Override
+            public List<Domain> getDomains() {
+                FolderPropertyImpl prop = (FolderPropertyImpl) abstractFolder.getProperties().get(FolderPropertyImpl.class);
```
stephenc: No, your property should have its own domain with the domain specification that it is exposing the proxy credentials on... and that is the domain you should be returning here... not the domain object instance from the user's store

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanCredentialsProvider.java
```diff
@@ -0,0 +1,307 @@
...
+
+            @Nonnull
+            @Override
+            public ModelObject getContext() {
+                return abstractFolder;
+            }
+
+            @Override
+            public boolean hasPermission(@Nonnull Authentication a, @Nonnull Permission permission) {
+                // its read only so for all permissions other than READ, we return false
```
stephenc: +1

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanDomainSpecification.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @Nonnull
+    @Override
+    public Result test(@Nonnull DomainRequirement scope) {
+        if(scope instanceof BlueOceanDomainRequirement){
+            if(((BlueOceanDomainRequirement)scope).getValue().equals(DOMAIN_SPECIFICATION)){
+                return Result.POSITIVE;
+            }
+        }
+        return Result.NEGATIVE;
```
stephenc: FYI by returning  NEGATIVE we prevent the credential from being visible to the user until they are searching with the `BlueOceanDomainRequirement`. In other words, `BlueOceanDomainRequirement` is *MUST HAVE*.

If you want the user to be able to use the credential in other contexts without the context being aware of your requirement then you may want to return `UNKNOWN`.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanDomainSpecification.java
```diff
@@ -0,0 +1,26 @@
...
+ * @author Vivek Pandey
+ */
+public class BlueOceanDomainSpecification extends DomainSpecification {
+    public static final String DOMAIN_SPECIFICATION = "BlueOcean Credentials Domain Specification";
+
+    @Nonnull
+    @Override
+    public Result test(@Nonnull DomainRequirement scope) {
+        if(scope instanceof BlueOceanDomainRequirement){
+            if(((BlueOceanDomainRequirement)scope).getValue().equals(DOMAIN_SPECIFICATION)){
```
stephenc: Redundant test IIUC

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/BlueOceanDomainSpecification.java
```diff
@@ -0,0 +1,26 @@
...
+                return Result.POSITIVE;
```
stephenc: Probably should be `PARTIAL`

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/CredentialsUtils.java
```diff
@@ -0,0 +1,180 @@
...
+            @Override
+            public boolean apply(@Nullable CredentialsStore s) {
+                return s!= null && s.hasPermission(CredentialsProvider.CREATE) && s.hasPermission(CredentialsProvider.UPDATE);
+            }
+        });
+
+    }
+
+    private static List<DomainSpecification> generateDomainSpecifications(@Nullable URI uri){
+        if (uri == null) {
```
stephenc: You are aware of https://github.com/jenkinsci/credentials-plugin/blob/master/src/main/java/com/cloudbees/plugins/credentials/domains/URIRequirementBuilder.java

vivek:
@stephenc I think I have all in place. Give a bee if it all looks good, once its in place I can ask Cliff to give it a try before calling it done:)

- proxy specific domain (with DomainSepcification creating using uri) attached to folder property.
- test case added to pick correct BO specific key and not from system. some other test enhancements.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
```diff
@@ -11,8 +11,10 @@
  * @author Vivek Pandey
  */
 public class GithubEnterpriseScm extends GithubScm {
-    private static final String DEFAULT_ENTERPRISE_API_SUFFIX = "/api/v3";
-    private static final String ID = "github-enterprise";
+    static final String DEFAULT_ENTERPRISE_API_SUFFIX = "/api/v3";
+    static final String ID = "github-enterprise";
```
vivek: Missed, that let me add that one.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
```diff
@@ -28,6 +30,11 @@ public GithubEnterpriseScm(Reachable parent) {
         return super.getUri()+DEFAULT_ENTERPRISE_API_SUFFIX;
     }
 
+    @Override
+    public String getCredentialDomainName() {
+        return DOMAIN_NAME;
```
vivek: better to keep these id's in their own domain, for example GitHub enterprise is in it's own domain so is consistent that way.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
```diff
@@ -55,35 +66,102 @@ public BluePipeline create(Reachable parent) throws IOException {
...
+                    //add the properties otherwise simply use it
+                    Domain domain = CredentialsUtils.findDomain(credentialId, authenticatedUser);
+                    if(domain == null){ //this should not happen since validateCredentialId found the credential
+                        throw new ServiceException.BadRequestExpception(
+                                new ErrorMessage(400, "Failed to create pipeline")
+                                        .add(new ErrorMessage.Error("scm.credentialId",
+                                                ErrorMessage.Error.ErrorCodes.INVALID.toString(),
+                                                "No domain in user credentials found for credentialId: "+ scmConfig.getCredentialId())));
+                    }
+                    if(domain.test(new DomainRequirement())) {
```
vivek: yes it is, good catch.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
```diff
@@ -55,35 +66,102 @@ public BluePipeline create(Reachable parent) throws IOException {
...
+                        ((OrganizationFolder) item)
+                                .addProperty(
```
vivek: good point, will add validation of this.

vivek: No duplicate test needed, this code creates a new org folder and is add property right after creation.

vivek:
@stephenc the one issue we should open is handling case where admin turns off access to user credential provider or disables storing credentials in these providers. and another one is BO provider icon.

cliffmeyers:
@vivek testing this on my end looks good. Proper access tokens appear to be used during both org scanning and branch scanning, and I've avoided any rate limit exceptions at least for the modest orgs I've been testing with. :)

PR #786, opened 2017-02-03
Only need it once

michaelneale:
# Description

cc @i386 @vivek I saw this warning, given it is required now, can be removed from test scope right? 

@reviewbybees


michaelneale:
not actually sure why we had to put this in there, but we did

vivek:
Good catch 🐝

PR #791, opened 2017-02-06
[FIX JENKINS-41280] Hide organization prefix from UI

tfennelly:
# Description

See [JENKINS-41280](https://issues.jenkins-ci.org/browse/JENKINS-41280).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigStatePreloader.java
+++ b/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigStatePreloader.java
@@ -55,11 +55,17 @@ public class BlueOceanConfigStatePreloader extends PageStatePreloader {
                         .key("loginUrl").value(jenkins.getSecurityRealm() == SecurityRealm.NO_AUTHENTICATION ? null : jenkins.getSecurityRealm().getLoginUrl())
                         .key("authorizationStrategy").object()
                             .key("allowAnonymousRead").value(allowAnonymousRead)
-                            .endObject()
-                        .key("enableJWT").value(BlueOceanConfigProperties.BLUEOCEAN_FEATURE_JWT_AUTHENTICATION)
                         .endObject()
+                        .key("enableJWT").value(BlueOceanConfigProperties.BLUEOCEAN_FEATURE_JWT_AUTHENTICATION)
                     .endObject()
-                .endObject();
+                .endObject()
+                // Environment properties. If more "env" vars are added, we could just iterate the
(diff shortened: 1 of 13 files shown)


i386:
Awesome!

May want to hold on merging this until Josh lands the header

i386:
@tfennelly  Safe to go ahead with now - @sophistifunk has landed his changes.

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigStatePreloader.java
```diff
@@ -55,11 +55,17 @@ public String getStateJson() {
...
-                        .key("enableJWT").value(BlueOceanConfigProperties.BLUEOCEAN_FEATURE_JWT_AUTHENTICATION)
                         .endObject()
+                        .key("enableJWT").value(BlueOceanConfigProperties.BLUEOCEAN_FEATURE_JWT_AUTHENTICATION)
                     .endObject()
-                .endObject();
+                .endObject()
+                // Environment properties. If more "env" vars are added, we could just iterate the
+                // system props and add any starting with "blueocean.". However, lets not do that
+                // unless there are more than a few.
+                .key("env").object()
```
i386: How about `features` instead of `env`? Can reuse this for feature flags later?

michaelneale:
Needs a bit of merge conflict resolving, and we don't have an ATH to check it with

michaelneale:
code & functionality - NICE

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigStatePreloader.java
tfennelly: Yeah, I was thinking of doing that but decided to just go with "env". Happy to change it to "features".

michaelneale:
nice one!

@i386 I think this is all ok, does this look right: 

screen shot 2017-02-08 at 1 35 40 pm

(you can see the Keith organization in there). 

LGTM - would like to see it updated to master if possible of course, but LGTM

i386:
Nice one Tom :)

tfennelly:
@michaelneale hey Mic .... any idea where the "Keith Zantow" org came from? It looks like it's from the favourites, but it should be gone from those (as you can see from the other cards).

michaelneale:
@tfennelly that is because I added github.com/kzantow as github org (default name it picks) so that is 100% correct.

PR #793, opened 2017-02-06
[FIX JENKINS-41638] "Open Blue Ocean" popup for when the jenkins url doesn't match current url

tfennelly:
# Description

See [JENKINS-41638](https://issues.jenkins-ci.org/browse/JENKINS-41638).

Basically ... a check on the current browser url, comparing it to the expected Blue Ocean url (which is constructed based on the configured Jenkins URL. If the bit before "/blue" doesn't match, then show a "configure the Jenkins URL" popup when the user clicks to open blue ocean.

See this little recording: https://youtu.be/tTV4wuoxEeA

I didn't make any real effort wrt styling the little popup. I'm not the designer, so not going to waste time. Maybe @brody could direct please?

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-web/src/main/js/try.js
+++ b/blueocean-web/src/main/js/try.js
@@ -9,17 +9,56 @@ function getAppUrl() {
     }
 }
 
-$(document).ready(() => {
+function aHref(tryBlueOceanUrl) {
     var tryBlueOcean = $('<a id="open-blueocean-in-context" class="try-blueocean header-callout">Open Blue Ocean</a>');
+    tryBlueOcean.attr('href', tryBlueOceanUrl);
+    return tryBlueOcean;
+}
+
+function configureJenkinsUrl() {
+    var tryBlueOcean = $('<span id="open-blueocean-in-context" class="try-blueocean header-callout">Open Blue Ocean</span>');
+    var tryBlueOceanClickHandler = function() {
+        // Remove the click handler, preventing it from picking up clicks
+        // on the the popup (below).
+        tryBlueOcean.off();
+
+        var popup = $(`<span class="unable-to-open"><h3>Unable to Open Blue Ocean</h3>The configured Jenkins URL does not match the current URL.<p/><a href="${getAppUrl()}/configure">Please configure the Jenkins URL</a></span>`);
+        tryBlueOcean.append(popup);
+        popup.css({
+            top: tryBlueOcean.outerHeight() + 10,
+            right: (0 - (popup.outerWidth()/2 - tryBlueOcean.outerWidth()/2))
+        });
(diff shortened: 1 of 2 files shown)


michaelneale:
ok will try it out - it shouldn't pop up in case of the localhost though - how did you get it to show up?

tfennelly:
@michaelneale look at the little video.

I don't think there's a problem with that error appearing even if `localhost` not matching `127.0.0.1`. Yes, you could start getting clever there, but I think lets just insist on them being the same in order fort hat button to work.

michaelneale:
no I don't think it will fly making people set it for just the tyre kicking

tfennelly:
@michaelneale no ... that's not what's happening ... I modified my jenkins setting the URL to `127.0.0.1`. It was only then that the popup appears ... this will not click in for normal tyre kicking .. will only happen if the URL used to get to jenkins somehow changes (e.g. a proxy) wrt what's configured in the system config. You get it?

michaelneale:
ah ok - so ideally we only want this to happen in cases where we *know* this won't work, so I think that may cover it. Let me think on it a bit.

michaelneale:
hrm @i386 brings up a good point, the person seeing this probably doens't have access to fix it in all cases. The ideal is that it could limp along with a warning that they need to set the url. Otherwise we aren't really at parity with classic (classic shows a warning, but only on manage screen). Given we want to replace that some time, I am not sure if this is right (as a blue ocean manage screen would never show up). Didn't think of that :(

tfennelly:
Yeah, we could actually ignore where `localhost` and `127.0.0.1` doesn't match, but I don't think there's any need really. There's no reason why anyone would configure that (other than me in this case for testing 😄 ) and even if they do, we can just ask them to make things match up (even if it's `localhost` Vs `127.0.0.1`). Imo that's easier than trying to get overly clever in the code.

michaelneale:
no my point is that at some point, blue ocean will still have to work with a mismatched config, just like classic does, if it is to offer a config screen itself.

tfennelly:
ok ... show something other than the link if the user doesn't have admin privs (or whatever is required)?

tfennelly:
screenshot 2017-02-06 21 47 25

brody:
@tfennelly Could we use the toast design, in the top right hand corner? Similar interaction to [this](http://codepen.io/brody/pen/ggxLVO) but no dark overlay and reversing the transition fade in & down, instead of up.

blue ocean url issue

tfennelly:
@brody thanks Brody ... I'll do that if we come back to this approach.

PR #812, opened 2017-02-10
Feature/jenkins 38598 create pipeline from GitHub integration

cliffmeyers:
# Description
- There are some API changes in this PR, notably:
   - Different steps of the flow are referred to as "States" rather than "statuses" since that term is used by the visual components and it was easy to confuse.
   - `FlowManager` changed a good bit. Notably the new "renderStep" method that requires that the `stateId` ("name") be specified in addition to the `stepElement` (actual React element). It also has a `afterStateId` parameter that indicates which state/step it should be placed after (destroying anything already there). The old API used simple `push` or `pop` semantics, but this was a problem when the user scrolled up and started to fiddle with previous steps, as it throw extra views onto the bottom. The new API is tighter and throws errors when not used properly.
- `MessageEnricher` changes can be ignored as they being handled in #810. 
- There is a lot of JSON in here that was used for mocking the services while they were in development. It can be ignored.
- This feature will be behind a feature flag and has blessing from @michaelneale and @i386 to get this in even though it's not in a final state yet.  (I am also trying to keep the PR from getting any larger than it already is. Things that follow on:
   - i18n
   - Making the repository list filterable
   - Automatically scrolling the browser vertically as new steps are added
   - Refactor of Git and Github code into appropriate submodules as defined here: https://issues.jenkins-ci.org/browse/JENKINS-41139
- See [JENKINS-38598](https://issues.jenkins-ci.org/browse/JENKINS-38598).
- To test, navigate to `/create-pipeline` and choose "Github"

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


blueocean-events/src/main/java/io/jenkins/blueocean/events/JobIndexingMessageEnricher.java
```diff
@@ -0,0 +1,106 @@
...
+            }
+        }
+    }
+}
+
+// TODO: move to pubsub-light's EventProps?
+enum OrgFolder {
+    job_orgfolder_indexing_result,
+    job_orgfolder_indexing_status,
+}
```
cliffmeyers: These changes are being tracked in #810 but included in this branch to make the "completion" flow work.

michaelneale:
So some comments from a first pass review: 

* really need to work on the "0 pipelines" thing as discussed - in the case when I picked a repo that has nothing, it could be clearer to say there were none discovered yet. 
* when I picked one that DID have Jenkinsfiles, and it picked up branches, it still shows as zero at the end: 

screen shot 2017-02-14 at 11 51 13 am

* Edge case: there is a repo I like to sometimes play with: kzantow/failure-project, yet I am not a member of "kzantow". There is no way for me to add that repo as a pipeline (as it is in a personal account). Perhaps this is ok cc @i386 ? but it does seem like a gap

michaelneale:
Some other feedback: 

* Really does need that searchable list
* Loading a list of repos takes "too long" (probably due to it paging before displaying? makes me 3ant to have a text box to just type things in if I was in a hurry)
* Tested private, public, private in user, private in org, discover in user, discover in org, one at a time and adding more to the repo - all seem good: combination wise. 

The comments above still apply - there is a lot of confusing messages that say things like: `Saving was successful. Pipeline creation is in progress. 0 pipelines have been created.` which could probably be reworded (even if things don't otherwise change) as I am yet to see it say anything but 0 - but it always works. 

* I tested removing my personal access token, this broke it when it checked the credential: 

```
http://localhost:8080/jenkins/blue/rest/organizations/jenkins/scm/github/organizations/?credentialId=github Failed to load resource: the server responded with a status of 500 (Server Error)
blueocean.js:2107 Uncaught (in promise) Error: Server Error
    at checkStatus (blueocean.js:2107)
```

which it looks like it is trying to cater for but failing (it should offer to take a new credential from me, which would make sense) - so that probably needs to be looked into.

michaelneale:
I think there are a few possibly follow on tickets above, but I think fixing the "copy" a bit and dealing with the disabled access token probably should be addressed ASAP (probably along with taking people to the proper place at the end) - the rest can be follow on tickets, as it is probably time to get this in master.

blueocean-dashboard/src/main/js/creation/ScmProvider.js
```diff
@@ -23,4 +23,11 @@ export default class ScmProvider {
         throw new Error('must implement getFlowManager');
     }
 
+    /**
+     * Called when the creation flow is about to end. Perform cleanup here.
+     */
+    destroyFlowManager() {
+        throw new Error('must implement destroyFlowManager');
+    }
+
```
sophistifunk: Is this called based on some user action that will never happen if the user just closes their tab, for example? We don't want to encourage implementers to put important stuff in here in that case.

blueocean-dashboard/src/main/js/creation/flow2/FlowStepStatus.js
```diff
@@ -5,8 +5,14 @@ const status = {
     ACTIVE: 'active',
     COMPLETE: 'complete',
     INCOMPLETE: 'incomplete',
+    ERROR: 'error',
 
-    values: () => [status.ACTIVE, status.COMPLETE, status.INCOMPLETE],
```
sophistifunk: Seems this functionality should go into a constructor if it's a pattern, or we want to make sure `values(`) stays correct over the life of the code? Something like:

```
const status = new FakeEnum({
    ACTIVE: 'active',
    COMPLETE: 'complete',
    INCOMPLETE: 'incomplete',
    ERROR: 'error',
});
```

blueocean-dashboard/src/main/js/creation/git/GitCompletedStep.jsx
```diff
@@ -4,7 +4,7 @@ import { observer } from 'mobx-react';
 import { buildPipelineUrl } from '../../util/UrlUtils';
 import FlowStep from '../flow2/FlowStep';
 import StepStatus from '../flow2/FlowStepStatus';
-import FlowStatus from './GitCreationStatus';
+import STATE from './GitCreationState';
```
sophistifunk: Why is this uppercase here, but lowercase elsewhere?

blueocean-dashboard/src/main/js/creation/git/GitCreationState.js
```diff
@@ -1,20 +1,20 @@
 /**
  * Valid statuses for GitFlowManager.
  */
-const status = {
+const state = {
```
sophistifunk: ....Like here, it's lowercase

blueocean-dashboard/src/main/js/creation/github/api/mocks/ApiMock.js
```diff
@@ -0,0 +1,31 @@
+import { Promise } from 'es6-promise';
```
sophistifunk: Does this belong in `/src/main/` or `/src/test/`?

sophistifunk:
I don't see any major issues, just a few nitpicks commented inline.

cliffmeyers:
@michaelneale some quick responses:
- Not handling the "revoked token" use case correctly is definitely a bug. Working on that today.
- I will fix the issue where "single repo" mode after completion doesn't link the user to the Activity screen. That was an oversight.

cliffmeyers:
@michaelneale regarding the completion steps' copy and the display of pipeline quantity, this is something I was experimenting with in the absence of clear direction on how to provide meaningful feedback to the user. @i386 and I have a TODO to go through the flow together and refine all of the copy.

The quantity might ultimately have to be abandoned. The challenge is that all of this is based on SSE and the events are received in an unpredictable order (understandable given how async all this stuff is). The core problem is that we are keying off two different events:  "indexing_complete" for the child multi-branch pipelines that get created, and "indexing_complete" for the parent org folder itself. The code is keying off completion of the org folder to reflect the process truly being complete, and that point I unsubscribe from listeners which means the pipeline quantity might not be accurate if those events are still flowing in after the org folder completes. I could look at keying off the "crud_created" event for pipelines instead, perhaps those are fired before the indexing_complete ones.

There's also a problem in the re-entrant flow where we will receive indexing events for pipelines that are already created (I think).

So maybe it was a bad idea and we should just key off the org folder indexing and ignore anything else. Or we could just key off the first child multibranch indexing event we receive and do a subtle UI transition at this point. Transitioning from a message like "Waiting for pipelines to be created" to "Pipelines are being created." Ultimately I don't know if this subtlety will actually mean anything to users.

blueocean-dashboard/src/main/js/creation/ScmProvider.js
cliffmeyers: This would mainly be used for cleanup of timers or event listeners... similar to what is typically done in `componentWillUnmount`. I will add more details to the docs around its intended use. Is there anything else you'd like me to address on this, or concerns you have?

cliffmeyers: See commit 43d8c39

blueocean-dashboard/src/main/js/creation/github/api/mocks/ApiMock.js
cliffmeyers: While these could be used in automated tests, I was using them to get around GH rate limits and build / test the UI in advance of backend services being available. Do you think `src/test/js/mock` is a good location?

blueocean-dashboard/src/main/js/creation/flow2/FlowStepStatus.js
cliffmeyers: Fixed in 1e4b525

blueocean-dashboard/src/main/js/creation/git/GitCompletedStep.jsx
cliffmeyers: Fixed in 1e4b525

blueocean-dashboard/src/main/js/creation/git/GitCreationState.js
cliffmeyers: Fixed in 1e4b525

blueocean-dashboard/src/main/js/creation/github/api/mocks/ApiMock.js
sophistifunk: Hrm.... I'm not really sure in that case, so I say leave it where it is :)

blueocean-dashboard/src/main/js/creation/flow2/Enum.js
```diff
@@ -0,0 +1,31 @@
+/**
```
sophistifunk: 👍

cliffmeyers:
@michaelneale summary of updates:
- After saving "Just one repo" you should now get a button that links directly to that MBP's activity page.
- I improved the logic that handles the SSE's received after the initial save. It seemed to behave pretty well for me, but maybe your scenarios will still give it trouble. As I mentioned in an earlier comment, if it's still fickle perhaps we remove the numbers in favor of "better copy" that still gives the user a sense of what's going on.
- I am still working on the "revoked access token" scenario but am pretty close. The fix will definitely be ready tomorrow.

michaelneale:
ok look forward to it!

blueocean-dashboard/src/main/js/LoadingIndicator.js
```diff
@@ -44,7 +44,10 @@ export default {
     },
 
     hide() {
-        loadingCount--;
```
kzantow: wonder how this ended up being a problem?

cliffmeyers: If you navigate directly to `/blue/create-pipeline` the loading worm displays by default. I was hiding the indicator on mount, which set the value to -1, which wouldn't actually hide it. Maybe the better fix is for it not to display by default. WDYT?

kzantow: Right, shouldn't be an issue, since people wouldn't go directly to it typically. Although, this seems like a fine change to make sure.

blueocean-dashboard/src/main/js/creation/flow2/Enum.js
kzantow: seems like this should go in core-js or something

blueocean-dashboard/src/main/js/creation/flow2/FlowManager.js
```diff
@@ -21,43 +22,153 @@ export default class FlowManager {
...
-        this._stepsChanged();
+    changeState(stateId) {
+        if (!stateId) {
+            this._throwError('stateId is required');
+        }
+
+        const currentStep = this.steps[this.steps.length - 1];
+
+        if (currentStep.stateId === stateId) {
+            console.warn(`stateId already set to ${stateId}`);
```
kzantow: debugging?

blueocean-dashboard/src/main/js/LoadingIndicator.js
michaelneale: I think it is reasonable that people may go directly right? if it has a URI, it can be bookmarked and shared.

michaelneale:
LGTM - would like the "0 pipelines" dropped - if there are 0, don't say anything as it always looks wrong anyway but that can be done here or follow on

i386:
Don't worry about the zero Pipelines case for the moment - we need to come back and revisit the messaging anyway.

Thanks Cliff and everyone who's helped out to get this PR reviewed.

Ship it!!

michaelneale:
@cliffmeyers James things that we should drop the feature switch and get this in. However, before we cut a release, we need to do: https://issues.jenkins-ci.org/browse/JENKINS-41087 - ok?

i386:
@cliffmeyers though not in this PR ;)

cliffmeyers:
Yep, just waiting for CI to finish on this one. Should have a PR for 41087 tomorrow. Pretty simple change.

cliffmeyers:
ATH passed:  https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/master/1523/pipeline/18

PR #815, opened 2017-02-13
Use non alpine variant

michaelneale:
# Description

See https://issues.jenkins-ci.org/browse/JENKINS-41582

@rtyler - would it be ok to use non alpine variant? Will this work as simply as this? 

cc @ydubreuil

### DIFF
--- a/docker/official/Dockerfile
+++ b/docker/official/Dockerfile
@@ -1,5 +1,5 @@
 # This image is based off the latest Jenkins LTS
-FROM jenkins:alpine
+FROM jenkins:latest
 
 USER root
 


rtyler:
@michaelneale From the original ticket

> The reason is lack of gcc and other tools that affect the first users experience of people doing things outside of the jvm: eg nodejs. 

This change still does absolutely nothing to address this "requirement." Not only that, the "requirement" is easily satisfied both on alpine and debian-based images. (`apk add` vs. `apt-get install`)


Both the alpine and debian-based official images from the Jenkins project also run inside the container as "jenkins", as I mentioned in [this JIRA comment](https://issues.jenkins-ci.org/browse/JENKINS-41582?focusedCommentId=286378&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-286378). Meaning a user would still need to build their own custom container anyways.

I feel like there's still a fundamental misunderstanding of Docker norms and practices here, or what I have written isn't being read.

crazy pills

michaelneale:
@rtyler oh I was under the impression that getting tools on alpine is NOT that easy (eg try running nodejs on alpine - you can't just install build tools via apk, is my understanding). Otherwise yes.

rtyler:
> @rtyler oh I was under the impression that getting tools on alpine is NOT that easy (eg try running nodejs on alpine - you can't just install build tools via apk, is my understanding). Otherwise yes.

```
➜  ~  docker run --rm -ti alpine sh
/ # apk update
fetch http://dl-cdn.alpinelinux.org/alpine/v3.5/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.5/community/x86_64/APKINDEX.tar.gz
v3.5.1-34-g1d3b13bd53 [http://dl-cdn.alpinelinux.org/alpine/v3.5/main]
v3.5.1-29-ga981b1f149 [http://dl-cdn.alpinelinux.org/alpine/v3.5/community]
OK: 7958 distinct packages available
/ # apk add nodejs
(1/8) Installing ca-certificates (20161130-r0)
(2/8) Installing libcrypto1.0 (1.0.2k-r0)
(3/8) Installing libgcc (6.2.1-r1)
(4/8) Installing http-parser (2.7.1-r0)
(5/8) Installing libssl1.0 (1.0.2k-r0)
(6/8) Installing libstdc++ (6.2.1-r1)
(7/8) Installing libuv (1.9.1-r0)
(8/8) Installing nodejs (6.9.2-r1)
Executing busybox-1.25.1-r0.trigger
Executing ca-certificates-20161130-r0.trigger
OK: 43 MiB in 19 packages
/ # node --version
v6.9.2
/ # 
```

michaelneale:
@rtyler to run an npm install will require a bunch of other compiler tools etc

I was looking at the docker socket solution, which requires the docker client installed (and the installation instructions for alpine don't seem to work)

michaelneale:
@rtyler what about including the docker binary in the image? (ads 60M to size it seems) - worth it?

michaelneale:
Discussed on IRC - won't be doing this.

PR #817, opened 2017-02-13
Bump pipeline and git dependencies

i386:
# Description

@michaelneale @vivek a number of new pipeline updates are available we should integrate and test with.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-pipeline-api-impl/pom.xml
+++ b/blueocean-pipeline-api-impl/pom.xml
@@ -109,7 +109,10 @@
             <artifactId>unirest-java</artifactId>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
--- a/pom.xml
+++ b/pom.xml
@@ -265,7 +265,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-multibranch</artifactId>
-            <version>2.13</version>
+            <version>2.12</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,7 @@
 
   <properties>
     <java.level>7</java.level>
-    <jackson.version>2.2.3</jackson.version>
+    <jackson.version>2.8.7</jackson.version>
     <jenkins.version>2.7.1</jenkins.version> <!-- Should be kept in sync with Dockerfile FROM statement-->
     <javadoc.exec.goal>javadoc-no-fork</javadoc.exec.goal> <!-- stop initialize phase plugins executing twice -->
     <node.version>6.4.0</node.version>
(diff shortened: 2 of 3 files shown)


michaelneale:
Still isn't happy with it - probably something else is pulling in 2.10.. perhaps hold off until workflow-api has been fixed?

pom.xml
```diff
@@ -302,6 +302,11 @@
             <artifactId>github-organization-folder</artifactId>
             <version>1.6</version>
         </dependency>
+        <dependency>
```
michaelneale: Is there a reason this one is explicitly set?

michaelneale:
passes ATH https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/1718/

blueocean-pipeline-api-impl/pom.xml
```diff
@@ -109,7 +109,10 @@
             <artifactId>unirest-java</artifactId>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
```
michaelneale: @i386 any reason this is here and not test scoped?

i386: Scopes are inherited

michaelneale: nice.

pom.xml
```diff
@@ -230,27 +230,27 @@
...
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.25</version>
+            <version>2.28</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.7</version>
+            <version>2.12-alpha-2</version>
```
svanoort: @i386 We're ready to take this up to 2.12 now

michaelneale:
@i386 I can't tell what your latest changes are - but they have broken both unit tests and the ATH

pom.xml
```diff
@@ -17,7 +17,7 @@
 
   <properties>
     <java.level>7</java.level>
-    <jackson.version>2.2.3</jackson.version>
+    <jackson.version>2.8.7</jackson.version>
```
michaelneale: why?

michaelneale:
:bug: some failures and a lot of dependency changes that need a bit more description/work.

michaelneale:
I don't think we have the luxury of bumping deps all at once unless we have time to baby sit all the tests/ATH. wouldn't it make sense to focus on workflow-api on its own?

blueocean-pipeline-api-impl/pom.xml
michaelneale: so - why is this dependency here? It wasn't used before right?

michaelneale:
@i386 did you squash commits here ? as I Can't see changes between today and yesterday to know what caused breakage (best to not squash when pushing to a branch for this reason).

michaelneale:
:bee: LGTM

jglick:
Revert 45aa4f0da66b7439023464bb960d0fcca7a8e139 and fix your tests to `checkout scm`.

michaelneale:
@jglick ?? commenting on long closed PRs?

jglick:
> commenting on long closed PRs?

Two days! Well within the range of my review backlog, unfortunately…

> you mean that changing tests to not use git but use checkout scm

No, IIUC the problem is that there are tests which create a `Jenkinsfile` which is just, say,

```groovy
echo 'hi!'
```

and as of 2.13 (using the now-default lightweight mode—controllable via system property) successive builds of that branch project will not display any SCM changelog, because Jenkins is not running all the machinery of  `SCM.checkout` and its changelog parsing—it is just grabbing the `Jenkinsfile` straight out of the selected revision (in the case of GitHub, via API). So the test should actually do something with the source code:

```groovy
node {
  checkout scm
  echo 'hi!'
}
```

Now I *might* try to implement changelog generation even in lightweight mode—in principle the `SCMFileSystem` APIs make it possible, though AFAIK this code path is untested. But it would be a fair amount of work for what seems like a corner case so it is not exactly a high priority.

michaelneale:
@jglick haha yes 2 days - ancient history!

OK I will take a look at this today, seems pretty simple. If that works, makes sense. They are contrived cases. They are only "echo" so they would in theory work on windows (wishful thinking there though).

jglick:
> they would in theory work on windows

`checkout scm` should be safe on Windows too. But rather than guess, why not patch your own `Jenkinsfile` to run your tests on the label `windows`, which ci.jenkins.io supports?

michaelneale:
@jglick we aren't yet running on ci.jenkins.io - but a dogfood instance, but otherwise yes. What we are planning is some windows + browser permutations first before building js on windows (I believe it can work though, but its more of a pain than it should be with npm).

vivek:
@jglick yeah, we will fix the test to do that and upgrade to workflow-multibranch 2.13. thanks.

PR #820, opened 2017-02-13
[FIX-JENKINS-41421] Extend StepImpl to pass along submitterParameter

scherler:
# Description

See [JENKINS-41421](https://issues.jenkins-ci.org/browse/JENKINS-41421).

Basic change in PipelineStepImpl.java rest is to adopt to the v. 2.5 of pipeline-input-step

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
@@ -43,6 +43,7 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.TimeoutException;
 
 /**
  * @author Vivek Pandey
@@ -178,13 +179,13 @@ public class PipelineStepImpl extends BluePipelineStep {
                     " have an InputAction.");
         }
 
-        InputStepExecution execution = inputAction.getExecution(id);
-        if (execution == null) {
-            throw new ServiceException.BadRequestExpception(
-                    String.format("Error processing Input Submit request. This Run instance does not" +
-                    " have an Input with an id of '%s'.", id));
-        }
         try {
+            InputStepExecution execution = inputAction.getExecution(id);
+            if (execution == null) {
+                throw new ServiceException.BadRequestExpception(
+                        String.format("Error processing Input Submit request. This Run instance does not" +
+                        " have an Input with an id of '%s'.", id));
+            }
             //if abort, abort and return
             if(body.get(ABORT_ELEMENT) != null && body.getBoolean(ABORT_ELEMENT)){
                 return execution.doAbort();
@@ -200,7 +201,7 @@ public class PipelineStepImpl extends BluePipelineStep {
                 listener.onStepContinue(execution.getInput(), run);
             }
             return response;
-        } catch (IOException | InterruptedException | ServletException e) {
+        } catch (IOException | InterruptedException | TimeoutException e) {
             throw new ServiceException.UnexpectedErrorException("Error processing Input Submit request."+e.getMessage());
         }
     }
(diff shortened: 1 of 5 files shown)


scherler:
@i386 I tried to add the `submitterParameter` to the inputParameter step as shown in https://github.com/scherler/jenkins-testfiles/commit/8745249d6d5b619e75bedddfa6b014c5906ec1a4 but that is no good. Can you provide an example to pass `submitterParameter` to the inputParameter?

michaelneale:
@jglick where is it copied from? What should be all rewritten?

michaelneale:
@scherler the ticket linked talks about text areas - which I don't think this addresses. I am confused as to what this is about.

jglick:
> where is it copied from?

[`InputStepExecution`](https://github.com/jenkinsci/pipeline-input-step-plugin/blob/1031ad8008f2b84e1ca331e3eff7839d97922169/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/InputStepExecution.java)

scherler:
@jglick yeah, since we have our own implementation of it. Maybe @vivek has some thoughts.

@michaelneale the ticket talks about that the `submitterParameter` is not passed to the flow and in case you only have one parameter that leads to passsing a string not a map to the pipeline api.

michaelneale:
@scherler but not about Text input (text area?) which seemed one of the other subjects? 

@vivek will be reviewing this, and this may be related to: https://issues.jenkins-ci.org/browse/JENKINS-41044 which he also wanted to look at, so lets see what he thinks.

scherler:
@michaelneale textarea had been supported by the first implementation. The problem is that without the PR the linked pipeline cannot be submitted in BO without causing an error.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
```diff
@@ -241,6 +243,12 @@ private Object parseValue(InputStepExecution execution, JSONArray parameters, St
             }
             mapResult.put(name, convert(name, v));
         }
+        // If a destination value is specified, push the submitter to it.
+        String valueName = input.getSubmitterParameter();
+        if (valueName != null && !valueName.isEmpty()) {
```
vivek: Perhaps use StringUtils.isNotBlank(valueName)?

scherler: @vivek as @jglick pointed out, this comes from https://github.com/jenkinsci/pipeline-input-step-plugin/blob/1031ad8008f2b84e1ca331e3eff7839d97922169/src/main/java/org/jenkinsci/plugins/workflow/support/steps/input/InputStepExecution.java how could we reduce that redundancy?

scherler:
@michaelneale BTW this PR fixes https://issues.jenkins-ci.org/browse/JENKINS-41044

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
vivek: Yeah, possibly InputStepImpl.parseValue() should be refactored to avoid working off form submission and work just off JSONArray and also expose this and some other method public. I noted that time in PR/comment but missed to follow it up with JIRA/PR in pipeline-onput-step plugin. I will follow it up.

vivek:
@scherler LGTM 🐝 pending ATH run. Thanks for fixing:)

michaelneale:
@scherler NICE!

Checked it too, looks good. ATH is happy - and @vivek gave it blessing. Nice catch! 

And yes ideally wouldn't be repeating code in InputStepExecution but that was existent before this PR. 

:bee:

PR #834, opened 2017-02-16
hotfix/time harmonizer defend npe

cliffmeyers:
# Description
- Defend against the function not being made available on `context.config` - this was breaking storybooks in personalization, and also requires some mocking in `PipelineCard-spec` that would be nice to avoid.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-core-js/src/js/components/TimeHarmonizer.jsx
+++ b/blueocean-core-js/src/js/components/TimeHarmonizer.jsx
@@ -13,7 +13,8 @@ export const TimeHarmonizer = ComposedComponent => {
             super(props, context);
             logger.warn(props, context);
             const { startTime } = this.props;
-            this.skewMillis = this.context && this.context.config ? this.context.config.getServerBrowserTimeSkewMillis() : 0;
+            this.skewMillis = this.context && this.context.config && this.context.config.getServerBrowserTimeSkewMillis ?
+                this.context.config.getServerBrowserTimeSkewMillis() : 0;
             this.durationMillis = startTime ? this.getTimes(props).durationMillis : 0;
             this.getI18nTitle = this.getI18nTitle.bind(this);
             this.getDuration = this.getDuration.bind(this);


cliffmeyers:
cc @scherler 

I have a larger question which is why we are passing this function down in the React context? Couldn't we just export the function from `core-js`? And let `blueocean-web` import it and set the skew value in appropriate place? Then we wouldn't have to worry about mocking it out in different places, or worry about it being null.

michaelneale:
:bee:

scherler:
@cliffmeyers https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-core-js/src/js/config.js in there yeah makes lot of sense

michaelneale:
any change this fixes this one? https://issues.jenkins-ci.org/browse/JENKINS-42318

michaelneale:
@cliffmeyers looks like this one is worth it (no biggie)

cliffmeyers:
@michaelneale looking at the code that produces that error message, I don't think it's related to this change. The error this was fixing was mostly coming up in a unit test or storybook environment if I recall. It would probably be better to make the broader fix discussed above once higher priority items get wrapped up.

michaelneale:
@cliffmeyers SGTM

michaelneale:
still seeing harmonizer errors - so this and more is still relevant IMO

cliffmeyers:
@michaelneale if we merge this now, that's fine, but let's add a ticket to describe the better fix and slap it into the tech debt epic? If you agree w/ that approach, I'll make the ticket later today and the merge this sucker.

michaelneale:
@cliffmeyers we can leave this if you like - just ensure there is a labelled ticket for it. 
@scherler may have some changes making this irrelevant anyway, so I think can close this but keep a link to the branch behind it in the ticket.

michaelneale:
Closing this - feel free to resurrect one day if important (just too many olde PRs hanging around)

PR #835, opened 2017-02-16
[FIX JENKINS-42068] Ignore queue-enter MBP indexing event in `core-js:services/DefaultSSEHandler.js`

tfennelly:
# Description

[See JENKINS-42068](https://issues.jenkins-ci.org/browse/JENKINS-42068)

Side effect of the new MBP indexing events. Results in an "undefined" row inserted into the list of branches in a MBP.

screenshot 2017-02-16 17 39 11

@reviewbybees esp @cliffmeyers

### DIFF
--- a/blueocean-core-js/npm-shrinkwrap.json
+++ b/blueocean-core-js/npm-shrinkwrap.json
@@ -1,6 +1,6 @@
 {
   "name": "@jenkins-cd/blueocean-core-js",
-  "version": "0.0.74",
+  "version": "0.0.75",
   "dependencies": {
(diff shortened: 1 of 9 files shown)


cliffmeyers:
@tfennelly actually, thinking about this a little more... would this prevent queue items corresponding to freestyle jobs being added?

michaelneale:
oh good catch :bee: and ATH is happy

tfennelly:
@cliffmeyers I expanded the condition for ignoring the event, so should be fine now.

PR #839, opened 2017-02-16
[FIX-JENKINS-38523] mobx driven rendering of karaoke

scherler:
# Description

See [JENKINS-38523](https://issues.jenkins-ci.org/browse/JENKINS-38523).
[Acceptance PR130](https://github.com/jenkinsci/blueocean-acceptance-test/pull/130)

# Submitter checklist
https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/karaoke_reloaded/1/pipeline

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees


michaelneale:
Build passes but ATH is choking on ping test: 

https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/1871/console

I'll try and take a look today sometime

michaelneale:
Nice to see this move along.. 

I can still see flashing of the empty state when it moves along, it seems worse than before: 


This was using the https://github.com/cloudbeers/blueocean-demo-project project

If I use a project like: 

https://github.com/multibranchorg/app-store-demo and load up completed runs, the flashing is much worse than before (even when finished).

michaelneale:
Another :bug:

Following along a pipeline that has parallel branches, and I clicked back to an earlier stage that was completed, and it stole focus while I was reading when a branch finished executing, vs leaving me on that stage reading the step logs. 

I can confirm that you can click on different parallel nodes at least while running, however the above may be a regression. 

Also, when a run starts, before anything happens, it shows the final node highlighted, which doesn't seem right.

michaelneale:
Another :bug: when running log-karaoke/freestylePing.js seems to do with a race condition: 


Note how it goes green and halts the progressive log - so it can never see the end message which the test is looking for. 

In this case, it is getting the success early, before the log has come in. ie the header goes green, and that stops the progressive log. Perhaps this is due to the change

michaelneale:
Well if you run `nightwatch src/test/js/log-karaoke`

✖ log-karaoke/freestyle

✖ log-karaoke/freestylePing

✖ log-karaoke/input

✖ log-karaoke/noStages

 ✖ log-karaoke/stages-block

 ✖ log-karaoke/stages-new

 ✖ log-karaoke/stages-old

So this is picking up quite few things, but likely just a few bugs all showing up in different ways.

scherler:
@michaelneale 

> Following along a pipeline that has parallel branches, and I clicked back to an earlier stage that was completed, and it stole focus while I was reading when a branch finished executing, vs leaving me on that stage reading the step logs.

When you refer to the `job` has complete then the view get updated, that is currently the same as we have. If you refer to follow along karaoke and then stop it and then click earlier node. That should prevent further refreshes until the overall job has finished. Will have a closer look while testing

> I  confirm that you can click on different parallel nodes at least while running, however the above may be a regression.

See above, no if we talk of overall job has finished.

> Also,  when a run starts, before anything happens, it shows the final node highlighted, which doesn't seem right.

Yeah, I saw that as well and I should have a fix for that now.

> the flashing is much worse than before (even when finished, which implies somehting is quite wrong, as it flashes a few times even when loading a completed run).

That is based on the pending states and I still finding the sweet spot ;)

> In this case, it is getting the success early, before the log has come in. ie the header goes green, and that stops the progressive log. Perhaps this is due to the change

Need to look into that one but the desired behaviour (as we have now) is refetch when the freestyle ends and display full log.

scherler:
I pushed some changes that should fix (or much enhance them) all your observations @michaelneale beside:

> Following along a pipeline that has parallel branches, and I clicked back to an earlier stage that was completed, and it stole focus while I was reading when a branch finished executing, vs leaving me on that stage reading the step logs.

that I need to look into a bit more, but I will jump on https://github.com/jenkinsci/blueocean-plugin/compare/bs-poc and come back to this

michaelneale:
@scherler ok great - I will take another look today.

michaelneale:
Fixed some things - I can still see it flashing once when you open a new job. 

I also have it in a state where I open a finished job and it won't let me select anything but the last tage. Clicking on an earlier stage just doesn't work: https://media.giphy.com/media/l4FGyucK0BAfSn2nu/giphy.gif

Also - clicking on an earlier stage while running still has the focus yanked when another event comes in for a later stage (ie isn't pausing follow along). 

I haven't retested other things or looked at the ATH yet though, but still some :bug: but not sure if new since last version

michaelneale:
Some progress, my testing results from last week: 

Using a project created from: https://github.com/multibranchorg/app-store-demo (you can fork that, and try running a few branches to see the following):

It flashes "waiting for ..." when I open a completed run (very briefly, hard to see)
When following along a pipeline with parallel I wasn't able to click back to an earlier stage - it kept steeling focus, or it wouldn't let me click at all sometimes.
When following along the pipeline above it didn't seem to get logs for some steps in the first stage (ie would just show one line for the build, and then that stage would complete)
When waiting for a run to start it says nothing - no empty state (it seems that empty state is mistakenly in the completed run - maybe some backwards logic there).
When I run master from branches tab - follow along to completion and close it, activity screen is not updated (always shows as running)
Clicking on activity seems to reload the page each time (ie switching tabs) - which is definitely not in master
Also - what are the changes specifically to the ATH to make it pass - can that be documented in that PR? (only have the commits to go by, but just want to see a little more detail).
So getting there, nice on the ATH branch, but above seems to still be a problem when I compare with master, but that is all I could see. Flashing is gone, and it feels snappy, just above 🐛's

blueocean-dashboard/package.json
```diff
@@ -39,10 +39,11 @@
     "skin-deep": "0.16.0"
   },
   "dependencies": {
-    "@jenkins-cd/blueocean-core-js": "0.0.89",
+    "@jenkins-cd/blueocean-core-js": "0.0.90-reload",
     "@jenkins-cd/design-language": "0.0.125",
     "@jenkins-cd/js-extensions": "0.0.33",
     "@jenkins-cd/js-modules": "0.0.8",
+    "@jenkins-cd/preferences": "0.0.2",
```
michaelneale: what is this? This doesn't seem to resolve it seems. Is this new?

scherler:
The preference module https://www.npmjs.com/package/@jenkins-cd/preferences can be found https://github.com/scherler/js-preferences but should go into the jenkins-ci organisation.

michaelneale:
so re:reviewing, haven't tried all the above but I now see an error in the console when following along the repo: 

```
blueocean.js:18697 [ERROR - io.jenkins.blueocean.dashboard.karaoke.Pager] Error fetching page TypeError: Already read
    at jenkins-js-extension.js:54082
```

https://github.com/multibranchorg/app-store-demo

scherler:
@michaelneale yeah I saw that when I refreshed a page that where already loading. seems to come from some fetch code. Need to investigate that some more.

michaelneale:
@scherler I think it may me exposing some of the other problems I saw - I posted above the things I still see that are a problem, better than thursday/friday but still those ones above.

scherler:
@michaelneale ok I will use the above repo and will fix those findings

scherler:
* It flashes "waiting for ..." when I open a completed run (very briefly, hard to see)

-> when pending on backend answer it will show "waiting for backend to response". You can deactivate this by setting `jenkins-instance/preferences:runDetails.pipeline.showPending` to `never` in the localStorage

* When following along a pipeline with parallel I wasn't able to click back to an earlier stage - it kept steeling focus, or it wouldn't let me click at all sometimes. 

-> that should be fixed now

* When following along the pipeline above it didn't seem to get logs for some steps in the first stage (ie would just show one line for the build, and then that stage would complete)

-> that could be a problem of the server response I could not reproduce this 

* When waiting for a run to start it says nothing - no empty state (it seems that empty state is mistakenly in the completed run - maybe some backwards logic there). 

-> that should be fixed now

* When I run master from branches tab - follow along to completion and close it, activity screen is not updated (always shows as running)

I cannot reproduce the always shows as running but I can reproduce the running state in activity on current master on https://ci.blueocean.io/blue/organizations/jenkins/freestyle/activity when you start a build do to the details, wait until finish and then close the modal, you see that the run is still running and then gets updated. Trying to say that is nothing special to the branch as I understand it.

* Clicking on activity seems to reload the page each time (ie switching tabs) - which is definitely not in master

-> I could not reproduce this 

* Also - what are the changes specifically to the ATH to make it pass - can that be documented in that PR? (only have the commits to go by, but just want to see a little more detail). 

They are related to ending a run and updating the details we did not do that before so I had to adopt the ATH a bit.

michaelneale:
Testing again, a few things that haven't changed on this branch, or may be new: 

* still not seeing logs reliably run. Did you try the repo: https://github.com/multibranchorg/app-store-demo it rarely works for me. Not at all showing now: 

screen shot 2017-03-28 at 6 42 39 pm

Master does not to this for exact same workspace/branch. 

* The "waiting for backend to respond" needs to go. It looks bad and there are other ways to show things are loading (look at creation, with the worm) - or it is fine to show nothing as it is clear anyway, there is no need for a big message here (the preference isn't really a solution to this, more of a demo it as I see it)

* Activity is STILL not updating: 

screen shot 2017-03-28 at 6 46 32 pm

It even has the toast at the bottom left still, most odd. Master does not do this for same workspace and branch

* I was able to see activity reload the whole page. Even opening a results screen once I saw that. Just leave it running for a while, then click on activity tab, or somewhere else. From time to time I see full page reloads which is not happening elsewhere - what could be triggering this? 

* New regression: it now shows "waiting" for a run to start, but it should also show the planned pipeline if there is one (ie if that branch has been run before), like this on master: 


screen shot 2017-03-28 at 6 51 40 pm

 

If you just use the above repo, and drive it lots and lots, you should see the above. There aren't errors in the console now which is good.

scherler:
@michaelneale https://youtu.be/o73QiOgNbbc I am using that branch and as you can see it works as expected. Regarding the activity that occurred in the video the first time for me, I am looking into that.

The waiting for server screen I deactivated now.

Regarding the showing the node and the waiting for run to start I will fix that.

scherler:
@michaelneale all issues should be fixed (as well the activity one) It seems that it was caused by 

```
this.props.augmenter.setKaraoke(false);
```

Which is a mobx controlled component, not sure why but as soon that the above was used the activity page did not work anymore. On a side note the activity page is as well driven by mobx, but I did not used the same objects (well the run, but I did not change that).

michaelneale:
Getting much better, almost there: 

* When you are waiting for a run to start - it shows an additional "Steps Build" text on the left that shouldn't be there (should be like the screenshot above) - that should be removed: 

screen shot 2017-03-29 at 2 14 12 pm


* I am still seeing something flashing in blue when loading steps now for a finished run - it looks like it says "no steps" but hard to see (it may be a loading message) - in any case, there shouldn't be any flashing of blue boxes visible I think

* When I open runs from time to time, it triggers a full page load. 
* Occasionally clicking on branches triggers a full page load (like the above, seems to happen only after a while, after opening and closing completed runs)
* When I opened a completed run, and close it, sometimes I can't scroll the activity list to the bottom, or to the top - it is like it has pinned my view somehow (this one is subtle)

These are now getting more subtle, so if it isn't clear how to repro, perhaps we should have a hangout one morning/evening and I can just take you through it (really hard to see some, but they are there). Main ones are the screen reloading which indicates osmething fishing is going on. 

good progress!

scherler:
@michaelneale regarding the flashing please set `jenkins-instance/logging/categories:io.jenkins.blueocean.dashboard.karaoke.Pipeline` to `DEBUG` and then filter for `EarlyOut - ` Further please check whether your localStorage has the `jenkins-instance/preferences:runDetails.pipeline.showPending`set to `never` if not either set it or remove it and reload.

I removed the logBar on queued pipelines as it had been before.

For the reloading of the "unrelated" pages you mention I think a screenshare is needed. I will try a small tweak now but I need to see what you see.

scherler:
@michaelneale I tested the following in the Augmenter.jsx

```
+ import { Immutable } from 'immutable';// NOTE: remove when you remove logger in setRun

    @action
    setRun(run) {
+       logger.debug('is run an immutable?', Immutable.Iterable.isIterable(run));
        this.run = run;
    }

```

The result is:

```
 io.jenkins.blueocean.dashboard.karaoke.Augmenter] is run an immutable? true
```

So the data we are using in the details as run is an Immutable. Trying to say it is not an MOBX object so I do not see any direct connection. However we do alter the state of the Augmenter when a new run is started, in Pipeline.jsx:

```
...
            if (nextProps.run !== this.props.run) {
                logger.debug('Need to set new Run. Happens when e.g. re-run.');
                nextProps.augmenter.setRun(nextProps.run);
            }
```

michaelneale:
I can see the logBar is gone - but all other problems remain. Still flashing, and logs did not show up the first time I ran a pipeline. I am seeing occasional reloads when opening  run. 

Probably best to screenshare. 

I can't follow the instructions for local storage as I doin't know what to do with the results or what to look for. I did clear it all out and start fresh though - but no change.

scherler:
@michaelneale please set the following logger settings
- io.jenkins.blueocean.dashboard.karaoke.Pager.Step	￼ DEBUG 
- io.jenkins.blueocean.dashboard.karaoke.RestApi	￼ DEBUG 
- io.jenkins.blueocean.dashboard.karaoke.Step	￼ DEBUG 

to track down the problem you see that the log is not loaded for the first step when starting the RUN from activity or branch tab.

When you first see log appearing (in the second node), pause karaoke by scrolling up and then export the output of the log console. Best to start with a clean console.

I do not show anymore the noSteps component when the pager is pending.

I reduced further the load to the server via debounce in the KaraokeApi.

BTW I did not merged master yet, since I will do it just before merge back.

michaelneale:
@scherler you are golden god!



So flashing gone, and the log thing isn't happening now (may be to do with that other change?). 

So I think this is ready for @cliffmeyers and @imeredith to have a closer look at code/try it out (and also getting this up to date etc) - and we can probably move this forward soon!

i386:
Looking great but I found some issues 🐛 
* Selecting text does not pause karaoke
selecting-text
* Opening/closing a step does not pause karaoke
collapsing
* Clicking on other nodes does not pause karaoke if the nodes are yet to be executed and will always take you back to the first node.
clicking-on-nodes

I love love love seeing all the flashes gone!!

michaelneale:
@i386 the last one above - I am not sure pausing by clicking on a future one makes sense... the other 2 are valid though.

i386:
@michaelneale @scherler OK, in that case we should make "not run" nodes not selectable.

scherler:
@i386 your points are more then valid however they are feature requests since if you see current master it does not behavior like you want currently.

In current Implementation (both this PR and karaokeV1) stopping karaoke is done by three things:

1. scroll up mouse
2. arrow key up
3. click on a finished node in karaoke

The following points are all new features and I would rather do them not in the PR

- Selecting text does not pause karaoke
- Opening/closing a step does not pause karaoke
- Not built nodes should not be selectable.

michaelneale:
@scherler since last change - https://ci.blueocean.io/job/ATH-Jenkinsfile/job/karaoke_reloaded/15/console is failing (well I think last change... may have been earlier). 

If this has been updated to master, then the ATH will also need to be kept up to date with its master branch.

i386:
@scherler fair point. Ive raised different issues for these:
* https://issues.jenkins-ci.org/browse/JENKINS-43316
* https://issues.jenkins-ci.org/browse/JENKINS-43317
* https://issues.jenkins-ci.org/browse/JENKINS-43318

scherler:
@michaelneale I will have a look now

scherler:
@i386 thank you very much

scherler:
@michaelneale Actually ATH caught a valid error :) I removed one import too much in my earlier commit which let to an error when clicking/focus on different nodes. Should be fixed now. I will hold on the merge of master until the code review has enough :bee: to be mergable.

cliffmeyers:
@michaelneale I am actively reviewing this now but I probably won't have good coherent feedback until sometime tomorrow. Big PR and I wasn't super familiar w/ the old karaoke code, so taking my time with this one.

kzantow:
I'm also reviewing and testing this, a lot of changes to grok

i386:
I've been playing around with this PR. So very good stuff @scherler 😎😎😎

michaelneale:
@cliffmeyers @kzantow ok - that's great if you can take this on. Lets get this in for 1.1

kzantow:
I'm seeing some issues with this. I'm using this simple pipeline:

```
pipeline {
  agent any
  stages {
    stage('Build') {
      steps {
        sh '''for e in asdf fwef oijewf lwejf lwekfj wlekfjwlekfj wef lwekfj welkf jwlefkj wlefj; do
echo $e
sleep 1
done
'''
      }
    }
    stage('Test') {
      steps {
        parallel(
          "Chrome": {
            sh '''for e in asdf fwef oijewf lwejf lwekfj wlekfjwlekfj wef lwekfj welkf jwlefkj wlefj; do
echo $e
sleep 1
done
'''
            
          },
          "Firefox": {
            sh '''for e in asdf fwef oijewf lwejf lwekfj wlekfjwlekfj wef lwekfj welkf jwlefkj wlefj; do
echo $e
sleep 3
done
'''
            
          },
          "IE": {
            retry(count: 5) {
              sh '''for e in asdf fwef oijewf lwejf lwekfj wlekfjwlekfj wef lwekfj welkf jwlefkj wlefj; do
echo $e
sleep 2
done
'''
            }
            
            
          }
        )
      }
    }
    stage('Deploy') {
      steps {
        sh '''for e in asdf fwef oijewf lwejf lwekfj wlekfjwlekfj wef lwekfj welkf jwlefkj wlefj; do
echo $e
sleep 1
done
'''
      }
    }
  }
}
```

* I am unable sometimes to click a particular node - e.g. while it's running, it wouldn't let me click IE sometimes, it just shows me the running Firefox log
* The logs are not always fetched fully. E.g. I got to the deploy step but it had 5 lines in the log, I had to do a full page refresh to get it to fetch all of the log

blueocean-core-js/src/js/config.js
```diff
@@ -15,6 +15,8 @@ export default {
             // Look up where the Blue Ocean app is hosted
             config.blueoceanAppURL = headElement.getAttribute('data-appurl');
 
+            // load server skewTime
+            config.serverBrowserTimeSkewMillis = headElement.getAttribute('data-servertime') - Date.now();
```
kzantow: I thought we were already tracking this somewhere?

cliffmeyers:
I noticed an error in the console where a fetch appears to be made with an empty URL. This seems to happen at the very end of the pipeline:

```
Uncaught TypeError: Cannot read property 'startsWith' of undefined
    at Object.fetch (blueocean.js:2700)
    at jenkins-js-extension.js:53264
    at invokeFunc (jenkins-js-extension.js:11247)
    at trailingEdge (jenkins-js-extension.js:11294)
    at timerExpired (jenkins-js-extension.js:11282)
```

It appears to be invoked from a debounce; perhaps when the pipeline ends there is a pending fetch that needs to be canceled?

You should be able to repro with this branch: https://github.com/cliffmeyers/jenkins-design-language/tree/docker-test

blueocean-dashboard/src/main/js/components/karaoke/components/Pipeline.jsx
```diff
@@ -0,0 +1,272 @@
...
+    }
+    componentWillMount() {
+        // starting pipeline service when we have an augmenter
+        if (this.props.augmenter) {
+            const { augmenter, params: { node } } = this.props;
+            this.pager = KaraokeService.pipelinePager(augmenter, { node });
+        }
+        // get sse listener to react on the different in sse events
+        this.listener.ssePipeline = sseConnection.subscribe('pipeline', this.sseEventHandler);
+        this.listener.sseJob = sseConnection.subscribe('job', this.sseEventHandler);
```
cliffmeyers: I think for added safety these subscriptions should be made in `componentDidMount`, or perhaps add a "cleanup" function that's called first, and will unsubscribe from any listeners that exist.

Per Abramov, in future versions of React `componentWillMount` could be called multiple times and this might leak SSE listeners.

see: http://stackoverflow.com/questions/41612200/in-react-js-should-i-make-my-initial-network-request-in-componentwillmount-or-co

blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
```diff
@@ -26,19 +25,20 @@ export class LogConsole extends Component {
     }
 
     componentWillMount() {
-        this._processLines(this.props.logArray);
+        // We need a shallow copy of the ObservableArray to "cast" it down to normal array
+        const lineArray = this.props.logArray !== undefined && !Array.isArray(this.props.logArray) ? this.props.logArray.slice() : this.props.logArray;
+        logger.debug('isArray props', Array.isArray(this.props.logArray), 'isArray after', Array.isArray(lineArray));
```
cliffmeyers: Would it be easier to just do this?

`this.props.logArray && this.props.logArray.slice();`

I would imagine that `slice()` is still quite fast even for larger (~10K) arrays?

blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
```diff
@@ -26,19 +25,20 @@ export class LogConsole extends Component {
...
     componentWillReceiveProps(nextProps) { // eslint-disable-line
-        const newArray = nextProps.logArray;
-        // we only want to update if we have an array and if it is new
-        if (!newArray || (newArray && newArray === this.props.logArray)) {
-            return null;
-        }
+        logger.debug('newProps isArray', Array.isArray(nextProps.logArray));
+        // We need a shallow copy of the ObservableArray to "cast" it down to normal array
+        const newArray = nextProps.logArray !== undefined && !Array.isArray(nextProps.logArray) ? nextProps.logArray.slice() : nextProps.logArray;
+        const oldArray = this.props.logArray !== undefined && !Array.isArray(this.props.logArray) ? this.props.logArray.slice() : this.props.logArray;
```
cliffmeyers: Same comment regarding `slice()`

kzantow:
Another 🐛  if I explicitly select a stage that hasn't started, when a new stage starts, it jumps to that one to follow along.

kzantow:
Also, scrolling up doesn't seem to cancel the following along

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
```diff
@@ -89,401 +33,93 @@ export class RunDetailsPipeline extends Component {
             document.addEventListener('keydown', this._handleKeys, false);
```
cliffmeyers: Per @kzantow 's comment about scroll up not canceling karaoke, I think the problem is that attaching the `wheel` event to the scrollArea ref is that it will not get events if the user's cursor is to the right of the log console. Maybe attach this one to `document` too?

kzantow: I just did something like this for the setup wizard, i think you just need to listen for scroll events on the scroll area and when following along maybe set a flag so those are ignored. This was thanks to @cliffmeyers 's awesome suggestion but seems to work well: https://github.com/jenkinsci/jenkins/pull/2825/files#diff-1b14e0dbd9ab104cdbcea2c1f823bcacR473

kzantow: ... this way you don't have to worry about key presses or scrolling with the wheel or scrollbar, etc.

blueocean-dashboard/src/main/js/components/karaoke/components/FreeStyle.jsx
```diff
@@ -0,0 +1,89 @@
...
+        this.pager = KaraokeService.generalLogPager(augmenter, location);
+    }
+    render() {
+        if (this.pager.pending) {
+            logger.debug('abort due to pager pending');
+            return null;
+        }
+        const { t, router, location, scrollToBottom, augmenter } = this.props;
+        const { data: logArray, hasMore } = this.pager.log;
+        logger.debug('props', scrollToBottom, this.pager.log.newStart, augmenter.generalLogUrl);
```
kzantow: might avoid logging things during render

blueocean-dashboard/src/main/js/components/karaoke/components/Pipeline.jsx
```diff
@@ -0,0 +1,272 @@
...
+        if (this.pager.pending && this.showPending) { // we are waiting for the backend information
+            logger.debug('EarlyOut - abort due to pager pending');
+            const pendingMessage = t('rundetail.pipeline.pending.message', { defaultValue: 'Waiting for backend to response' });
+            return <QueuedState message={pendingMessage} />;
+        }
+        // here we decide what to do next if somebody clicks on a flowNode
+        // Underlying tasks are fetching nodes information for the selected node
+        const afterClick = (id) => {
+            logger.debug('clicked on node with id:', id);
+            this.showPending = false; // Configure flag to not show pending anymore -> reduce flicker
```
kzantow: is there a reason this is set in a number of different places? is there just some condition that render could check - e.g. if there are nodes?

blueocean-dashboard/src/main/js/components/karaoke/components/Pipeline.jsx
```diff
@@ -0,0 +1,272 @@
...
+                        scrollToBottom,
+                        router,
+                        location,
+                    }}
+                />
+            }
+
+            { !this.pager.pending && !isPipelineQueued && noResultsToDisplay && <NoSteps
+                message={t('rundetail.pipeline.nosteps',
+                { defaultValue: 'There are no logsrrr' })}
```
kzantow: yarrr

michaelneale:
@kzantow yes James opened a few issues above for behavior changes, including selecting future nodes (I think the idea was to make them non selectable). 

The most serious thing seems to be logs not arriving - I think that needs to be dived into more. Is there any way to reliably see this? I have seen this too but not of late.

kzantow:
@michaelneale no, I've been trying to use different Jenkinsfiles and such but can't figure out a reliable way to reproduce it, I've only seen it happening randomly 2 or 3 times, but definitely seen it

blueocean-core-js/src/js/config.js
scherler: yes, but IMO we should move it to here. This way we can drop the `this.context.config` usage

blueocean-dashboard/src/main/js/components/karaoke/components/Pipeline.jsx
scherler: good catch will move that

blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
scherler: There had been a reason to not always do `slice` ...but do not remember. Will try and see if it starts ringing a bell.

blueocean-dashboard/src/main/js/components/karaoke/components/FreeStyle.jsx
scherler: why?

blueocean-dashboard/src/main/js/components/karaoke/components/Pipeline.jsx
scherler: Not sure what you mean.

blueocean-dashboard/src/main/js/components/karaoke/components/Pipeline.jsx
scherler: jeje yeahrrr

blueocean-dashboard/src/main/js/components/karaoke/components/Pipeline.jsx
kzantow: `showPending` is set `false` in multiple places, why not just figure out the right condition to check in the render method?

blueocean-dashboard/src/main/js/components/karaoke/components/FreeStyle.jsx
kzantow: render should be artifact-free and could be called repeatedly, even if it's not logging things to console due to the level, it is still creating objects that just get destroyed for no real reason. better to log things when state changes, I think, than just logging what is happening. that's what debugging is for.

blueocean/src/test/java/io/jenkins/blueocean/config/JenkinsJSExtensionsTest.java
```diff
@@ -55,7 +55,7 @@ public void test() {
             Assert.assertNotNull(extensionPoints);
 
             if ("blueocean-dashboard".equals(pluginId)) {
-                Assert.assertEquals(7, extensionPoints.size());
+                Assert.assertEquals(9, extensionPoints.size());
```
kzantow: what the heck kind of a test is this? better to just remove this line

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
```diff
@@ -115,6 +117,7 @@ public BluePipelineNodeContainer getNodes() {
     }
 
     @Override
+    @Navigable
```
kzantow: why remove this from the "interface" and add it to the impl?

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
scherler: I am not 100% sure but the behaviour is still the same as in current implementation in aspect of stopping.

I tried 

```
window.addEventListener('scroll', (event) => console.log(event), false);
```

but this did not kicked in.

I followed @cliffmeyers and attached it to `document` and dropped the `ref`

scherler:
@kzantow I am looking into the click problem you see, I believe it is because we allow future nodes to add a node parameter to the url

scherler:
@kzantow I have addressed your observation and now clicking in future nodes simple starts the karaoke again (waiting for the next sse event to kick in) it will *not* change the url anymore as it did before so the weirdness should be gone.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
scherler: That is to only add links for steps and nodes on things that can actually have it. Freestyle e.g. implements the interface but will never have steps and nodes. Hence moving it to the implementation that actually can provide it

blueocean/src/test/java/io/jenkins/blueocean/config/JenkinsJSExtensionsTest.java
scherler: That is in the age of times. There are even notes about it https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-dashboard/src/main/js/jenkins-js-extension.yaml#L3

blueocean-dashboard/src/main/js/components/karaoke/components/FreeStyle.jsx
scherler: Actually I only partly agree, since especially in support you may need to know those properties in `render`

kzantow: no, you need to know those properties _when they change_

scherler: yes but all observables are not covered by the method you mentioned!

blueocean-dashboard/src/main/js/components/karaoke/components/Pipeline.jsx
scherler: Since if pending is activated in general, we do not want to show pending when clicking on finished nodes. You can see pending when you set `runDetails.pipeline.showPending`  to `always` in the localStorage.

scherler:
@cliffmeyers I could not reproduce the error you reported. I added a check that should now output a warning in the KaraokeApi when href is null. This way you may be able to debug it.

kzantow:
@scherler @cliffmeyers yeah, I haven't actually seen that error, either and I've been looking for it. I rebuild like 20 times before I start testing, though :)

blueocean/src/test/java/io/jenkins/blueocean/config/JenkinsJSExtensionsTest.java
michaelneale: yeah - it is ancient as the rocks... I think it is fine to delete it if it helps. I probably add it so accept full responsibility.

*stabs self with sword*

scherler: https://issues.jenkins-ci.org/browse/JENKINS-43419 ;) will do it in a another PR to isolate that change.

michaelneale:
so the logs seem ok now, but some of the "pause/resume" stuff seems a little odd. 
1) a simple pipeline
```
pipeline {
    agent any
    stages {
        stage('first') {
            steps {
                sh 'ping -c 5 localhost'
            }
        }
        stage('second') {
            steps {
                sh 'ping -c 5 localhost'
            }
        }
        stage('final') {
            steps {
                sh 'ping -c 5 localhost'
            }
        }   
    }    
}
```
When I follow along and click on a completed stage, it doesn't show any further progress on the graph (odd) until it all finishes, then it all suddenly finishes all stages at once (despite them all taking the same time) - and then steals focus. this seems odd. What I was expecting: the stage graph would keep updating, but nothing to do with the steps I was looking at would change (ie not steal focus). 



2) Selecting parallel branches doesn't work (this is a regression from master, where you can do this, and is much needed, without it you can't really use parallel: 
Video: 


And the source for this simple pipeline: 
```
pipeline {
    agent any
    stages {
        stage('first') {
            steps {
                sh 'ping -c 4 localhost'
            }
        }
        stage('second') {
            steps {
                 parallel(
                  "Firefox": {
                    sh 'echo \'setting up selenium environment\''
                    sh 'ping -c 7 localhost'
                    
                  },
                 "Chrome": {
                    sh 'echo \'setting up selenium environment\''
                    sh 'ping -c 7 localhost'
                    
                 })

            }
        }
        stage('final') {
            steps {
                sh 'ping -c 4 localhost'
            }
        }

        
    }
    
}
```

michaelneale:
@scherler @imeredith any chance we could get this updates to master again? There is only one file actually in conflict (RunDetailsPipeline jsx) but not sure why - the diff seems confusing. There was only one small change to add in organisation recently

scherler:
@michaelneale will now have a look on your reported issue and then merge master into the pr

scherler:
@michaelneale  regarding the stopping karaoke mode, the current master would update the pipelineGraph but as you observed we do that neither anymore. If you want I am looking into how we can fetch the nodes again but ATM that is connected in fetching as well new steps. Regarding the stealing focus, that is the  `runDetails.pipeline.updateOnFinish` properties which will update the run on finish. We can set that to false, but that needs a "hard-refresh" of the page.

michaelneale:
@scherler thanks for getting this up to date. 

So I think the 2 remaining issues are the same as mentioned before: https://github.com/jenkinsci/blueocean-plugin/pull/839#issuecomment-293173473

ie - 

1) if the graph could keep updating even while paused - that wouldn't look "broken". 
2) with running parallel branches, I should be able to click on each one (that should be the same as master now)

But other than that - I think that might be enough. I updated the ATH branch (I hope you don't mind) with the master of the ATH - and it passes now (ran out of disk space today too! I had to tweak the job to log rotate!). 

As for `runDetails.pipeline.updateOnFinish` - not sure what you mean by that? I don't think the job updating to finished is so bad (just a bit annoying it collapses away, but no worse than now)

So really: as above - the parallel thing and updating the graph and we are good to go!

i386:
@scherler on @michaelneale's urging I went ahead and updated this branch against master because I merged #968 and this branch conflicted quite a bit.

Few things:
* I rebased all the changes from master to `karaoke_reloaded`
* Republished corejs and took it for a spin - works great!
* I dropped some commits that were not worth saving where you were just updating `package.json`
* I pushed a new branch that contained your previous branch state to `karaoke_reloaded_backup` incase you wanted to get to them again.
* Updated the empty states to use `Alerts` instead of `EmptyState` as this is what @brody has told us to use. Was on #968 but your changes needed to be updated to reflect it.
* I did notice a 🐛  where the `causeOfBlockage` message was not being updated. Ill fix that after this PR lands on maser (its my fault). You have to reload to get that message which is a bit odd.

michaelneale:
yikes, this doesn't work at all now - we need to reset the state of this, and try again, or ??

i386:
@michaelneale this has been reset. Refresh the PR to see the original history back.

i386:
@scherler so some additional screwing around and it looks like @michaelneale was able to get the tests running against this PR with updated master. new commits have been pushed to the branch 🎉

michaelneale:
ok, having fixed this up - it is now up to date with master, including the queue changes (and I checked ATH is happy for karaoke)... so this should be ready for the final "regression fixes" of

1) making sure graph updates even when karaoke paused
2) that you can click on parallel while they are running 

as per my comment: https://github.com/jenkinsci/blueocean-plugin/pull/839#issuecomment-293173473

But the merge pain has been done with master, given a bit of messing around

michaelneale:
OK, looked at master again, I don't think #2 above is new anyway for that sample pipeline. 

Given that - if we can make the graph update even when paused - I think we should just merge this!

scherler:
@michaelneale will look into the issues and take it for a spin afer @i386 went ahead and did the merge dance (thanks to you both!)

scherler:
BTW awesome backup strategy @i386 git branches are so great. :)

i386:
@scherler Git is the shizzle :)

How did you go tracking those issues down? So keen to see this PR drop :)

michaelneale:
OK passes ATH now, and stage graph updates nicely. 

@i386 the only thing that may cause pause on this - is the regression (against master) that doesn't let you click on executing parallel nodes. Should we merge this in and then follow that up as a REGRESSION?

i386:
@scherler so @michaelneale did some testing and we did find one regression with clicking parallels. We decided to go ahead and merge this as the ATH was green to save having to keep this branch updated. We will find someone to pick up that ticket - YOU'RE FREE!! (unless you want to fix it 😉 )

Really nice job mate. Thank you 👍 👍 👍

michaelneale:
yes - awesome.

scherler:
:) nice! Thanks everyone involved

scherler:
@michaelneale I think the parallel is easy I just figured https://github.com/jenkinsci/blueocean-plugin/blob/41936a8b5a1a33c8ea5ab69483f946397b5bc130/blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx#L366 We did test for parallel and allow click on them regardless of they state.

Should I prepare a preview branch of that?

i386:
@scherler yes please :)

scherler:
@i386 found https://issues.jenkins-ci.org/browse/JENKINS-43747 will try to fix it now

PR #840, opened 2017-02-17
why not include editor, the proper way

michaelneale:
We can include the editor this way, worked for me locally. See how this works. 

cc @i386 @kzantow I think this is how it could happen - right? 

https://issues.jenkins-ci.org/browse/JENKINS-42134

### DIFF
--- a/blueocean/pom.xml
+++ b/blueocean/pom.xml
@@ -81,6 +81,12 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>blueocean-autofavorite</artifactId>
         </dependency>
+        
+        <dependency>
+          <groupId>io.jenkins.blueocean</groupId>
+          <artifactId>blueocean-pipeline-editor</artifactId>        
+        </dependency>
+
 
         <!-- Test deps -->
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -415,6 +415,12 @@
             <artifactId>favorite</artifactId>
             <version>2.0.4</version>
         </dependency>
+        <dependency>
+          <groupId>io.jenkins.blueocean</groupId>
+          <artifactId>blueocean-pipeline-editor</artifactId>        
+          <version>0.1-preview-4</version>
+        </dependency>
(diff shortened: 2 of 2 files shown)


i386:
I believe this is the way it's done :)

michaelneale:
good to know we broke the circular dependency

tfennelly:
@michaelneale I'd have expected that the user would need to be logged in before they could get to the pipeline editor ...

screenshot 2017-02-17 10 32 30

And same for the "New Pipeline" button.

tfennelly:
I was also saying to @michaelneale that I (as a user) find it a bit odd that atm that when I'm in a pipeline and click on the editor it opens something completely different i.e. doesn't open that pipeline for editing. That's fine, but is not what I'd expect as a user, so feels a bit "weird" because I end up not knowing what I'm looking at.

I think it would make sense to somehow indicate to the user that this feature is evolving atm and will eventually work in the way they're expecting.

michaelneale:
@tfennelly right yes ultimately should only be when logged in. For now it is a bit of a read-only UI for editor, you have to copy paste things in (which isn't clear, yes) but that is only for the next week or so, then it will open an editor in context of a branch etc. 

Don't need to merge this now, I just watned to check things were happy with including the editor and we didn't have some unexpected dependency cycle (we don't!)

michaelneale:
@tfennelly oh - the new pipeline button should not show up - that is probably a regression

i386:
@tfennelly good eye! I believe @cliffmeyers raised this as [JENKINS-41434](https://issues.jenkins-ci.org/browse/JENKINS-41434) but I don't see this as a blocker for beta 24.

i386:
@tfennelly when we have it integrated into creation and allow you to edit from a branch, we should move that link into a different spot.

i386:
@michaelneale this missed the b24 window :( Can we merge this one now?

michaelneale:
I would rather have first cut of round tripping in, otherwise people will be confused by the link as @tfennelly was

PR #841, opened 2017-02-18
JENKINS-42163# SCM save API for github

vivek:
# Description

See [JENKINS-42163](https://issues.jenkins-ci.org/browse/JENKINS-42163).

### Get SCM file content of a pipeline (Multibranch or OrganizationFolder)

```
curl -v -u xxx:yyy "http://127.0.0.1:8080/jenkins/blue/rest/organizations/jenkins/pipelines/hk/branches/master/scm/content/?path=Jenkinsfile"

{
   "content" : {
      "name" : "Jenkinsfile",
      "sha" : "f13b26341cf403aa7d697bc252908de092ee279d",
      "_class" : "io.jenkins.blueocean.blueocean_github_pipeline.GithubContent",
      "repo" : "test-no-jenkins-file",
      "size" : 7,
      "owner" : "vivek",
      "path" : "Jenkinsfile",
   },
   "_class" : "io.jenkins.blueocean.blueocean_github_pipeline.GithubFile"
}
```

#### Save file to an OrganizationFolder

SCM owner and credentials are computed from the OrganizationFolder. Request must include **repo** element.

```
curl -H 'Content-Type: application/json' -u user:password -XPUT http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/vivek/scm/content/ -d 

'{
  "content" : {
    "message" : "first commit",
    "path" : "Jenkinsfile",
    "branch" : "test1",
    "repo" : "test-no-jenkins-file",
    "sha" : "9e82c4011cd70446a2f44881a6c288c59b4abac0",
    "base64Data" : "VGVzdDEyMw=="
  }
}'
```

#### Save file to a MultiBranchProject

SCM owner, repo and credentials are computed from the MultiBranchProject.

```
curl -H 'Content-Type: application/json' -u user:password -XPUT http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/vivek/pipelines/test-no-jenkins-file/scm/content/ -d 

'{
  "content" : {
    "message" : "first commit",
    "path" : "Jenkinsfile",
    "branch" : "test1",
    "sha" : "9e82c4011cd70446a2f44881a6c288c59b4abac0",
    "base64Data" : "VGVzdDEyMw=="
  }
}'
```

- If file **path** doesn't exist in SCM then a new file will be created
- If **sha** is provided and file **path** exists then it must match with the sha of existing file, else 400 (Bad Request) error will be returned. 
- If **sha** matches then the file will be updated with the content provided in the request
- If **branch** element is not provided file will be saved on default branch (typically maser)
- If **branch** element is present and this branch doesn't exist then a new branch will be created off default branch HEAD

Response

```
{
   "_class" : "io.jenkins.blueocean.blueocean_github_pipeline.GithubFile",
   "content" : {
      "_class" : "io.jenkins.blueocean.blueocean_github_pipeline.GithubContent",
      "sha" : "f13b26341cf403aa7d697bc252908de092ee279d",
      "path" : "Jenkinsfile4",
      "owner" : "vivek",
      "repo" : "test-no-jenkins-file",
      "name" : "Jenkinsfile4"
   }
}
```

- [x] Auto-create branch if it doesn't exist in Github
- [x] API to load file from GitHub

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

cc: @michaelneale @kzantow 

@reviewbybees


i386:
@vivek certainly based on the design we might want to create or update a file in any new or existing branch.

vivek:
@i386 in this implementation existing branch is supported. If content.branch element is provided file will be created or updated on that branch.

i386:
@vivek ahh sorry I misread

michaelneale:
Looking good so far, I guess some questions: 

* Can we expand this to save AND create a new branch optionally (so we know we can create a branch)? 
* When you are updating a file, how do you specify what branch it is updating in? (and is the sha for the content or the version that you are updating?). 

The UI will need to report if a save failed due to a concurrent edit by someone else, so we will need to tell this error apart from others.

vivek:
>Can we expand this to save AND create a new branch optionally (so we know we can create a branch)?

I am not sure if we want to include as part of this API. Creating new branch needs to be explicit action on user part, guess UI is already taking care of that? There could be error conditions and clubbing it with Save API might be confusing. 

>When you are updating a file, how do you specify what branch it is updating in? (and is the sha for the content or the version that you are updating?).
A load Jenkinsfile probably could be included as part of this? or is this exposed elsewhere that I haven't seen?

Right, we need API to read it, that should have been first:) It will be separate API than save.

kzantow:
I think this should be a single API that we pass a branch, source branch, file name, and content; then it:
a) creates a branch if it doesn't exist
b) creates the file if it doesn't exist, or update it if it does

The UI shouldn't be keeping track of the status of what should be an atomic operation from the BO point of view.

stephenc:
@jglick well assuming we can come up with an abstraction that is generally meaningful... right now that seems far off

i386:
@stephenc if we ended up with one abstraction for all the Git services and native Git, that would be a win.

i386:
@jglick thats right. We want to get this working so we can ship sooner and inform any future API requirements for a possible SCM API.

michaelneale:
@vivek @kzantow would it make more sense for this code to be in the editor plugin itself? Is this a) possible and b) a good idea? 

Given this api exists for that plugin, does it belong there logically and to ease development?

kzantow:
Doesn't really ease development, there need to be coordinated changes with BO core, so I'll be working on it all together (hooray for hpi:hpl!)...

michaelneale:
@kzantow ok with me as long as you are happy. Hopefully from this the shape of an extension point will become clear, making it less work to take on bitbucket and others. 

So Vivek has a plan to allow this to create branches and also to load the Jenkinsfile - changes will be added to this branch (as discussed earlier today)

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/HttpRequest.java
```diff
@@ -0,0 +1,135 @@
...
+import java.lang.reflect.Array;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.Collections;
+import java.util.zip.GZIPInputStream;
+
+/**
+ * @author Vivek Pandey
+ */
+class HttpRequest {
```
kzantow: Would rename this to `GitHubHttpRequest` or something

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/scm/Scm.java
```diff
@@ -73,4 +79,34 @@
...
+    @TreeResponse
+    public ScmFile saveFile(StaplerRequest staplerRequest) throws IOException {
+        JSONObject body = JSONObject.fromObject(IOUtils.toString(staplerRequest.getReader()));
+        String cls = (String) body.get("$class");
+        if(cls == null){
+            throw new ServiceException.BadRequestExpception(new ErrorMessage(400, "Failed to save file")
+                    .add(new ErrorMessage.Error("$class", ErrorMessage.Error.ErrorCodes.MISSING.toString(),
+                            "$class is required")));
+        }
+        ScmSaveFileRequest request = staplerRequest.bindJSON(ScmSaveFileRequest.class, body);
```
kzantow: We really should have an automatic way of doing this, no? Thinking of other users implementing APIs, there are much easier ways available with automatic object bindings, don't they exist in stapler?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/scm/ScmFile.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.kohsuke.stapler.export.ExportedBean;
+
+/**
+ * Content stored in SCM
+ *
+ * @author Vivek Pandey
+ */
+@ExportedBean(defaultVisibility = 999999)
+public abstract class ScmFile{
```
kzantow: Why not `interface` ?

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/HttpRequest.java
vivek: Why?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/scm/Scm.java
vivek: This is closest you got in Stapler. We do not expect users to deal with it, its there in abstract class for a reason:)

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/scm/ScmFile.java
vivek: Prefer abstract class over interface unless it describes behavior, gives leverage in case of binary incompatible change later on.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/HttpRequest.java
kzantow: Since HttpRequest is an extremely common class name and this is specific to github.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/scm/Scm.java
kzantow: Seems like this should be dirt simple to provide a couple annotations to just automatically do the bindings in and out, without needing `@Exported...` on things, maybe it's just a preference, but I prefer service methods to reference the objects they use with clear signatures. It pains me when I have to explain to people how to write REST services currently, but that's a topic for another day, nothing to do with this PR.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/HttpRequest.java
vivek: There isn't really anything specific to GitHub in it though:) Anyways its package private, so is fine IMO.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/scm/Scm.java
vivek: @kzantow Can't agree more:) However specific to this API, doc needs work, for people wanting to extend this API, the thing they need to look for is `ScmSaveFileRequest`.

vivek:
@kzantow @michaelneale branch auto-create and load file off GitHub is ready for review, PTAL.

kzantow:
Nice! I'm working off your branch, will give it a spin with the editor and give it a look over.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/HttpRequest.java
```diff
@@ -0,0 +1,141 @@
...
+        return this;
+    }
+
+    public <T> T to(Class<T> type) throws IOException {
+        HttpURLConnection connection = connect();
+        if (methodNeedsBody()) {
+            connection.setDoOutput(true);
+            connection.setRequestProperty("Content-type", contentType);
+            if (body == null) {
+                GithubScm.om.writeValue(connection.getOutputStream(), Collections.emptyMap());
```
kzantow: ... but there is something specific to Github in it. Just a NIT, though.

vivek: Ha ha yeah references this ObjectMapper from that class:) Good catch, that ObjectMapper should have belonged to HttpRequest. There is another PR I am working on where I am touching that class as that needs cleanup anyways.

michaelneale:
@vivek keith was wondering if the save url could also skip the github specific bit - as it seems redundant...

michaelneale:
@kzantow viveks' comment is that the URIs are different as in one case - it is loaded from the pipeline/SCM - in the other - to the specific provider (github). The load should in thoery work with any SCM provider, but save is specific to github (and needs specific github fields).

kzantow:
@michaelneale well, there's no telling what additional metadata or behavior is needed for other SCMs to save, necessarily... but we've already configured the repository at the org folder level, so it would seem scm save would be agnostic to the back-end, with any additional metadata along for the ride as needed for the particular scm implementation.

kzantow:
also, just pointing out that we'll need loading of the jenkinsfile from a run. probably doesn't have to be part of this PR at all, but doing replay and such, it's actually loaded from somewhere other than the scm.

vivek:
> there's no telling what additional metadata or behavior is needed for other SCMs to save, necessarily... but we've already configured the repository at the org folder level, so it would seem scm save would be agnostic to the back-end, with any additional metadata along for the ride as needed for the particular scm implementation.

Right, the metadata needed for a specific SCM is going to be specified by that SCM. 

> just pointing out that we'll need loading of the jenkinsfile from a run. probably doesn't have to be part of this PR at all, but doing replay and such, it's actually loaded from somewhere other than the scm.

SCM configuration are attached to pipeline/org folder/MBP build configuration and not Run so not sure what you mean by 'loading of the jenkinsfile from a run'? Or maybe I need to see mockup of replay case. Anyways if its targeted for another PR then I won't drag this discussion in this, am just curios:)

michaelneale:
@kzantow opening from a run for edit means loading it from scm. When for replay - not as clear, as it may be loading it from a specific commit (but we won't have replay for a little while anyway). 

@vivek @kzantow I suppose if there was some api that "saved to the pipeline" it could calculate what scm to delegate to and then the front end doesn't have to care if it is github etc (although it needs to know to pass a sha etc etc - some things might not call them branches etc)

vivek:
> I suppose if there was some api that "saved to the pipeline" it could calculate what scm to delegate to and then the front end doesn't have to care if it is github etc (although it needs to know to pass a sha etc etc - some things might not call them branches etc)

At present the API is in context of scm. There is reason for that, this is to take care of case where there is no MBP pipeline as given GitHub repo doesn't have Jenkins file. In this case, there was never indexing and SCM API (SCMFileSystem) doesn't exist. So the only way we can handle it is that: we know the repo name where user wants to create a Jenkinsfile, it gets saved up directly over there to SCM. for this to work, UI needs to provide SCM in context, GitHub/bitbucket etc. (this comes straight off creation flow)

kzantow:
Well, there's already an org folder created by the time we get to the save action. It's already been indexed and determined there is no Jenkinsfile, so no MBP has been created. It already has the Github connection information, just not the repository & branch to use.

michaelneale:
@kzantow right - but in that case it isn't saving *to* a pipeline+branch. as there is no existant pipeline.

michaelneale:
@kzantow in the case of "save to a pipeline that jenkins doesnt yet know about" - it may be possible to translate between a generic type URL and the github specific one if you really think it is important ( @vivek may disagree...) - as like you say, all the information is there...

vivek:
@michaelneale @kzantow sure, we can save/create a file in the context of a org folder as well,  but client still needs to know what scam its talking about as it still needs to send scm specific payload. If this makes things simpler it's easy to do, let me know. I think better I chat with @kzantow to understand what's needed and resolve it.

michaelneale:
@vivek well it would know it is github from either creation flow (empty repo case) or they clicked on a branch to edit - in which case there was a load event to tell it what type... so its reasonable. Up to @kzantow if this helps things out or not.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScmContentProvider.java
```diff
@@ -0,0 +1,88 @@
...
+import org.apache.commons.codec.digest.DigestUtils;
+import org.jenkinsci.plugins.github_branch_source.GitHubSCMSource;
+
+import javax.annotation.Nonnull;
+import java.io.IOException;
+
+/**
+ * @author Vivek Pandey
+ */
+@Extension(ordinal = -100)
```
kzantow: Does ordinal matter here?

i386:
Wow thats great @kzantow 💃  🎆

kzantow:
So, let me just try to rephrase my point a bit. This is more food for thought, as in the interest of time, I don't think there are any changes we necessarily need.

However, from the perspective of a consumer, I think I should either be a) interacting with the SCM specifically or b) interacting with Jenkins specifically. As it stands, I'm interacting with Jenkins in one case (fetching Jenkinsfile from an existing MBP branch) and interacting with SCM specifically in the other case (saving).

Interacting with Jenkins is more limiting and might not always serve the right purpose, not that we support this exactly, but say for argument's sake I had a custom `git checkout` step and a custom Jenkinsfile that was called downstream from the main MBP. I wouldn't have a way to edit this without adding another MBP just so I could get a Jenkins endpoint. Or, if in the future we wanted to support generating some other stuff for whatever reason, again no way to really do it without having a MBP configured a specific way to read things.

If, instead, we could just get a reference to the SCM provider (sorry poor terminology) and most appropriate credential from somewhere, e.g. a MBP or an Org Folder, if configured, it would be more flexible for read and save actions as more implementations are added. For Org folders this is somewhat important, as I could specifically configure it to use a particular GHE endpoint, so we'd need to have the correct reference to that configured SCM provider.

Anyway, hope that makes sense. For now I'm all good with merging this and moving on, though!

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScmContentProvider.java
michaelneale:
```
10 REM It's like lines of code in BASIC. Gives it some space for future. 
20 GOTO 10
```

vivek: Haha, Yes it is to allow others to provide their own version of GitHub content provider.

michaelneale:
@kzantow that's great to hear that it worked "round trip". 

>As it stands, I'm interacting with Jenkins in one case (fetching Jenkinsfile from an existing MBP branch) and interacting with SCM specifically in the other case (saving).

Yes, this is actually how this works right now, by design. It isn't ideal, or the end game, but more a short term measure for specific github support. So yeah, I think medium term it needs to be interacting with Jenkins.

>but say for argument's sake I had a custom git checkout step and a custom Jenkinsfile that was called downstream from the main MBP. I wouldn't have a way to edit this without adding another MBP just so I could get a Jenkins endpoint

Yeah not sure of a good way to support that - but jenkins doens't know about the different scms involved until there is a fingerprint (ie run once at least). 

So do you think the proper way is to talk to a "reference" to scm provider for loading and saving in all cases - or to Jenkins? 

>For Org folders this is somewhat important, as I could specifically configure it to use a particular GHE endpoint, so we'd need to have the correct reference to that configured SCM provider.

Oh good point - how does it work now when save? is it only github.com specific the way ti is in the api or does it go via the org folder and it knows the api to talk to?

vivek:
Chatted with @kzantow. We agreed, its better to move save in the context of folder/mbp. so that load and save will be at same place, will be consistent. I am going to make this change in this PR.

michaelneale:
ok SGTM - as long as you both are happy

vivek:
@kzantow done with my changes, docs updated with changes in URL and request body of save API. PTAL.

blueocean-rest/README.md
```diff
@@ -2195,7 +2195,7 @@ curl -v -u xxx:yyy http://localhost:8080/jenkins/blue/rest/organizations/jenkins
 ### Get SCM file content of a pipeline (Multibranch or OrganizationFolder)
 
 ```
-curl -v -u xxx:yyy "http://127.0.0.1:8080/jenkins/blue/rest/organizations/jenkins/pipelines/hk/branches/master/scmcontent/?path=Jenkinsfile"
+curl -v -u xxx:yyy "http://127.0.0.1:8080/jenkins/blue/rest/organizations/jenkins/pipelines/hk/branches/master/scm/content/?path=Jenkinsfile"
```
michaelneale: ah ignore me - just saw it

michaelneale:
there is some annoying ATH failure which needs investigation... has failed a few times there.. taking a closer look

michaelneale:
Ah, I think this is a bit out of date with master and there was some changes that went in...

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScmContent.java
```diff
@@ -0,0 +1,84 @@
...
+    private final String sha;
+    private final boolean autoCreateBranch;
+
+    @DataBoundConstructor
+    public GithubScmContent(String owner, String repo, String path, String message, String base64Data, String sha, String branch, Boolean autoCreateBranch) {
+        this.owner = owner;
+        this.repo = repo;
+        this.path = path;
+        this.message = message;
+        this.base64Data = base64Data;
```
kzantow: This should be consistent with the read operation, either `encodedContent` or they should both be this.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScmContent.java
```diff
@@ -0,0 +1,84 @@
...
+import io.jenkins.blueocean.commons.ErrorMessage;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * @author Vivek Pandey
+ */
+public class GithubScmContent {
```
kzantow: Seems like this could be generic, too... similarly with `payload` instead of `sha`

kzantow:
Also, I merged with master, so make sure to pull, sorry if that causes any inconvenience

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScmFileContent.java
```diff
@@ -0,0 +1,139 @@
+package io.jenkins.blueocean.blueocean_github_pipeline;
+
+import org.kohsuke.stapler.export.Exported;
+import org.kohsuke.stapler.export.ExportedBean;
+
+/**
+ * @author Vivek Pandey
+ */
+@ExportedBean
+public class GithubScmFileContent{
```
vivek: Save operation is scm dependent. Maybe I can wrap the scm response in other element such as:

```
{
    "content":{
       "sha":"xxxx",
       ....
    }
}
```

kzantow: Right, I understand that but all the other stuff is going to be pretty much needed. It doesn't even need to be wrapped up. This would make implementing other load/save with other providers easier and more consistent, and allow the UI to just delegate to the back end what to do without having any provider specific stuff in it (e.g. gitlab doesn't have/use the sha)

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScmContent.java
vivek: Its' echoing GitHub save request and response and did not want to get in to business of transforming messages talking to GitHub apis. Let me think over it.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScmContent.java
vivek: ok.

vivek:
@kzantow no problem, I pulled this branch so all good. thanks:)

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScmFileContent.java
vivek: Ah, thats already taken care at `ScmContentProvider.saveContent()` level. For every provider, like gitlab, you have gitlab specific saveContent() implemented and it knows what to expect from  read and respond with gitlab specific save content response. Tomorrow if we have a totally different scm whose request/response looks very different from GitHub that will work too. I am assuming UI bit is implemented in purely extensible way as well to allow different implementations to load/save in fairly pluggable way.

michaelneale:
:bee: go for it - nice one!

Get this one in!

vivek:
@michaelneale request and response looks consistent, some minor adjustment. But this means @kzantow needs to change his code to process response. Will Wait for Keith's bee before merging:)

vivek:
>If a MBP is not created, but a Jenkinsfile exists I cannot save due to 'no sha' since it wasn't able to load from the repo
... this seems to be a compelling reason to take the SCM approach vs. the Jenkins approach we discussed

Hmm, thats right no guarantee indexing has been done/failed etc. we can't load using scm plugin API, need to talk to GitHub directly. That should fix this issue.

> There appears to be no indexing/rebuilding taking place after the save operations
... probably best to add an extension point for SCM save listeners

Right, because the listener in this case is UI, it can always call API to trigger it after receiving 200 from save.

kzantow:
I thought it was discussed that indexing should be triggered on the server side? Pushing behavior like this to the client is not at all ideal.

vivek:
Discussed yes but no agreement on this. These are generic SCM content load save API. Putting such behavior that otherwise exists in other REST API is purely redundant and makes it convoluted. For example you have no way to know whether indexing was successful or failed or get other information related to it.

kzantow:
If this was a SCM-based API (e.g. `.../rest/organization/:organization/scm`), I'd buy indexing should be done elsewhere, but this is in context of an organization folder or multibranch project, which are based on indexing happening and Jenkinsfile changes, calls to these methods in this context should definitely trigger re-indexing or re-builds when the `Jenkinsfile` is updated.

vivek:
If it were Jenkins load/save API then yes. but its not, these are generic file load/save API to SCM configured in a jenkins job.

kzantow:
Not really, this is "save to this org folder or multibranch project". This is the point I had before about interacting with Jenkins vs. interacting with the SCM. The path we've taken is "interacting with Jenkins", which can be fine, but should also "do Jenkins stuff" as appropriate.

kzantow:
... It's like pushing changes to a branch or PR. People want Jenkins to re-run the build automatically, if they have it set up in Jenkins. This is just a more extreme example of it, something they've already tried to set up in Jenkins.

vivek:
Well, here are some of concerns for adding indexing in save API: 

1. No easy way for client to distinguish indexing error vs save error. Cases where save succeeds, indexing fails.
2. In case of successful new jenkinsfile save and failed indexing, sha may not get back to client and only way for client to check is call load API, get sha and then call  API to trigger indexing. At this point it has sha so it doesn't want to save API again to trigger (expensive) indexing, it can call api to trigger indexing directly.
3. Some clients might not like to trigger indexing

For 1, and 3 something can be done but for 2 or who knows some other cases it might make indexing as part of save redundant.

vivek:
@kzantow load api implemented with GitHub api to support use cases such as loading file in context of org folder or in context of unindexed MBP. Updated [docs](https://github.com/jenkinsci/blueocean-plugin/blob/task/JENKINS-42163/blueocean-rest/README.md#get-github-file-content-of-a-pipeline-multibranch-or-organizationfolder).

kzantow:
Sorry, should have swapped approval. I *still* don't really agree with having the UI trigger a re-index, but I don't care much, since I got it working...

kzantow:
MERGE THIS!! :)

michaelneale:
So perhaps a future enhancement is to have an option to trigger a reindex on save? but not default to it? 

Also on @kzantow's other point - that sounds like an unpleasant case where there is no pipeline but is a jenkinsfile

kzantow:
@michaelneale right, I feel like using this API should attempt to "keep Jenkins in sync" as best it can. In the case we're saving a Jenkinsfile, it would have indexed and created a job and tried building. If we have web hooks set up, it would index and build. I can see a case you just want to save something and not build, though, but probably not when saving a Jenkinsfile. Anyway, I am happy to have this merged, I have the basics working with it already, including indexing and such. We can always revisit this later.

PR #842, opened 2017-02-20
[FIX JENKINS-41425] Fix UI issues where Job name is not encoded (has slashes etc)

tfennelly:
# Description

See [JENKINS-41425](https://issues.jenkins-ci.org/browse/JENKINS-41425).

The pages are rendering now with these changes. Needs some tests etc.

I think @vivek needs to have a close look at this since he obviously has the bigger picture here wrt the REST API etc. Maybe he'll want to take this altogether.

To test, I used the script console to programmatically create a Job with slashes in the name e.g.

```groovy
def jenkins = Jenkins.getInstance();

jenkins.createProject(org.jenkinsci.plugins.workflow.job.WorkflowJob, "a/job/with/slashes");
```

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/BlueUrlTokenizer.java
+++ b/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/BlueUrlTokenizer.java
@@ -33,6 +33,7 @@ import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
+import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
@@ -269,6 +270,75 @@ public class BlueUrlTokenizer {
             return getPart(this.lastPart).equals(value);
         }
         return false;
+    }
+
+    public static String reconstructJobNamePath(String name, String fullName) {
+        return reconstructJobNamePath(name, fullName, "/");
+    }
+
+    public static String reconstructJobNamePath(String name, String fullName, String separator){
+        StringBuilder pipelinePath = new StringBuilder();
+        String fullPath;
+        List<String> pathTokens;
+
+        //
+        // The Job "fullName" is typically suffixed with the "name".
+        // The name is NOT part of the path to the job, so lets remove
+        // it from the fullName before we parse and reconstruct the path.
+        // This, along with reencoding of the name, helps us to eliminate
+        // the problems where a Job is programatically created with a name
+        // that's not URL encoded and contains characters that should be
+        // encoded e.g. slashes (see JENKINS-41425).
+        //
+        if (fullName.endsWith(name)) {
+            fullPath = fullName.substring(0, fullName.length() - name.length());
+            pathTokens = new ArrayList<>();
+            pathTokens.addAll(Arrays.asList(fullPath.split("/")));
+            pathTokens.add(urlReencode(name));
+        } else {
+            fullPath = fullName;
+            pathTokens = Arrays.asList(fullPath.split("/"));
+        }
+
+        for (String n : pathTokens) {
+            if (n.length() == 0) {
+                // Ignore empty strings. Can be caused by
+                // leading/trailing slashes.
+                continue;
+            }
+            // Don't add the separator to the start of the path.
+            if(pipelinePath.length() > 0){
+                pipelinePath.append(separator);
+            }
+            pipelinePath.append(n);
+        }
+
+        return pipelinePath.toString();
+    }
+
+    /**
+     * Re-encode the supplied string.
+     * <p>
+     * Use this on strings that may already be encoded, but are not guaranteed to
+     * be e.g. Job names (see JENKINS-41425).
+     * @param string The string to be re-encoded.
+     * @return The re-encoded string.
+     */
+    private static String urlReencode(String string) {
+        try {
+            // Decode the string in case it's already encoded, or partially
+            // encoded. This will be a no-op if the string is not encoded.
+            string = URLDecoder.decode(string, "UTF-8");
+            // And encode ...
+            string = URLEncoder.encode(string, "UTF-8");
+            // The Java URLEncoder encodes spaces as "+", while the javascript
+            // encodeURIComponent function encodes them as "%20". We need to make them
+            // consistent with how it's done in encodeURIComponent, so replace the
+            // "+" with "%20".
+            return string.replace("+", "%20");
+        } catch (UnsupportedEncodingException e) {
+            throw new IllegalStateException("Unexpected exception re-encoding the pipeline name.", e);
+        }
     }
 
     private static String urlDecode(String string) {
--- a/blueocean-dashboard/src/main/java/io/jenkins/blueocean/BlueOceanWebURLBuilder.java
+++ b/blueocean-dashboard/src/main/java/io/jenkins/blueocean/BlueOceanWebURLBuilder.java
@@ -124,7 +124,7 @@ public class BlueOceanWebURLBuilder {
         } else if (classicModelObject instanceof Item) {
             Resource blueResource = BluePipelineFactory.resolve((Item) classicModelObject);
             if (blueResource instanceof BlueMultiBranchPipeline) {
-                return getOrgPrefix() + "/" + encodeURIComponent(((BluePipeline) blueResource).getFullName()) + "/branches";
+                return getOrgPrefix() + "/" + encodeURIComponent(decodeURIComponent(((BluePipeline) blueResource).getFullName())) + "/branches";
             }
         }
 
@@ -149,13 +149,13 @@ public class BlueOceanWebURLBuilder {
             return new BlueOceanModelMapping(
                 multibranchJob,
                 multibranchJobResource,
-                getOrgPrefix() + "/" + encodeURIComponent(multibranchJobResource.getFullName())
+                getOrgPrefix() + "/" + encodeURIComponent(decodeURIComponent(multibranchJobResource.getFullName()))
             );
         } else {
             return new BlueOceanModelMapping(
                 job,
                 blueResource,
-                getOrgPrefix() + "/" + encodeURIComponent(blueResource.getFullName())
+                getOrgPrefix() + "/" + encodeURIComponent(decodeURIComponent(blueResource.getFullName()))
             );
         }
     }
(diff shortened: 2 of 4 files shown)


blueocean-dashboard/src/main/java/io/jenkins/blueocean/BlueOceanWebURLBuilder.java
```diff
@@ -124,7 +124,7 @@ private BlueOceanWebURLBuilder() {
         } else if (classicModelObject instanceof Item) {
             Resource blueResource = BluePipelineFactory.resolve((Item) classicModelObject);
             if (blueResource instanceof BlueMultiBranchPipeline) {
-                return getOrgPrefix() + "/" + encodeURIComponent(((BluePipeline) blueResource).getFullName()) + "/branches";
+                return getOrgPrefix() + "/" + encodeURIComponent(decodeURIComponent(((BluePipeline) blueResource).getFullName())) + "/branches";
```
tfennelly: All this messing about with URLs is surely going to come back and bite us on the ass.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractPipelineImpl.java
```diff
@@ -178,23 +181,15 @@ public Link getLink() {
         return OrganizationImpl.INSTANCE.getLink().rel("pipelines").rel(getRecursivePathFromFullName(this));
     }
 
-    public static String getRecursivePathFromFullName(BluePipeline pipeline){
```
vivek: @tfennelly This method should not change, it computes recursive path from full name. This expects pipeline.getFullName() has already been encoded properly and no changes needed. So if there is pipeline named 'p/p1' inside folder named 'f/f1'. folderPipeline.getFullName() should give `f%2Ff1/p%2Fp1`, this would result in return from this method as `/pipelines/f%2Ff1/pipelines/p%2Fp1`.

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/BlueUrlTokenizer.java
```diff
@@ -271,6 +272,75 @@ public boolean lastPartIs(@Nonnull UrlPart urlPart, @Nonnull String value) {
...
+        //
+        // The Job "fullName" is typically suffixed with the "name".
+        // The name is NOT part of the path to the job, so lets remove
+        // it from the fullName before we parse and reconstruct the path.
+        // This, along with reencoding of the name, helps us to eliminate
+        // the problems where a Job is programatically created with a name
+        // that's not URL encoded and contains characters that should be
+        // encoded e.g. slashes (see JENKINS-41425).
+        //
+        if (fullName.endsWith(name)) {
```
vivek: How does it distinguish legal '/' from the one present in the job name. That is a job nested inside folders and folders are created with illegal chars. For example job name 'p/p1/ inside folder 'f/f1' should be encoded as full name  `f%2Ff1/p%2Fp1`.

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/BlueUrlTokenizer.java
```diff
@@ -271,6 +272,75 @@ public boolean lastPartIs(@Nonnull UrlPart urlPart, @Nonnull String value) {
...
+
+    /**
+     * Re-encode the supplied string.
+     * <p>
+     * Use this on strings that may already be encoded, but are not guaranteed to
+     * be e.g. Job names (see JENKINS-41425).
+     * @param string The string to be re-encoded.
+     * @return The re-encoded string.
+     */
+    private static String urlReencode(String string) {
```
vivek: Safer to use hudson.Util.rawEncode(name)?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractPipelineImpl.java
```diff
@@ -143,7 +142,11 @@ public Link getLink() {
 
     @Override
     public String getFullName(){
-        return job.getFullName();
+        // Lets make sure that the fullName is properly encoded etc.
+        // The name part of the fullName can contain unencoded characters
+        // if the Job was programmatically created, causing screwups
+        // when mapping those URLs to Blue Ocean web/rest URLs. See JENKINS-41425.
+        return BlueUrlTokenizer.reconstructJobNamePath(job.getName(), job.getFullName());
```
vivek: I think we need to override implementation of job.getFullName() and in there do safe-url-encoding, that is ensure we are not double encoding. For encoding perhaps we should use Util.rawEncode().

vivek: BluePipeline.getName() should be encoded as well. This is to keep things consistent between job name created using UI vs creating them programatically.

vivek:
@tfennelly Left few comments. I guess this is odd behavior thanks to core not checking/encoding job name from inside Jenkins.createProject(), instead it does it on form submission in Jenkins.createTopLevelItem(). Overall I think it should take in to account of fall out from different cases and avoid regression. Some test needs to be added as well, one with top level item and other with nesting including MBP tests. Maybe I should take it from you, if its ok with you?

jglick:
> To test, I used the script console to programmatically create a Job with slashes in the name

That is illegal, and branch projects will never have slashes in their names. So what are you actually fixing here?

vivek:
@jglick It might be illegal but you can create from script console. If its illegal perhaps `jenkins.createProject(org.jenkinsci.plugins.workflow.job.WorkflowJob, "a/job/with/slashes");` should throw exception. Maybe fix lies there but I guess we need to handle it in BO (like classic) regardless as users are somehow able to create jobs with slashes in them.

tfennelly:
@jglick legal/illegal is irrelevant. The fact that it's possible and is screwing BO up and we need to accommodate it is all thats relevant here. 

@vivek hey Vivek, I think it makes sense for you to take this and own it. Wdyt?

blueocean-dashboard/src/main/java/io/jenkins/blueocean/BlueOceanWebURLBuilder.java
michaelneale: yeah feels bad :( stapler has to do all sorts of things like this too (ie branches have to be double encoded for some reason, if you use classic you will see that). The proposed alternative was to totally change the branch name -> URI mapping (which stephen proposed) to something opaque. Certainly easier, but viscerally uglier.

vivek:
@tfennelly sure, will take care of it. thanks:)

michaelneale:
yes I am surprised this actually is a thing to fix, if it isn't really allowed (but there is a linked ticket so someone did see it in the wild...)

i386:
The bitbucket branch source is at fault here - it puts slashes in the names of branches I believe.

michaelneale:
@i386 right and using git flow style names is common, BUT, this only impacts blue ocean, right?

tfennelly:
It's Jenkins core that's at fault here since it allows it to happen. If you can do it, then it's legal. Unwritten/unenforced rules are a waste of time. There are multiple ways of doing it i.e. not just via the Bitbucket plugin.

jglick:
> legal/illegal is irrelevant. The fact that it's possible and is screwing BO up and we need to accommodate it is all thats relevant here.

No I think that is completely off base. What you can do in the script console includes replacing `jenkins-core.jar` with a sad emoji GIF and making it crash. The UI will never allow you to create an `Item.name` which includes `/`. And AFAIK `branch-api` will never do so either—it encodes slashes in branch names (regardless of the branch source). (The *display name* of the job may well includes slashes, or any Unicode text.)

So, again, what is the actual user scenario you are solving? I suspect you are mishandling URL encoding somewhere in Blue Ocean but the test methodology is wrong. @stephenc may be able to assist.

tfennelly:
> The UI will never allow you to create an Item.name which includes /. And AFAIK branch-api will never do so either.

But Jenkins core (`Jenkins.createProject`) does allow creation of a job with slashes in the name. Anyone can write a plugin that generates Jobs with dodgy names. So long as `Jenkins.createProject` allows this (or BO needs to be backward compatible with a version that does), then Blue Ocean is exposed to it and is going to hit issues when it happens (and shit always does happen e.g. [JENKINS-41425](https://issues.jenkins-ci.org/browse/JENKINS-41425) ). 

Relying on higher layers (in the UI, or in other plugins e.g. branch source etc) to enforce these rules is just bad/wrong imo. If we disagree on that, then we can just agree to disagree.

[JENKINS-41425](https://issues.jenkins-ci.org/browse/JENKINS-41425) does suggest there's something leaking through from the Bitbucket plugin. Otoh, could be that there's some driver error there too (the reporter is checking more things for us tomorrow). Personally though, I don't think it matters ... so long as core allows it, then we on BO should protect users from it if we can (for when plugin-xyz comes along and decides to call `Jenkins.createProject` without encoding the name).

This is a big part of what BO is all about i.e. filling/bypassing potholes before the user drives into them Vs just saying "driver error". It's not just about making things look nicer.

Anyway ... @vivek is the one that will ultimately decide how/if this needs to be addressed in BO.

vivek:
+1 @tfennelly 

Fix is going to be generating correct links and generating consistent and correct JSON response with name/fullName etc. In terms of actual fix, if BluePipeline.getName() and BluePipeline.getFullName() returns the name with correct encoding (encode if name had illegal chars and were not encoded by core). Basically correct handling of Item.getName() and Item.getFullName().

We also need to scan our code for wherever its calling Item.getName() or Item.getFullName() as they are going to give inconsistent result depending on whether project was created from UI or elsewhere.

jglick:
> Anyone can write a plugin that generates Jobs with dodgy names.

We can tighten up core, but as far as I know no one _did_ write such a plugin. I think JENKINS-41425 is being misanalyzed somehow. Again: did you reproduce this, and investigate what went on, or not?

`Item.name` must never contain `/`. All sorts of things will break at a much lower level than B.O.

michaelneale:
@jglick I think  JENKINS-41425 is being misdiagnosed, I agree. Looking into it. Tom opened: https://issues.jenkins-ci.org/browse/JENKINS-42249 which seems a bit clearer to reproduce.

tfennelly:
Sorry, I'm on vacation these few days.

No, didn't manage to reproduce it with Bitbucket directly, but did manage to easily do it programmatically. Difference of approach I suppose ... I'd prefer to make BO immune to screwed up name encodings in the URLs if we can (maybe we can't) Vs going around trying to  make sure everything else behaves correctly all the time (bound to failure at some point imo).

Anyway, as I said, @vivek is looking at this now since all the REST stuff is his baby 😄

tfennelly:
In fact ... I'm going to close this PR for now anyway.

michaelneale:
Yes, as it turned out it was nginx decoding URLs confusingly, so it wasn't the root cause. Core does allow invalid job names but that is not something we can correct or should correct like this necessarily. So closing this is the right decision I think. Thanks all!

PR #849, opened 2017-02-23
[JENKINS-41434] Hide link when no create permission

imeredith:
# Description

To test create a user without admin and job create permissions.

See [JENKINS-41434](https://issues.jenkins-ci.org/browse/JENKINS-41434).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-core-js/src/js/User.js
+++ b/blueocean-core-js/src/js/User.js
@@ -72,7 +72,7 @@ export class User {
     }
 
     get _class() {
-        return this._blueUser.class;
+        return this._blueUser._class;
     }
 
     get _links() {
--- a/blueocean-core-js/src/js/User.js
+++ b/blueocean-core-js/src/js/User.js
@@ -4,14 +4,91 @@
 
 import { blueocean } from './scopes';
 
-export class User {
-
+class PipelinePermissions {
     constructor(blueUser = {}) {
-        this._class = blueUser._class;
-        this._links = blueUser._links;
-        this.email = blueUser.email;
-        this.fullName = blueUser.fullName;
-        this.id = blueUser.id || 'anonymous';
+        if (blueUser.permission && blueUser.permission.pipeline) {
+            this._permissions = blueUser.permission.pipeline;
+        } else {
+            this._permissions = {};
+        }
+    }
+
+    create() {
+        return !!this._permissions.create;
+    }
+
+    read() {
+        return !!this._permissions.read;
+    }
+
+    start() {
+        return !!this._permissions.start;
+    }
+
+    stop() {
+        return !!this._permissions.stop;
+    }
+
+    configure() {
+        return !!this._permissions.configure;
+    }
+}
+
+
+class CredentialPermissions {
+    constructor(blueUser = {}) {
+        if (blueUser.permission && blueUser.permission.credential) {
+            this._permissions = blueUser.permission.credential;
+        } else {
+            this._permissions = {};
+        }
+    }
+
+    create() {
+        return !!this._permissions.create;
+    }
+
+    view() {
+        return !!this._permissions.view;
+    }
+
+    delete() {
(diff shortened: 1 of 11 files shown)


michaelneale:
This upsets a bunch of the ATH: 

https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/1575/#showFailuresLink

michaelneale:
Hrm... i think this can't work with the ATH until it has user identity, this also stops blue ocean working for a whole bunch of people who used full anon access if I read it correctly. This may not be the time for this in retrospect.

michaelneale:
ok have looked closer, yes this cases failures in the expected way. We can't do this one right now unfortunately without a bit of work to the ATH, or else putting "special casing" in for the ATH (which is not good). We don't know what else this may do yet either - the ATH runs with full anon access (and some people do) - and this will also affect them. 

So this is a :bug: for me from now, I think we need to postpone this (yes, we should have known this before starting). 

@imeredith tried to get you on various chat things but can't seem to reach you in the day - so can cover this later if you have any ideas.

michaelneale:
@imeredith some other options: James suggests that the security check should allow that if they are admin (even anon but still admin) - but button could show. 

The other way is to alter the ATH just a little bit to not look for that button (short term thing).

imeredith:
I think I have a 'fix' for ath which will let us login, will try it tomorrow.

michaelneale:
https://github.com/jenkinsci/blueocean-acceptance-test/pull/122

should fix the test failure for this (very minor).

blueocean-core-js/src/js/User.js
```diff
@@ -21,6 +98,18 @@ export class User {
...
+        const permission = this._blueUser._permission || {};
+        return !!permission.administrator;
+    }
+
+    get permissions() {
+        return {
+            pipeline: new PipelinePermissions(this._blueUser),
+            credential: new CredentialPermissions(this._blueUser),
+        };
+    }
```
cliffmeyers: Do we want to add a getter or function for every single property in this graph? It seems like a maintenance problem as new permissions are added, we need to keep adding them here, and republishing every time. Should we consider doing a deep clone from the passed in `blueUser` directly onto the `User` instance?

michaelneale:
@cliffmeyers good point - yes that wouldn't be a good experience to just let them fail at the end.

michaelneale:
I think this should be postponed and ticket put back in queue until we have ATH - OR, as @i386 suggested, make this still show up in the case of security turned off (so it still is hidden when a logged in user doesn't have create permissions) - then ATH will be happy

michaelneale:
This passes the ATH and is likely in better shape now - if updated then I think it can go in.

cliffmeyers:
@michaelneale so we won't show an error message to the user if they've bookmarked creation and are not logged in? Could someone at least log that as a separate ticket that we can fix later?

i386:
Could we just make it redirect?

i386:
I raised https://issues.jenkins-ci.org/browse/JENKINS-42375 for this FYI

blueocean-core-js/src/js/User.js
imeredith: I agree I was just maintaining the api.

michaelneale:
@imeredith this seems to fail when looking at the translation: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/JENKINS-41434_creation_permissions/2/testReport/junit/(root)/loadGerman/Step_02/ - which is most odd... that shouldn't really be affected here - worth checking out though.

michaelneale:
OK, after an "underarm bowl" incident where I made a branch on the ATH which ended up with the one failure tricking @imeredith  - this now passes ATH, and sniff test. 

:bee: from me.

PR #847, opened 2017-02-22
[FIX JENKINS-38751] Escape back to Jenkins classic

tfennelly:
# Description

See [JENKINS-38751](https://issues.jenkins-ci.org/browse/JENKINS-38751).

Can be manually tested. Look for the "escape to app" widget in the headers (main and run details).

screenshot 2017-02-22 16 44 16

ATH tests: https://github.com/jenkinsci/blueocean-acceptance-test/pull/123

# Submitter checklist
- [x] Link to JIRA ticket in description, if appropriate.
- [x] Change is code complete and matches issue description
- [x] Appropriate unit or acceptance tests or explanation to why this change has no tests
- [x] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below.
- [x] Ran Acceptance Test Harness against PR changes.

# Reviewer checklist
- [ ] Run the changes and verified the change matches the issue description
- [ ] Reviewed the code
- [ ] Verified that the appropriate tests have been written or valid explanation given

@reviewbybees

PR #859, opened 2017-02-27
needs to know if it is set to a string or not

michaelneale:
# Description

See https://issues.jenkins-ci.org/browse/JENKINS-42291

No published or anything, just proposed... 

@reviewbybees

### DIFF
--- a/blueocean-core-js/src/js/config.js
+++ b/blueocean-core-js/src/js/config.js
@@ -65,12 +65,13 @@ export default {
     showOrg() {
         return this.isFeatureEnabled('organizations.enabled', false);
     },
-
+    
     getJenkinsRootURL() {
         if (!config.isLoaded) {
             this.loadUrls();
         }
-        return config.jenkinsRootURL || '/jenkins';
+        const { jenkinsRootURL } = config;
+        return typeof jenkinsRootURL === 'string' ? jenkinsRootURL : '/jenkins';
     },
 
     getRestRoot() {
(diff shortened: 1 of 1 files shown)


blueocean-core-js/src/js/config.js
```diff
@@ -65,12 +65,13 @@ export default {
     showOrg() {
         return this.isFeatureEnabled('organizations.enabled', false);
     },
-
+    
     getJenkinsRootURL() {
         if (!config.isLoaded) {
             this.loadUrls();
         }
-        return config.jenkinsRootURL || '/jenkins';
```
michaelneale: this will be set to '' in the case of it running on root (from jenkins), so this never would have worked.

blueocean-core-js/src/js/config.js
```diff
@@ -65,12 +65,13 @@ export default {
...
+        const { jenkinsRootURL } = config;
+        return typeof jenkinsRootURL === 'string' ? jenkinsRootURL : '/jenkins';
```
michaelneale: Instead, just check if it is a string or not.

tfennelly:
@michaelneale I encountered this the other day too doing https://github.com/jenkinsci/blueocean-plugin/pull/847. See https://github.com/jenkinsci/blueocean-plugin/pull/847/commits/59368db9049dd2bf44b486ee45bd708f472966a6

michaelneale:
Yes can close this one in favour of that IMO

tfennelly:
Closing this PR. Same fix is already in #847 so will make into upstream there.

PR #867, opened 2017-02-28
[FIX JENKINS-42137] Don't fetch run data on an event if it not already in the store

tfennelly:
# Description

Only fetch job activity data if we have a pager instance for it. I think this is the right thing to do. It seems to work + eliminates that unwanted REST call for pipelines not yet visited by the UI.

See [JENKINS-42137](https://issues.jenkins-ci.org/browse/JENKINS-42137).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-core-js/src/js/services/DefaultSSEHandler.js
+++ b/blueocean-core-js/src/js/services/DefaultSSEHandler.js
@@ -66,12 +66,18 @@ export class DefaultSSEHandler {
 
         for (const key of this.branchPagerKeys(event)) {
             const pager = this.pagerService.getPager({ key });
-            this.activityService.fetchActivity(runSelf, { overrideQueuedState }).then(d => {
-                if (pager && !pager.has(runSelf)) {
-                    pager.insert(runSelf);
-                }
-                this.pipelineService.updateLatestRun(d);
-            });
+            // If we don't have a pager, then the app in the browser has never visited that
+            // Job, which means there's no need to fetch the data associated with it because
+            // there's nothing in the app that needs to be kept up-to-date. The data will
+            // be fetched when the user visits that job/pipeline.
+            if (pager) {
+                this.activityService.fetchActivity(runSelf, { overrideQueuedState }).then(d => {
+                    if (!pager.has(runSelf)) {
+                        pager.insert(runSelf);
+                    }
+                    this.pipelineService.updateLatestRun(d);
+                });
+            }
         }
     }
     queueCancel(event) {
@@ -121,7 +127,7 @@ export class DefaultSSEHandler {
         };
 
         this.activityService.setItem(newRun);
-        
+
         for (const key of this.branchPagerKeys(event)) {
             const pager = this.pagerService.getPager({ key });
             if (pager) {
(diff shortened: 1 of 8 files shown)


michaelneale:
Looking good - does what it says, but it seems to have mucked up something in the ATH, which I was able to reproduce locally: 

https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/1648/

the failing tests are failing.js and failingStage.js in src/test/js

They fail in the same way: a failing test is run, and then the header never updates (you can see it if it runs locally). Switching back to master and it is ok. So perhaps there is some event being dropped that shouldn't be, or else there is something else going on. Those tests can be a little race-ey, so maybe they could be better, as when I run a failing build like that by hand, it does update the result. 
But for now, either a :bug: here or in the ATH

tfennelly:
@michaelneale yeah I saw that, but didn't go about fixing it yet. I did try reproduce it on a local job though and failed - the header etc loaded fine. So must be more to causing it than I had thought. I'll look more closely, or maybe get more instructions from you on how you reproduced it.

michaelneale:
@tfennelly oh I just ran ATH in `--dev` mode locally and then: `nightwatch src/test/js/failing.js` and immediately saw it (I had a VNC session open at the time, I saw header never changed).

Was not able to reproduce it by human clicking in chrome. Switched back to master - and that doesn't happen. very interesting, this may show up something interesting. Or an annoying test. Hard to say!

blueocean-core-js/src/js/services/DefaultSSEHandler.js
```diff
@@ -66,12 +66,18 @@ export class DefaultSSEHandler {
...
+            // If we don't have a pager, then the app in the browser has never visited that
+            // Job, which means there's no need to fetch the data associated with it because
+            // there's nothing in the app that needs to be kept up-to-date. The data will
+            // be fetched when the user visits that job/pipeline.
+            if (pager) {
+                this.activityService.fetchActivity(runSelf, { overrideQueuedState }).then(d => {
+                    if (!pager.has(runSelf)) {
+                        pager.insert(runSelf);
+                    }
+                    this.pipelineService.updateLatestRun(d);
```
kzantow: I think this is the issue, the run may not have been fetched, but in the case it is a new, latest run it won't update stuff...

kzantow: ... or something of the sort.

michaelneale: @kzantow we were thinking that this change (this whole PR) may not be warranted just yet, as it may have other subtle bugs introduced (ATH picked up just one). So unless we/you are sure - do you think it should be shelved? it is a valid improvement, but not a critical thing IMO (do you agree?)

kzantow: I'm just pointing out that I think this is what could be causing the ATH failure you mentioned. I like the change, but there probably needs to be some more logic to determine what to update (this is exactly why I've mentioned not being a fan of this manual client cache we have). I think it's a valid improvement, but as you said, probably not critical.

PR #869, opened 2017-02-28
Task/jenkins 42119 creation design tweaks

cliffmeyers:
# Description
- See [JENKINS-42119](https://issues.jenkins-ci.org/browse/JENKINS-42119).

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/creation/github/steps/GithubCredentialStep.jsx
+++ b/blueocean-dashboard/src/main/js/creation/github/steps/GithubCredentialStep.jsx
@@ -59,7 +59,7 @@ export default class GithubCredentialsStep extends React.Component {
                 </p>
 
                 <FormElement errorMessage={errorMessage}>
-                    <TextInput className="text-token" placeholder="Enter access token" onChange={val => this._tokenChange(val)} />
+                    <TextInput className="text-token" placeholder="Your Github access token" onChange={val => this._tokenChange(val)} />
 
                     <button className="button-connect" disabled={buttonDisabled} onClick={() => this._createToken()}>Connect</button>
                 </FormElement>
--- a/blueocean-dashboard/src/main/less/creation/git.less
+++ b/blueocean-dashboard/src/main/less/creation/git.less
@@ -3,6 +3,10 @@
         margin-top: 0;
     }
 
+    .text-repository-url {
+        width: 100%;
+    }
+
     .credentials > .FormElement-heading {
         padding: 0;
     }
@@ -14,6 +18,7 @@
 
 .git-step-rename {
     .text-pipeline {
+        width: 100%;
         margin-right: 10px;
     }
 }
--- a/blueocean-dashboard/src/main/less/creation/index.less
+++ b/blueocean-dashboard/src/main/less/creation/index.less
@@ -29,10 +29,16 @@
         display: flex;
         flex-flow: row wrap;
 
-        .provider-button .IconButton {
-            margin-right: 20px;
-            min-width: 220px;
-            height: 56px;
+        .provider-button {
+            .IconButton {
+                margin-right: 20px;
+                min-width: 220px;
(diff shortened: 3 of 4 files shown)


cliffmeyers:
Screenies. Perhaps @brody can have a look.

creation-1-initial

creation-2-github

i386:
Oh that is very nice :)

i386:
@cliffmeyers  I had just tried this out - it's beautiful.

One thing I noticed is that the branch sources buttons never get displayed. Should that be done along side [JENKINS-42206](https://issues.jenkins-ci.org/browse/JENKINS-42206)? I believe this is assigned to @imeredith 

-blueocean-2017-02-28-18-11-18

michaelneale:
:bee: and ATH happy - nice one @cliffmeyers !

kzantow:
why would you disable the buttons? how could you undo clicking the wrong thing then?

cliffmeyers:
Right, @i386 I think @kzantow makes a valid point. If those are disabled the whole time, then there's no easy way to start a new flow without refreshing this screen or navigating away and back.

kzantow:
makes sense to highlight one as active, but i'd avoid disabling them.

i386:
If we are not disabling the first one in this scenario then we shouldn't be disabling any steps.

kzantow:
agree... unless there's a way to go back to the previous step, i'd avoid disabling them at all; if there's a way to go back to the prior step, then disabling them is just fine

kzantow:
the source and single/multiple repository steps seem to be missing selected states, though

i386:
@brody can you weigh in?

brody:
@cliffmeyers Correct, previously selected buttons shouldn't be disabled, so the user can change their mind and change any selection in the flow... which obviously would have knock on affects unselecting any of the below selections.

cliffmeyers:
@i386 @brody @kzantow to clarify, here is how it works today:

1. User can move through the steps in flow. If the user picks the wrong org or the wrong auto-discover option, they can obviously go back up to an early step and change their mind. The UI will re-render the next step appropriately.
2. Once the user presses "Create Pipeline", the buttons in the preceding steps are disabled. This is because while the system is working, listening for events, etc we probably don't want the user fiddling with selections. I mean we could, just realize it will open a whole new class of bugs to defend against.
3. There is a bug and oversight that when "Create Pipeline" is clicked, the Git and Github buttons at the top aren't disabled. This can be fixed, it will just take a modest effort - only because how those buttons are handled via extension point mechanism are subtly different than how the steps themselves are.
4. We could further enhance the UI so that when the creation process "completes" that the buttons at the top become enabled. If the user wants to do a new creation flow for Git or Github they just need to click one of those buttons and the old steps get wiped. (To clarify further, "completes" means that creation finishes with success, with an error, or after the 60s timeout currently in place if the right events aren't received).

kzantow:
Ah, right... after clicking create. Yeah, makes sense how it is now and to disable the top selections after clicking create, as you said.

brody:
1. 👍
2. Sweet. Disabling the entire form when "Create Pipeline" is clicked is perfect.
3. 👍
4. Once the creation process successfully completes, the intended action is to launch the pipeline. Reenabling the buttons on failure, while showing the error, would be good.

cliffmeyers:
@michaelneale per your question in Gitter, I think this PR satisfies the work items in the ticket. If we want to fiddle with button enabling or disabling, I'm happy to cut another ticket and handle in another PR.

michaelneale:
ok great @cliffmeyers - Ivan made a small change to creation so may want to update to master and be happy this is still ok (should be) - but LGTM

cliffmeyers:
Created https://issues.jenkins-ci.org/browse/JENKINS-42507 to track the disabling issue.

PR #872, opened 2017-03-01
[FIX JENKINS-42206] Add selected state for git/github buttons

imeredith:
# Description

See [JENKINS-42206](https://issues.jenkins-ci.org/browse/JENKINS-42206).

@cliffmeyers wdyt ?

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


i386:
@imeredith nice to see the selection state but it is not to the style of the other big buttons.

jenkins blue ocean 2017-03-02 10-51-49

imeredith:
screen shot 2017-03-02 at 1 30 03 pm

kzantow:
Not really part of this PR, but I wish the icons on those buttons were equidistant from the top/bottom/left, like the repository icons are

i386:
@kzantow not sure what you mean? Perhaps @brody knows

i386:
Looking great @imeredith - gets a 🐝  from me pending a code review.

kzantow:
@i386 
image

i386:
@kzantow nice find 👍

imeredith:
There is a larger bucket of work to do here I think. Some of these things need to be moved out of the provider and genefied, either in JDL, or just creation generallly

i386:
@imeredith these buttons could all share the same component?

michaelneale:
@i386 with a fair bit of work, yes (given how expensive it is to do common components right now).

i386:
@michaelneale ack. We need to pay that down post-1.0.

kzantow:
just to hijack @imeredith 's PR even more, the extensibility stuff will solve a lot of this: shared components between plugins, APIs for stuff like name/icon, etc...

imeredith:
The lists are uisng <List /> not button. Agreed they should be the same, but its not so simple as "using the same thing"

i386:
@kzantow mentally note this for the extensibility investigation

brody:
Screenshot from Zeplin
image

blueocean-dashboard/src/main/js/creation/git/GitDefaultOption.js
```diff
@@ -15,9 +15,11 @@ export default function GitDefaultOption(props) {
         }
     }
 
+    const styles = `git-creation monochrome ${props.isSelected ? 'active' : ''}`;
```
cliffmeyers: minor nit : I'd name this `classNames` as we use that convention throughout, and at a glance most folks might assume that `styles` would be a React style object rather than list of classes.

michaelneale:
passed the ATH (build #5) so over to @cliffmeyers to bless

cliffmeyers:
@brody as far as the styling, the "scm buttons" at the top use the "large button" style which has different paddings for the top and sides. If we agree that the padding should be uniform, we can make the change to the global button styles and it will "just work." If we feel strongly the padding for buttons with and without icons should be different, that's fine too, just a global style change we need to make to `IconButton`.

@imeredith note for later: there is some inconsistency in styling between the buttons at the top and the org buttons because the latter are not buttons. The `List` uses radio buttons under the hood, so you can use arrow keys to go through the options but also bypass the entire list with a single press of tab. If `List` used buttons, then a list with "n" buttons will require "n" keypresses of tab on certain platforms, and for long lists it will make keyboard accessibility totally unusable (imagine pressing tab 1900+ times to go through the repo list for jenkinsci).

cliffmeyers:
All that said, I think this PR fixes the original bug / issue and is good to go 🐝

brody:
@cliffmeyers where else has "large button" been used? 

Padding for icon & non-icon buttons shouldn't be different. 
screen shot 2017-03-06 at 10 23 47 am

cliffmeyers:
@brody per the Zeplin style guide...

style-guide-1-buttons

brody:
@cliffmeyers ah ok. I designed and have thought of Large buttons & Radio grid buttons as very different things.

PR #875, opened 2017-03-02
[FIX JENKINS-41662] Input submissions not working if another browser loads the same input step

tfennelly:
# Description

See [JENKINS-41662](https://issues.jenkins-ci.org/browse/JENKINS-41662).

`io.jenkins.blueocean.rest.impl.pipeline.InputStepImpl` has some unusual code, resetting the ID of the underlying pipeline `InputStep` instance to a random ID. The result of this ... two browsers load the same input step ... the first browser is screwed because the id it is using to submit against no longer corresponds to a valid ID because the load from the second browser caused it to be changed via `io.jenkins.blueocean.rest.impl.pipeline.InputStepImpl`.

@vivek hey Vivek ... can you look at this and tell us what was the intention behind the original impl, if you can still remember (I know you are like me and getting old ... memory failing ;) ).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/InputStepImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/InputStepImpl.java
@@ -19,7 +19,10 @@ public class InputStepImpl extends BlueInputStep {
 
     public InputStepImpl(InputStep inputStep, Reachable parent) {
         this.inputStep = inputStep;
-        this.inputStep.setId(UUID.randomUUID().toString().replaceAll("-", ""));
+        if (inputStep.getId() == null) {
+            // TODO: Why are we resetting the input ID to something random?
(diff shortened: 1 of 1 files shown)


vivek:
LGTM 🐝 Thanks for fixing:)

tfennelly:
@vivek hey Vivek ... any idea why it was setting the ID?

vivek:
@tfennelly because you might get id as null and we need to look for one when it's submitted. See `PipelineStepImpl.submitInputStep()`. Your fix makes it do right thing, that is set id when its null.

tfennelly:
@vivek ok ... thanks Vivek. Just wanted to make sure.

michaelneale:
WOW what a great catch... hahaha wow.

i386:
Nice one!

When would it not have an ID?

vivek:
@i386 Oh, well, inputStep.setId() allows null and its marked optional in InputStep.

PR #877, opened 2017-03-02
Consume new workflow-api fixes

svanoort:
# Description

Test the fixes in https://github.com/jenkinsci/workflow-api-plugin/pull/33 against Blue Ocean CI coverage to see that there are no regressions introduced.

See the wad of linked JIRA issues there, which are tied to Blue Ocean consumption of workflow-api.

Special case here, since it's just a dependency bump.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -245,7 +245,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.7</version>
+            <version>2.12-alpha</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -245,7 +245,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.12-alpha</version>
+            <version>2.12-alpha-2</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>


svanoort:
@vivek @recampbell Pending results from this to see if the mega-fix in workflow-api plays nicely with the BlueOcean implementation.

svanoort:
I already see kabooms, so clearly something was either not back-compatible or unexpected from the API consumer.

i386:
Might want to consder updating https://github.com/jenkinsci/blueocean-plugin/tree/topic/dependency-update-12-02-2017 with your new dependency. We've been checking the tests against all the new changes coming out of oss team and I believe the failures on this one are due to the workflow api regression

michaelneale:
yes those test failures look similar to what was seen before with 2.8 and onwards. Perhaps some other dependencies need to be updated at the same time?

vivek:
@svanoort looking at the failures.

i386:
We definitely need to pull in and test all the other changes from the branch I quoted. Without those we don't know if things work if someone upgrades all heir plugins ATM.

vivek:
@i386 +1

i386:
Here it is with @svanoort and @stepehenc's latest https://github.com/jenkinsci/blueocean-plugin/pull/817

vivek:
@svanoort I looked in to failing tests, there seems to be a regression. parallelEnd event is not fired for the pipeline scripts in these tests. If you try the script below, you can see parallelEnd event is not send. This causes bug in computing each branch's status and timing correctly. 

```
node {
    stage("parallelStage"){
      parallel left : {
            echo "left running"
        }, 
        right : {
            sh 'echo "right running"'
        }
    }
}
```

I will send some more details on the failing tests with more specific details.

i386:
@svanoort @vivek all the unit tests seem to pass on my dependency bump branch now https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/topic%2Fdependency-update-12-02-2017/14/tests

vivek:
@i386 thanks. I will try this branch out locally to ensure it all works fine or which dependency is making it work alright.

vivek:
@i386 narrowed it down to workflow-cps. Upgrading it to 2.8 from 2.5 made all the tests pass and in log I can see parallelEvent is fired correctly. @svanoort can you update this PR with workflow-cps 2.8. This also needs workflow-support upgraded to 2.13?

svanoort:
@recambell @vivek  @i386 All green!  I removed the need to configure the parallel start predicate (API consumers with a legacy version of workflow-cps should have had to do that anyway, but it appears this was somehow missed and did not cause a test failure due to blind luck).

Tests will pass with any workflow-cps version, but analysis proceeds considerably faster when using a current version of workflow-cps, due to the use of reflection to work around a previous dependency issue.

vivek:
@svanoort awesome! Thanks for the quick fix!

michaelneale:
cool - running ATH over this now. It passed on @i386 version which bumped a whole lot of deps.

michaelneale:
There are now a bunch of failures in the ATH. 

I think there are more dependencies to update: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/1716/

@svanoort @vivek if this was updated the way https://github.com/jenkinsci/blueocean-plugin/pull/817 was - it may be in a better shape, but its not ready yet.

michaelneale:
Yes this has broken things in other ways, as of alpha-2 - need to look into what, but it is less happy than before in the ATH (unit tests are happier, but then they were happy with other depedencies updated). Something new happened in alpha 2.

svanoort:
@michaelneale What's failing in the ATH?  This *should* be more or less back-compatible except where tests were expecting subtly incorrect results in some of the more involved parallel cases.  Are the failing tests hardcoded on something highly specific to the flow structure, or timing based?

michaelneale:
@svanoort likely just a bum steer - I am looking into it. The errors don't look legit to me.

michaelneale:
@svanoort ok, updating to master https://github.com/jenkinsci/blueocean-plugin/pull/817 and it all passes (I think there were some features merged over the weekend which are missing from this fork): https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/1718/ - perhaps we should close this and shift to that branch on James PR and continue work there?

michaelneale:
@svanoort if this was updated with master - I bet it would be just fine and dandy. From what I can tell from other branch - things look great with this new alpha. The ATH failure is due to this fork being a little stale I think. 

Eyeballing and it looks good too - @vivek and @i386 ideally would cast eyes over it, but for the purposes of that alpha, I think we have a winner!

svanoort:
Thanks @michaelneale - I've cut the release (after review passed), so I'm going to close this one out in preference to the more-comprehensive update.

PR #886, opened 2017-03-06
[FIX JENKINS-41966] load and save using the editor from Github

kzantow:
# Description

Only real changes here are:

* refactored loading indicator to `blueocean-core-js`
* modified back behavior of `RunDetails` to use router history
* added an extension point for the `Create Pipeline` button

Before the `pipeline-editor` work is completed and the plugin is installed (and we'll update the version in the POM), we'll get this for creation, which I think is acceptable:

image

See [JENKINS-41966](https://issues.jenkins-ci.org/browse/JENKINS-41966).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


pom.xml
```diff
@@ -133,49 +133,49 @@
         <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>blueocean-commons</artifactId>
-            <version>${project.version}</version>
+            <version>${project.parent.version}</version>
```
michaelneale: oh? What difference does this make vs project?

kzantow: I should probably revert this, but the issue was if you use the blueocean parent (which I was trying for another reason) you have to set your plugin to the same as the blueocean verasion. But one shouldn't need to use the blueocean parent, so...

michaelneale: ok

kzantow:
ATH passing: https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/master/1744/pipeline

blueocean-dashboard/src/main/js/components/RunDetails.jsx
```diff
@@ -64,6 +64,7 @@ class RunDetails extends Component {
 
     componentWillMount() {
         this._fetchRun(this.props, true);
+        this.opener = locationService.previous;
```
cliffmeyers: Good call, this is much simpler code. Can remove the `true` as second param up above though.

blueocean-dashboard/src/main/js/components/RunDetails.jsx
```diff
@@ -138,18 +135,13 @@ class RunDetails extends Component {
...
-         this.opener would then be location and we the above location = this.opener || { pathname: fallbackUrl]
-         */
-        location.query = null;
-        router.push(location);
+        const { router, params } = this.context;
+        if (this.opener) {
+            router.goBack();
+        } else {
+            const fallbackUrl = buildPipelineUrl(params.organization, params.pipeline);
+            router.push(fallbackUrl);
```
cliffmeyers: I am not sure if `push` or `replace` is more appropriate here.

blueocean-core-js/src/js/LoadingIndicator.js
```diff
@@ -0,0 +1,84 @@
...
+                clearTimeouts();
+                setLoaderClass('complete', 10);
+                timeouts.push(setTimeout(() => {
+                    // The Element.classList is a read-only property
+                    const classList = loadbar.classList;
+                    if (classList && classList.length && classList.length > 0) {
+                        const classListAsArray = new Array(classList.length);
+                        for (let i = 0, len = classList.length; i < len; i++) {
+                            classListAsArray[i] = classList[i];
+                        }
```
cliffmeyers: You may be able to simplify this code to:
`const classListAsArray = [... classList]`

blueocean-dashboard/src/main/js/components/RunDetails.jsx
kzantow: hmm, I'm not really sure, either. replace might be more consistent, but if i go to the page directly, and accidentally hit esc or something, it would be nice to go back in that case.

kzantow:
@cliffmeyers the `RunDetails` changes are necessary to solve the 'link to editor page' from there. Otherwise, we get into a cycle, as you navigate to the pipeline editor, and when you navigate back to the run details, it stores the `opener` of the pipeline editor. it seems to make a lot more sense to just pop this off the history, as it's supposed to be a modal for now, anyway. this gives it more 'modal' type behavior.

blueocean-core-js/src/js/LoadingIndicator.js
kzantow: good call, updated.

kzantow:
@michaelneale happy to have you give this a smoke test before merging, if you had some time - especially making sure loading indicators aren't acting really crazy, though you might see more, and the run details seems to work as expected wherever you navigate from

blueocean-dashboard/src/main/js/components/RunDetails.jsx
```diff
@@ -198,18 +190,24 @@ class RunDetails extends Component {
...
                           onNavigation={ switchRunDetails }
                           autoNavigate
             />,
-            <RunButton className="dark"
+            <RunButton className="icon-button dark"
                        runnable={ this.props.pipeline }
                        latestRun={ currentRun }
                        buttonType="stop-only"
             />,
+            <Extensions.Renderer extensionPoint="jenkins.blueocean.rundetails.top.widgets"
```
michaelneale: Nice

michaelneale:
@kzantow nice - yes I can see loading worm on creation too - which is a nice addition. Of course when it scrolls down it isnt visible but that doesn't matter. 

LGTM :bee: 

This is a tour de force of necessary additions and improvements.

PR #887, opened 2017-03-06
just upgrade workflow api essentials

michaelneale:
# Description

Just upgrade workflow as recommended 
@reviewbybees


michaelneale:
not using this one

PR #889, opened 2017-03-07
JENKINS-41372# Fix to show correct Link header start parameter

vivek:
# Description

See [JENKINS-41372](https://issues.jenkins-ci.org/browse/JENKINS-41372).

GET http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/?start=0

should print Link header start parameter to be 0.

```
Link:</jenkins/blue/rest/organizations/jenkins/pipelines/?start=0&limit=100>; rel="next"
```

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/PipelineApiTest.java
+++ b/blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/PipelineApiTest.java
@@ -2,6 +2,8 @@ package io.jenkins.blueocean.service.embedded;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.mashape.unirest.http.HttpResponse;
+import com.mashape.unirest.http.Unirest;
 import com.mashape.unirest.http.exceptions.UnirestException;
 import hudson.Extension;
 import hudson.FilePath;
@@ -42,7 +44,6 @@ import org.jvnet.hudson.test.ExtractResourceSCM;
 import org.jvnet.hudson.test.MockFolder;
 import org.jvnet.hudson.test.TestBuilder;
 import org.jvnet.hudson.test.ToolInstallations;
-import org.jvnet.hudson.test.recipes.WithPlugin;
 import org.kohsuke.stapler.export.Exported;
 
 import java.io.IOException;
@@ -53,6 +54,7 @@ import java.util.Stack;
 import java.util.concurrent.ExecutionException;
 
 import static junit.framework.TestCase.assertFalse;
+import static org.junit.Assert.*;
 
 /**
  * @author Vivek Pandey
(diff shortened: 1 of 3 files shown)


michaelneale:
passes ATH but up to @cliffmeyers to bless as he would know what impact if any.

cliffmeyers:
@michaelneale we don't actually use the "Link" response header in the UI code, so per my comments in the ticket the risk is low.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/pageable/PagedResponse.java
```diff
@@ -54,7 +54,7 @@ public void generateResponse(StaplerRequest req, StaplerResponse rsp, Object nod
                     }
                     Object[] page = Iterators.toArray(resp.iterator(start, limit), Object.class);
                         String separator = (req.getQueryString() != null) ? "&" : "?";
-                        rsp.setHeader("Link", "<" + req.getRequestURIWithQueryString() + separator + "start=" + (start + limit) + "&limit="+limit + ">; rel=\"next\"");
+                        rsp.setHeader("Link", "<" + req.getRequestURIWithQueryString() + separator + "start=" + start + "&limit="+limit + ">; rel=\"next\"");
```
cliffmeyers: I am not sure this is the correct change. Isn't this just going to return the same values for `start` and `limit` as were specified in the original request? The Link relation should point to  "next" page, however.

I think the old logic adding `start` and `limit` was correct but there needs to be some logic or tokenizer to prevent duplicate `start` and `limit` parameters from being set.

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/PipelineApiTest.java
```diff
@@ -68,6 +70,20 @@ public void getFolderPipelineTest() throws IOException {
         validatePipeline(p, response);
     }
 
+    @Test
+    public void linkStartLimitTest() throws IOException, UnirestException {
+        MockFolder folder = j.createFolder("folder1");
+        Project p = folder.createProject(FreeStyleProject.class, "test1");
+
+        HttpResponse<String> response = Unirest.get(getBaseUrl("/organizations/jenkins/pipelines/folder1/pipelines/"))
```
cliffmeyers: Perhaps a better test would actually specify values for `start` and `limit` here?

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/pageable/PagedResponse.java
vivek: ack.

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/PipelineApiTest.java
vivek: ack

vivek:
@cliffmeyers duplicates removed and more tests added. PTAL.

cliffmeyers:
Tested and LGTM! 🐝

PR #891, opened 2017-03-07
Minor fixes to Docker file and Jenkins file

halkeye:
# Description

* Add UID and GID support so it can be built on any jenkins install. Jenkins doesn't always run as user 1000
* Add latest npm so it never fails on fsevents (optional dep)
* Add phantomjs to speed up runtime

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -2,6 +2,9 @@ FROM ubuntu:16.04
 
 ENV MAVEN_VERSION 3.3.3
 ENV NODE_VERSION 6.4.0
+ENV PHANTOMJS_VERSION 2.1.1
+ARG UID=1000
+ARG GID=1000
 
 USER root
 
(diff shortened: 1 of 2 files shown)


michaelneale:
@halkeye nice - so where else are you planning to use this .build file? 
With "mvn install" and js-builder, it will pull its own version of npm anyway - but I assume you are doing something else with this image?

michaelneale:
I think the way MBP works, I will have to put this in a branch so it "trusts" the Jenkinsfile change (I think in forks they are not used, for security reasons). or @halkeye I can give you repo access so you can work on a branch if you like (I suspect you will do PRs from time to time anyway).

halkeye:
wasn't planning on using it anywhere else. Both blueocean-plugin and ATH use the exact same dockerfile (but named differently), so I have another PR coming for that too.

mvn install pulls npm? I never saw that behavior but i can easily be mistaken.

michaelneale:
I checked this out on a branch of mine and LGTM - not sure if the phantomjs stuff is needed though - as it would be pull as part of the normal build anyway.

michaelneale:
Doesn't seem faster, yet. Probably as it is building the docker image each time until warm. I have not used the build params to docker build, so not entirely confident how they affect build behavior.

michaelneale:
thanks for this @halkeye !

PR #907, opened 2017-03-16
[FIX JENKINS-39770] Karaoke: steps and nodes are limited to 100 - need to increase limit

scherler:
# Description

See [JENKINS-39770](https://issues.jenkins-ci.org/browse/JENKINS-41897).

This change allows to render the pipeline graph when not enough nodes are returned in parallel

https://github.com/jenkinsci/jenkins-design-language/pull/158

# test resources

- https://github.com/scherler/blueocean-shared-library/blob/master/vars/lotsParallel.groovy
- https://github.com/scherler/jenkins-testfiles/blob/lotsParallel/Jenkinsfile

# Preview

screenshot from 2017-03-16 09-45-45

screenshot from 2017-03-16 09-46-20

screenshot from 2017-03-16 09-50-04

screenshot from 2017-03-16 09-50-14

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
Is this meant to include the change to include the jdl version? 

Also does this bump the limit of nodes as well?

scherler:
@michaelneale no it did not raise the initial limit. I can do that as well but IMO even with 100 nodes it is hardly usable so add more will not help,

michaelneale:
Nice one. Interesting - I ran out of time to try this, but doesn't seem too bad approach. I think @i386 would need to inspect/try to check it out.

scherler:
Some observations, the horizontal nodes have a big problem with "earlier" nodes, they are basically unreachable since they are in the left corner and unreachable by scroll (screenshot nodes < 46). There would need to be some overlay applied to the svg to scroll from one extreme to the other.

The vertical is ordering the nodes by name which leads to an order like p-1, p-10, p-100, p-101, ...  further it leaves a long space under the graph. Other then that it is 100% functional.

michaelneale:
I think this is ok, it certainly stops it breaking. 

I used a simple pipeline: 

```
def lotsParallel(max = 110) {
        
    def build_tasks = [:]
        
    for (ii = 0; ii < max; ii++) {
        build_tasks["parallel-${ii}"] = {
          println "Number ${ii}"
        }
    }

    parallel build_tasks
}

node { 
    lotsParallel() 
    stage 'NoSteps' 
} 
```

If this can be updated, ATH still happy etc - lets get this in!

@scherler nice one - also - you didn't use your usual VERY LONG BRANCH NAMES? is this is a new trend? ;)

scherler:
@michaelneale I interpret that as :bee: ;)

...and yeah but it was because I did not followed "my usual routine" 
- take ticket
- create branch 

;)

michaelneale:
Yes as long as all the merging/publishing and ATH work out - yes that is a :bee:

PR #908, opened 2017-03-16
bug/jenkins 42784 github creation single repo optimization changes

cliffmeyers:
# Description
- See [JENKINS-42784](https://issues.jenkins-ci.org/browse/JENKINS-42784).
- Currently if the user creates a single repo Github org folder that is empty or contains no Jenkinsfiles, there are no events sent back to the UI and it shows the very generic "pipeline creation pending" state (aka "timeout" state).
- This changes the logic so that org folder scanning is initiated in all cases, so that the "job_orgfolder_indexing_result" will propagate back to the UI and trigger the "create pipeline" state. 
- There are some concerns about this being wasteful that are totally legit, but trying to work within the limitations exposed by the upstream APIs.
- One other small advantage to this change is that the org folder now properly displays the Github org avatar (in classic). The prior code that was skipping the `organizationFolder.scheduleBuild()` call for single repo cases resulted in an org folder with a generic Github icon.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
+++ b/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
@@ -117,11 +117,16 @@ public class GithubPipelineCreateRequest extends AbstractPipelineCreateRequestIm
                 OrganizationFolder organizationFolder = (OrganizationFolder) item;
                 organizationFolder.getNavigators().replace(gitHubSCMNavigator);
 
+                // for single repo scenario, aggressively scan just that repo to trigger 'job_multibranch_indexing_result' sooner
+                // note this will only occur if that repo actually contains a Jenkinsfile
                 if(repos.size() == 1){
                     SCMSourceEvent.fireNow(new SCMSourceEventImpl(repos.get(0), item, apiUrl, gitHubSCMNavigator));
-                }else {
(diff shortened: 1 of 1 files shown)


cliffmeyers:
cc @vivek @kzantow @michaelneale @i386 

Some folks have reservations about this approach and rightly so. If we discover that it's too wasteful or leads to strange inconsistent state in Jenkins then we can abandon it, and just change the UI to assume to show the "create pipeline" state when the event timeout is reached.

kzantow:
I think this basically just undoes the optimization @vivek was trying to do...

michaelneale:
@kzantow does it - it still fires the "shortcut" event it looks like? I'll give it a try.

michaelneale:
This worked well for me with both the blank case and adding subsequent pipelines. Did not seem to slow it down. 

So assuming no one has objections  - :bee: from me (ATH was happy as of time of writing).

kzantow:
I looked at this again, I think in the create case, we need to trigger the normal org scan, anyway otherwise it gets created incorrectly, without some of the data from github like the display name. I think we should just get rid of the single event for the create case, then it looks like they both would ultimately work correctly.

michaelneale:
@kzantow isn't that what this is doing? Without the single event it is 60s to create a single repo still?

michaelneale:
I believe after a discussion earler with Stephen C, @kzantow and @vivek were goign to look at some other approaches - is that fair to say? 

cc @cliffmeyers

kzantow:
@michaelneale yes, going to see about adapting what we talked about earlier, probably can just close this PR

PR #909, opened 2017-03-16
Adding some classNames to ExtensionRenderer

sophistifunk:
Something I need in order to do the table migration nicely; add some classNames to ExtensionRenderer and the ability to specify additional classes from client code.

# Description

See [JENKINS-41833](https://issues.jenkins-ci.org/browse/JENKINS-41833).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/js-extensions/spec/ExtensionRenderer-spec.js
+++ b/js-extensions/spec/ExtensionRenderer-spec.js
@@ -77,7 +77,7 @@ describe('ExtensionRenderer', function () {
         assert.isTrue(result.is('ExtensionRenderer'), 'should be ExtensionRenderer');
         assert.equal(result.length, 1, 'length');
         assert.equal(result.children().length, 0, 'children.length');
-        assert.equal(result.html(), '<div></div>', 'html');
+        assert.equal(result.html(), '<div class="ExtensionPoint foo-bar-baz"></div>', 'html');
         // Fixme: ^^^^ figure out how to test the rendered element name other than html() string comparison
     });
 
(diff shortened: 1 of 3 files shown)


js-extensions/spec/ExtensionRenderer-spec.js
```diff
@@ -77,30 +77,35 @@ describe('ExtensionRenderer', function () {
         assert.isTrue(result.is('ExtensionRenderer'), 'should be ExtensionRenderer');
         assert.equal(result.length, 1, 'length');
         assert.equal(result.children().length, 0, 'children.length');
-        assert.equal(result.html(), '<div></div>', 'html');
+        assert.equal(result.html(), '<div class="ExtensionPoint foo-bar-baz"></div>', 'html');
```
kzantow: huh? at least better naming...

sophistifunk:
Huh? Context? Not outputting divs? I am confuse.

scherler:
@kzantow I not sure whether I understand you neither. As I understand the change is that we now allow and add a default class to the wrapping element. The Renderer still does not know about the context.

js-extensions/src/ExtensionRenderer.jsx
```diff
@@ -82,7 +82,22 @@ export default class ExtensionRenderer extends React.Component {
...
+        
+        if (className) {
+            classNames.push(className);
+        }
+
+        const newProps = {
+            className: classNames.join(' ')
+        };
+
+        return React.createElement(this.props.wrappingElement, newProps, newChildren);
```
scherler: I personally would have extracted `this.props.wrappingElement` as well in the earlier statement, but I guess personal preference

michaelneale:
OH this is interesting... could this possibly mean we could flick a switch and highlight extension points? (or can easily spot them anyway)? That seems a rather good idea, but I am sure there were other reasons. 

I am curious what @kzantow redacted now... I bet it was something saucy. :bee: as others seem happy!

ATH is also happy:

kzantow:
@michaelneale it was the fact I don't think we should be adding any HTML for extension points (it's currently a requirement because they are react components), this is furthering that idea. Although, it was in some incomprehensible english. It really doesn't matter too much, but the extensions would typically be used in some context... regarding styling, I'd bake in some special rendering mode instead of just the css.

js-extensions/src/ExtensionRenderer.jsx
sophistifunk: Nice catch, churbrew.

sophistifunk:
Just bumped the version no and made Thorsten's suggested change, but I'll wait til green before merge :)

michaelneale:
rebuilding as npm is a stupid piece of craaaaaaaaap and is SHOCKED by a download that didn't work.

michaelneale:
@kzantow oh I just thought this may be a nice side benefit of this.

PR #912, opened 2017-03-17
Task/jenkins 42123 GitHub access token progress

cliffmeyers:
# Description
- See [JENKINS-42123](https://issues.jenkins-ci.org/browse/JENKINS-42123).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/creation/github/Button.jsx
+++ b/blueocean-dashboard/src/main/js/creation/github/Button.jsx
@@ -67,7 +67,7 @@ export class Button extends React.Component {
         } = this.props;
 
         const { result } = this.state;
-        const disabled = result === 'running';
+        const disabled = result === 'running' || this.props.disabled;
         const statusClass = result ? 'Button-status' : '';
         const transitionClass = this.state.transition ? 'Button-transitioning' : '';
 
@@ -100,6 +100,7 @@ Button.propTypes = {
     className: PropTypes.string,
     style: PropTypes.object,
     children: PropTypes.string,
+    disabled: PropTypes.bool,
     status: PropTypes.shape({
         result: PropTypes.string,
         reset: PropTypes.bool,
--- a/blueocean-dashboard/src/main/js/creation/github/Button.jsx
+++ b/blueocean-dashboard/src/main/js/creation/github/Button.jsx
@@ -49,7 +49,7 @@ export class Button extends React.Component {
             });
         }
 
-        // schedule a reset of internal "state.result"
+        // schedule a reset of internal "state.result" to restore button text
         if (statusChanged && shouldReset) {
             this.timeoutId = setTimeout(() => {
                 this.setState({
--- a/blueocean-dashboard/src/main/js/creation/github/Button.jsx
+++ b/blueocean-dashboard/src/main/js/creation/github/Button.jsx
@@ -14,44 +14,49 @@ export class Button extends React.Component {
 
         this.state = {
             result: null,
-            transition: false,
+            pendingReset: false,
         };
     }
 
     componentWillReceiveProps(nextProps) {
-        // TODO: probably need to check against state's result instead of props?
         const statusChanged = !this.props.status || !nextProps.status ||
-            this.props.status.result !== nextProps.status.result;
-        const resetAfterDelay = nextProps.status && nextProps.status.reset;
+            this.props.status.result !== nextProps.status.result ||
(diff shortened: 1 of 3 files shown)


cliffmeyers:
github-creation-token-progress

cliffmeyers:
@brody lmk what you think. I added an error state for the scenario where the access token is invalid.

michaelneale:
Looks good - @brody says that if you could drop the "X" on the button when has failed. There is enough feedback on the text field, so no need for that - but otherwise go for it (as long as build happy etc)

cliffmeyers:
@michaelneale all right, I've removed the "X" state, although it was a useful exercise that helped make the Button API more flexible.

cliffmeyers:
ATH passed:  https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/master/1869/pipeline/17

cliffmeyers:
@sophistifunk maybe worth having you take a look, as this is a new `Button` component in dashboard that we might eventually port into JDL and merge with other `IconButton` functionality.

sophistifunk:
I'm not sure I understand the point of the button? Why does it have a reset parameter, which then erases the status? What happens if you set the status to success but not the reset parameter? Why does this weird logic need to go into a button component rather than a container that either shows some sort of message or a button?

michaelneale:
:bee: from me

cliffmeyers:
@sophistifunk fair questions :) So I'd like our behavior in forms to be as consistent as possible, without a lot of boilerplate to accomplish it.

Without any components, I'd have to do the following within `GitCredentialsStep` to accomplish the desired UX:
1. Render a normal button w/ text by default,
2. While in "pending" state, hide the text and render a spinner icon in the button,
3. On error, flip the button back to just text,
4. On success, flip the button to a "success" icon and leave it as such.

It'd be nice not to have to manage that every place we need it, but I take your point that maybe a "smart button" isn't the best way to do it.

You make a good point about `reset` too: when I first built this, I wanted to give some error feedback in the button, but then restore the button text after a small delay. So for 3, it originally would display the error "X" for 2s and then flip back to the default button text. Having to manage that w/ a timer inside of the `GitCredentialsStep` seemed crap to me, and it would be tedious to have to impl that same logic in every place. Ultimately Brody said it wasn't necessary, so I removed it from the step, but not from the button impl itself.

So, all that said, I'm open to a different approach... maybe a container component that can hold every different buttons and switch between them? Or maybe it could even be further formalized with a `Form` component and some wrapper around a submit button. I'm totally open to ideas, but I would hate to implement all this logic in the step and not be able to reuse it since it seems a pattern for forms that will come up a lot.

michaelneale:
@sophistifunk @cliffmeyers the editor may be another place a similar button would be used. It has a bunch of action buttons that do "expensive" things that currently use the worm to indicate network traffic (which is fine) but could do with similar behavior to this. Not saying we should do that prematurely, or even if a component isn't overkill, but there are other places for this, (input may be another one).

blueocean-dashboard/src/main/js/creation/github/Button.jsx
```diff
@@ -0,0 +1,130 @@
...
+            const result = nextProps.status && nextProps.status.result;
+            this.setState({
+                result,
+                pendingReset: shouldReset,
+            });
+        }
+
+        // schedule a reset of internal "state.result" to restore button text
+        if (statusChanged && shouldReset) {
+            this.timeoutId = setTimeout(() => {
```
kzantow: Is there a reason for not using the css transition group?

kzantow:
Although it may not be ideal, we could make this sort of button listen to the `loadingIndicator` and potentially intercept the worm when a button like this is present and clicked. That would reduce a bunch of code dealing with it in specific cases.

cliffmeyers:
@sophistifunk just a ping on this one. More than willing to make changes or even scrap the entire idea, although if we do it'd be good for it to lead somewhere that makes this kind of behavior "easy" so we can use it in many places w/o boilerplate.

sophistifunk:
I say roll with it for now, and when we need to use it somewhere else we'll have a better idea of what to do.

i386:
I love that button.

@cliffmeyers maybe raise a task on jira to extract it and we can revisit later?

cliffmeyers:
@i386 captured here:  https://issues.jenkins-ci.org/browse/JENKINS-43096

PR #923, opened 2017-03-23
Hotfix/loading worm preloader

cliffmeyers:
# Description
- When opening certain pages directly, the loading worm would display on the screen indefinitely. Normally the worm is hidden by the success/failure handler for XHR calls. There are some screens however that get data exclusively from the preloader, which doesn't actually fire XHR calls.
- The change is that we now call "show" for the indicator for any fetch that hasn't opt'd out, and if the data is pulled from the preloader then we immediately make a call to "hide".
- You should be able to repro this scenario by deep-linking to the "Branches" tab for a multi-branch project with no branches, or the "Branches" tab for a non-multibranch project. "PR's" tab should also illustrate it.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
if with and ATH happy - sounds good. 

Are there specific pages to see this on?

michaelneale:
:bee: if you are comfortable with it and @kzantow is

cliffmeyers:
ATH passed: https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/master/2025/pipeline/17

cliffmeyers:
@kzantow mostly showed up in empty state cases. When a screen had data - even if it was provided just by the preloader - there were using some an extra fetch or two afterwards (e.g. for capabilities)  that would trigger hiding of the status indicator.

cliffmeyers:
ATH passed: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/2027/

PR #924, opened 2017-03-24
Pipeline editor -> 0.2.0

kzantow:
See title.


michaelneale:
works for me - as long as ATH doesn't barf :bee: (ran with latest editor fine)

michaelneale:
yes ATH passes so :shipit:

PR #925, opened 2017-03-24
Feature/jenkins 41737 empty state dashboard

cliffmeyers:
# Description
- See [JENKINS-41737](https://issues.jenkins-ci.org/browse/JENKINS-41737).

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code


cliffmeyers:
placeholder-dashboard

cc @brody

i386:
@cliffmeyers this looks good to me

michaelneale:
nice - probably needs a little work to pass the ATH, can look at this monday

i386:
@brody @cliffmeyers is the little pipeline animated?

michaelneale:
Looking good - it won't flash up that empty state if dashbaord takes a while to load will it? as that would be B.A.D ;)

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages.properties
```diff
@@ -68,6 +68,10 @@ home.pipelineslist.header.name=Name
 home.pipelineslist.header.pullrequests=PR
 home.pipelineslist.row.failing={0} failing
 home.pipelineslist.row.passing={0} passing
+home.placeholder.linkhref=/create-pipeline
```
michaelneale: will this really be localised? I don't think it will be for the time being. We dont have l10n of routes at the moment anyway

brody:
@cliffmeyers @i386 Having it animate might be a little distracting. Happy to have it static and possible look to add animation down the line.

i386:
I was just asking rather than wanting :)

michaelneale:
I updated this to master, and fixed some eshit violations, and checked it passed the ATH: 

Passes ATH: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/1957/

And looks good to me

:bee:

This is a nice addition to polish

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages.properties
cliffmeyers: I agree that internal routes should `not` be in the i18n bundles. I will fix this in the next empty state PR.

For external URL's to Jenkins docs, is it fair to leave those in? It seems like those might get i18n's within a reasonable time frame, and it does seem like a "UI string" that someone might want to change easily without mucking through the React code.

michaelneale: yes external ones are fine I think.

cliffmeyers: Fixed this in 2d5a809 in PR #929

PR #926, opened 2017-03-27
[JENKINS-42872] Use input step to check submission rights

imeredith:
# Description

See https://issues.jenkins-ci.org/browse/JENKINS-42872

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
@@ -272,18 +272,7 @@ public class PipelineStepImpl extends BluePipelineStep {
         }
     }
     private boolean canSubmit(InputStep inputStep){
-
-        Authentication a = Jenkins.getAuthentication();
-        String submitter = inputStep.getSubmitter();
-        if (submitter==null || a.getName().equals(submitter)) {
-            return true;
-        }
-        for (GrantedAuthority ga : a.getAuthorities()) {
-            if (ga.getAuthority().equals(submitter)) {
-                return true;
-            }
-        }
-        return false;
+        return inputStep.canSubmit();
     }
 
     @Override


blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
```diff
@@ -272,18 +272,7 @@ private Object convert(String name, ParameterValue v) throws IOException, Interr
...
-        if (submitter==null || a.getName().equals(submitter)) {
-            return true;
-        }
-        for (GrantedAuthority ga : a.getAuthorities()) {
-            if (ga.getAuthority().equals(submitter)) {
-                return true;
-            }
-        }
-        return false;
+        return inputStep.canSubmit();
```
michaelneale: Do you know for sure this honours the currently logged in users permissions somehow?

michaelneale:
cc @vivek  need your input (no pun intended!)

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
imeredith: Yeah i checked the source code. Its basically the same as what was replaced with some extra stuff.

michaelneale:
I am ok with this, but slightly worried about the deprecation warning, and why @vivek did it that other way in the first place, lets see what he says. 

cc @jglick any idea what is mean't to happen to check input submit permissions (that isn't a deprecated call).

michaelneale:
hey @i386 I know you tested this in some non trivial way before, any chance you would find time to try this change for input permissions?

i386:
Seems to work for me.

michaelneale:
OK lets check with @vivek before merging this in. This works well for me too.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
vivek: InputStep.canSubmit() is marked `@Deprecated`, these methods have been moved over to InputStepExecution, but they are privates so can't use. I thought its better to copy that code instead of using deprecated ones in  InputStep as its going to lag behind vs how it evolves in InputStepExecution. Right fix is to get them made public in InputStepExecution and use it directly. 

I am fine using deprecated ones in the meantime.

PR #930, opened 2017-03-28
[JENKINS-38460] Move toaster from bottom left to bottom right

ModuloM:
# Description

See [JENKINS-38460](https://issues.jenkins-ci.org/browse/JENKINS-38460).

Warning: depends on https://github.com/jenkinsci/jenkins-design-language/pull/161.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/less/connection-lost.less
+++ b/blueocean-dashboard/src/main/less/connection-lost.less
@@ -3,7 +3,7 @@
         .toast {
             position: absolute;
             bottom: 16px;
-            left: 16px;
+            right: 16px;
             padding: 16px;
             border-left: 5px solid;
 
--- a/blueocean-web/src/main/less/toaster.less
+++ b/blueocean-web/src/main/less/toaster.less
@@ -3,14 +3,19 @@
     position: fixed;
     z-index: 4000;
     bottom: 0;
-    left: 0;
+    right: 0;
     width: 400px;
     height: 210px;
-    padding-left: 10px;
+    padding-right: 10px;
     pointer-events: none;
     background: none !important;
 
     .toast {
         pointer-events: all;
+    }
+
+    /* override @jeninks-language/design-language/component/toast.less */
+    & > span {
+        align-items: flex-end;
     }
 }


i386:
Nice job and thanks for contributing!

Could you add a screenshot here just so I can usually confirm please?

michaelneale:
I'll spin this up later and show...

michaelneale:
ATH passes: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/1975/

michaelneale:
Here is what it looks like: 

screen shot 2017-03-29 at 6 06 12 pm

i386:
LGTM 🙌

michaelneale:
Does this really require the PR to the JDL? As it works just fine without it, as it is in this PR alone. I am confuse.

ModuloM:
Initially there was a case where the toaster wasn't exactly to the right. This happened when restarting a pipeline for example. 
First I had overwritten the CSS rule in toaster.less. But it seemed like a hack to me. So I have submitted an other PR in JDL repo to properly change the rule.
Maybe I'm wrong or maybe you don't have the last modified version  of this PR.
Hope this helps.

brody:
🐝

cliffmeyers:
@michaelneale this raises an interesting point that contributors won't be able to publish a beta JDL so that those changes can be easily incorporated into PR's against blueocean-plugin.

@ModuloM Right now there is a PR with `align-items: flex-end` (right-aligned) in the JDL, and then CSS in this PR that places the `Toaster` on the right side of the screen. I think we want to ensure that the CSS which handles `Toaster` alignment is in the same repo.

I think perhaps we ought to close jenkinsci/design-language#161 and add the `align-items` CSS in this PR. You should be able to override it in `blueocean-web` by default, or can use a more specific selector if necessary.

michaelneale:
@cliffmeyers yes good Q on JDL - I guess with any future solution we would have to treat it closer to a normal plugin? (I am not sure if that is the intent of the JDL though...). Not sure of the longer term solution to this...

ModuloM:
Hi,

@cliffmeyers Ok, I have pushed an update that overwrite the JDL behavior. In this case, jenkinsci/jenkins-design-language#161 can be close. Do you want me to do it?

cliffmeyers:
Tested this locally and looks good!

@ModuloM thank you for the contribution. I am merging this now. I also closed the JDL PR earlier today.

PR #932, opened 2017-03-29
Trigger the run for docker hub daily

i386:
@rtyler is it OK for us to run this daily?


rtyler:
This has been picked up by trusted.ci

michaelneale:
nice one.

PR #935, opened 2017-03-31
Bug/jenkins 43217 run details modal madness

cliffmeyers:
# Description
- See [JENKINS-43217](https://issues.jenkins-ci.org/browse/JENKINS-43217).
- Has matching ATH jenkinsci/blueocean-acceptance-test#133

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/RunDetails.jsx
+++ b/blueocean-dashboard/src/main/js/components/RunDetails.jsx
@@ -65,6 +65,7 @@ class RunDetails extends Component {
     componentWillMount() {
         this._fetchRun(this.props);
         this.opener = locationService.previous;
+        this.initialHistoryLength = history.length;
     }
 
     componentWillReceiveProps(nextProps) {
@@ -137,10 +138,11 @@ class RunDetails extends Component {
     afterClose = () => {
         const { router, params } = this.context;
         if (this.opener) {
-            router.goBack();
+            const offsetToInitialRoute = this.initialHistoryLength - history.length;
+            router.go(offsetToInitialRoute - 1);
         } else {
             const fallbackUrl = buildPipelineUrl(params.organization, params.pipeline);
-            router.push(fallbackUrl);
+            router.replace(fallbackUrl);
         }
     };
 
--- a/blueocean-dashboard/src/main/js/components/RunDetails.jsx
+++ b/blueocean-dashboard/src/main/js/components/RunDetails.jsx
@@ -138,9 +138,11 @@ class RunDetails extends Component {
     afterClose = () => {
         const { router, params } = this.context;
         if (this.opener) {
+            // step back in the history to the item that's prior to initial load of RunDetails
             const offsetToInitialRoute = this.initialHistoryLength - history.length;
             router.go(offsetToInitialRoute - 1);
         } else {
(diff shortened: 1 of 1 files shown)


blueocean-dashboard/src/main/js/components/RunDetails.jsx
```diff
@@ -137,10 +138,13 @@ class RunDetails extends Component {
     afterClose = () => {
         const { router, params } = this.context;
         if (this.opener) {
-            router.goBack();
+            // step back in the history to the item that's prior to initial load of RunDetails
+            const offsetToInitialRoute = this.initialHistoryLength - history.length;
+            router.go(offsetToInitialRoute - 1);
```
kzantow: why not keep popping off history elements until we're away from the 'modal' ? damn modal...

kzantow: the problem is the same as tracking the opener. you could navigate around, close the modal because you navigate to the editor for example, then go back to the modal and all of a sudden the `initialHistoryLength` is recalculated, since it remounted...

cliffmeyers: What API exactly are you thinking of calling? You can't read from the browser history. We could create our own history array in the `LocationService`; I'm not opposed to that.

kzantow: ok, so maybe we need this. a prior app i worked on had a 'navigation stack', maybe we maintain that separately from the router, we could easily go back in these sorts of cases. it actually worked well, allowed for breadcrumbs and such

cliffmeyers: I'm not fully up to speed on all the navigation flows involving in the editor. Above, are you describing the following flow?

1. User opens modal for a run,
2. User clicks "edit" to navigate to editor,
3. User presses back button

Need to make sure I test more wacky scenarios with this flow. After talking to @michaelneale I'm also going to file a ticket for us to consider the design of this modal and how it works. These interactions are causing a tremendous amount of development and testing pain.

michaelneale: damn modal indeed...

michaelneale:
Nice one: running ATH for the change: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/bug%252FJENKINS-43217-run-details-modal-madness/

passes with main ATH already - so this is ok with me. @kzantow your call? this is an annoying problem, and we need a bigger discussion on how to rework this hotmess of modal that isn't modal anyway, but for now - this seems a small, useful change to me. :bee: from me.

michaelneale:
I tried this out in a few ways with the editor as well and it seemed just fine.

michaelneale:
yes it does work. :bee: from me at least for funcitonality. Nice.

cliffmeyers:
@michaelneale sorry I missed your comment the other day. It does work, and there is an existing ATH test to cover that scenario.

michaelneale:
:bee:

cliffmeyers:
Going to merge this now, as we have some ideas for larger router improvements that will take a while, and this bug is IMO nasty enough that it needs a fix sooner than that. Thanks all.

PR #937, opened 2017-03-31
Fix dialog form text color for first time run

kxepal:
Currently, if project has no builds, the Dialog class inherits text
color from class:

    .empty-state .empty-state-container .empty-state-content

Which is white now. And it happens for dialog form to have white
text on white background.

This change ensures that the Dialog always has readable text color
for the white background.

Before: http://i.imgur.com/lciqAyK.png
After: http://i.imgur.com/ZLvhclUl.png

### DIFF
--- a/blueocean-dashboard/src/main/less/forms/index.less
+++ b/blueocean-dashboard/src/main/less/forms/index.less
@@ -74,6 +74,7 @@
             margin: 24px 32px 16px 32px;
         }
 
+        color: #4A4A4A;
         min-width: 450px;
 
         * > .FormElement {


i386:
Thanks for this! Hmm, I think theres something more strange going on there - the buttons aren't styled correctly. Can you provide some reproduction steps please?

kxepal:
Sure! Steps to reproduce:
1. Create a new project. In my case it's Pipeline one, but shouldn't matter.
2. Make build parameterized. Add any parameter with a text description. The rest options are irrelevant.
3. Make sure there were no builds for this project made. Otherwise form will inherit right styles from build log.
4. Open project in Blue Ocean and hit the button Run. You should see no description text in dialog form, but you can select it to make visible for you.

i386:
@kxepal so I think something strange is going on here - the button styles on the form haven't loaded. While the change proposed superficially fixes some of it I'd say there is probably something very odd going on that is the root cause. Ive raised [JENKINS-43259](https://issues.jenkins-ci.org/browse/JENKINS-43259) to track this issue.

I favourited my parameterised job and clicked run on the dashboard - and got it to display correctly. Could you try the same on your server? Perhaps the style is not getting applied on the Activity tab where you are running your job from.
jenkins blue ocean 2017-03-31 16-07-00

kxepal:
You miss step 4. You need to navigate to project page and click on Run button there. Notice a shoes on my screenshots. From main page it will works fine because of the same reason: dialog form will inherit colors from backlog.

But you're right, buttons lost their styles as well. Hm. I can pick their styles to Dialog class as well, but I suppose there should be some more solid solution and doesn't involves copy-paste style. I fear, I don't know less such good to make that /:

i386:
Right so it seems things aren't loading there. Yeah, probably need to get one if he developers involved. I mentioned them on the JIRA ticket so perhaps one of them knows what's up.

kxepal:
Thank you! I'm closing this PR them.

PR #938, opened 2017-03-31
Hotfix/savingorgs

michaelneale:
URGENT fix for saving github orgs correctly. as it is now, they are not persisted so a restart loses the pattern in the github org folder. 
"pattern" is the pattern field in the classic config screen for the github org folder (click on config, and look for the pattern field, which determines what repos make pipelines in this organisation).

No ticket as hard to describe and I am too tired. 

To test this:: 

```
Create a single repo, check pattern in multibranch folder config that it has that single repo
Restart, check it is still in folder pattern
Add a repo, check folder pattern has new one and previous one
Restart, check both in pattern still
Switch to discover, check pattern is ".*"
restart, check pattern is still ".*"
Switch back to single, check that the one you picked is in the pattern, along with previous ones
Restart, check pattern is the same as before you retsarted

Remove whole org folder 
Add as auto discover, check pattern
restart, check pattern
```
If this sniffs ok, we need to cherry pick it to the release branch urgently.

### DIFF
--- a/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
+++ b/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
@@ -179,8 +179,10 @@ public class GithubPipelineCreateRequest extends AbstractPipelineCreateRequestIm
                         }
                     });
                 }else {
+                    gitHubSCMNavigator.setPattern(".*");
                     organizationFolder.scheduleBuild(new Cause.UserIdCause());
                 }
+                organizationFolder.save();
                 return githubOrganizationFolder;
             }
         } catch (Exception e){
--- a/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
+++ b/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
@@ -148,6 +148,13 @@ public class GithubPipelineCreateRequest extends AbstractPipelineCreateRequestIm
                         }
                     }
 
+                    // Add any existing discovered repos
+                    for (MultiBranchProject<?,?> p : organizationFolder.getItems()) {
+                        if (!repos.contains(p.getName())) {
+                            repos.add(p.getName());
+                        }
+                    }
+
                     if (credentialId == null) {
                         credentialId = gitHubSCMNavigator.getScanCredentialsId();
                     }
--- a/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineUpdateRequest.java
+++ b/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineUpdateRequest.java
@@ -55,6 +55,7 @@ public class GithubPipelineUpdateRequest extends BluePipelineUpdateRequest {
                 folder.scheduleBuild(new Cause.UserIdCause());
             }
         }
+        item.save();
         return pipeline;
     }
 


michaelneale:
this passes the ATH: https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/2022/ 

and locally looks good with the release branch when cherry picked and manually tested with above steps.

vivek:
@michaelneale LGTM, but this code got updated to serve update case as well by @kzantow, lets wait for him to review the changes.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
```diff
@@ -148,6 +148,13 @@ public BluePipeline create(Reachable parent) throws IOException {
                         }
                     }
 
+                    // Add any existing discovered repos
```
michaelneale: Without this, if you switch from "discover all" to "add a single pipeline" - it actually wipes out all other discovered pipelines. This simply iterates through the existent ones, and ensures they end up added to the pattern expression - this matches up with the description on the GUI which is the correct design intention.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
```diff
@@ -179,8 +186,10 @@ public boolean meetsIndexingCriteria() {
                         }
                     });
                 }else {
+                    gitHubSCMNavigator.setPattern(".*");
```
michaelneale: Without this, you can never set it to "discover all"

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
i386: Whoa very good catch

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineUpdateRequest.java
```diff
@@ -55,6 +55,7 @@ public BluePipeline update(BluePipeline pipeline) throws IOException {
                 folder.scheduleBuild(new Cause.UserIdCause());
             }
         }
+        item.save();
```
michaelneale: and again:

michaelneale:
@kzantow @cliffmeyers I am pretty happy with my extensive testing on both this branch, and cherry picking the 2 commits to release. I think the fixes are clear, but a double check and code review appreciate. 

We need to get this into master AND cherry picked to release - and check the release build runs (and its ATH branch runs after it ok) on friday, as this is pretty important !

i386:
Good find @michaelneale - you saved our collective asses here. I haven't manually verified this (I will in the morning) but code wise LGTM.

@vivek @kzantow we do need some tests written here ASAP. I realise we should cannot mock out Github in a short amount of time. However, we can write a test to ensure that the patterns are updated in the correct way for the scenarios and that we always call `item.save()` without having Github mocked out.

Scenarios that should be covered:
* Creating the org folder initially with a single pipeline
* Adding an additional Pipeline 
* Creating the org folder with auto discover
* Switching to/from pipeline/discover

michaelneale:
@i386 I have no idea how to even possible test the fact that it saved things correctly ... but still can test a lot more.

i386:
@michaelneale this stuff is mockable with Mockito and a few helpful abstractions.

i386:
giphy 1

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineUpdateRequest.java
cliffmeyers: Question for @kzantow: is this class still being used? I thought all operations had been unified under "create" so perhaps best to delete this?

michaelneale:
awaiting @kzantow to bless and merge and cherry pick and an appropriate meme

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineUpdateRequest.java
michaelneale: good point, when I was debugging (and by debuggin I mean print statements with swears in them) I never saw this one called. only put it in there for completeness.

vivek: item.save() is not needed, in fact should be removed - extra cost of saving, since folder.getNavigators().replace(gitHubSCMNavigator); does save underneath.

vivek: Maybe we should but better to check with @kzantow, last time I discussed with @kzantow on this, he thought there might be some use cases where we want to update existing orgs outside of create flow.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
vivek: Should be moved in the else condition up there at https://github.com/jenkinsci/blueocean-plugin/blob/da60f3b1903de17b5008e60b0e76633a00aaca5d/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java#L164.

```
                if (sb.length() > 0) {
                    gitHubSCMNavigator.setPattern(sb.toString());
                }else{
                    gitHubSCMNavigator.setPattern(".*");
                }
```
This will also avoid doing extra save you have added in this change. It's optimization of code and will ensure save happens before scheduleBuild() is called  - save happens in `organizationFolder.getNavigators().replace(gitHubSCMNavigator);`. Your code will work fine, its more of optimization that we can do later given its all been tested already. Up to @kzantow as I might be hitting road in couple hours.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineUpdateRequest.java
kzantow: Right, maybe down the road, so I just left this method but it's currently unused.

kzantow:
@i386 @michaelneale @vivek I'll work today getting some tests written for this; don't have a lot else urgent on my plate AFAIK

kzantow:
FYI this has been cherry-picked into the release branch now, I'm going to build it and do some more testing there as well as see if I can add some tests.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
michaelneale: sure - not really worried about optimisation here as not exactly a warm code path!

With testing we can rework this, but yeah as you said, laborious to test!

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineUpdateRequest.java
michaelneale: I needed the save() in teh create thought right? (It certainly worked better when I added it). 

yes there will be an update one day I guess - to change some other settings.

michaelneale:
@kzantow great - yes some extra hand testing and then working on automated tests would be well worth while use of time at this stage.

i386:
Thanks @kzantow :)

PR #947, opened 2017-04-06
[FIXED JENKINS-42636] Fix duration and start time logic

abayer:
# Description

See [JENKINS-42636](https://issues.jenkins-ci.org/browse/JENKINS-42636).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
@@ -567,7 +567,7 @@ public class PipelineNodeGraphVisitor extends StandardChunkVisitor implements No
 
         long duration = 0;
         long pauseDuration = 0;
-        long startTime = 0;
+        long startTime = System.currentTimeMillis();
         TimingAction timingAction = null; //= parallelStartNode.getAction(TimingAction.class);
         if(timingAction != null){
             startTime = timingAction.getStartTime();
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
@@ -274,9 +274,13 @@ public class PipelineNodeGraphVisitor extends StandardChunkVisitor implements No
                     status = new NodeRunStatus(genericStatus);
                 }
             }else{
-                times = new TimingInfo(TimingAction.getStartTime(branchStartNode)+System.currentTimeMillis(),
+                long startTime = System.currentTimeMillis();
+                if (branchStartNode.getAction(TimingAction.class) != null) {
+                    startTime = TimingAction.getStartTime(branchStartNode);
+                }
+                times = new TimingInfo(System.currentTimeMillis() - startTime,
                         chunk.getPauseTimeMillis(),
-                        TimingAction.getStartTime(branchStartNode));
+                        startTime);
                 status = new NodeRunStatus(BlueRun.BlueRunResult.UNKNOWN, BlueRun.BlueRunState.RUNNING);
             }
 


michaelneale:
thanks @abayer giving this a spin now. this one is fickle - so will see if it pops up. If it passes the gauntlett of tests and @vivek says ok - worth a shot to see if it solves for all.

michaelneale:
functionally seems ok to me - see what ATH and @vivek say....

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
```diff
@@ -274,9 +274,13 @@ public void parallelStart(@Nonnull FlowNode parallelStartNode, @Nonnull FlowNode
                     status = new NodeRunStatus(genericStatus);
                 }
             }else{
-                times = new TimingInfo(TimingAction.getStartTime(branchStartNode)+System.currentTimeMillis(),
+                long startTime = System.currentTimeMillis();
```
vivek: @abayer Doesn't look like right fix to me. This can be misleading, that is branchStartNode.getAction(TimingAction.class) will be null only in cases when this branch node has not started executing yet. Only case I see it working is if the assumption that if branchStartNode is available (branch endTime is not) that means its already started executing. Don't think this is true. This is situation where startBranchNode is created but it's timing info is not recorded yet, maybe it's waiting for executor to be available or just delay in writing to file system.

startTime of null or 0 feels like correct behavior, it tells client that we do not know it's start time vs giving current time without knowing its actually executing.

abayer: See the JIRA - the `TimingAction` is only missing for a very brief window between creation of the `FlowNode` and attaching the `TimingAction`, always, so far as I can tell. @svanoort - does that sound right?

abayer: A little more detail - https://github.com/jenkinsci/workflow-job-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/job/WorkflowRun.java#L895 gets called from https://github.com/jenkinsci/workflow-cps-plugin/blob/92019ea288775c5d2e357f701472b52e40a7cc99/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsThreadGroup.java#L383, right after the node gets created. So the `TimingAction` should only be null during that tiny window between node creation and the action being added. I don't see any other way it could be null.

vivek: Ok, looks fine🐝  Let's wait for @svanoort to confirm though.

michaelneale:
paging Dr @svanoort - will the patient die with this treatment?

scherler:
@abayer sehr gut, sieht gut aus. :bee:

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
svanoort: That's reasonable.  I think there's a problem here if start time is null but end time isn't -- but this fixes the B.O. logic at least.

svanoort: And I think in general the issue would be some sort of transitory condition, so not too concerned.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
```diff
@@ -563,7 +567,7 @@ protected String getTypeDisplayName() {
 
         long duration = 0;
         long pauseDuration = 0;
-        long startTime = 0;
+        long startTime = System.currentTimeMillis();
```
svanoort: Oof.  Yeah, never assume startTime = 0 -- this is NOT the first bug I've seen from this exact statement (in other places).

michaelneale:
@vivek you want to get this one in?

PR #949, opened 2017-04-06
Bug/jenkins 41331 favorites stack missing animation

cliffmeyers:
# Description
- See [JENKINS-41331](https://issues.jenkins-ci.org/browse/JENKINS-41331).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-personalization/src/main/js/components/DashboardCards.jsx
+++ b/blueocean-personalization/src/main/js/components/DashboardCards.jsx
@@ -18,6 +18,26 @@ import { PipelineCard } from './PipelineCard';
 
 const t = i18nTranslator('blueocean-personalization');
 
+function CardStack(props) {
+    const { cards, message } = props;
+    return (
+        <div key={message} className="favorites-card-stack">
+            <div className="favorites-card-stack-heading"> {message}</div>
+            <TransitionGroup transitionName="vertical-expand-collapse"
+                             transitionEnterTimeout={300}
+                             transitionLeaveTimeout={300}
+            >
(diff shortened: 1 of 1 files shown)


cliffmeyers:
cc @sophistifunk @scherler @kzantow @imeredith @tfennelly as this root cause here wasn't obvious and was a learning for me.

The cause of the bug was that ReactCSSTransitionGroup was unmounting and remounting with each addition or removal of a favorite, so the enter and leave anims were never being played. That part wasn't too hard to figure out once I debugged into RCSSTG. The true fix for this bug is in 1da79ae. 

The reason this happened is because a component was being defined within the render method. This means each time `DashboardCards.render` is called, a new function is created. The React reconciler will treat them a different element types, therefore doing the unmount/remount dance, rather than just reusing the same element instances and passing in new props.

I would argue that we should avoid this style of "inline components" as it 1. leads to non-obvious behavior like this, and 2. probably isn't the best thing for perf since it's triggering a spurious unmount and remount. I am not sure if that guarantees a DOM destroy and recreate in all cases but in this case I definitely see a full destroy/create in the element inspector.

michaelneale:
works nice for me - as long as everyone else happy - :bee:

scherler:
:bee: Nice find and nice explanation. I second the no inline component!!!

cliffmeyers:
ATH passed:  https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/master/2076/pipeline/17

PR #952, opened 2017-04-07
[JENKINS-43375-Implement_extension_point_on_dashboard_to_add_addition…

scherler:
…al_data_like_navigation_dropdown] create new extensionPoint jenkins.pipeline.header and pass current implementation as child

# Description

See [JENKINS-43375](https://issues.jenkins-ci.org/browse/JENKINS-43375).

# example
screenshot from 2017-04-08 22-11-50

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code


blueocean-core-js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@jenkins-cd/blueocean-core-js",
-  "version": "0.0.95-SNAPSHOT",
+  "version": "0.0.95-brand",
```
batmat: for my culture, what does `-brand` mean? My google-fu does not seem to help.

blueocean-core-js/src/js/components/ContentPageHeader.jsx
```diff
@@ -26,7 +27,9 @@ export const SiteHeader = props => {
     return (
         <BasicHeader className="ContentPageHeader">
             <TopNav>
-                <BlueLogo href={props.homeURL} />
+                <Extensions.Renderer extensionPoint="jenkins.header.logo">
+                    <BlueLogo href={props.homeURL} />
```
batmat: for my culture, again, or can you point to some docs about it: IIUC, the tag in the middle is the one rendered if no extension is found? Like a default value?

blueocean-dashboard/src/main/js/components/Pipelines.jsx
```diff
@@ -61,13 +59,15 @@ export class Pipelines extends Component {
...
                 <ContentPageHeader>
                     <div className="u-flex-grow">
-                        <h1>
-                            <Link to="/" query={ location.query }>
-                                { translate('home.header.dashboard', { defaultValue: 'Dashboard' }) }
-                            </Link>
-                            { organization && ' / ' }
-                            { organization && orgLink }
-                        </h1>
+                        <Extensions.Renderer extensionPoint="jenkins.pipeline.header">
```
batmat: ok, seems like this is it reading this. Interesting.

i386:
@scherler extension point location LGTM

blueocean-core-js/src/js/components/ContentPageHeader.jsx
michaelneale: I would like to know this too (!) but I think you are right.

michaelneale:
Seems to work as advertised, I need @kzantow to review the code though. @batmat had a few questions above, but looking good!

blueocean-core-js/src/js/components/ContentPageHeader.jsx
scherler: Well I guess there is no docu and this. I looked at the code to find out ;)

What happens here is that yes @batmat some kind of fallback but every extensionPoint implementation will get the the `<BlueLogo href={props.homeURL} />` as props.children passed to it. So it can decide what to do with it. In case there are no implementors the `Ext.Renderer` will render those children.

blueocean-core-js/package.json
scherler: That is the new version of the plugin. We use https://docs.npmjs.com/getting-started/using-tags to publish a "snapshot" version of the lib, so JenkinsCD can install them from the npm repo

blueocean-core-js/src/js/components/ContentPageHeader.jsx
scherler: BTW https://github.com/jenkinsci/blueocean-plugin/tree/master/js-extensions

kzantow: yes, the inner contents are rendered when there are no matching extensions

blueocean-web/src/main/js/main.jsx
```diff
@@ -78,9 +78,13 @@ class App extends Component {
         });
 
         const topNavLinks = [
-            <Link query={location.query} to="/pipelines">{pipeCaption}</Link>,
+            <Extensions.Renderer extensionPoint="jenkins.blueocean.top.pipelines">
+                <Link query={location.query} to="/pipelines">{pipeCaption}</Link>
+            </Extensions.Renderer>,
             <Extensions.Renderer extensionPoint="jenkins.blueocean.top.links"/>,
-            <AdminLink t={translate} />,
+            <Extensions.Renderer extensionPoint="jenkins.blueocean.top.admin">
```
kzantow: There is one _very_ annoying thing about this, but nothing you can really do about it yet... the entire set of extensions is wrapped in a `<div>`, so you end up with a structure like:
```
<AdminLink>
<div>
    <AdminLink>
</div>
```

This can easily cause issues with styling if it's not taken into consideration

michaelneale:
@kzantow if by run without any extensions - you mean just normal blue ocean? yes seems all ok to me. And the ATH passes for this too. 

So @scherler - you have permission to land (need to do the js publish dance etc etc).

blueocean-core-js/src/js/components/ContentPageHeader.jsx
sophistifunk: Yes. Initially I put that functionality in for an editor demo, but I definitely had logo-replacement in mind, I'm glad it's going in.

cliffmeyers:
Looks good from me too, given the limitations of the current extension point API. We definitely need to think about better ways of handling the navigation down the road: an XP for every item doesn't seem ideal; maybe there should be points for the "main links" and "icon links" and plugins have the ability to add / remove / replace items.

PR #961, opened 2017-04-12
Remove blueocean-web dependency from blueocean-rest

vivek:
This makes blueocean-rest pure model. Other components including
blueocean-web can depend on rest model to interconnect model and UI
components.

For example this will allow plugin developers to implement blueocean extensions without depending on UI components.

@i386 PTAL

# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
Similarly to https://github.com/jenkinsci/blueocean-plugin/pull/956 - what is the impact on the editor plugin? Having trouble understanding all these changes and the large effects it may have

michaelneale:
This will need to be checked with @kzantow as I am not sure of impact on these dependency changes with the editor (external) plugin.

i386:
Seems reasonable - thanks for doing this.

Echoing @michaelneale we need to ensure the editor works and it's tests pass before we do this

michaelneale:
also looked at this some more - as long as @kzantow is ok - this is probably fine, go ahead if so.

PR #962, opened 2017-04-12
Task/libify time harmonizer

sophistifunk:
# Description

Ignore this for now, I just want this branch to run against ATH to make sure I've got the first part of the fix without breaking client code, so probably have enough tests :)


michaelneale:
@sophistifunk and it passes - nice

cliffmeyers:
@sophistifunk curious if you'd considered that we could handle some of these time transforms upstream - like closer to the fetch response layer? If not, that's cool, just throwing it out there.

sophistifunk:
Yes, the very concept is ridiculous, I have NFI why on earth we have the concept of "server skew" at all. If your sever thinks it's 1970, you should get bad dates in BO like you will in everything else, and you should hire a sysadmin.

michaelneale:
I see lots of new tests here - is nice right @cliffmeyers ? @sophistifunk when do you want this in? who else needs to review?

sophistifunk:
I'm not sure, maybe @scherler because he wrote a lot of the original stuff IIRC. But what I found had the feel like it had been messed around with over time since his original work, so I'm not sure. I just needed to have it not as a wrapper for some table stuff, which is why it mostly moved into a helper class.

michaelneale:
@cliffmeyers any objections on this one?

michaelneale:
:bee: go go go

PR #965, opened 2017-04-17
Task/jenkins 43248 Remove hardcoding of the jenkins organization from the UI

vivek:
# Description

See [JENKINS-43248](https://issues.jenkins-ci.org/browse/JENKINS-43248).

To summarize changes:

- Routing changes to allow serving create-pipeline from organization route
- Organization name is served as preloaded data from backend, this organization is served to all of UI thru config.js
- Changes in creation code to call backend API with configured organization
- Fixes elsewhere to remove hardcoded organization
- Backend organization name 'jenkins' can be changed using system property to help testing with different org name.

@cliffmeyers  PTAL as most changes are in create pipeline related. 

 

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
This passes ATH last I checked - and works as advertised. Just needs a review from @cliffmeyers to bless it

cliffmeyers:
@vivek I just want to understand the intent of the PR a little better. This allows the organization name to be set by the backend and eliminate the hard-coding. I remember when we originally created the URL schemes for the UI and REST API that there was discussion of supporting multiple organizations from within the same UI. Is that on the radar anytime soon? If it is, we might want to have a brief discussion about how the UI would accomplish that switch.

blueocean-core-js/src/js/config.js
```diff
@@ -35,6 +36,10 @@ export default {
         return config.jenkinsConfig || {};
     },
 
+    getOrganization() {
+        return encodeURIComponent(organization.name);
```
cliffmeyers: Would there be any value to having separate methods for the organization and organizationName? We might also want to consider whether the organization name should always be encoded, in case we'd like to display it in the UI.

vivek: @cliffmeyers There are different ways to do it. We can either keep different method names or pass encoding true or false via method param. e.g.  getOrganization(encoded=true)?

vivek:
> My more general comment above was really driving toward whether we expect to mutate the organization on the fly (i.e. without reloading the application). If we do, putting it in "AppConfig" might not be the best place for it, as we don't tend to mutate data in there but rather drive it exclusively from the preloader.

organization is immutable more of singleton, in backend as well as front end so do not expect it to change. So I guess AppConfig seems like right place for it to be.

blueocean-core-js/src/js/config.js
vivek: @cliffmeyers changed to getOrganizationName(encoded=true). PTAL.

vivek:
ATH passes https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/master/2162/pipeline.

PR #966, opened 2017-04-17
Remove redux from tests, support pagination

i386:
See [JENKINS-38054](https://issues.jenkins-ci.org/browse/JENKINS-38054)

@cliffmeyers I was having some issues with getting stdout/stderr fetch to update the `TestCaseResultRow` - any ideas?

Pagination using mobx and ShowMore works perfectly.


blueocean-dashboard/src/main/js/components/testing/TestService.js
```diff
@@ -0,0 +1,57 @@
+import { BunkerService } from '../../../../../../blueocean-core-js/src/js/services/BunkerService';
```
i386: @cliffmeyers I think I need to publish corejs so I can just import in the line below?

cliffmeyers:
@i386 I assume this PR will be a bit easier to read if #950 is merged first?

i386:
@cliffmeyers I merged #950 and rebased this PR so this should be easier to read.

Stuff I need help with:
- I can't see to make `TestLogService` update the UI when we have a successful fetch of stdout/err.
- Not sure how to rewrite the tests now that pager is used etc

blueocean-core-js/src/js/index.js
```diff
@@ -9,6 +9,18 @@ import { RunApi } from './rest/RunApi';
 
 import { SseBus } from './sse/SseBus';
 import { ToastService } from './ToastService';
+import {
```
i386: I think my IDE is being sneaky and reformatting this file. can probably just revert.

blueocean-dashboard/src/main/js/components/RunDetailsTests.jsx
```diff
@@ -40,33 +39,15 @@ NoTestsPlaceholder.propTypes = {
  * Displays a list of tests from the supplied build run property.
  */
 export class RunDetailsTests extends Component {
-    componentWillMount() {
```
i386: @cliffmeyers death blow for redux

blueocean-core-js/src/js/index.js
```diff
@@ -90,17 +100,15 @@ sseService.registerHandler(defaultSSEhandler.handleEvents);
...
 // Export some debugging stuff client code may need
 
-import { enableMocksForI18n, disableMocksForI18n } from './i18n/i18n';
-
 export const DEBUG = {
     enableMocksForI18n,
     disableMocksForI18n,
 };
 
+export { BunkerService } from './services/BunkerService';
```
i386: @cliffmeyers I did need to export this type so I can use `BunkerService` itself in `blueocean-dashboard`

michaelneale:
unit tests failing on the tests feature now: 

```
[INFO]   TestResults
[INFO]     1) Test fixed included
[INFO]     2) Handles REGRESSION case
[INFO]     3) All passing shown
[INFO]     4) All passing and fixed shown
[INFO]     5) unstable renders with no error message
```

cliffmeyers:
@michaelneale @i386 do either of you gents have a suitable test repo w/ hundreds of tests?

i386:
@cliffmeyers no, but I tested with app-store-demo and set page size of 1 to achieve the same effect.

cliffmeyers:
@i386 I spent some time with this PR on Fri and it seems to behaving okay in my environment. I forked your test project into a new org I'm using just for test repos, and added some tests. See here:

https://github.com/i386/app-store-demo/compare/master...histos:failures-skips-passes

If you are still seeing errors on your end - you mentioned something about an error related to organization that I don't see - could you put some time on my calendar on Tue the 24th EDT  (basically 24 hours from the time of this post) so we could debug this together?

i386:
@cliffmeyers got the unit tests to work 👍

i386:
@cliffmeyers can I get a review on this now please :)

blueocean-dashboard/src/main/js/components/testing/TestLogService.js
```diff
@@ -0,0 +1,55 @@
...
+export default class TestLogService extends BunkerService {
+
+    loadStdOut(test) {
+        if (this.hasItem(test)) return;
+        Fetch.fetch(test._links.stdout.href)
+            .then(res => res.text())
+            .then((data) => {
+                this.setItem(new Item(test, data, true));
+            })
+            .catch(TestLogService.ignoreNotFound);
```
cliffmeyers: Use of `catch` like this can swallow errors caused by any prior `then` call.  If `this.setItem` triggers a React render and an exception occurs (either in `observer` component or one of its descendants) then this catch block will fire, even though it's really intended just to handle a 404. If we are just concerned about handling a 404 from the REST API, the `ignoreNotFound` method should be passed as the second argument to the first `then` call.

blueocean-dashboard/src/main/js/components/testing/TestLogService.js
```diff
@@ -0,0 +1,55 @@
...
+        return item && item.value && item.value.log;
+    }
+
+    bunkerKey(data) {
+        const links = data.value.test._links;
+        return data.value.isStdOut ? links.stdout.href : links.stderr.href;
+    }
+
+    static ignoreNotFound(e) {
+        if (e.response.status !== 404) {
```
cliffmeyers: I think this method should add an entry via `setItem` which records the fact the fetch occurred and no data was available. You could probably even just store `null` as the value and the logic that calls `hasItem` to suppress the fetch will continue to work fine. That way repeated open/close for test cases without `stdout` and `stderr` won't cause repeated fetches.

blueocean-dashboard/src/test/js/testResult-spec.js
```diff
@@ -1,60 +1,118 @@
...
+            '_links': {
+                tests: {
+                    href: '/pipeline/runs/1/tests'
+                }
+            }
+        };
+
+        const testService = new MockTestService({fixed: fixed, skipped: skipped, existingFailed: existingFailed});
+
+        const wrapper = mount(<TestResults t={t} run={run} testService={testService} pipeline={pipeline} />);
```
cliffmeyers: Will this test continue to work if we use `shallow(...).html()` rather than `mount`? The latter requires some `jsdom` bootstrap to happen which would be nice to avoid. Unfortunately when we first started to use Enzyme I didn't know about the `html()` method, so some of the tests used `mount` to do deep rendering.

blueocean-dashboard/src/test/js/testResult-spec.js
```diff
@@ -1,60 +1,118 @@
...
-        { age: 0, name: 'failure.TestThisWillFailAbunch', duration: 0, status: 'PASSED', state: 'FIXED' },
-        { age: 1, name: 'failure.TestThisWillFailAbunch', duration: 0.003, errorDetails: '<some exception here>', status: 'FAILED' },
-        { age: 4, name: 'failure.TestThisWillFailAbunch', duration: 0.001, status: 'SKIPPED' },
-        { age: 0, name: 'failure.TestThisWontFail', duration: 0, status: 'PASSED' },
-        { age: 0, name: 'failure.TestThisWontFail', duration: 0, status: 'PASSED' },
-        { age: 0, name: 'failure.TestThisWontFail', duration: 0, status: 'PASSED' },
-        { age: 4, name: 'failure.TestThisWillFailAbunch', duration: 0.003, errorDetails: '<some exception here>', status: 'FAILED' },
-        { age: 4, name: 'failure.TestThisWillFailAbunch', duration: 0.003, errorDetails: '<some exception here>', status: 'FAILED' },
-    ];
+    class MockTestService {
```
cliffmeyers: 👍

i386:
Passes ATH https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/master/2306/pipeline

i386:
@cliffmeyers switching to `shallow` fails some tests (so does not just work :( )

i386:
Took into account all @cliffmeyers feedback, doing the last corejs dance and ATH run

PR #967, opened 2017-04-18
I18n loading - readmedocs + a tweak to support loading from other plugins

tfennelly:
Some `README.md` docs around `@jenkins-cd/js-extensions` i18n resource pre-loading + a tweak to allow plugins to pre-load i18n bundles defined in other plugins.

See [js-extensions#i18n-resource-pre-loading](https://github.com/jenkinsci/blueocean-plugin/tree/i18n-loading-readmedocs/js-extensions#i18n-resource-pre-loading).

For @scherler i.e. **he owns this PR now** 😄


scherler:
:bee: Awesome stuff Tom. Thank you very much. I did not know about 

```
+ i18nBundles:
 +  - jenkins.plugins.aaa.Messages
 +  - jenkins.plugins.bbb.Messages
```

tfennelly:
@scherler beta published with the loading from other plugins option ... `@jenkins-cd/js-extensions@0.0.36-tf-beta1` ... give it a try and let me know

alvarolobato:
@tfennelly thanks for the PR.

One question. Will this allow us to load a bundle just by the bundle name?
When you create the i18NTranslator we pass the plugin name, but that is to decide the class loader isn't it? With this change we should be able to just pass the Name Space.

Our problem is that some bundles are declared on a jar which is included on a plugin:

common-jar -> Contains translation bundles and some jsx resources
plugin-> Contains the jar

With this:
```
+ i18nBundles:
 + - hpiPluginId: plugin
 +    resource: jenkins.plugins.acme.common.Messages
```

We should be able to load the plugin correctly but when instantiating  the i18NTranslator on the resources that come from the jar we don't know the plugin name `new i18NTranslator("plugin-name")` we don't know the name of the plugin where this jar has been included and we should be able to just load it from the resource name: jenkins.plugins.acme.common.Messages

Maybe we would need some changes to i18NTranslator what do you think?

scherler:
works in my set up :bee:

michaelneale:
nice one :bee: 

very cool work !

tfennelly:
@alvarolobato no, this change is purely to tell `js-extensions` that the generated bundle needs to ensure that the i18n resources need to be pre-loaded before the bundle executes. You still need to specify the plugin name for `i18NTranslator`.

I'm sure there's some `falafel` build time analysis we could do in order to find calls to the `i18NTranslator` code (or some decorators perhaps) and so remove the need to manually define shit in the `.yaml`, but that's another days work. These changes don't stop us doing that in the future.

scherler:
@kzantow and @michaelneale  I actually changed some more things after you :+1: 

Basically I extended the i18n java plugin to work on JAR as well and not only on real registered plugins.

You can now request something like `http://localhost:8080/jenkins/blue/rest/i18n/KEITH/dummyVersion/jenkins.plugins.myplugin.common.Messages/en` and if you have somewhere `jenkins.plugins.myplugin.common.Messages` as resource defined it will resolve that file. In case it does not find the file we return 404 as it is now.

cc @tfennelly

scherler:
@michaelneale and @kzantow if you or @tfennelly do not oppose my changes, l will merge it within 8 hours

blueocean-core-js/src/js/i18n/i18n.js
```diff
@@ -28,7 +28,9 @@ function newPluginXHR(pluginName, onLoad) {
     let pluginVersion = store.getPluginVersion(pluginName);
 
     if (!pluginVersion) {
-        throw new Error(`Unable to create an i18n instance for plugin "${pluginName}". This plugin is not currently installed, or is disabled.`);
+        // if we do not have a version we may have an alias to resolve a resourceBUndle
+        pluginVersion = 'dummy';
```
tfennelly: How about basing it on an `undefined` `pluginName`? I don't like these magic ID ("dummy") style solutions.

blueocean-i18n/src/main/java/io/jenkins/blueocean/i18n/BlueI18n.java
```diff
@@ -136,12 +136,19 @@ public HttpResponse doDynamic(StaplerRequest request) {
     private JSONObject getBundle(BundleParams bundleParams, Locale locale) {
         PluginWrapper plugin = bundleParams.getPlugin();
```
tfennelly: So if `BundleParams` supported the notion of an undefined plugin, then I think we could change the JS code to send that instead of "dummy".

tfennelly:
@scherler Added some comments.

I'm really not a fan of this `.jar` stuff ye are trying to do (and subsequent classpath shenanigans). I currently can't put my finger on how this is going to break, but I just feel in my gut that it's all too loosey goosey and bound to blow up in your/our faces. I know the "common plugin" approach might seem less convenient to ye (I'm not sure I agree with the logic behind it tbh), but at least we should be able to have a bit more confidence in it not having weird side effects.

blueocean-core-js/src/js/i18n/i18n.js
scherler: @tfennelly that is the version, not the ID. I was tempted to use 6.6.6 instead but yeah. ;)

blueocean-i18n/src/main/java/io/jenkins/blueocean/i18n/BlueI18n.java
```diff
@@ -136,12 +136,19 @@ public HttpResponse doDynamic(StaplerRequest request) {
...
-        if (plugin == null) {
-            return null;
-        }
 
         try {
-            ResourceBundle resourceBundle = ResourceBundle.getBundle(bundleParams.bundleName, locale, plugin.classLoader);
+            ResourceBundle resourceBundle;
+            if (plugin == null) {
+                Jenkins jenkins = Jenkins.getInstance();
+                if (jenkins!=null && jenkins.pluginManager!=null && jenkins.pluginManager.uberClassLoader!=null) {
```
tfennelly: As I said to @scherler ... whenever I see code with `uberClassLoader`, a big flashing light comes on ... "uber hack in progress here"

blueocean-i18n/src/main/java/io/jenkins/blueocean/i18n/BlueI18n.java
scherler: @tfennelly as said in the earlier comment the "dummy" is a fake version of that not existing plugin

blueocean-i18n/src/main/java/io/jenkins/blueocean/i18n/BlueI18n.java
scherler: ...like in the real world ;) https://www.nytimes.com/2016/12/21/technology/san-francisco-california-uber-driverless-car-.html?_r=0

blueocean-core-js/src/js/i18n/i18n.js
tfennelly: @scherler oh yeah, sorry ... replace "name" with "version" in what I said ... whatever .... it's a hack

blueocean-i18n/src/main/java/io/jenkins/blueocean/i18n/BlueI18n.java
kzantow: well, you need this to load from all plugins

kzantow:
Mega-question regarding `http://localhost:8080/jenkins/blue/rest/i18n/KEITH/dummyVersion/jenkins.plugins.myplugin.common.Messages/en` is `KEITH` the organization? What is the version for? Seems like this is a little odd since there can't be multiple versions of the plugins... my bigger question is why do we have to do this at all? _Where_ are plugins requesting the bundles in Blue Ocean, again? - in each file they're used?

michaelneale:
a few questions from @tfennelly and @kzantow

blueocean-i18n/src/main/java/io/jenkins/blueocean/i18n/BlueI18n.java
scherler: Actually we need it for the resource bundle which is stored in the classpath. Our common code is in a jar and contains the translations. Since it is not hpi we had to workaround that. 

Having said all this, we are trying another approach now.

Since we have our jsx in the common and create an instance like

```
Translator('commons')
```

However since this code will be imported to a real hpi plugin we could change it to something like

```
const callingPlugin = AppConfig.getCallerPlugin();
translator(callingPlugin)
```

And could revert the change to the java code.

blueocean-i18n/src/main/java/io/jenkins/blueocean/i18n/BlueI18n.java
jtnord: so is this because when using a sharable module the plugin is the name in the jar and hard coded at build time (so we serve js files that reference the jar's artifact ID rather than the plugins artifact IDs instead of the plugin that includes the jar?)

blueocean-i18n/src/main/java/io/jenkins/blueocean/i18n/BlueI18n.java
jtnord: I think this links in to my [comment](https://github.com/jenkinsci/blueocean-plugin/pull/967/files#diff-eb19343eb8fa1a34c9981c57f23d1f92L137) above.  Given that we could have clashes in jar resources no one says that within a jar resource everything needs to be unique) - looking in the uber class loader first could cause some issues...

blueocean-core-js/src/js/i18n/i18n.js
```diff
@@ -28,7 +28,9 @@ function newPluginXHR(pluginName, onLoad) {
...
+        // throw new Error(`Unable to create an i18n instance for plugin "${pluginName}". This plugin is not currently installed, or is disabled.`);
```
batmat: Do we really need this? Can we better delete instead of commenting out code? :-)

scherler:
I reverted all changes which came in after the initial bees. Will merge this as soon Jenkins gave the green light

PR #968, opened 2017-04-18
Abstract queuedness of runs in the REST API

i386:
All look upon this refactor and despair no more
Holy be the client, for now it is blessed with innocence
Of architectural sins visited by the father
For now what is run is queued
And for what is run is queued no more

# Description

Removed all the knowledge of queued items from the client so that non-pipeline queued runs isn't so broken anymore. Bonus feature: `causeOfBlockage` [JENKINS-41540](https://issues.jenkins-ci.org/browse/JENKINS-41540).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


i386:
Updated unit tests now that semantics of starting runs has changed and `/activities` is no more.

i386:
ATH https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/2135/

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
```diff
@@ -83,7 +82,7 @@ public BlueRunState getStateObj() {
         } else if(!run.isLogUpdated()){
             return BlueRunState.FINISHED;
         } else {
-            return BlueRunState.QUEUED;
```
vivek: I thought QUEUED is right state to return, why change to RUNNING? Also if RUNNING is really the right state to return the conditions could be further simplified.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueuedBlueRun.java
```diff
@@ -0,0 +1,159 @@
...
+        this.parent = parent;
+    }
+
+    @Override
+    public String getOrganization() {
+        return item.getOrganization();
+    }
+
+    @Override
+    public String getId() {
```
vivek: Its quasi run id, its not run yet. This can be confusing to client as expected run id, might be reused and is simply a guess. Better to call it out what it is, that is `getExpectedRunId()` and client should treat it as such. Otherwise I am pretty sure someone later on is going to re-write this whole thing again:)

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunContainerImpl.java
```diff
@@ -59,8 +62,22 @@ public BlueRun get(String name) {
...
             }
+
+            int number;
+            try {
+                number = Integer.parseInt(name);
+            } catch (NumberFormatException e) {
+                throw new NotFoundException(String.format("Run %s not found in organization %s and pipeline %s",
+                    name, pipeline.getOrganization(), job.getName()));
+            }
+            for (BlueQueueItem item : QueueContainerImpl.getQueuedItems(job)) {
```
vivek: Note: `QueueContainerImpl.getQueuedItems(job)` returns active queued items. If a queue item was aborted, it won't be returned here. Maybe its not any different than current implementation.

i386:
> assumes expected run id to be a guaranteed run id, which it's not

My argument for this change is that this is the exact same assumption the client makes several times in the frontend and fails to handle properly. 

If we take the opinion that in the future the architecture of the queue changes and items have guaranteed run numbers (numbers are no longer "expected") then my proposed way of handling queued items would continue to work and the client would be none the wiser.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunContainerImpl.java
i386: the client does not handle this today and that comes at a high price of complexity there. It depends where we want to wear the ugliness.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
i386: Agreed - I need to look into this. The non-pipeline version of this should never return queued as it will receive a `BlueQueuedRun` instead of this model

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueuedBlueRun.java
```diff
@@ -0,0 +1,159 @@
...
+        return Integer.toString(item.getExpectedBuildNumber());
+    }
+
+    @Exported(name = QUEUE_ID)
+    public String getQueueId() {
```
i386: @vivek to allay your concern the client can always find out if the Run is actually a queued run by the presence of this property in the fetch response.

i386:
ATH https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/master/2141/pipeline

i386:
Found what broke the ATH test. Simple enough to fix :)

michaelneale:
@i386 what is the conflict with what is going on in this branch: https://github.com/jenkinsci/blueocean-plugin/pull/839 - is there a way to tell if there is overlap?

i386:
@michaelneale there was a small conflict merging this with @scherler's karaoke reloaded branch around displaying the causeOfBlockage message instead of "waiting for run to start" but I don't mind if that gets stomped

michaelneale:
Ah, so the ATH failure(s) was legit. 

I ran commitMessages.js locally, and noticed activity tab was not updating when a new run happened: 



which it is rightfully picking up as a regression (you only see the first run, not the second) - so clealy a :bug: there which may be tripping up other things. no console errors, and pipeline did run, perhaps events being dropped on the floor?

i386:
@michaelneale yeah there are a few events that we receive that we drop that actually need to be pumped to the client to make things work. Matter of rearranging a switch statement...

michaelneale:
AH, that would explain it then.

i386:
BOOM! ATH passes https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/master/2151/pipeline/

i386:
@vivek would you mind giving the java changes a once over again? There were a few things that I had to change along the way.

michaelneale:
This seems to work nicely - and ATH is happy. I tried it with queuing and so on. 

Any specific areas we should look for @i386 to test this more? 

@cliffmeyers was going to take a quite look (code only). Nice!

i386:
@michaelneale @cliffmeyers heres a list of things to try for both multibranch and freestyle:

- Run from dashboard until completed
- Run from dashboard then stop
- Run from activity page (freestyle)
- Run from branch tab until completed (switching back to activity often)
- Run from branch tab then stop
- Rerun a freestyle run
- Rerun a branch run

blueocean-core-js/src/js/services/DefaultSSEHandler.js
```diff
@@ -59,26 +56,18 @@ export class DefaultSSEHandler {
...
-
-        for (const key of this.branchPagerKeys(event)) {
-            const pager = this.pagerService.getPager({ key });
-            this.activityService.fetchActivity(runSelf, { overrideQueuedState }).then(d => {
-                if (pager && !pager.has(runSelf)) {
-                    pager.insert(runSelf);
-                }
-                this.pipelineService.updateLatestRun(d);
-            });
-        }
```
cliffmeyers: This code here seems responsible for adding a new run to the pager which holds runs for that job. Can it be safely removed?

cliffmeyers: Ah, looks like this same logic was being done in `_updateRun` anyways. Good to remove it, then.

i386:
Woohoo thanks @cliffmeyers! 

How does one go about releasing corejs?

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
```diff
@@ -3,29 +3,28 @@ import ReactDOM from 'react-dom';
 import Extensions from '@jenkins-cd/js-extensions';
 
 import {
+    buildClassicInputUrl,
```
scherler: We will have fun when merging karaoke and this file

i386: Feel free to stomp this file - Ill come back and update the changes to `QueuedState` when your branch is merged

blueocean-core-js/src/js/services/DefaultSSEHandler.js
i386: @cliffmeyers yep - all that mutation is in a single place now :)

blueocean-core-js/src/js/services/ActivityService.js
```diff
@@ -74,7 +62,7 @@ export class ActivityService extends BunkerService {
      * @param {boolean} overrideQueuedState Hack to make SSE work. Not use unless you know what you are doing!!!
```
i386: Removed

blueocean-core-js/src/js/paths/rest.js
```diff
@@ -20,7 +20,7 @@ export default {
 
     activities(organization, pipeline, branch) {
```
i386: Updated but I will leave `ActivityService` as is.

blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
michaelneale: @scherler yeah - @i386 had a look - the karaoke branch side can stomp it

PR #971, opened 2017-04-18
Enforce exclusion of hudson.* jenkins.* imports with some exemptions

i386:
# Description

Enforce exclusion of `hudson.*` `jenkins.*` imports with some exemptions

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest/pom.xml
+++ b/blueocean-rest/pom.xml
@@ -24,4 +24,47 @@
           <artifactId>capability-annotation</artifactId>
       </dependency>
   </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <dependencies>
+                    <dependency>
+                        <groupId>de.skuzzle.enforcer</groupId>
+                        <artifactId>restrict-imports-enforcer-rule</artifactId>
+                        <version>0.7.0</version>
+                    </dependency>
+                </dependencies>
+                <executions>
+                    <execution>
+                        <id>check-no-jenkins-imports</id>
+                        <phase>process-sources</phase>
+                        <goals>
+                            <goal>enforce</goal>
+                        </goals>
+                        <configuration>
+                            <rules>
+                                <restrictImports implementation="de.skuzzle.enforcer.restrictimports.RestrictImports">
+                                    <basePackage>**</basePackage>
+                                    <bannedImports>
+                                        <bannedImport>jenkins.**</bannedImport>
+                                        <bannedImport>hudson.**</bannedImport>
+                                    </bannedImports>
+                                    <allowedImports>
+                                        <allowedImport>hudson.Extension</allowedImport>
+                                        <allowedImport>hudson.ExtensionPoint</allowedImport>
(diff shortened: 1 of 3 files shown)


blueocean-rest/src/main/java/io/jenkins/blueocean/Routable.java
```diff
@@ -12,7 +10,7 @@
  */
 public interface Routable {
     /**
-     * Follows the same convention as {@link Action#getUrlName()}
+     * Follows the same convention as Action.getUrlName()
```
i386: Did not want to add `Action` to whitelist as nothing depends on it except for javadoc.

blueocean-rest/pom.xml
```diff
@@ -24,4 +24,47 @@
...
+                                        <bannedImport>jenkins.**</bannedImport>
+                                        <bannedImport>hudson.**</bannedImport>
+                                    </bannedImports>
+                                    <allowedImports>
+                                        <allowedImport>hudson.Extension</allowedImport>
+                                        <allowedImport>hudson.ExtensionPoint</allowedImport>
+                                        <allowedImport>hudson.ExtensionList</allowedImport>
+                                        <allowedImport>hudson.ExtensionListListener</allowedImport>
+                                        <allowedImport>hudson.security.csrf.CrumbExclusion</allowedImport>
+                                        <allowedImport>hudson.util.AdaptedIterator</allowedImport>
```
i386: Needed for `Container`

blueocean-rest/pom.xml
```diff
@@ -24,4 +24,47 @@
...
+                    <execution>
+                        <id>check-no-jenkins-imports</id>
+                        <phase>process-sources</phase>
+                        <goals>
+                            <goal>enforce</goal>
+                        </goals>
+                        <configuration>
+                            <rules>
+                                <restrictImports implementation="de.skuzzle.enforcer.restrictimports.RestrictImports">
+                                    <basePackage>**</basePackage>
```
i386: We could add a second rule to allow certain imports for a new `io.jenkins.blueocean.rest.factory` package. All factory classes would be moved from `blueocean-rest-impl` to `blueocean-rest`. Then plugins could safely depend on `blueocean-rest` and use the factories.

blueocean-rest/src/main/java/io/jenkins/blueocean/Routable.java
vivek: +1

blueocean-rest/pom.xml
vivek: Looks good. Wish these could be avoided.

blueocean-rest/pom.xml
vivek: I guess we can do that, Item is one of the thing most blue extensions depend on, although it pulls in lots of other dependencies but not too bad. Where as Run (needed by BlueRunFactory) has too many dependencies and its like opening up to broader set of dependencies in Jenkins core.

vivek:
LGTM 🐝 as long as test/ATH pass:)

PR #978, opened 2017-04-19
JENKINS-40084 PipelineRunImpl.replay assumes the task is still in the…

i386:
… queue

# Description

Tries to find the item in the queue, if does not exist tries to find the run that is executing else throws an exception.

That should cover most of the cases. Better now that `replay` expects a run to be returned.

See [JENKINS-40084](https://issues.jenkins-ci.org/browse/JENKINS-40084).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


i386:
PTAL @vivek

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueUtil.java
```diff
@@ -0,0 +1,78 @@
+package io.jenkins.blueocean.service.embedded.rest;
```
i386: @vivek moved all of these methods off the queue container impl.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
```diff
@@ -22,7 +22,7 @@
 import io.jenkins.blueocean.rest.model.Containers;
```
i386: @vivek guts of this change is here.

PR #981, opened 2017-04-20
Improvement/jenkins 43411 build blueocean modules with yarn instead of npm

cliffmeyers:
# Description
- Depends on jenkinsci/plugin-pom#57
- See [JENKINS-43411](https://issues.jenkins-ci.org/browse/JENKINS-43411).
- Work that should definitely be done as part of this PR:
   - [ ] Pair down "checkshrinkwrap" script and rename it to describe what it actually does
   - [ ] Write a document on (1) known yarn issues and their workarounds, (2) yarn tips for those of us accustomed to npm
   - [ ] Update to non-SNAPSHOT parent pom once plugin-pom is released
- Work that could be done in this PR?
   - [ ] Port blueocean-core-js to use yarn; strictly speaking, this could be done in a separate PR, and perhaps there's value to show that the app still works okay when consuming packages we build / publish via npm
   - [ ] Port JDL to use yarn (same considerations as above, although since in separate repo needs to be done in a separate PR anyways

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


cliffmeyers:
At the request of @michaelneale , opened this new PR based off branch rather than a fork.

cliffmeyers:
@michaelneale I added two definite TODO's for myself that I will work on today.

As for the other two, what are your thoughts as to the timing for tackling them? I can see value into doing them now (make sure building all things w/ yarn play nicely together) or doing them later (make sure yarn consumes things we build with npm without issues). Although arguably, we've already proven most of the latter since yarn is already pulling in hundreds of modules already built and published via npm (albeit from third parties, not from our team).

casz:
@cliffmeyers Missing this to do 👍 

- [ ] Update to non-SNAPSHOT dependency once plugin-pom is released 🚀

cliffmeyers:
@casz thanks :)

michaelneale:
This is looking pretty good, going to leave this as WIP as it will probably be a little while before we merge it right @cliffmeyers ?

pom.xml
```diff
@@ -22,6 +22,7 @@
     <javadoc.exec.goal>javadoc-no-fork</javadoc.exec.goal> <!-- stop initialize phase plugins executing twice -->
     <node.version>6.4.0</node.version>
     <npm.version>3.10.3</npm.version>
```
jglick: Still needed?

pom.xml
```diff
@@ -22,6 +22,7 @@
...
+    <yarn.version>0.23.0</yarn.version>
```
jglick: Why the override?

cliffmeyers: Just to be explicit about the version that `blueocean-plugin` uses, in case the `plugin-pom` gets ticked forward down the road.

pom.xml
cliffmeyers: Removed in 47aa6b1

pom.xml
michaelneale: also people tend to use it when they are checking they have a new enough version for doing things outside of mvn (but yeah, redundant technically)

cliffmeyers:
As a note this PR is blocked for two reasons:
* plugin-pom:2.28 causes some of our Java unit tests to fail, possible because of a new version of the Jenkins test harness coming in. @vivek is investigating.
* I think we should hold-off on landing this until after @kzantow extensibility changes land, since it should lead to a lot less npm/yarn operators in general. Also, despite all the work this PR entailed, it's possible that npm 5 is worth evaluating as well, given a number of significant improvements:  see npm/npm#16244

jglick:
Consider picking up https://github.com/jenkinsci/plugin-pom/pull/63 once you deal with any test harness issues. (Possibly just that you depend on Jenkins 2.x and you were assuming that detached plugins would be loaded implicitly in tests. This is no longer true; there is a `TestPluginManager` method to explicitly load them, but much better to have all plugins you expect to test against be `test`-scoped dependencies.)

michaelneale:
going to close this - can resurrect it later if we want.

PR #983, opened 2017-04-20
Support docker-from-docker more easily

rtyler:
This makes it easy to demo and use the container for Dockery workloads, which makes the Pipeline Editor muy bueno.

For example, with this commit: `docker run -p 8080:8080 -v /var/run/docker.sock:/var/run/docker.sock jenkinsci/blueocean`

Will allow the user to get started right away with Jenkinsfiles such as:

```groovy
   pipeline {
      agent { docker { image 'maven:3-alpine' }
      stages {
         stage('Build') {
            steps {
               sh 'mvn'
            }
         }
      }
   }
```

### DIFF
--- a/docker/official/Dockerfile
+++ b/docker/official/Dockerfile
@@ -1,7 +1,10 @@
 # This image is based off the latest Jenkins LTS
-FROM jenkins:alpine
+FROM jenkinsci/jenkins:lts-alpine
 
 USER root
+
+# Add the docker binary so running Docker commands from the master works nicely
+RUN apk -U add docker
 
 RUN install-plugins.sh antisamy-markup-formatter matrix-auth pipeline-model-definition blueocean:$BLUEOCEAN_VERSION
 


docker/official/Dockerfile
```diff
@@ -1,8 +1,11 @@
 # This image is based off the latest Jenkins LTS
-FROM jenkins:alpine
+FROM jenkinsci/jenkins:lts-alpine
```
i386: LTS is good 👍

i386:
🐝  I like that we are using the LTS here!!

docker/official/Dockerfile
rtyler: A more current and correct LTS too, `jenkinsci/jenkins` is always updated automatically after releases. Whereas the "official library" image (`jenkins`) requires a manually submitted and approved pull request :face_with_head_bandage:

i386: Sounds more reasonable to me. Merge when ready 💯

PR #979, opened 2017-04-20
when a run is not replayable then do not show the replay button

i386:
# Description

When a run is not replayable then do not show the replay button

# Testing instructions
Start and stop a pipeline build very quickly and the run will be unplayable (this happens before Pipeline can store flow nodes). Increase your chances by making the number of executors zero.

The run should not have a rerun button.

See [JENKINS-42997](https://issues.jenkins-ci.org/browse/JENKINS-42997).

# Submitter checklist
- [x] Link to JIRA ticket in description, if appropriate.
- [x] Change is code complete and matches issue description
- [x] Appropriate unit or acceptance tests or explanation to why this change has no tests
- [x] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below.
- [ ] Ran Acceptance Test Harness against PR changes.

# Reviewer checklist
- [ ] Run the changes and verified the change matches the issue description
- [ ] Reviewed the code
- [ ] Verified that the appropriate tests have been written or valid explanation given

PTAL @imeredith @cliffmeyers

PR #985, opened 2017-04-20
JENKINS-43256# SSE subscribe event includes organization

vivek:
# Description

See [JENKINS-43256](https://issues.jenkins-ci.org/browse/JENKINS-43256).

To test check the network tab, sse subscribe event should include jenkins_org element:
```
{
   "subscribe" : [
      {
         "jenkins_org" : "jenkins",
         "jenkins_channel" : "job"
      },
      {
         "jenkins_channel" : "job",
         "jenkins_org" : "jenkins"
      }
   ],
   "unsubscribe" : [],
   "dispatcherId" : "jenkins-blueocean-core-js-1492722251735-3cxgmk"
}
```
@tfennelly @cliffmeyers PTAL. For some reason there are duplicate job channel subscription but thats how it's always been and beyond scope of this PR.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-core-js/src/js/sse/SseBus.js
```diff
@@ -44,7 +45,7 @@ export class SseBus {
         if (!this.sseListeners.job) {
             const sseListener = this.connection.subscribe('job', (event) => {
                 this._handleJobEvent(event);
-            });
+            }, { jenkins_org: AppConfig.getOrganizationName() });
```
michaelneale: so how does this work - is this filtering events coming back or throwing away events not for the current org - if the latter, this wont' really be a scalable approach if there are lot of orgs right?

vivek: Each UI instance knows one org. All we are telling is backend SSE pub-sub to only return events for this org. So in multi-master case, there can be special pub sub implementation that would filter events accordingly. Why you think scalability could be an issue?

michaelneale:
This will need to be updated to master please

blueocean-core-js/src/js/sse/SseBus.js
michaelneale: ok cool

michaelneale:
Can this be updated to master? also - if the branch name can have a little slug of what the ticket is about, helps with reviewing.

Does this pass ATH? I think that and @tfennelly should be enough to allow this in.

vivek:
Its late for me, tomorrow will update it to master.

>also - if the branch name can have a little slug of what the ticket is about, helps with reviewing.

Heh,  doesn't PR has link to associated ticket and PR description is already there, no?

vivek:
Merged latest master. Lets see CI, ATH.

michaelneale:
@vivek oh I just have a lot of things checked out - and see messages about runs, but only has the ticket number so I don't remember what one is what when looking at a listing.

vivek:
@michaelneale ah I see, right checked out branches its hard. doesn't hurt to include some description.

i386:
Please wait for #979 to land in master.

michaelneale:
if I get a better memory for numbers I wouldn't have this problem... there is also some shell script foo that makes it show up (but also in hipchat/other messages too)

michaelneale:
Looks ok - updaste to master and check ATH is ok and :bee:

cliffmeyers:
@vivek LGTM to me too. I believe the reason there are two subscriptions is that personalization makes it own subscription to SSE as at the time there was no mechanism to share a model or data between the two. Once we get ATH for personalization we can refactor to share state from dashboard / core-js and can eliminate this subscription.

vivek:
@cliffmeyers sounds good. 

@i386 #979 looks like ready to land?

michaelneale:
once updated and builds happy etc - sure. Go for it.

vivek:
ATH passes https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/master/2303/pipeline.

PR #990, opened 2017-04-21
JENKINS-43742# Return actions only if queried with tree params

vivek:
# Description

See [JENKINS-43742](https://issues.jenkins-ci.org/browse/JENKINS-43742).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ActionProxiesImpl.java
```diff
@@ -86,13 +91,17 @@ public Link getLink() {
             @Override
             public boolean apply(Action action) {
                 return action != null
-                    && (action.getClass().getAnnotation(ExportedBean.class) != null || alwaysAllowAction.apply(action))
-                    && !BANNED_ACTIONS.contains(action.getClass().getName());
+                    && (action.getClass().getAnnotation(ExportedBean.class) != null || alwaysAllowAction.apply(action));
```
kzantow: Don't you still need `BANNED_ACTIONS` until the client explicitly requests actions?

vivek: No, we don't need it since actions are not returned by default. When we will have actions thru extensions then they will pick what they want. See https://issues.jenkins-ci.org/browse/JENKINS-43742

kzantow: ah, saw the `?tree=*[*]`, missed that it was in a test... d'oh

michaelneale:
so - does this mean it won't work for logs/karaoke?

vivek:
@michaelneale step/node apis still fetch actions but only LogAction like they used to. So Karaoke is all good. I tested locally to make sure as well.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
```diff
@@ -107,7 +111,12 @@ public Link getLink() {
 
     @Override
     public Collection<BlueActionProxy> getActions() {
-        return ActionProxiesImpl.getActionProxies(node.getNode().getAllActions(), this);
+        return ActionProxiesImpl.getActionProxies(node.getNode().getActions(), new Predicate<Action>() {
```
i386: @vivek please make sure you create a ticket to get this removed

i386:
🐝

michaelneale:
if you can update to master now, it will run with correct ATH etc...

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ActionProxiesImpl.java
```diff
@@ -70,7 +71,11 @@ public Link getLink() {
      * @return actionProxies
      */
     public static Collection<BlueActionProxy> getActionProxies(List<? extends Action> actions, Reachable parent){
-        return getActionProxies(actions, Predicates.<Action>alwaysFalse(), parent);
+        if(isTreeRequest()){
+            return getActionProxies(actions, Predicates.<Action>alwaysFalse(), parent);
```
michaelneale: Will this return ALL actions if it happens to be tree? vs ones specified?

michaelneale:
Seems  to work nicely - if you could update to master and check ATH still happy - I think we can :bee: this. Nice.

i386:
Keen to see the page weight dropped

michaelneale:
hrm no this seems to have a huge :bug:

When I build and run this the first time, open an existing workspace, it shows up a multibranch project as a non multibranch: 

screen shot 2017-04-21 at 6 39 37 pm

If I reload the page, or restart, it is ok. 

So this looks rather dire - can you work out why it would do that? 

rebuild this branch. Restart - open a multibranch project... its broken.

vivek:
@michaelneale This is strange, front end definitely doesn't use actions to determine multi-branch or not, it might be using project type or capabilities. I would expect it to reproduce on master as well, but who knows.  Anyways, I followed your steps but can't reproduce on this PR branch. Maybe missing something. Best if you can show me next week.

vivek:
Added label do not merge till we resolve regression reported by @michaelneale.

michaelneale:
@vivek I have seen this wierdness elsewhere, trying to chase down with keith, nothing to do with this branch. 

I removed my retraction, :bee: from me.

PR #996, opened 2017-04-23
JENKINS-40979 add trigger cause messages

i386:
# Description

* Adds trigger cause to Run detail
* If there are no commits the cause is displayed in Activity and Branch message column

See [JENKINS-XXXXX](https://issues.jenkins-ci.org/browse/JENKINS-40979).

screencapture-localhost-8080-jenkins-blue-organizations-jenkins-dropspace-2fwhimsy-activity-1492914891859

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/FilteringTreePruner.java
+++ b/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/FilteringTreePruner.java
@@ -13,6 +13,9 @@ class FilteringTreePruner extends TreePruner {
     private final TreePruner base;
 
     FilteringTreePruner(Predicate<String> predicate, TreePruner base) {
+        if (predicate == null) throw new IllegalArgumentException();
+        if (base == null) throw new IllegalArgumentException();
+
         this.predicate = predicate;
         this.base = base;
     }
@@ -25,7 +28,7 @@ class FilteringTreePruner extends TreePruner {
 
         // for merge properties, the current restrictions on the property names should
         // still apply to the child TreePruner
-        if (prop.merge)
+        if (prop.merge && child != null)
             child = new FilteringTreePruner(predicate,child);
 
         return child;
--- a/blueocean-dashboard/src/main/js/components/Branches.jsx
+++ b/blueocean-dashboard/src/main/js/components/Branches.jsx
@@ -1,13 +1,9 @@
 import React, { Component, PropTypes } from 'react';
-import {
-    CommitHash,
-    ReadableDate,
-    WeatherIcon,
-} from '@jenkins-cd/design-language';
-import { RunButton, LiveStatusIndicator } from '@jenkins-cd/blueocean-core-js';
+import { CommitHash, ReadableDate, WeatherIcon } from '@jenkins-cd/design-language';
+import { LiveStatusIndicator, RunButton } from '@jenkins-cd/blueocean-core-js';
 import Extensions from '@jenkins-cd/js-extensions';
 import { observer } from 'mobx-react';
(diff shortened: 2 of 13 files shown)


blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/FilteringTreePruner.java
```diff
@@ -25,7 +28,7 @@ public TreePruner accept(Object node, Property prop) {
 
         // for merge properties, the current restrictions on the property names should
         // still apply to the child TreePruner
-        if (prop.merge)
+        if (prop.merge && child != null)
```
i386: Use of `merge` in the 'Cause' `@Exported(name="cause", merge = true)` caused this NPE. This check fixes it but perhaps this should go upstream?

i386:
@michaelneale this is going to make it 1,000,000 times easier to find the correct runs of the ATH.

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/Export.java
```diff
@@ -158,6 +160,9 @@ public Object getValue(Property property, Object model, ExportConfig config) thr
                     printError(model.getClass(), e);
                     return SKIP;
                 }
+            } else if (model instanceof Item || model instanceof Run) {
```
i386: @vivek I have skipped serialisation of `Item` and `Run` objects in case they are returned by a `Cause` object. These are evil for performance reasons anyway.

i386:
@cliffmeyers this is the PR with the hanging tests

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/Export.java
vivek: Its a hack but should be fine until we find something else pulling in something else during serialization thats not item or run. We can revisit if its trouble later on.

i386:
> We could have used causeAction directly instead of wrapping in a backend abstraction, no?
@vivek only if you are happy with `Cause` leaking out into blueocean-rest's public API - which I assume we do not?

vivek:
@i386 I meant using action mechanism, causeAction is served via actions element. Well, its now available only via tree query, which we don't have it yet.  I guess, we can deal with Cause this way and handle other future actions via action mechanism.

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/FilteringTreePruner.java
i386: Fixed https://github.com/stapler/stapler/pull/112

vivek: 👍

i386:
@vivek we would need to pass the cause through to the client everywhere in the system... (everywhere we use `BlueRun`) would be easier when we get @kzantow decorated extensions that automatically fetch action's for us.

vivek:
@i386 Sounds good 👍

i386:
@vivek fixed the comments about naming here.

i386:
@cliffmeyers @tfennelly in order to merge I need to get this change to jsbuilder merged and a release made https://github.com/tfennelly/jenkins-js-builder/pull/2

i386:
Requires jsbuilder PR to be merged, released then bumped across all modules.

i386:
ATH https://ci.blueocean.io/job/ATH-Jenkinsfile/job/master/2273/

michaelneale:
:bee: ATH is happy

PR #1000, opened 2017-04-24
Update build.sh

michaelneale:
cc 
fix for: https://issues.jenkins-ci.org/browse/INFRA-1166

to only publish non beta. 

@rtyler @i386


rtyler:
If it works? *shrug*

PR #1001, opened 2017-04-24
Remove deprecated plugin-dev reference

EcutDavid:
As we talked in gitter, `blueocean-sample-pipeline-result-ext-plugin` is deprecated😃


i386:
LGTM 💯

PR #1004, opened 2017-04-26
Make aggregator work with new plugins extensing blueocean

vivek:
New plugins that define dependency on blueocean aggregator fail to start
duing `mvn hpi:run` with errors:

```
WARNING: Failed to load
org.jenkinsci.plugins.workflow.steps.scm.GenericSCMStep$DescriptorImpl
java.lang.NoClassDefFoundError:
org/jenkinsci/plugins/structs/describable/UninstantiatedDescribable
  at java.lang.Class.getDeclaredMethods0(Native Method)
```
Reason it fails is due to order in which dependencies are resolved by
mvn hpi:run where it resolves structs 1.2 during loading scm-api which
depends on junit and it depends on structs 1.2. But structs 1.2 doesn't
have UninstantiatedDescribable.

I have this github project to demonstrate that it works: https://github.com/vivek/sample-blueocean-extension/blob/master/pom.xml#L70-L76

# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/pom.xml
+++ b/blueocean-rest-impl/pom.xml
@@ -35,6 +35,17 @@
             <groupId>${project.groupId}</groupId>
             <artifactId>blueocean-jwt</artifactId>
         </dependency>
+
+        <!--
+            Junit plugin loads structs 1.2 during mvn hpi:run for plugins depending on blueocean aggregator.
+            We tell mvn hpi:run to load later version of structs.
+
+            TODO: remove when junit plugin dependency is removed
+        -->
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>structs</artifactId>
+        </dependency>
         <dependency>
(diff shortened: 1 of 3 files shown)


michaelneale:
@vivek could we perhaps also have a PR for the junit plugin to update its dependency so this possibly could be temporary? Longer term would be nice to not need to do this ;)

vivek:
@michaelneale Agreed, I think there is no pressing need to make this change, better to get it fixed properly or without such side effects. I will try couple more things and get back.

michaelneale:
So - worth getting this in now? Or we should hope for an update to junit to correct this?

vivek:
We should close it. There is PR to move junit specific code over to junit plugin: https://github.com/jenkinsci/junit-plugin/pull/68. Once it's merged we can remove dependency on junit plugin and that should fix this issue.

michaelneale:
that other PR looks like it is a like it is a little way away. In any case, this isn't urgent.

PR #1005, opened 2017-04-26
JENKINS-37324# Step description feature

vivek:
# Description

See [JENKINS-37324](https://issues.jenkins-ci.org/browse/JENKINS-37324).

Follow up PR for https://github.com/jenkinsci/workflow-api-plugin/pull/26. Its a WIP.

@svanoort  looks great. As you mentioned metastapes/build wrappers do not work yet. Probably for this reason GitSCM class name appears in scm checkout phase. Also in post build actions in another image below.

Below images are result of Jenkinsfile at https://github.com/i386/app-store-demo/blob/master/Jenkinsfile.

image

image

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


vivek:
cc: @i386 @michaelneale

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
```diff
@@ -68,6 +69,10 @@ public String getId() {
 
     @Override
     public String getDisplayName() {
+        String desc = ArgumentsAction.getArgumentDescriptionString(node.getNode());
```
i386: Could we instead move thus to a `getDisplayDescription` so that we can separately display the type of the step and the description?

e.g. the UI should show "Shell: mvn clean install"

michaelneale:
nice - the scm checkout looks wrong though: 

screen shot 2017-04-26 at 11 46 57 am

Is this a problem with that step ?? or is it how blue ocean is using this api (it is not a helpful description, so I would hope it was null and then fall back to saying "Checkout scm" or whatever it was)

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
i386: You can then change `Step.jsx` line no. 148 to display that format.

svanoort:
Looks reasonable to me so far.

$class buildsteps and wrappers should be fixed if you pull in the new workflow-cps alpha I cut: 2.30-stepstorage2-alpha2

Mind you, if an argument takes multiple steps or does not return a useful toString result you'll need to implement the interface in your StepDescriptor to make it fully useful (this will be the case for the Git step).  

As a workaround until appropriate PRs can be merged for Steps, one can read parameters directly.  I expect that to be needed for the Git plugin due to the volume of PRs to the maintainer and the impact on review timelines.

vivek:
@svanoort I tried 2.30-stepstorage2-alpha2 but scm checkout still gives GithubSCM class's toString() value. In debugger descriptor in this case is not StepArgumentsFormatter, and then later on the value being GithubSCM instance, it's toString() is returned.

image

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
vivek: Done, UI changes went in logDisplayHelper.js. Probably it needs bit of styling, if we want to show step type as bold or differently to distinguish from description, so maybe ResultItem in JDL needs to change.

image

i386:
@vivek much nicer :) where do the names "general scm" etc come from?

vivek:
@i386 it's step type, comes from step.getDisplayName().

michaelneale:
The `toString()` is somehow still in play with the general scm step which kind of harshes the vibe

i386:
@svanoort whats the method for adding more descriptions to other steps?

svanoort:
@i386 Make the StepDescriptor implement StepArgumentsFormatter -- as you requested, this is an API that can be implemented by plugin authors, rather than centralized within BlueOcean.

i386:
@svanoort sweet!! 👍

svanoort:
@michaelneale It's a consequence of having a single argument supplied to GenericSCMStep (the SCM).   To solve it, the GeneralSCMStep StepDescriptor needs to implement StepArgumentsFormatter and do something more with the SCM.

The basic implementation in ArgumentsAction.getArgumentDescriptionString is just supposed to handle the simplest case (single argument) and reduce the amount of work needed for steps.  It's an added extra to reduce the level of effort needed, but **not** intended to be a catch-all.

brody:
success
## Logic
### Has Description
- Description on left, black @ 100% opacity
- Type to the right, with em dash, black @ 50% opacity (or grey color)
### Has no Description
- Type on left, black @ 100% opacity

https://zpl.io/1MtiWN

michaelneale:
@svanoort @vivek right - so I guess that is a fairly "big" edge case, there is no case in which a generic toString() will ever be useful that I can think of. Should toString even be used at all?

svanoort:
@vivek @michaelneale Handled now.    Pull in the stepstorage4-beta versions of the plugins -- and StepArgumentsFormatter interface is no longer used. 

Instead the StepDescriptor implements `argumentsToString(Map)`

jglick:
> It's a consequence of having a single argument supplied to GenericSCMStep (the SCM). To solve it, the GeneralSCMStep StepDescriptor needs to implement StepArgumentsFormatter and do something more with the SCM.

Yes (for example calling `SCM.getKey()`), though also cf. [JENKINS-37227](https://issues.jenkins-ci.org/browse/JENKINS-37227).

vivek:
@svanoort With updated dependencies it looks good. 

image

i386:
@vivek no strange flashing of the names or anything like that when we run the pipeline? If so 🐝

Nice work @svanoort!

vivek:
@i386 Nothing I noticed, there is this bit about clicking on step makes it disappear/appear but thats not related to this change.

i386:
@vivek @svanoort if this looks good can we spin a release?

michaelneale:
@vivek @svanoort so are the dependencies ready yet? we can continue oni this any time (just need some UI tweaks to finalise this).

vivek:
@michaelneale @i386 https://github.com/jenkinsci/workflow-api-plugin/pull/26 not merged yet, is approved though. As soon as it's merged and a release is made for all related workflow api we can proceed. Are we doing UI tweaks as part of this PR or different?

michaelneale:
@vivek either or - once they are ready, update this to master as well, then we can look at it first thing (probably next week).

vivek:
@michaelneale sounds good.

svanoort:
@michaelneale  @vivek  After much ado due to getting stonewalled by review, the Step API and Workflow-api should be releasable (as was noted above).   From your end there are some class/method renames to consume the finalized version but nothing else.   If I'm permitted to release the APIs without the concrete impl, I'll aim to do that tomorrow, if not when workflow-cps is done. 

The final workflow-cps implementation has absorbed a few more torpedoes in review due to needing to support future things Jesse wants to do (but which don't exist as of yet), which mandate a deeply recursive implementation.  However so far it is manageable and shouldn't be too much longer.

vivek:
@svanoort Sounds good. I will wait for implementation bits to arrive then test them about together.

michaelneale:
@svanoort so -there are some bits published now?

michaelneale:
@vivek may want to get this up to date with master

michaelneale:
@vivek ok this looks like it may be ready to continue. Sam has published: 

```
workflow-step-api: 2.10
workflow-api: 2.15
workflow-cps: 2.31

The core APIs: 
* ArgumentsAction.getArguments(FlowNode) - returns the Map<String,Object> of user-supplied arguments to the Step you ran in the FlowNode (empty if not stored).

* ArgumentsAction.getStepArgumentsAsString(FlowNode) - return a handy String for UI display.  Provides an initial implementation for the simplest case (single parameters which toString gives useful results for).  More complex cases are handled by overriding StepDescriptor#argumentsToString in the StepDescriptor implementation. 

* Additional APIs on ArgumentsAction to support returning single Step arguments by argument name. 
```

vivek:
@michaelneale @svanoort I will pick it up tomorrow, great work!

vivek:
Looks good. There is one change, declarative 'junit' step `junit '**/target/surefire-reports/TEST-*.xml'` appears as `General Build Step`. It used to be `@junit(<anonymous>=**/target/surefire-reports/TEST-*.xml)`, it was not great either but something changed that made it appear as `General Build Step`.

@i386 @michaelneale feel free to try it out, I guess there is some style/tweaking needed?

svanoort:
@vivek That's because of a change @jglick requested to support future planned changes.  Specifically he wanted to remove automatic unwrapping of the nested step arguments with metasteps (such as Core steps and CoreWrapperSteps that internally wrap a BuildStep or SimpleBuildWrapper).  

The reason was that there are plans to add additional parameters besides the internal Step, and unwrapping would create an ambiguous situation. 

HOWEVER it's fixable by a simple PR to the workflow-basic-steps plugin, so don't sweat it yet (the reason I agreed to this, otherwise would probably have pushed back due to the API usability impact).

michaelneale:
@vivek I have updated this to master - but there are various failures that may be worth looking in to, certainly seems an unstable build with these changes. The fact the errors happen in different places is a bit unsettling.

vivek:
@michaelneale Not sure which failures you are referring to. Can you paste them? Are you referring to the same error as below?

I fired up agains existing workspace and notice there are warning reported (incorrectly) as there is existing heads for previous builds. Although nodes/steps etc appear ok.

@svanoort Looks like there is timing issue where `SyntheticStageGraphListener.GraphListenerAction.onLoad(Run)` is called before pipeline heads are read off disk? Our visualization appear fine, although it looks confusing due to these stack traces appearing for no apparent reason.

```
31, 2017 4:22:32 PM org.jenkinsci.plugins.workflow.cps.CpsFlowExecution getCurrentHeads
WARNING: null
java.lang.IllegalStateException: List of flow heads unset for CpsFlowExecution[null]
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.getCurrentHeads(CpsFlowExecution.java:794)
	at org.jenkinsci.plugins.workflow.flow.FlowExecution.isComplete(FlowExecution.java:176)
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.isComplete(CpsFlowExecution.java:1016)
	at org.jenkinsci.plugins.pipeline.modeldefinition.SyntheticStageGraphListener$GraphListenerAction.attachListener(SyntheticStageGraphListener.java:102)
	at org.jenkinsci.plugins.pipeline.modeldefinition.SyntheticStageGraphListener$GraphListenerAction.onLoad(SyntheticStageGraphListener.java:87)
	at hudson.model.Run.onLoad(Run.java:349)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.onLoad(WorkflowRun.java:562)
	at hudson.model.RunMap.retrieve(RunMap.java:224)
	at hudson.model.RunMap.retrieve(RunMap.java:56)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.load(AbstractLazyLoadRunMap.java:479)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.load(AbstractLazyLoadRunMap.java:461)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.getByNumber(AbstractLazyLoadRunMap.java:367)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.search(AbstractLazyLoadRunMap.java:332)
	at jenkins.model.lazy.LazyBuildMixIn$RunMixIn.getPreviousBuild(LazyBuildMixIn.java:355)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.getPreviousBuild(WorkflowRun.java:204)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.getPreviousBuild(WorkflowRun.java:123)
	at hudson.model.RunMap$1.next(RunMap.java:109)
	at hudson.model.RunMap$1.next(RunMap.java:98)
	at io.jenkins.blueocean.service.embedded.rest.RunSearch.collectRuns(RunSearch.java:123)
```

michaelneale:
@vivek looking back through the failures for the last few runs (I have rerun it a few times) - sometimes the failure is ATH, sometimes integraiton tests. Running another one now.

vivek:
@michaelneale Ok, I ran tests locally and they seem to run fine. Let's see if this run goes thru. Also good to get that warning and stack trace printed in console.

michaelneale:
@vivek right - not back to ATH failures which are probably not unexpected (log-karaoke tests - should be able to verify locally without too much trouble).

michaelneale:
ah yes, many of the tests look for step names - which now fail as they names are a little different, hence ATH failures. Easy enough to adjust.

michaelneale:
yay @vivek I have updated this to master, fixed those small tests. Looking good

@i386 can take a look sometime - I think there are a few small visual tweaks but this is working fantastic IMO

michaelneale:
@i386 this is looking quite good - given the time, I would consider it for 1.1 if you think acceptable. Mostly its a bump of a few versions.

i386:
@michaelneale your call. I'm happy to ship without the front end change (still needs to be completed for maximum awesomeness)

abayer:
@vivek That's due to https://issues.jenkins-ci.org/browse/JENKINS-43055, which has a fix pending that'll be in before we do a non-beta release of Declarative with the parallel stages stuff.

abayer:
Also, yes, that `properties` step...I need to wedge it inside a synthetic stage. Will deal with that.

michaelneale:
@i386 if you gave it more testing - I think it could next week depending on when the release goes out. I don't mind either way. If not sure - leave it for a point release.

i386:
@michaelneale @vivek I tried this out and it works like a charm 🐝

PR #1010, opened 2017-04-27
JENKINS-42809# Fix for correctly showing commitId

vivek:
# Description

See [JENKINS-42809](https://issues.jenkins-ci.org/browse/JENKINS-42809).

## Branch with correct commit id:

image

## Activity with correct commit id

image

## Dashboard with favorites
image

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
@@ -3,11 +3,10 @@ package io.jenkins.blueocean.rest.impl.pipeline;
 import hudson.Extension;
 import hudson.model.Queue;
 import hudson.model.Run;
-import hudson.plugins.git.Revision;
-import hudson.plugins.git.util.BuildData;
 import hudson.scm.ChangeLogSet;
 import hudson.scm.ChangeLogSet.Entry;
 import io.jenkins.blueocean.commons.ServiceException;
+import io.jenkins.blueocean.rest.Navigable;
 import io.jenkins.blueocean.rest.Reachable;
 import io.jenkins.blueocean.rest.annotation.Capability;
 import io.jenkins.blueocean.rest.factory.BlueRunFactory;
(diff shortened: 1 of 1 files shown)


michaelneale:
LGTM - :bee: - go for it.

i386:
@vivek did you want a review from @jglick before merging?

vivek:
@i386 yeah, want to make sure we are using the correct action to retrieve revision. Back then I was told use BuildData action but it doesn't work.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
```diff
@@ -155,13 +154,9 @@ public void stop() {
...
     public String getCommitId() {
-        BuildData data = run.getAction(BuildData.class);
-
+        SCMRevisionAction data = run.getAction(SCMRevisionAction.class);
         if (data != null){
-            Revision revision = data.getLastBuiltRevision();
-            if(revision != null){
-                return revision.getSha1String();
-            }
+            return data.getRevision().toString();
```
jglick: Of course the details of what `SCMRevision.toString` will return is up to the SCM. For Git this will generally be a commit hash, though it will be something else in the case of a GitHub PR merge.

vivek: @jglick in that case what's the right thing to use? How can we get checked out repo's commit hash?

vivek:
Summary of discussing this PR with @jglick:

- [SCMRevision.toString()](https://github.com/jenkinsci/scm-api-plugin/blob/9e3e34ede2c371c1411e951be0808247a87c18de/src/main/java/jenkins/scm/api/SCMRevision.java#L65-L72) use is fine as long as it's not used to do any SCM related operation. That is its used for display/informational purpose.
- Github PR implementation of `SCMRevision.toString()` might return combination of [base branch and pull branch hash](https://github.com/jenkinsci/github-branch-source-plugin/blob/24c1589157b118e21c925630375dc8d29d48c255/src/main/java/org/jenkinsci/plugins/github_branch_source/PullRequestSCMRevision.java#L83)
- Mercurial implementation of `SCMRevision.toString()` didn't have implementation of it, Jesse opened a PR https://github.com/jenkinsci/bitbucket-branch-source-plugin/pull/45
- Other SCMs when supported by blueocean will need to have proper implementation of `SCMRevision.toString()`
- `WorkflowRun.getChangeSet()` won't work as it returns empty in many conditions (after reindexing, first checkout etc.). Perhaps after https://issues.jenkins-ci.org/browse/JENKINS-33274 is fixed, we can use it somehow to detected checked out repo's revision. 

@michaelneale @i386 Basically for git/github case we are ok, mercurial after this PR merge will be fine as well. Since activity/branch tab show repo checkout commit hash, this is the only way at present to do. If we agree with this approach we can go ahead with this fix.

i386:
LGTM

i386:
🐝

PR #1019, opened 2017-04-28
Jenkins 43935 fix language tag of head to output returns a well formed ietf bcp 47 language tag representing this locale

scherler:
# Description

See [JENKINS-43935](https://issues.jenkins-ci.org/browse/JENKINS-43935).

Exposes as well the static resource of jenkins `data-resurl` in a new method in core-js

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
this has a corejs change but the dependency is not bumped (also ATH didn
t seem totaly happy last I looked) - confused?

scherler:
@michaelneale Yeah correct if you want we can bumb up the version and do the version dance, but in the end I did not need that method published and could wait until the next publish cycle. 

Regarding ATH I will have a look, but if we do not validate the language attribute (what would surprise me) the ATH failure shall not be related.

michaelneale:
might just need updating to master as there was a change due to activity tables that the ATH is likely picking up - that is probably all.

scherler:
@michaelneale will do, thanks for the headsup

scherler:
ATH test: https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/master/2335/pipeline

scherler:
@michaelneale ATH passes 
buc6whmcaaascgs

blueocean-core-js/src/js/config.js
```diff
@@ -85,6 +87,13 @@ export default {
         return (typeof config.jenkinsRootURL === 'string' ? config.jenkinsRootURL : '/jenkins');
     },
 
+    getResourceURL() {
+        if (!config.isLoaded) {
+            this.loadUrls();
+        }
+        return (typeof config.resourceUrl === 'string' ? config.resourceUrl : '/jenkins/static');
```
michaelneale: for some reason, typeof + string makes me remember some really exotic bug we had... and undefined... wish I could remember.

michaelneale:
@scherler what is IETF_BCP_47 ? as you are multilingual, I will take your word for it that it is the Right Thing To do, but curious. 

Otherwise. LGTM :bee:

blueocean-core-js/src/js/config.js
scherler: @michaelneale the above is a replicate from e.g. `getJenkinsRootURL` https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-core-js/src/js/config.js#L85 I guess that it was the solution to the weird bug ;)

scherler:
@michaelneale https://tools.ietf.org/html/bcp47 is a working paper where you find all references to the related ISO standard. It is a working paper on "best current practise" BCP

# Abstract

   This document describes the structure, content, construction, and
   semantics of language tags for use in cases where it is desirable to
   indicate the language used in an information object.  It also
   describes how to register values for use in language tags and the
   creation of user-defined extensions for private interchange.

PR #1024, opened 2017-05-01
Revert "Fix bug where tests where not being filtered by state correctly."

RahmadaniBayuAji:
Reverts jenkinsci/blueocean-plugin#1021


michaelneale:
this account will shortly be banned by github

PR #1032, opened 2017-05-05
ATH reorganize

imeredith:
# Description
This PR puts the ath in `acceptance-tests` folder. To run it, first `npm install` in that folder (will be handled automatically with additions in the pipe but for now just need to do it once at the start). Use the normal ath command `./run.sh -a ...`.  If running in dev mode, then you can run nightwatch manually as normal from the `acceptance-tests` folder.

I am mainly looking for a sanity check. The tests were copied verbatum so there should be no problem there. I have made changes to Jenkinsfile which now runs ath in the main build. This means PR's automatically get tested against ath and reported back to github. 

 ## Changes

- Moved ath to main repo
- Split out ATH runner code from tests
_ ATH infra is now in acceptance-tests/runner
- run.sh has been split into multiple files
- Logged in user support exists but is disabled for now due to existing creation tests.
- Jenkinsfile updated to run ath directly.

See [JENKINS-42953JENKINS-42953](https://issues.jenkins-ci.org/browse/JENKINS-42953X).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


Jenkinsfile
```diff
@@ -4,25 +4,36 @@
...
-
+  stage 'Setup'
   deleteDir()
   checkout scm
   sh 'docker build -t blueocean_build_env --build-arg GID=$(id -g ${USER}) --build-arg UID=$(id -u ${USER}) - < Dockerfile.build'
 
   configFileProvider([configFile(fileId: 'blueocean-maven-settings', targetLocation: 'settings.xml')]) {
 
-  docker.image('blueocean_build_env').inside {
+  sh "./acceptance-tests/runner/scripts/start-selenium.sh"
```
michaelneale: how come this runs here? why not later?

Jenkinsfile
```diff
@@ -4,25 +4,36 @@
...
-        triggerATH();
+      
+        stage 'Sanity check dependancies'
+        sh "node ./bin/checkdeps.js"
+        stage 'Sanity check shrinkwrap'
+        sh "node ./bin/checkshrinkwrap.js"
+        stage 'Archive results'
+        
+        stage 'ATH'
+        sh 'cd acceptance-tests && npm install'
```
michaelneale: shoudnt' selenium start here?

Jenkinsfile
```diff
@@ -31,6 +42,7 @@ node {
         }
       } finally {
         sendhipchat()
+        sh "./acceptance-tests/runner/scripts/stop-selenium.sh"
```
michaelneale: *really* want to make sure it stops this hey ;) this seems the right place... I need to compare with ATH jenkinsfile

Jenkinsfile
imeredith: It starts a docker cotnainer which has to share the network with the main container

Jenkinsfile
imeredith: see above

michaelneale:
@imeredith 
```
./acceptance-tests/runner/scripts/start-selenium.sh: line 3: ./stop-selenium.sh: No such file or directory

 Starting Selenium Docker container...

docker: Error response from daemon: Conflict. The name "/blueo-selenium" is already in use by container 
```

I wonder if there is something that didn't clean it up - I think it needs to somehow shut it down. 
Looking good, but the Jenkinsfile changes are actually the riskiest part here and need a lot of review. I am not sure it is cleaning up after itself right yet.

Jenkinsfile
michaelneale: ah ok. That is failing right now due to a name conflict, so I think it isn't shutting it down correctly (the start could be more defensive and stop it if it is running)

michaelneale:
Some TODOs as discussed on hangout and more: 

* Stop selenium needs to have correct path (and work!) so build doesn't fail to start
* .gitignore needs to include .blueocean-ath-jenkins-url (or it needs to be cleaned up) 
* PULL_REQUEST_TEMPLATE from ath dir can be deleted
* README probably needs a rewrite to talk about running npm install once (not mvn) etc. 

I *think* that was all...

Jenkinsfile
michaelneale: Nice

acceptance-tests/runner/.blueocean-ath-jenkins-url
```diff
@@ -0,0 +1 @@
+http://127.0.0.1:62969
```
michaelneale: this probably should be ignored (or cleaned up)

acceptance-tests/runner/.gitignore
```diff
@@ -0,0 +1,23 @@
+#Ignore all dot files
```
michaelneale: not sure if this file works - can you have multiple gitignores? (as there is on at the root for blue ocean)

acceptance-tests/runner/Jenkinsfile
```diff
@@ -0,0 +1,226 @@
+#!groovy
```
michaelneale: delete this file as not used and not needed now

acceptance-tests/runner/PULL_REQUEST_TEMPLATE
```diff
@@ -0,0 +1,16 @@
+# Description
```
michaelneale: delete this file as not needed

acceptance-tests/README.md
```diff
@@ -0,0 +1,115 @@
+Acceptance tests for Blue Ocean
```
michaelneale: This needs some editing so it is now correct for this structure (it is mostly ok though)

acceptance-tests/run.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
michaelneale: as discussed, perhaps this could run npm install? (maybe best to let user run it?)

i386:
Nice job

Jenkinsfile
```diff
@@ -4,25 +4,36 @@
 properties([buildDiscarder(logRotator(artifactNumToKeepStr: '20', numToKeepStr: '20'))])
 
 node {
-
+  stage 'Setup'
```
hanleyt: Using the 'stage' step without a block argument is deprecated. See [blog post here](https://jenkins.io/blog/2016/10/16/stage-lock-milestone/). Any reason why its still used here?

michaelneale:
@imeredith it looks like it runs the ATH twice lookigna t that stage view: ie shows up a large time for archiving results as well as running ATH - what is that about?

acceptance-tests/runner/scripts/maven.sh
```diff
@@ -0,0 +1,33 @@
...
+                echo "*****"
+                exit 1
+            fi
+            popd
+        fi
+    fi
+    echo asdfasdfad $PWD
+    echo "Assembling ATH dependency plugins (non Blue Ocean) ..."
+    pushd runner/runtime-plugins
+    mvn clean install -DskipTests
```
michaelneale: this should use settings.xml from the Jenkinsfile if at all possible - as it will use a faster cache. As it is now, it seems to be really really slow

Jenkinsfile
```diff
@@ -4,25 +4,36 @@
...
         step([$class: 'JUnitResultArchiver', testResults: '**/target/surefire-reports/TEST-*.xml'])
         step([$class: 'ArtifactArchiver', artifacts: '*/target/*.hpi'])
-
-        triggerATH();
+      
+        stage 'Sanity check dependancies'
+        sh "node ./bin/checkdeps.js"
+        stage 'Sanity check shrinkwrap'
+        sh "node ./bin/checkshrinkwrap.js"
+        stage 'Archive results'
```
michaelneale: this stage seems empty... a mistake? 

Also using block scope may help

acceptance-tests/runner/scripts/maven.sh
michaelneale: If this uses -s ../settings.xml (or whatever) then it will run a lot faster, most likely (I have noticed ATH can slow down a lot probably due to this).

acceptance-tests/runner/nightwatch.conf.js
```diff
@@ -0,0 +1,65 @@
+module.exports = (function (settings) {
```
cliffmeyers: This file already exists in the "acceptance-tests" dir. Can one of them be deleted?

acceptance-tests/runner/scripts/maven.sh
michaelneale: hrm - actually this needs to depend on context. on CI we need it to use proxy, but not locally (to be consistent in both cases).

Jenkinsfile
imeredith: The last time I used stages it was the normal way :) Will fix

acceptance-tests/runner/nightwatch.conf.js
michaelneale: yes I think this one can be deleted.

acceptance-tests/runner/runtime-plugins/fake-securitly-realm-plugin/src/main/java/org/jenkinsci/plugins/fakerealm/Listerner.java
```diff
@@ -0,0 +1,87 @@
...
+import org.acegisecurity.userdetails.UserDetails;
+import org.acegisecurity.userdetails.UsernameNotFoundException;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.springframework.dao.DataAccessException;
+
+import java.io.IOException;
+import java.util.*;
+
+@Extension
+public class Listerner  extends ComputerListener{
```
michaelneale: I also often make typos on this name ;)

cliffmeyers:
@imeredith @michaelneale  it would also be nice if there was a single command we could run that (1) builds blueocean-plugin running all JS and Java unit tests, and (2) runs the ATH suite. That's the kind of command I'd like to run when I break for lunch, or finish up for the day, since it's time-consuming.

I suppose we could easily do it with a shell script but then we always have cross platform issues. A Node.js script might be more portable. Or could we conditionally include acceptance-tests as a Maven module via a profile activated via a command line arg? I guess since all the heavy listing is done via a shell script anyways we'd need to do some kind of exec-maven-plugin binding; maybe that's too fancy?

michaelneale:
@cliffmeyers I think ATH will always require some hand holding cross platform anyway. It won't run on windows at the moment and probably wont' for some time due to selenium needs.

michaelneale:
@imeredith I think this is close - really need to have the proxy setup for maven.sh when running in CI, and the other doco/tweaks mentioned above to make it clearer (and deleting unneeded files). Once that - I think we can get this in and hopefully have smoother sailing.

Jenkinsfile
```diff
@@ -30,7 +30,7 @@ node {
         stage 'Archive results'
         
         stage 'ATH'
-        sh "cd acceptance-tests && ./run.sh -a=../blueocean/ --no-selenium"
+        sh "cd acceptance-tests && ./run.sh -a=../blueocean/ --no-selenium --settings=settings.xml"
```
michaelneale: will that pass it through to the mvn that run calls??

Jenkinsfile
michaelneale: yes this is def a :bug:

michaelneale:
@imeredith it isn't picking up settings.xml - build failing with: 

```
[ERROR] Unknown lifecycle phase "../settings.xml". You must specify a valid lifecycle phase or a goal in the format <plugin-prefix>:<goal> or <plugin-group-id>:<plugin-artifact-id>[:<plugin-version>]:<goal>. Available lifecycle phases are: validate, initialize, generate-sources, process-sources, generate-resources, process-resources, compile, process-classes, generate-test-sources, process-test-sources, generate-test-resources, process-test-resources, test-compile, process-test-classes, test, prepare-package, package, pre-integration-test, integration-test, post-integration-test, verify, install, deploy, pre-clean, clean, post-clean, pre-site, site, post-site, site-deploy. -> [Help 1]
```

michaelneale:
```
[blueocean_neo-ath-U2IJUSMDP6OWYTZG6JVV4XKRX57W7RRFKFACIJXBJRXCY7NRMQQA] Running shell script
/home/ubuntu/slave/workspace/blueocean_neo-ath-U2IJUSMDP6OWYTZG6JVV4XKRX57W7RRFKFACIJXBJRXCY7NRMQQA@tmp/durable-7b2efff4/script.sh: 2: /home/ubuntu/slave/workspace/blueocean_neo-ath-U2IJUSMDP6OWYTZG6JVV4XKRX57W7RRFKFACIJXBJRXCY7NRMQQA@tmp/durable-7b2efff4/script.sh: Syntax error: Unterminated quoted string
```

michaelneale:
lol maybe next time @imeredith ?

```
[ERROR] Error executing Maven.
[ERROR] The specified user settings file does not exist: /home/ubuntu/slave/workspace/blueocean_neo-ath-U2IJUSMDP6OWYTZG6JVV4XKRX57W7RRFKFACIJXBJRXCY7NRMQQA/acceptance-tests/runner/runtime/../settings.xml
```

Jenkinsfile
```diff
@@ -30,7 +30,7 @@ node {
         stage 'Archive results'
         
         stage 'ATH'
-        sh "cd acceptance-tests && ./run.sh -a=../blueocean/ --no-selenium --settings='-s ../settings.xml''"
+        sh "cd acceptance-tests && ./run.sh -a=../blueocean/ --no-selenium --settings='-s ../settings.xml'"
```
michaelneale: perhps this should be expanded to be the full path so it is the same whereever it runs

michaelneale:
just checking this is ok with master...

PR #1036, opened 2017-05-08
Simplify create process for into AbstractMultiBranchCreateRequest

i386:
# Description

This change makes it very easy for people to add create requests for multibranch pipelines with non-git SCMSources 

See https://github.com/jenkinsci/blueocean-plugin/pull/1037 for an implementation.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/AbstractPipelineCreateRequest.java
```diff
@@ -41,6 +48,6 @@
         if(!acl.hasCreatePermission(a, p, descriptor)){
             throw new ServiceException.ForbiddenException("Missing permission: " +Item.CREATE.group.title+"/"+Item.CREATE.name + Item.CREATE + "/" + descriptor.getDisplayName());
         }
-        return parent.createProject(descriptor, name, true);
+        return Jenkins.getInstance().createProject(descriptor, name, true);
```
vivek: `parent.createProject()` is fine, at the abstract level, better for client to decide where to create this project.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/AbstractMultiBranchCreateRequest.java
```diff
@@ -0,0 +1,130 @@
...
+import org.jenkinsci.plugins.workflow.multibranch.WorkflowMultiBranchProject;
+
+import javax.annotation.Nonnull;
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Creates {@link MultiBranchProject}s with a single {@link SCMSource}
+ */
+public abstract class AbstractMultiBranchCreateRequest extends AbstractPipelineCreateRequest {
```
vivek: 👍

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/AbstractPipelineCreateRequest.java
i386: @vivek can always replace that call with a 
```
Folder getItemRoot() { ... }
```
... in the abstract class and ask the client to set it as a string on posting this request?

vivek: That works too.

blueocean-pipeline-api/src/main/java/io/jenkins/blueocean/pipeline/creation/AbstractMultiBranchCreateRequest.java
```diff
@@ -0,0 +1,132 @@
+package io.jenkins.blueocean.pipeline.creation;
```
vivek: Guess all these could be inside `api` package, like `io.jenkins.blueocean.pipeline.api`?

i386: @vivek sounds reasonable. Ive made the change. I also moved the credentials back into `io/jenkins/blueocean/rest/impl/pipeline/credential` package as I think they their fully qualified names get baked into xstream

vivek: Hmm,  I am going to check. We are using credentials apis for everything, why are our classes getting in there? I will check sometime tonight or tomorrow morning.

i386:
@vivek I think our Credential classes need to be moved out to another module

i386:
ATH https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/master/2376/pipeline

vivek:
@i386 LGTM 🐝 We can move credentials in to another module in separate PR.

PR #1043, opened 2017-05-10
JENKINS-38982 fast search

NicuPascu:
# Description

I included the magnifying glass icon as a base64 string in less, is that ok or should I include the svg as a separate asset

See [JENKINS-38982](https://issues.jenkins-ci.org/browse/JENKINS-38982).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
Ah, this involves a change to core-js - so won't be able to review this. 

@imeredith what is best way forward? this is kind of an area we want to avoid until we have the new plugin code sharing. 

An alternative, do the code in dashboard before moving it? That pager isn't use elsewhere (so not sure why it is in core-js)

blueocean-core-js/src/js/paths/rest.js
```diff
@@ -12,12 +12,11 @@ export default {
         return '/blue/rest';
     },
 
-    organizationPipelines(organizationName) {
-        return `${this.apiRoot()}/search/?q=type:pipeline;organization:${encodeURIComponent(organizationName)};excludedFromFlattening:jenkins.branch.MultiBranchProject,hudson.matrix.MatrixProject&filter=no-folders`;
-    },
+    pipelines(organizationName, searchText) {
```
michaelneale: hrm... this isn't really a very practical "first time" PR - (to @imeredith) - does the pipeline/page stuff really need to be in core-js? There doesn't seem to be another other modules using it (eg only dashboard changed). Is this premature abstraction?

blueocean-core-js/src/js/paths/rest.js
```diff
@@ -12,12 +12,11 @@ export default {
...
+        const startsWith = searchText ? `;startsWith:${encodeURIComponent(searchText)}` : '';
+        const organization = organizationName ? `;organization:${encodeURIComponent(organizationName)}` : '';
 
-    allPipelines() {
```
michaelneale: if changing the api, why not move the code into dashboard for now?

blueocean-core-js/src/js/services/PipelineService.js
```diff
@@ -25,27 +25,16 @@ export class PipelineService extends BunkerService {
     }
 
     /**
-     * Gets pager for /blue/pipelines
-     *
-     * @returns {Pager}
-     */
-    allPipelinesPager() {
```
michaelneale: Once again, this file is in core-js, but is only used by pager - why is it in core-js?

kzantow:
Regarding the SVG: we've been embedding them as react components instead of image assets. I'd just make a simple component that just returns that icon (there are also the Material icons.. but I can't seem to find a magnifying glass in them, what the heck?)

blueocean-core-js/src/js/services/PipelineService.js
kzantow: Yeah, this should not be in core-js - it just seems to be providing data for a specific page...

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineSearch.java
```diff
@@ -40,6 +44,7 @@
  */
 @Extension
 public class PipelineSearch extends OmniSearch<BluePipeline>{
+    private static final String STARTSWITH_PARAM="startsWith";
```
vivek: We do not need new param, there is already `pipeline` param, it should be used regex. Current code does equals comparison, should be using it as regex and collect all pipelines whose name match this regex 🐛

NicuPascu:
@michaelneale the PR can now be tested, the corejs changes have been pushed

@kzantow I want to use the icon as a background image so what is visible is actually the input, when clicked it expands because of the CSS :focus rule, it makes the code shorter and doesn't involve any js that would be necessary if we use a different element to trigger the visibility of the input. I base64ed the icon and it is in the dashboard.less file, it cannot be reused but for this case I think it is the cleanest way to implement the desired behaviour

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineSearch.java
imeredith: I dont see this in pipeline search.

blueocean-core-js/src/js/services/PipelineService.js
imeredith: I disagree, its its pulling it from a cache for a specific pipeline/search.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineSearch.java
vivek: @imeredith regex related code doesn't exists, thats what I am suggesting to be implemented instead of adding startsWith param.

vivek: @imeredith @NicuPascu  for example: if client wants to get all pipeline starting with 'blue' then it can send `/rest/search?q=type:pipeline;pipeline=^blue.*;...` then following code example can work:
```java
       String pipeline = q.param(getType()); //it should give string "^blue.*"
        if(pipeline != null) {
            Pattern pattern = Pattern.compile(pipeline);
            while (pipelineIterator.hasNext()) {
                BluePipeline p = pipelineIterator.next();
                if (!pattern.matches(pipeline).matches()) {
                    continue;
                }
                pipelines.add(p);
            }
            return Pageables.wrap(pipelines);
       }
```

i386:
@vivek @imeredith @NicuPascu perhaps if we enabled [glob pattern matching](https://en.wikipedia.org/wiki/Glob_(programming)) here along side regex? I think [org.apache.oro.text.GlobCompiler](https://svn.apache.org/repos/asf/jakarta/oro/trunk/src/java/org/apache/oro/text/GlobCompiler.java) can do the trick.

Given this list of Pipeline full names:
* Folder 1/Sub Folder/Foo
* Folder 1/Sub Folder/baz
* Folder 1/Another Sub Folder/Foobar'

You could search with `/rest/search?q=type:pipeline;pipeline=Folder 1/*` and get back
* Folder 1/Sub Folder/Foo
* Folder 1/Sub Folder/baz
* Folder 1/Another Sub Folder/Foobar'

You could search with `/rest/search?q=type:pipeline;pipeline=*/*Sub*` and get back
* Folder 1/Sub Folder/Foo
* Folder 1/Sub Folder/baz

etc

vivek:
@i386 👍

i386:
Given:
* Apps/Sub Folder/Foo
* Project/Sub Folder/baz
* Util/Another Sub Folder/Foobar'

and I type `F` that should translate to `*F*` which gets back only
* Apps/Sub Folder/Foo

i386:
Here's a basic implementation of a glob search https://github.com/jenkinsci/blueocean-plugin/pull/1056 

We still need to modify search so that we can rank the result.

i386:
@NicuPascu you should be able to use the [search](https://material.io/icons/#ic_search) icon that is in the `react-material-icons` package.

michaelneale:
looking good - the style loading seems broken though. Somehow blueocean-dashboard/src/main/less/dashboard.less is being loaded too late - and it looks really bad. 
Is there somewhere else this style should live?

michaelneale:
Some other feedback: 
* favourites need to be hidden when searching and in future filtered to match

i386:
Examples https://docs.google.com/a/cloudbees.com/document/d/1riU62MGFF7A9cInJx04gIwZQ8PGwN9MkQQORC3dGq40/edit?usp=sharing 

To search with glob
```
/search?q=type:pipeline;pipeline:*userinput*
```

michaelneale:
@i386 wasn't clear - did you want nicu to revert the server side changes from this and bring them over from the glob branch?

cliffmeyers:
@michaelneale I've noticed this style loading problem too: if you deep-link directly to "create pipeline" it does an ugly reflow because the UI renders immediately (not depending any initial fetch).

I thought js-modules would guarantee that CSS and i18n assets were loaded before the bundle actually executes. It seems way more sensible that we address the problem there, rather than trying to push assets into different files so they load faster. Perhaps @tfennelly can provide some input?

michaelneale:
could @NicuPascu @imeredith you get this up to date with master? (including the ATH stuff?)

NicuPascu:
JDL PR: https://github.com/jenkinsci/jenkins-design-language/pull/169

blueocean-dashboard/src/main/js/components/Pipelines.jsx
```diff
@@ -68,6 +91,10 @@ export class Pipelines extends Component {
                                 { organization && orgLink }
                             </h1>
                         </Extensions.Renderer>
+                        
+                        {fastSearchFeatureFlag &&
+                            <TextInput className="search-pipelines-input" iconLeft="search" defaultValue={this.getSearchText()} placeholder="Search pipelines..." onChange={this.onChange} />
```
cliffmeyers: Based on the discussion in jenkinsci/jenkins-design-language#170 I don't think we can get the correct fix for `TextInput` in a fast timeframe and I wouldn't want to block this feature because of that. You may want to change this back to a raw HTML `input` that uses the styles for `TextInput` but binds the `value` attribute manually.

michaelneale:
@cliffmeyers yes - did the discussion this morning shed some light on the path forward? 

@NicuPascu if you can update this to master etc - will make the build more friendly I think.

michaelneale:
some lint failures: 

```
[INFO] /home/ubuntu/slave/workspace/e_JENKINS-38982_fast_search-TYIRS2QESATX7LNGEGRSRCQYFY5LCK36AVLKCHITCRYSPGOMWJMA/blueocean-dashboard/src/main/js/components/Pipelines.jsx
[INFO]   18:24   error    There should be no spaces inside this paren           space-in-parens
[INFO]   22:5    error    _initPager should be placed after getSearchText       react/sort-comp
[INFO]   34:13   error    'getParams' is never reassigned, use 'const' instead  prefer-const
[INFO]   37:18   error    'i' is never reassigned, use 'const' instead          prefer-const
[INFO]   39:23   error    Expected '===' and instead saw '=='                   eqeqeq
[INFO]   40:84   error    Strings must use singlequote                          quotes
[INFO]   96:1    warning  Line 96 exceeds the maximum line length of 160        max-len
[INFO]   96:190  error    Trailing spaces not allowed                           no-trailing-spaces
```

michaelneale:
test failures: 

```
[INFO]   Pipelines
[INFO]     basic table rendering
[ERROR] [mobx] An uncaught exception occurred while calculating your computed value, autorun or transformer. Or inside the render() method of an observer based React component. These functions should never throw exceptions as MobX will not always be able to recover from them. Please fix the error reported after this message or enable 'Pause on (caught) exceptions' in your debugger to find the root cause. In: 'Pipelines#0.render()'. For more details see https://github.com/mobxjs/mobx/issues/462
[INFO]       1) "before each" hook for "check header to be as expected"
[INFO]     pending state
[INFO]       2) should continue to render existing data while a fetch is pending
[INFO]     duplicate job names
[ERROR] [mobx] An uncaught exception occurred while calculating your computed value, autorun or transformer. Or inside the render() method of an observer based React component. These functions should never throw exceptions as MobX will not always be able to recover from them. Please fix the error reported after this message or enable 'Pause on (caught) exceptions' in your debugger to find the root cause. In: 'Pipelines#0.render()'. For more details see https://github.com/mobxjs/mobx/issues/462
[ERROR] [mobx] An uncaught exception occurred while calculating your computed value, autorun or transformer. Or inside the render() method of an observer based React component. These functions should never throw exceptions as MobX will not always be able to recover from them. Please fix the error reported after this message or enable 'Pause on (caught) exceptions' in your debugger to find the root cause. In: 'Pipelines#0.render()'. For more details see https://github.com/mobxjs/mobx/issues/462
[INFO]       3) should render two rows when job names are duplicated across folders
[INFO] 
[INFO]   PullRequest should render
[INFO]     ??? does renders the PullRequest with data
[INFO] 
[INFO]   PullRequest should not render
[INFO]     ??? does renders the PullRequest without data
[INFO] 
[INFO]   PullRequest
[INFO]     ??? opens correctly
[INFO] 
[INFO]   PullRequests should render
[INFO]     ??? does renders the PullRequests with data
[INFO] 
[INFO]   PullRequests should not render
[INFO]     ??? does render NotSupported the PullRequests without data
[INFO] 
[INFO]   push events - queued run tests
[INFO]     ??? currently displayed pipeline
[INFO]     ??? not currently displayed pipeline
[INFO]     ??? ignore multiple events with same queueId
[INFO] 
[INFO]   push events - started run tests
[INFO]     ??? run fetch ok
[INFO]     ??? run fetch failed
[INFO] 
[INFO]   RunDescription
[INFO]     ??? displays description
[INFO]     ??? displays nothing when description unavailable
[INFO]     ??? displays nothing when run unavailable
[INFO] 
[INFO]   RunDetailsArtifacts
[INFO]     bad data
[INFO]       ??? renders nothing
[INFO]     valid artifacts
[INFO]       ??? renders a Table with expected data
[INFO] 
[INFO]   RunDetailsChanges
[INFO]     empty runs / bad data
[INFO]       ??? renders nothing
[INFO]     empty changeSet
[INFO]       ??? renders NoChangesPlaceholder
[INFO]     valid changeSet
[INFO]       ??? renders a Table with expected data
[INFO] 
[INFO]   RunIdCell
[INFO]     ??? displays name
[INFO]     ??? displays id
[INFO] 
[INFO]   RunMessageCell
[INFO]     ??? displays description
[INFO]     ??? displays with multiple commits
[INFO]     ??? displays with single commit
[INFO]     ??? displays cause because more than 1 cause is more important
[INFO]     ??? displays cause
[INFO]     ??? displays nothing
[INFO] 
[INFO]   smart-fetch
[INFO]     ??? Callbacks work
[INFO]     ??? Promises work
[INFO] blad
[INFO]     ??? Sends fetch status notifications
[INFO]     ??? Fetch multiple works
[INFO]     ??? Pagination works
[INFO] 
[INFO]   StatusIndicator should render
[INFO]     ??? does render success
[INFO] 
[INFO]   SvgStatus should render
[INFO]     ??? does render FAILURE
[INFO] 
[INFO]   SvgSpinner should render
[INFO]     ??? does render RUNNING
[INFO] 
[INFO]   Store should work
[INFO]     Log Store should work
[INFO]       ??? create store with run data
[INFO]     Store should work with steps
[INFO]       ??? create store with step data
[INFO]     Store should work with nodes
[INFO]       ??? create store with node data
[INFO] 
[INFO]   Redux Store - 
[INFO]     ??? create store with branch data
[INFO]     ??? multi-branch job_crud_job_created blueocean_is_for_current_job=true
[INFO]     ??? multi-branch job_crud_job_created blueocean_is_for_current_job=false
[INFO] 
[INFO]   TestResults
[INFO]     ??? Test fixed included (105ms)
[INFO]     ??? Handles REGRESSION case
[INFO]     ??? All passing shown
[INFO]     ??? All passing and fixed shown
[INFO] <div><div class="TestSummary u-failed"><svg fill="#000000" height="24" viewBox="0 0 24 24" width="24" xmlns="http://www.w3.org/2000/svg" class="TestSummary-icon"><path d="M19 6.41L17.59 5 12 10.59 6.41 5 5 6.41 10.59 12 5 17.59 6.41 19 12 13.41 17.59 19 19 17.59 13.41 12z"></path><path d="M0 0h24v24H0z" fill="none"></path></svg><div class="TestSummary-content"><h1 class="TestSummary-title">rundetail.tests.results.summary.failing_title</h1><p class="TestSummary-message">rundetail.tests.results.summary.failing_message</p></div></div><!-- react-empty: 9 --><div class="test-result-block existing-failure-block"><h4>rundetail.tests.results.errors.existing.count</h4><div class="result-item failure"><div class="result-item-head"><span class="result-item-icon result-bg failure"><svg width="28" height="34"><g transform="translate(14 18)" class="result-status-glyph"><g class="result-status-glyph"><polygon points="4.67 -3.73 3.73 -4.67 0 -0.94 -3.73 -4.67 -4.67 -3.73 -0.94 0 -4.67 3.73 -3.73 4.67 0 0.94 3.73 4.67 4.67 3.73 0.94 0"></polygon></g></g></svg></span><span class="result-item-title"><svg width="28" height="24" class="result-item-expando"><g transform="translate(14 12)"><g class="expando-glyph"><polygon points="-1.7,-5 3.3,0 -1.7,5 -2.9,3.8 1,0 -2.9,-3.8"></polygon></g></g></svg><span class="result-item-label">failure.TestThisWontFail - aPassingTest2</span><span class="result-item-extra-info">a few seconds</span></span></div><span></span></div><!-- react-empty: 27 --></div><!-- react-empty: 28 --><!-- react-empty: 29 --></div>
[INFO]     ??? unstable renders with no error message (50ms)
[INFO] 
[INFO]   Enum
[INFO]     construction
[INFO]       ??? should populate the keys and values
[INFO]     values
[INFO]       ??? should return an erray with all values
[INFO] 
[INFO]   GitFlowManager
[INFO]     _createNameFromRepoUrl
[INFO]       ??? should use last part of repository URL
[INFO]       ??? should discard the extension
[INFO] 
[INFO] 
[INFO]   99 passing (591ms)
[INFO]   3 failing
[INFO] 
[INFO]   1) Pipelines basic table rendering "before each" hook for "check header to be as expected":
[INFO]      TypeError: Cannot read property 'query' of undefined
[INFO]       at Pipelines.getSearchText (src/main/js/components/Pipelines.jsx:23:16)
[INFO]       at Pipelines._initPager (src/main/js/components/Pipelines.jsx:28:33)
[INFO]       at Pipelines.render (src/main/js/components/Pipelines.jsx:56:14)
[INFO]       at Object.allowStateChanges (node_modules/mobx/lib/mobx.js:634:15)
[INFO]       at Reaction.<anonymous> (node_modules/mobx-react/index.js:282:44)
[INFO]       at trackDerivedFunction (node_modules/mobx/lib/mobx.js:901:20)
[INFO]       at Reaction.track (node_modules/mobx/lib/mobx.js:1230:9)
[INFO]       at reactiveRender (node_modules/mobx-react/index.js:278:17)
[INFO]       at Pipelines.initialRender (node_modules/mobx-react/index.js:272:15)
[INFO]       at ReactCompositeComponentMixin._renderValidatedComponentWithoutOwnerOrContext (node_modules/react/lib/ReactCompositeComponent.js:796:32)
[INFO] 
[INFO]   2) Pipelines pending state should continue to render existing data while a fetch is pending:
[INFO]      TypeError: Cannot read property 'query' of undefined
[INFO]       at Pipelines.getSearchText (src/main/js/components/Pipelines.jsx:23:16)
[INFO]       at Pipelines._initPager (src/main/js/components/Pipelines.jsx:28:33)
[INFO]       at Pipelines.render (src/main/js/components/Pipelines.jsx:56:14)
[INFO]       at Object.allowStateChanges (node_modules/mobx/lib/mobx.js:634:15)
[INFO]       at Reaction.<anonymous> (node_modules/mobx-react/index.js:282:44)
[INFO]       at trackDerivedFunction (node_modules/mobx/lib/mobx.js:901:20)
[INFO]       at Reaction.track (node_modules/mobx/lib/mobx.js:1230:9)
[INFO]       at reactiveRender (node_modules/mobx-react/index.js:278:17)
[INFO]       at Pipelines.initialRender (node_modules/mobx-react/index.js:272:15)
[INFO]       at ReactCompositeComponentMixin._renderValidatedComponentWithoutOwnerOrContext (node_modules/react/lib/ReactCompositeComponent.js:796:32)
[INFO] 
[INFO]   3) Pipelines duplicate job names should render two rows when job names are duplicated across folders:
[INFO]      TypeError: Cannot read property 'query' of undefined
[INFO]       at Pipelines.getSearchText (src/main/js/components/Pipelines.jsx:23:16)
[INFO]       at Pipelines._initPager (src/main/js/components/Pipelines.jsx:28:33)
[INFO]       at Pipelines.render (src/main/js/components/Pipelines.jsx:56:14)
[INFO]       at Object.allowStateChanges (node_modules/mobx/lib/mobx.js:634:15)
[INFO]       at Reaction.<anonymous> (node_modules/mobx-react/index.js:282:44)
[INFO]       at trackDerivedFunction (node_modules/mobx/lib/mobx.js:901:20)
[INFO]       at Reaction.track (node_modules/mobx/lib/mobx.js:1230:9)
[INFO]       at reactiveRender (node_modules/mobx-react/index.js:278:17)
[INFO]       at Pipelines.initialRender (node_modules/mobx-react/index.js:272:15)
[INFO]       at ReactCompositeComponentMixin._renderValidatedComponentWithoutOwnerOrContext (node_modules/react/lib/ReactCompositeComponent.js:796:32)
[INFO] 
```

imeredith:
@NicuPascu The easiest way to fix this is to do a null check

blueocean-dashboard/src/main/js/components/Pipelines.jsx
```diff
@@ -14,24 +15,47 @@ const translate = i18nTranslator('blueocean-dashboard');
...
 @observer
 export class Pipelines extends Component {
-    componentWillMount() {
-        this._initPager(this.props);
+    onChange = debounce(e => {
+        this.context.router.push(`${this.props.location.pathname}${this.updateGetParam('search', e)}`);
+    }, 200);
+
+    getSearchText() {
+        return this.props.location.query.search ? decodeURIComponent(this.props.location.query.search) : '';
```
imeredith: You can either do a null check here, or you acn pass in a location prop into the component in the test such that...

```javascript
{
    location: {
        query: {}
    }
}
```

michaelneale:
@NicuPascu in the meantime you could work on the conflict for ResourceLoadTracker.js introduced from this commit to master https://github.com/jenkinsci/blueocean-plugin/commit/15ffcab6ab582ad4c9c6119905a9ffe4f94c3bbd#diff-40178297c2a0beb0515596e634a974e8

blueocean-core-js/src/less/header.less
```diff
@@ -1,3 +1,7 @@
+html {
+    overflow-y: scroll;
+}
+
```
cliffmeyers: This could even be an inline style in the blueocean-web index.jelly.

blueocean-core-js/src/less/header.less
```diff
@@ -128,3 +132,63 @@
...
+            }
+            &::-moz-placeholder {
+                color: rgba(255, 255, 255, 0.4);
+            }
+            &:-ms-input-placeholder {
+                color: rgba(255, 255, 255, 0.4);
+            }
+        }
+    }
+}
```
cliffmeyers: If this code is going to be put in core-js to work around the loading / timing issue, could we add a TODO and corresponding JIRA ticket that describes fixing the underlying problem?  Long term we really don't want to put CSS for dashboard-specific components in core-js.

i386:
@NicuPascu when this PR is merged https://github.com/jenkinsci/blueocean-plugin/pull/1102 you will be able to respect the placement of `/` to send the right globbing patterns as per [this document](https://docs.google.com/a/cloudbees.com/document/d/1riU62MGFF7A9cInJx04gIwZQ8PGwN9MkQQORC3dGq40/edit?usp=sharing).

This is looking really great btw 👍

i386:
@NicuPascu @cliffmeyers last time I played with this PR (sometime last week) it was looking really close. What is left to do? Is it possible to get this over the line this week?

i386:
@imeredith @cliffmeyers can you please give @NicuPascu a review of this PR today.

i386:
@NicuPascu "There are no pipelines that match <i>$pattern</i>"

blueocean-core-js/src/less/header.less
cliffmeyers: Please add a TODO here as a reminder that this code needs to get moved out.. and file a JIRA ticket to track the work.

blueocean-core-js/src/less/header.less
imeredith: At the very least it should be in web. So yes +1 for index.jelly

blueocean-dashboard/src/main/js/components/Pipelines.jsx
```diff
@@ -68,18 +72,29 @@ export class Pipelines extends Component {
...
-                            store={ this.context.store }
-                            router={ this.context.router }
-                        />
+                        {!this.getSearchText() &&
+                            <Extensions.Renderer
+                                extensionPoint="jenkins.pipeline.list.top"
+                                store={ this.context.store }
+                                router={ this.context.router }
+                            />
+                        }
```
imeredith: Not sure we want to be excluding extension points like this. Probably ok for now but really we should be providing the seaqrch test or soemthing to this extension point so that it can use it if needed. Not sure exactly.

i386: @imeredith think it could wait until we've got a use case?

i386:
@nicupascu is this functionally complete now? If so I'll start doing some testing today

NicuPascu:
Yes, all features, bugs and suggestions have been implemented, let me know if you find anything that I might have missed

PR #1046, opened 2017-05-10
Automatically republish if a new LTS is available

michaelneale:
# Description

This tags the image additionally with the LTS id. Thus in theory, if a new LTS is published before a new blue ocean, it will then publish a the same blue ocean but with the new LTS. 

I don't have the environment to test this in, so up to @rtyler to approve or reject this.

### DIFF
--- a/docker/official/build.sh
+++ b/docker/official/build.sh
@@ -5,9 +5,18 @@ cd $HERE
 
 BLUEOCEAN_VERSION=$(git for-each-ref --sort=-taggerdate refs/tags/blueocean-parent-\*  | grep -v 'beta' | head -1 |awk '{print $3 }' | sed 's/refs\/tags\/blueocean-parent-//')
 
+# Check we have the latest LTS for blue ocean to build on and get its image ID
+docker pull jenkinsci/jenkins:lts-alpine
+LTS_IMAGE_ID=$(docker images jenkinsci/jenkins:lts-alpine | sed -n 2p | awk '{ print $3 }')
+
+# Make an explicit version tag for both blue and LTS, like 1.0.1-3321321
+# We need this so if the LTS image changes OR blue ocean, we rebuild and published
+# A new LTS means there may be security fixes
+FULL_VERSION="$BLUEOCEAN_VERSION-$LTS_IMAGE_ID"
+
 # Check if the image already exists
-if docker pull jenkinsci/blueocean:$BLUEOCEAN_VERSION; then
-  echo "Image jenkinsci/blueocean:$BLUEOCEAN_VERSION already exists in Docker Hub"
+if docker pull jenkinsci/blueocean:$FULL_VERSION; then
+  echo "Image jenkinsci/blueocean:$FULL_VERSION already exists in Docker Hub"
   exit 0
 fi
 
@@ -18,5 +27,10 @@ docker build --rm --no-cache --pull \
 # Consider this build is the latest
 docker tag "jenkinsci/blueocean:$BLUEOCEAN_VERSION" jenkinsci/blueocean:latest
 
+# Tag this build with the full version tag
+docker tag "jenkinsci/blueocean:$BLUEOCEAN_VERSION" "jenkinsci/blueocean:$FULL_VERSION"
+
+# push it real good
 docker push "jenkinsci/blueocean:$BLUEOCEAN_VERSION"
+docker push "jenkinsci/blueocean:$FULL_VERSION"
 docker push jenkinsci/blueocean:latest


ydubreuil:
Changes are fine with me. One enhancement would be to use the Jenkins version instead of the Docker image id. Assuming `jq` is installed on the slave, it is possible to extract JENKINS_VERSION from image metadata with:

```
JENKINS_VERSION=$(docker inspect jenkinsci/jenkins:lts-alpine | jq '.[0].ContainerConfig.Env|.[]' | sed -n '/"JENKINS_VERSION=/s/^.*=\(.*\)"$/\1/p')
```

michaelneale:
@ydubreuil yes that was one option - but I wasn't sure if jq was available where this runs, and how much the data structure may change the deeper one gets - but yeah, that could work.

i386:
FWIW version string LGTM

michaelneale:
the downside is that this way there is a tag published that you don't really want people to use - it is just to keep "state". I think @ydubreuil's approach is purer and doesn't publish a tag that is less sensical (although in theory, no one cares about the blue ocean + jenkins exact version in the tag)

ydubreuil:
@michaelneale I'm a bit worried about the potential future breakage too but I don't think the JSON structure will change soon and I expect that JENKINS_VERSION variable is here for a long time.

michaelneale:
@rtyler ok - is jq available where this runs?

rtyler:
@michaelneale No, `jq` is not available in the default Ubuntu 14.04 or 16.04 distributions so it's not available. Our agents are basically just stock Ubuntu with `docker-engine` installed

rtyler:
For what it's worth, the Pipeline was started automatically today (May 10) at around ~12:00 UTC:

```
Started by timer
Checking out git https://github.com/jenkinsci/blueocean-plugin.git to read docker/official/Jenkinsfile
Fetching changes from the remote Git repository
Checking out Revision a986f24023c4192f94ce6fc5b28b2f4b1e7b1626 (refs/remotes/origin/master)
Loading library pipeline-library@master
Setting origin to https://github.com/jenkins-infra/pipeline-library.git
Fetching origin...
Fetching changes from the remote Git repository
Checking out Revision b500b8160a6296a223dc52eadab52b686a85978a (master)
[Pipeline] properties
WARNING: The properties step will remove all JobPropertys currently configured in this job, either from the UI or from an earlier properties step.
This includes configuration for discarding old builds, parameters, concurrent builds and build triggers.
WARNING: Removing existing job property 'Discard old builds'
WARNING: Removing existing job property 'Build triggers'
[Pipeline] node
Running on trusted-agent-2 in /home/jenkins/trusted-workspace/workspace/Containers/blueocean
[Pipeline] {
[Pipeline] stage
[Pipeline] { (Checkout)
[Pipeline] deleteDir
[Pipeline] checkout
Cloning the remote Git repository
remote: Counting objects
remote: Compressing objects
Receiving objects
Resolving deltas
Updating references
Checking out Revision a986f24023c4192f94ce6fc5b28b2f4b1e7b1626 (refs/remotes/origin/master)
[Pipeline] }
[Pipeline] // stage
[Pipeline] stage
[Pipeline] { (Build and publish container)
[Pipeline] sh
[blueocean] Running shell script
+ docker/official/build.sh
++ dirname docker/official/build.sh
+ HERE=docker/official
+ cd docker/official
++ git for-each-ref --sort=-taggerdate 'refs/tags/blueocean-parent-*'
++ grep -v beta
++ sed 's/refs\/tags\/blueocean-parent-//'
++ awk '{print $3 }'
++ head -1
+ BLUEOCEAN_VERSION=1.0.1
+ docker pull jenkinsci/blueocean:1.0.1
1.0.1: Pulling from jenkinsci/blueocean
Digest: sha256:445f3ef27fcfb1dbd3298a9cdb9803210762a0c2c7a5ab22e4bb28871aaf2dd4
Status: Image is up to date for jenkinsci/blueocean:1.0.1
+ echo 'Image jenkinsci/blueocean:1.0.1 already exists in Docker Hub'
Image jenkinsci/blueocean:1.0.1 already exists in Docker Hub
+ exit 0
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
Finished: SUCCESS
```

ydubreuil:
@rtyler `jq` may not be available in the minimal Ubuntu cloud image but it's available on 16.04 as a system package (http://packages.ubuntu.com/xenial/jq). Do you think you could make it available in the slave?

rtyler:
@ydubreuil No, we don't make any updates whatsoever to these VMs, they're stock on purpose. If you need a system dependency, I suggest using a Docker-based environment for it.

michaelneale:
@rtyler @ydubreuil so is this one reasonable as it is or overkill?

rtyler:
@michaelneale I don't have a problem with it

michaelneale:
ok - @ydubreuil - can I get a bee from you if you think this is reasonable?

ydubreuil:
yeah, it's fine as it is :bee:

PR #1057, opened 2017-05-11
fix content jumping when vertical scrollbar appears/disappears

NicuPascu:
# Description
Made the html tag have a 100vw width that will ignore the appearance or disappearance of the vertical scrollbar on systems that have a mouse attached, this bug did not manifest when only a touchpad was present.

A possible problem with this fix would appear if we need to ever horizontal scroll the html tag. If we ever need to do that, an alternate fix to this bug is to make the vertical scrollbar always appear on systems that have a mouse connected but this fix is preferred because the the vertical scrollbar won't appear if it is not needed.

No tests are necessary because the fix consists of only css changes

See [JENKINS-44048](https://issues.jenkins-ci.org/browse/JENKINS-44048).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/less/core.less
+++ b/blueocean-dashboard/src/main/less/core.less
@@ -1,3 +1,8 @@
+html {
+    width: 100vw;
+    overflow-x: hidden;
+}
+
 #modal-snap-background {
     position: fixed;
     top: 0;
--- a/blueocean-dashboard/src/main/less/core.less
+++ b/blueocean-dashboard/src/main/less/core.less
@@ -1,6 +1,5 @@
 html {
-    width: 100vw;
-    overflow-x: hidden;
+    overflow-y: scroll;
 }
 
 #modal-snap-background {


sophistifunk:
Am I the only person in the room who thinks this is not a bug?

If we use viewport width, will we get a horizontal scroll bar before we should? Or will we *always* have one, when on a narrower display, even if there's room for all the content?

kzantow:
... also probably shouldn't have `overflow-x: hidden`, you never know if there are reports or something else that we eventually need to scroll

michaelneale:
@sophistifunk @kzantow this was a problem when people added extensions mainly - the video in the attached ticket shows a minor version of it which isn't really a problem. @scherler do you have a better video showing it looking worse with some extension? 

@kzantow so what should it be if not hidden?

sophistifunk:
I certainly can't see a good case for allowing the scroll bar to cover content, and when the content is changing so that things are different sizes, how can it not "jump"?

kzantow:
Is there a reason `html { overflow-y: scroll; }` doesn't work?

NicuPascu:
@sophistifunk The horizontal scroll bar is hidden by the overflow-x: hidden rule, I have tested on narrow viewports and everything looks good, the content will not jump because the html will always be 100% of the viewport regardless of scrollbar

@kzantow html { overflow-y: scroll; } works but will cause a greyed out vertical scrollbar to always appear even it there is not enough content to require scrolling, I implemented the fix this way because it will make the vertical scrollbar behave naturally (not appear at all when there is nothing to scroll)

sophistifunk:
Never showing the horizontal scrollbar is not how you fix making the content too wide for the browser window.

michaelneale:
summary so far: 

* ideally we don't want thing to jump, scroll showing or not
* We don't want to mess with scroll however as it is browsers domain, by hiding it or whatever

Does latest version address these?

NicuPascu:
Yes, the only thing it does is always shows the vertical scrollbar on devices that have a physical mouse, so no more jumping or hiding scrollbars

michaelneale:
ok - @scherler @sophistifunk @kzantow - acceptable? or need to think about something else?

kzantow:
@michaelneale @NicuPascu my view is... if it's ever going to scroll, just keep the scroolbar visible at all times since it's a SPA

kzantow:
side note: it doesn't show up on chrome on the mac, anyway

michaelneale:
@kzantow "ever going to scroll" - so always? (as any listing could). 

And yes, lol OSX... who needs scrollbars! (it shows little things just as you start to scroll)

kzantow:
... i mean, you could have some sales UI that just goes fullscreen and never scrolls the main panel, but this app does, so the jumping is just annoying and much of the time you'll see the scrollbar anyway, so just show it from the get-go

michaelneale:
so @kzantow is this a bee?

sophistifunk:
🐝

jlong:
Very nice. Thank you!

PR #1058, opened 2017-05-11
Ability to get/resolve extension resources (JS, CSS etc) from an external host (Vs the Jenkins master

tfennelly:
# Description

Feel free to __IGNORE__ for now.

No related Jenkins JIRA. Need this for something else I'm working on.

# Submitter checklist

### DIFF
--- a/js-extensions/spec/ResourceLoadTracker-spec.js
+++ b/js-extensions/spec/ResourceLoadTracker-spec.js
@@ -104,4 +104,42 @@ describe("ResourceLoadTracker.js", function () {
             done();
         });
     });
+
+    it("- test loading extension CSS from another location for a specific plugin", function (done) {
+        jsTest.onPage(function() {
+            var javaScriptExtensionInfo = require('./javaScriptExtensionInfo-02.json');
+            var jsModules = require('@jenkins-cd/js-modules');
+            var ResourceLocationResolver = require('@jenkins-cd/js-modules/js/ResourceLocationResolver');
+
+            // get the 'plugin-1' resources from another host
+            // continue to get 'plugin-2' resources from Jenkins adjuncts
+            jsModules.addResourceLocationResolver(
+                new ResourceLocationResolver(function(moduleSpec) {
+                    // 'plugin-1' is the hpiPluginId set in ./javaScriptExtensionInfo-02.json
+                    return (moduleSpec.namespace === 'plugin-1');
+                }, function(moduleSpec, srcPath) {
+                    return "http://resources.acme.com/javascript/" + srcPath;
+                })
+            );
+
+            // Initialise the load tracker with plugin extension point info.
+            ResourceLoadTracker.setExtensionPointMetadata(javaScriptExtensionInfo);
+
+            // Verify that there's no link elements on the page.
+            var cssElements = document.getElementsByTagName('link');
+            expect(cssElements.length).to.equal(0);
+
+            // Mounting ep-* should result in the CSS for both plugins being
+            // loaded ...
+            ResourceLoadTracker.onMount('ep-1');
+            ResourceLoadTracker.onMount('ep-2');
+            ResourceLoadTracker.onMount('ep-3');
+            cssElements = document.getElementsByTagName('link');
+            expect(cssElements.length).to.equal(2);
+            expect(cssElements[0].getAttribute('href')).to.equal('http://resources.acme.com/javascript/org/jenkins/ui/jsmodules/plugin-1/extensions.css');
+            expect(cssElements[1].getAttribute('href')).to.equal('/jenkins/adjuncts/908d75c1/org/jenkins/ui/jsmodules/plugin-2/extensions.css');
+
+            done();
+        });
+    });
 });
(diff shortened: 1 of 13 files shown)


michaelneale:
I know this is early, but this looks potentially quite handy - @kzantow @i386 you agree?

i386:
@tfennelly could this help us in the case where we might not want to use adjuncts?

tfennelly:
@michaelneale @i386 right, this is for when e.g. you want to load a plugins extensions (JS, CSS etc) from outside the Jenkins master e.g. a completely different host server.

js-extensions/spec/ResourceLoadTracker-spec.js
```diff
@@ -104,4 +104,42 @@ describe("ResourceLoadTracker.js", function () {
...
+            // Initialise the load tracker with plugin extension point info.
+            ResourceLoadTracker.setExtensionPointMetadata(javaScriptExtensionInfo);
+
+            // Verify that there's no link elements on the page.
+            var cssElements = document.getElementsByTagName('link');
+            expect(cssElements.length).to.equal(0);
+
+            // Mounting ep-* should result in the CSS for both plugins being
+            // loaded ...
+            ResourceLoadTracker.onMount('ep-1');
```
scherler: I do not follow what is `'ep-1'` 

`onMount` sounds to me that it would accept properties or a function, so I wonder what you do with the string?

scherler:
@tfennelly how do I define it? for me I have a common plugin that supposed to provide the css but I ended up to create a copy of the `less` in each plugin. As I understand correct with this change I can drop my copies and place the compiled `css` in the the `webapp/css` folder and then reference to the static route `...static/a6a62bc0/plugin/pluginA/css/common.css`

Where could I define the above route?

Actually I am more interested in only define `css/common.css` and the prefix generation to the path should be done automatic. ...but yeah that are details and features ;)

...awesome btw as stated in the beginning I need this too. :)

js-extensions/spec/ResourceLoadTracker-spec.js
tfennelly: That's just a test, with `ep-1` being "extension point 1". See the other tests in the same spec.

tfennelly:
@scherler I'd need to see the details of what you're doing. Atm, this is mostly geared towards BO extensions i.e. the bundle and css associated with the 

I'm working on a way for a plugin to declare that it wants it's bundle JS, css etc to come from some other location (e.g. a CDN). It's based on a plugin that defines the following `ExtensionPoint`. We'll probably publish that plugin once we have it nailed down.

```java
/**
 * Hosted Resource Location Jenkins Extension Point.
 *
 * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
 */
public abstract class HostedUIResourceLocation extends ExtensionPoint {

    /**
     * Get the UI resource namespace.
     * <p>
     * Typically, this would be the plugins ID. This is a js-modules concept,
     * where UI resources can "live" in different namespaces.
     *
     * @return The UI resource namespace.
     */
    public abstract String getResourceNamespace();

    /**
     * Get the base URL.
     * <p>
     * This is the URL of the server hosting the resources. The plugin can
     * decide where it gets this from e.g. from a config file, a system/env
     * property, from a DB, or other.
     *
     * @return The base URL.
     */
    public abstract URL getBaseURL();

    /**
     * Get all implementations of the HostedUIResourceLocation Extension Point.
     * @return All implementations of the HostedUIResourceLocation Extension Point.
     */
    public static ExtensionList<HostedUIResourceLocation> all() {
        return ExtensionList.lookup(HostedUIResourceLocation.class);
    }
}
```

tfennelly:
@michaelneale @i386 @kzantow Any comments on this or are ye okay with it?

js-extensions/src/ResourceLoadTracker.js
```diff
@@ -31,31 +32,34 @@ export default class ResourceLoadTracker {
...
-    
+
                 // Add the plugin CSS if it's not already in the list.
-                if (pointCSS.indexOf(pluginCSS) === -1) {
-                    pointCSS.push(pluginCSS);
+                if (pointCSS.filter((pluginCSSEntry) => pluginCSSEntry.url === pluginCSS).length === 0) {
+                    pointCSS.push({
+                        url: pluginCSS,
+                        hpiPluginId: pluginMetadata.hpiPluginId
+                    });
```
cliffmeyers: Small nit: this filter/push approach isn't super intuitive (to me at least). Perhaps just map it to a new array ref and then assign that to `this.pointCSSs[extensionPoint]`

tfennelly: @cliffmeyers yeah I agree with you on that.

TBH, I think a bit of refactor would be good there such that we extract the CSS "Resource" parts of `ResourceLoadTracker` into a type of its own (with all the support methods e.g. `_addCSS` etc). I'll have a look at that and if it's easy to do without making this PR explode into something a lot bigger then I'll do it.

i386:
I am ok with this change so long as @kzantow is 👍

michaelneale:
yes need an ok from @kzantow

michaelneale:
@tfennelly you have the go ahead - up to you if you still need this

PR #1061, opened 2017-05-12
JENKINS-44208# Generate url for organization

vivek:
# Description

See [JENKINS-44208](https://issues.jenkins-ci.org/browse/JENKINS-44208).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanWebURLBuilder.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanWebURLBuilder.java
@@ -33,6 +33,7 @@ import io.jenkins.blueocean.rest.factory.BluePipelineFactory;
 import io.jenkins.blueocean.rest.factory.organization.OrganizationFactory;
 import io.jenkins.blueocean.rest.impl.pipeline.BranchImpl;
 import io.jenkins.blueocean.rest.model.BlueMultiBranchPipeline;
+import io.jenkins.blueocean.rest.model.BlueOrganization;
 import io.jenkins.blueocean.rest.model.BluePipeline;
 import io.jenkins.blueocean.rest.model.Resource;
 import jenkins.model.ModifiableTopLevelItemGroup;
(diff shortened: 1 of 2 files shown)


vivek:
cc: @alvarolobato

vivek:
ATH passes https://ci.blueocean.io/blue/organizations/jenkins/ATH-Jenkinsfile/detail/master/2421/pipeline

alvarolobato:
Thanks for the PR @vivek 
There is a possible NPE that we should protect from when you call `OrganizationFactory.getInstance().getContainingOrg(i)` by definition you could get a null, in which case we should fall back to a no org url:

```
Caused by: java.lang.NullPointerException
	at io.jenkins.blueocean.BlueOceanWebURLBuilder.getOrgPrefix(BlueOceanWebURLBuilder.java:144)
	at io.jenkins.blueocean.BlueOceanWebURLBuilder.toBlueOceanURL(BlueOceanWebURLBuilder.java:133)
	at io.jenkins.blueocean.BlueOceanWebURLBuilder.getTryBlueOceanURLs(BlueOceanWebURLBuilder.java:82)
	at io.jenkins.blueocean.TryBlueOceanPageDecorator.getTryBlueOceanURLs(TryBlueOceanPageDecorator.java:43)
	... 103 more
```

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanWebURLBuilder.java
```diff
@@ -128,6 +129,8 @@ private BlueOceanWebURLBuilder() {
             if (blueResource instanceof BlueMultiBranchPipeline) {
                 return getOrgPrefix(item) + "/" + encodeURIComponent(((BluePipeline) blueResource).getFullName()) + "/branches";
             }
```
alvarolobato: This implementation will not work for orgs which are on a folder, bare in mind that most of the `ModifiableTopLevelItemGroup`s are also Items so they will come through this else if and never get the org ie. Folders. 
Maybe just getting the org and if not null return the org url instead of blue for all the cases and not only for `ModifiableTopLevelItemGroup`  that could be done a level above or checking all the if branches.

vivek: @alvarolobato 

>There is a possible NPE that we should protect from when you call OrganizationFactory.getInstance().getContainingOrg(i) by definition you could get a null, in which case we should fall back to a no org url:

Good catch, there is NPE at couple places, lucky it didn't encounter so far because there is always default impl of `OrganizationFactory`. I will fix that.

> Maybe just getting the org and if not null return the org url instead of blue for all the cases and not only for ModifiableTopLevelItemGroup that could be done a level above or checking all the if branches.

Ah, yeah, it won't work with how its done. Let me look in to this.

vivek:
@alvarolobato 

> There is a possible NPE that we should protect from when you call OrganizationFactory.getInstance().getContainingOrg(i) by definition you could get a null, in which case we should fall back to a no org url:

In case of null, we can't always map to no org url. All jobs, branches are rendered in context of organization. Only fail safe thing in case of null org, is to land the user at landing page, that is `/blue`.

vivek:
Replaced by https://github.com/jenkinsci/blueocean-plugin/pull/1063

PR #1063, opened 2017-05-12
JENKINS-44208 Web url builder extension

vivek:
# Description

See [JENKINS-44208](https://issues.jenkins-ci.org/browse/JENKINS-44208).

- Renders 'Open Blue Ocean' as Task action menu on all Jenkins pages
- See  [BlueOceanUrlFactory](https://github.com/jenkinsci/blueocean-plugin/blob/36d7dfd66a7585bd9df288b74e507871cc2c1bb1/blueocean-rest/src/main/java/io/jenkins/blueocean/rest/factory/BlueOceanUrlFactory.java#L16) and [BlueOceanUrlAction](https://github.com/jenkinsci/blueocean-plugin/blob/36d7dfd66a7585bd9df288b74e507871cc2c1bb1/blueocean-rest/src/main/java/io/jenkins/blueocean/rest/factory/BlueOceanUrlAction.java#L13)
- To control individual Jenkins `ModelObject` url mapping implement [BlueOceanUrlMapper](https://github.com/jenkinsci/blueocean-plugin/blob/web-url-builder-extension/blueocean-rest/src/main/java/io/jenkins/blueocean/rest/factory/BlueOceanUrlMapper.java#L16)
- See implementations [BlueOceanUrlActionFactoryImpl](https://github.com/jenkinsci/blueocean-plugin/blob/web-url-builder-extension/blueocean-rest/src/main/java/io/jenkins/blueocean/rest/factory/BlueOceanUrlActionFactory.java#L16) and [BlueOceanUrlActionImpl](https://github.com/jenkinsci/blueocean-plugin/blob/36d7dfd66a7585bd9df288b74e507871cc2c1bb1/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanUrlActionImpl.java#L13)

Basically, idea is for plugins interested in customizing blueocean url mapping for given ModelObject, they can implement `BlueOceanUrlMapper`. If there is need to customize `Open Blue Ocean` menu item/action, then implement `BlueOceanUrlAction` and `BlueOceanUrlActionFactory`.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


vivek:
@brody @i386 Ended up using 48x48, it looked better:

image

i386:
Nice 👍

vivek:
@alvarolobato I think I have everything you need. 

- Implement `BlueOceanUrlMapper` to control custom mapping of jenkins ModelObject to url
- Implement `BlueOceanUrlActionFactory` and `BlueOceanUrlAction` to customize the view of action menu to open blueocean. You might need to implement it for the `ModelObject` not handled by default implementation.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanUrlActionImpl.java
```diff
@@ -0,0 +1,55 @@
+package io.jenkins.blueocean;
+
+import hudson.Extension;
+import hudson.model.ModelObject;
+import io.jenkins.blueocean.rest.factory.BlueOceanUrlAction;
+import io.jenkins.blueocean.rest.factory.BlueOceanUrlMapper;
+
+import javax.annotation.Nonnull;
+
+import static io.jenkins.blueocean.BlueOceanUrlMapperImpl.getLandingPagePath;
```
alvarolobato: This probably should need to be moved here, as we may not have the dependency with pipeline plugins

alvarolobato:
:bee:

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanUrlActionFactoryImpl.java
```diff
@@ -0,0 +1,26 @@
...
+import io.jenkins.blueocean.rest.factory.BlueOceanUrlActionFactory;
+import jenkins.model.Jenkins;
+
+import javax.annotation.Nonnull;
+
+/**
+ * @author Vivek Pandey
+ */
+@Extension(ordinal = -9999)
+public class BlueOceanUrlActionFactoryImpl extends BlueOceanUrlActionFactory {
```
vivek: Agreed, it used to depend on this plugin but with this PR, we can pretty much move it inside blueocean-rest-impl.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/BlueOceanUrlMapperImpl.java
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * @author Vivek Pandey
+ */
+@Extension(ordinal = -9999)
+public class BlueOceanUrlMapperImpl extends BlueOceanUrlMapper {
+    @Override
+    public String getUrl(@Nonnull ModelObject modelObject) {
+        BlueOrganization organization = getOrganization(modelObject);
+        if(organization == null){ //no organization, best we can do is to land user on landing page
+            return getLandingPagePath();
```
vivek: @alvarolobato `getLandingPagePath()` is computed using `BlueOceanUIProvider`, so should be fine, anyways ordinal value is going to let other plugins override it.

alvarolobato: You are right, I miss read the method.

vivek:
@alvarolobato PTAL. I think I have answered your concerns.

There are ATH tests failing as `#open-blueocean-in-context` element is no more there. I will fix those tests after I figure how to test these action menu. Thats coming up next.

vivek:
@imeredith PTAL at ATH changes.

vivek:
@imeredith tests are failing with same error:

```
Timed out while waiting for element <body.page-head-removed> to be present for 20000 milliseconds.  - expected "found" but got: "not found"
    at Timeout._onTimeout (/Users/vivek/ws/blueocean/acceptance-tests/src/main/js/custom_commands/removePageHead.js:50:18)
```
Perhaps these changes expose this timing bug? @imeredith It will be great if you could try out this branch.

imeredith:
@vivek .useXpath() sets all selectors in the future to be xpath. A css one is failing so im guessing your use of xpath caused this (noticed in changes). You can do .useCss() to fix that

imeredith:
Also the stuff in addOpenBlueOceanLinkToFooter is just flaky in general. If you cant get a test to pass it, try restarting the ATH server. its a real PITA.

vivek:
@imeredith Ah, right. I reset selector to css. Lets see how this run goes. `addOpenBlueOceanLinkToFooter` is deleted because there is no need for this trick as that button on top is removed.

vivek:
@imeredith I changed openBlueOcean.js from command to page object. This way there is one place to open blueocean page and check it's visibility. All looks good now. PTAL before I merge:)

acceptance-tests/runner/.blueocean-ath-jenkins-url
```diff
@@ -1 +1 @@
-http://127.0.0.1:62969
\ No newline at end of file
+http://127.0.0.1:50313
```
michaelneale: probably should be ignored

michaelneale:
:bee: from me but need @imeredith to give this the nod - as it is a bit of a cross cutting change, but looks great!

acceptance-tests/runner/.blueocean-ath-jenkins-url
imeredith: It is in my latest stuff

imeredith:
Why a pageobject and not a command? For a start its not a page, but it exists on most of classic pages.

vivek:
@imeredith nightwatch page objects can be anything that appears on a page and tests want access to it, so it's fine for it to be page object without any side effect. 

Other reasons listed below:
- It exposes two methods, `open()` and `visible()`, if we go with command approach there will be two commands then. This makes it modular.
- Fixes flakiness. From what I experienced, command objects trying to `useXpath()` doesn't work well even if it gets reset with `useCss()`. OTOH, page object element's `locateStrategy` works quite reliably.

imeredith:
:bee:

PR #1064, opened 2017-05-14
JENKINS-44258 Replayed runs have same changeset as original

i386:
This change is fairly nuanced but results in a more relevant experience for users.

Included in this change:
* When we detect the replay cause on the current run we lookup the original run and return its changesets.
* It turns out the first cause in the array is the least significant - replayed runs never had the correct cause displayed and is now fixed.
* If there are two or more causes on the run then the last cause in the array is more important to the user than the changeset message when viewing Activity or Branches tabs.

# Testing
- Create a pipeline and run it so that it contains a few changesets
-  Message column on Activity should be the changeset
- Replay the run
- Message column on Activity should be "Replayed #?" where ? is the run number of the original run

# Description

See [JENKINS-44258](https://issues.jenkins-ci.org/browse/JENKINS-44258).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/RunMessageCell.jsx
+++ b/blueocean-dashboard/src/main/js/components/RunMessageCell.jsx
@@ -11,17 +11,31 @@ export default class RunMessageCell extends Component {
         const run = this.props.run;
         const t = this.props.t;
         let message;
-        if (run && run.description) {
+
+        // Note that the order that this is evaluated is important for providing a relevant message to the user
+
+        // 1. If the user has set a message then we always show it
+        const showUserDefinedMessage = run && run.description;
+
+        // 2. Show commit messages if available
+        //    however if a run has > 1 cause then the cause is likely more important than the change set (e.g. replay)
+        const showCommitMessage = run && run.changeSet && (run.changeSet && run.changeSet.length > 0) && (run.causes && run.causes.length <= 1);
+
+        // 3. Lastly if there are any causes, display the last cause available
+        const showCauses = run && (run.causes && run.causes.length > 0);
+
+        if (showUserDefinedMessage) {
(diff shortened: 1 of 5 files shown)


vivek:
@i386 this is good one 👍

blueocean-dashboard/src/main/js/components/RunMessageCell.jsx
```diff
@@ -11,17 +11,31 @@ export default class RunMessageCell extends Component {
...
+
+        // 1. If the user has set a message then we always show it
+        const showUserDefinedMessage = run && run.description;
+
+        // 2. Show commit messages if available
+        //    however if a run has > 1 cause then the cause is likely more important than the change set (e.g. replay)
+        const showCommitMessage = run && run.changeSet && (run.changeSet && run.changeSet.length > 0) && (run.causes && run.causes.length <= 1);
+
+        // 3. Lastly if there are any causes, display the last cause available
+        const showCauses = run && (run.causes && run.causes.length > 0);
```
cliffmeyers: This is well-commented. Would it be simpler though just to check `run.causes.length > 1` and move the `showCauses` check below up above the `showCommitMessage` case? This is a nit, so if that seems silly, please ignore.

blueocean-dashboard/src/test/js/run-message-cell-spec.js
```diff
@@ -30,12 +31,27 @@ describe('RunMessageCell', () => {
...
+                    'msg': 'Oops',
+                }],
+                causes: [
+                    { shortDescription: 'Cake is delicious' },
+                    { shortDescription: 'Have some cake' },
+                ]
+            };
+            const cell = render(<RunMessageCell run={run} t={t} />);
+            expect(cell.text()).to.equal('Have some cake');
+    });
```
cliffmeyers: 👍

i386:
@cliffmeyers thanks :)

PR #1068, opened 2017-05-16
decodeURI text in branch filter

NicuPascu:
The value of the input should be uri decoded because it comes from the "branch" url get param that is uri encoded

# Description

See [JENKINS-44260](https://issues.jenkins-ci.org/browse/JENKINS-44260).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
This shows "undefined" as the default value in the drop down: 

screen shot 2017-05-16 at 5 39 08 pm

Also there is a failure in the ATH (see the log linked) to do with this - probably just picking up the same thing.

i386:
It would be good if we could get a unit test here.

michaelneale:
yes agree - could that way check it is never undefined

NicuPascu:
Ivan says that making a test for this is quite involved and we need to talk to Cliff about making this easier, should we create another ticket for the test and merge this branch or wait for the test before merging?
We will try to talk to Cliff tomorrow morning.

michaelneale:
I think this is fine to go in as it is already :bee:

Nice one @NicuPascu !

PR #1069, opened 2017-05-16
Working with Nicu on css loading + JSX render coordination

tfennelly:
WIP: 

https://issues.jenkins-ci.org/browse/JENKINS-44431

### DIFF
--- a/js-extensions/src/ExtensionRenderer.jsx
+++ b/js-extensions/src/ExtensionRenderer.jsx
@@ -20,8 +20,9 @@ export default class ExtensionRenderer extends React.Component {
     }
 
     componentDidMount() {
-        ExtensionRenderer.resourceLoadTracker.onMount(this.props.extensionPoint);
-        this._renderAllExtensions();
+        ExtensionRenderer.resourceLoadTracker.onMount(this.props.extensionPoint, function() {
+            this._renderAllExtensions();
+        });
     }
 
     componentWillUpdate(nextProps, nextState) {
@@ -89,7 +90,7 @@ export default class ExtensionRenderer extends React.Component {
         } = this.props;
 
         const classNames = ['ExtensionPoint', extensionPoint.replace(/\.+/g,'-')];
-        
+
         if (className) {
             classNames.push(className);
         }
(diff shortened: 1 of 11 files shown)


js-extensions/src/ResourceLoadTracker.js
```diff
@@ -118,18 +133,28 @@ export default class ResourceLoadTracker {
...
+
+        jsModules.addCSSToPage(cssURL);
+
+        const linkElId = jsModules.toCSSId(cssURL);
+        const linkEl = document.getElementById(linkElId);
+        if (linkEl) {
+            linkEl.onload = onload;
+        } else {
+            onload();
+        }
```
tfennelly: @NicuPascu I just added the above check on `linkEl`. That code is not in the version o `js-extensions` that we published.

js-extensions/src/ExtensionRenderer.jsx
```diff
@@ -20,8 +20,9 @@ export default class ExtensionRenderer extends React.Component {
     }
 
     componentDidMount() {
-        ExtensionRenderer.resourceLoadTracker.onMount(this.props.extensionPoint);
-        this._renderAllExtensions();
+        ExtensionRenderer.resourceLoadTracker.onMount(this.props.extensionPoint, function() {
+            this._renderAllExtensions();
+        });
```
tfennelly: So now, we only call `this._renderAllExtensions()` after all CSS has ben loaded.

cliffmeyers:
Hi, I have two questions about this:
- I recall seeing a similar "re-render flash" when I would open the `/organizations/jenkins/create-pipeline` screen directly. Some of the content in this page renders via an extension point, so I expect it was rendering before its child CSS was loaded. @NicuPascu perhaps you could check that page too as part of this fix?
- I noticed that @scherler was working on a bug that involved translations sometimes not loading before initial page render. Could that be related to this same kind of problem?

tfennelly:
@cliffmeyers 

> I noticed that @scherler was working on a bug that involved translations sometimes not loading before initial page render. Could that be related to this same kind of problem?

I don't know what that is but I doubt it's related.

michaelneale:
closing this as a bit old now. It will live on in our hearts and git history

PR #1070, opened 2017-05-16
Bug/jenkins 43948 step header disappears clicked

cliffmeyers:
# Description
- See [JENKINS-43948](https://issues.jenkins-ci.org/browse/JENKINS-43948).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/karaoke/components/Step.jsx
+++ b/blueocean-dashboard/src/main/js/components/karaoke/components/Step.jsx
@@ -134,7 +134,7 @@ export class Step extends Component {
             liveFormat={t('common.date.duration.format', { defaultValue: 'm[ minutes] s[ seconds]' })}
             hintFormat={t('common.date.duration.hint.format', { defaultValue: 'M [month], d [days], h[h], m[m], s[s]' })}
         />);
-        return (<div className={logConsoleClass} key={this.pager.currentLogUrl}>
+        return (<div className={logConsoleClass}>
             <ResultItem {...{
                 extraInfo: time,
                 key: step.key,
--- a/blueocean-dashboard/src/main/js/components/karaoke/components/Step.jsx
+++ b/blueocean-dashboard/src/main/js/components/karaoke/components/Step.jsx
@@ -77,11 +77,6 @@ export class Step extends Component {
     }
 
     render() {
-        // early out
-        if (this.pager.pending) {
-            logger.debug('pending returning null');
-            return null;
-        }
         const { step, locale, router, location, t, scrollToBottom } = this.props;
         if (step === undefined || !step) {
             return null;


cliffmeyers:
@scherler would just like some feedback from you that these changes are appropriate. See my comments inline. Thanks!!

blueocean-dashboard/src/main/js/components/karaoke/components/Step.jsx
```diff
@@ -77,11 +77,6 @@ export class Step extends Component {
     }
 
     render() {
-        // early out
-        if (this.pager.pending) {
-            logger.debug('pending returning null');
-            return null;
-        }
```
cliffmeyers: This was causing the entire `ResultItem`  - header included - to disappear while data was loading. Removing this didn't seem to cause any issues and ATH passed.

blueocean-dashboard/src/main/js/components/karaoke/components/Step.jsx
```diff
@@ -139,7 +134,7 @@ export class Step extends Component {
             liveFormat={t('common.date.duration.format', { defaultValue: 'm[ minutes] s[ seconds]' })}
             hintFormat={t('common.date.duration.hint.format', { defaultValue: 'M [month], d [days], h[h], m[m], s[s]' })}
         />);
-        return (<div className={logConsoleClass} key={this.pager.currentLogUrl}>
+        return (<div className={logConsoleClass}>
```
cliffmeyers: After making the above change, I noticed that when the log data loaded, the slide/open animation would play again. It wasn't terrible but it looked a little funny. I noticed that the value of `this.pager.currentLogUrl` was undefined before the data fetch, and then populated after. This caused the component to be created/redestroyed, so the `ResultItem` was playing its animation again. This fix seems appropriate but I am curious if there is a specific reason you set a `key` on this component. It seems unnecessary since the parent `Steps` component is already defining a `key` on each Step?

cliffmeyers:
jenkins-49348-1-fixed

i386:
@cliffmeyers nice one 👍

blueocean-dashboard/src/main/js/components/karaoke/components/Step.jsx
michaelneale: ah yes, react accepts null as meaning "remove this component or don't bother rendering". Looks like left over debug code.

michaelneale:
:bee: nice one cliff on this critical issue!

blueocean-dashboard/src/main/js/components/karaoke/components/Step.jsx
scherler: @cliffmeyers I am not 100% sure but I think that has been there because of the effect it has on rendering,  I remember adding it when using the satges and steps and the steps had been from the old stage, but I guess that is not needed anymore then ;)

PR #1071, opened 2017-05-17
bug / JENKINS 44321 changes tab doesn't load

i386:
# Description
- See [JENKINS-44321](https://issues.jenkins-ci.org/browse/JENKINS-44321).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/records.jsx
+++ b/blueocean-dashboard/src/main/js/components/records.jsx
@@ -70,7 +70,7 @@ export class RunRecord extends Record({
     _class: null,
     _capabilities: [],
     _links: null,
-    changeSet: ChangeSetRecord,
+    changeSet: [],
     causes: [],
     causeOfBlockage: null,
     artifacts: null,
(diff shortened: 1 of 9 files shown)


blueocean-dashboard/src/main/js/components/records.jsx
```diff
@@ -70,7 +70,7 @@ export class RunRecord extends Record({
     _class: null,
     _capabilities: [],
     _links: null,
-    changeSet: ChangeSetRecord,
+    changeSet: [],
```
cliffmeyers: `changeSet` is actually an array of these records. If JSON response omitted the prop, then it would get initialized to an empty `ChangeSetRecord` which would defeat null checking in the UI layer but still cause an exception to be thrown.

michaelneale:
what is this one about? is that a go/no go @cliffmeyers ?

cliffmeyers:
This is a simple fix, just looking for a bee from someone.

cliffmeyers:
@michaelneale renamed and linked appropriate JIRA issue.

i386:
🐝

PR #1076, opened 2017-05-22
[FIX JENKINS-44419] Makes the organizations enabled flag configurable at …

alvarolobato:
…runtime.

# Description

See [JENKINS-44419]](https://issues.jenkins-ci.org/browse/JENKINS-44419]).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
+++ b/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
@@ -1,0 +1,14 @@
+package io.jenkins.blueocean.config;
+
+import org.kohsuke.stapler.export.ExportedBean;
+
+/**
+ * Holds the Blue Ocean configuration
+ *
+ */
+@ExportedBean
+public abstract class BlueOceanConfig {
+    public static final String FEATURE_PROPERTY_PREFIX = "blueocean.features.";
+    public static final String ORGANIZATION_ENABLED = "organizations.enabled";
+    public abstract <T> T get(String key, Class<T> type);
+}
--- a/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigFactory.java
+++ b/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigFactory.java
@@ -1,0 +1,45 @@
+package io.jenkins.blueocean.config;
+
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+
+import hudson.ExtensionList;
+import hudson.ExtensionPoint;
+
+/**
+ * Blue Ocean configuration factory which allows different extensions to return configuration values.
+ * 
+ */
+public abstract class BlueOceanConfigFactory implements ExtensionPoint {
+
+    /**
+     * Returns Blue Ocean configuration
+     * 
+     * @return the current configuration
+     */
+    @Nonnull
+    public abstract BlueOceanConfig getConfig();
+
+    /**
+     * Looks up for an specific config key on the available <code>BlueOceanConfigFactory</code> implementations
+     * 
+     * @param key the key to look for
+     * @param type the type of the value
+     * @return the value for the requested key
+     */
+    @CheckForNull
+    public static <T> T getConfig(String key, Class<T> type) {
+        for (BlueOceanConfigFactory configFactory : ExtensionList.lookup(BlueOceanConfigFactory.class)) {
+            BlueOceanConfig config = configFactory.getConfig();
+            T value = config.get(key, type);
+            if (value != null) {
+                return value;
+            }
+        }
+        return null;
+    }
+
+    public static ExtensionList<BlueOceanConfigFactory> all() {
+        return ExtensionList.lookup(BlueOceanConfigFactory.class);
+    }
+}
--- a/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigStatePreloader.java
+++ b/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigStatePreloader.java
@@ -71,8 +71,7 @@ public class BlueOceanConfigStatePreloader extends PageStatePreloader {
                 // If more "features" vars are added, we could just iterate the system props
                 // and add any starting with "blueocean.features.". However, lets not do that
                 // unless there are more than a few.
-                .key("features").object()
-                    .key("organizations.enabled").value(Boolean.getBoolean("blueocean.features.organizations.enabled"))
+                               .key("features").object().key(BlueOceanConfig.ORGANIZATION_ENABLED).value(BlueOceanConfigFactory.getConfig(BlueOceanConfig.ORGANIZATION_ENABLED, Boolean.class))
                 .endObject()
             .endObject();
 
--- a/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigStatePreloader.java
+++ b/blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigStatePreloader.java
@@ -8,6 +8,8 @@ import hudson.util.VersionNumber;
 import io.jenkins.blueocean.auth.jwt.JwtTokenServiceEndpoint;
 import io.jenkins.blueocean.commons.BlueOceanConfigProperties;
 import io.jenkins.blueocean.commons.PageStatePreloader;
+import io.jenkins.blueocean.rest.factory.BlueOceanConfigFactory;
+import io.jenkins.blueocean.rest.model.BlueOceanConfig;
 import jenkins.model.Jenkins;
 import net.sf.json.util.JSONBuilder;
 
--- a/acceptance-tests/README.md
+++ b/acceptance-tests/README.md
@@ -20,7 +20,12 @@ Please make sure that BlueOcean plugins are built correctly. If _funky_ things s
 mvn clean install -DskipTests -DcleanNode
 ```
 
+### Areas covered
 
+* Karaoke parallel: ParallelNavigationTest. This checks that users can navigate between concurrently executing parallel branches, including steps waiting for input.
+* Commit Messages: CommitMessagesTest checks to see that commit messsages of git commits are shown on a run
+* Folder Names: FoldersTest checks that folders with sperial characters work correctly in blueoean
+* Github integration: GithubCreationTests cover github integration testing of buth the creation flow and also simple editor round tripping.
 
 ### Run all tests (in one command)
  ```bash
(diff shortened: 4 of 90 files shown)


michaelneale:
@vivek I think you are best to review this one

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigStatePreloader.java
```diff
@@ -83,4 +88,22 @@ public String getStateJson() {
...
+    public static boolean isOrganizationsEnabled() {
+        return organizationsEnabled;
+    }
+
+    /**
+     * Sets the organizations enabled/disabled flag
+     * 
+     * @param organizationsEnabled true to enable the organizations
+     */
+    public static void setOrganizationsEnabled(boolean organizationsEnabled) {
```
vivek: Don't think such static setter on this class is good idea. Why not go with extension point approach, something like:

```java
public abstract class BlueOceanConfiguration implements ExtensionPoint {
    public abstract boolean isOrganizationsEnabled();
}
```
This way we can add more configuration elements as needed instead of adding methods around? 🐛

alvarolobato: Yes, I just thought there was going to be a rewrite of that part (per the comment) that's why I went for a simpler solution. Will do the extension.

alvarolobato:
@vivek PTAL. I think that now should be enough for any new configuration we may need to add allowing flexible override of the values by extension points.

michaelneale:
up to @vivek or @i386

i386:
I don't have enough context here to review... ill let @vivek own it

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
```diff
@@ -0,0 +1,22 @@
+package io.jenkins.blueocean.config;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Holds the Blue Ocean configuration
+ *
+ */
+public class BlueOceanConfig {
```
vivek: This should be abstract with pretty much no implementation.

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * Indicates if the Organizations feature should be enabled
+     */
+    public Boolean organizationEnabled;
```
vivek: should be a private final with a getter to be consistent with other model objects.

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
```diff
@@ -0,0 +1,22 @@
...
+
+    BlueOceanConfig() {
```
vivek: Not sure if we need these constructors. See my other comments.

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigFactory.java
```diff
@@ -0,0 +1,62 @@
...
+
+import hudson.ExtensionPoint;
+import jenkins.model.Jenkins;
+
+/**
+ * Blue Ocean configuration factory which allows different extensions to override the default behavior.
+ * 
+ */
+public abstract class BlueOceanConfigFactory implements ExtensionPoint {
+    public static final String FEATURE_PROPERTY_PREFIX = "blueocean.features.";
```
vivek: These constants should go in implementation.

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigFactory.java
```diff
@@ -0,0 +1,62 @@
...
+    public static final String ORGANIZATIONS_ENABLED = "organizations.enabled";
+
+    private static BlueOceanConfig config;
```
vivek: Looks like intent is to cache config instance, why not use cache such as guava? Or maybe use read/write lock instead of synchronization block on the method?

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigFactory.java
```diff
@@ -0,0 +1,62 @@
...
+        }
+        return new BlueOceanConfig(config); //Immmutable
+    }
+
+    /**
+     * Force a re-load of the Blue Ocean configuration
+     * 
+     */
+    @Nonnull
+    public static synchronized void reloadConfig() {
```
vivek: Not sure if I understand why we need loading/reloading configuration. Configuration should be singleton instance on each Jenkins boot or when blueocean plugin is updated.  When do you think config needs to be reloaded?

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigFactory.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+public abstract class BlueOceanConfigFactory implements ExtensionPoint {
+    public static final String FEATURE_PROPERTY_PREFIX = "blueocean.features.";
+    public static final String ORGANIZATIONS_ENABLED = "organizations.enabled";
+
+    private static BlueOceanConfig config;
+
+    /**
+     * Extension method to be implemented for each Blue Ocean configuration factory. Any value left <code>null</code> on
+     * the <code>config</code> parameter will be initialized by {@link BlueOceanConfigurationFactoryDefault},
```
vivek: Remove reference to implementation. Factory/model classes should go to blueocean-rest model package, in this case both BlueOceanConfig and BlueOceanConfigFactory.

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigFactory.java
```diff
@@ -0,0 +1,62 @@
...
+     * initialized values won't be overriden.
+     * 
+     * @param config to initialize
+     * @return the initialized {@link BlueOceanConfig}
+     */
+    @Nonnull
+    protected abstract BlueOceanConfig configure(@Nonnull BlueOceanConfig config);
```
vivek: This is redundant. With higher ordinal you can override method such as 'BlueoceanConfig getConfig()'.

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigFactory.java
alvarolobato: This is a base constant for all the configuration. My intent was to make sure all the configuration starts with the same prefix. Anyway, I can move it.

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigFactory.java
alvarolobato: I didn't intend it like a cache, but as the loaded config, no expiration needed or any of the features a cache would give you.

If we only load this at boot, and the synchronized method is only called at that moment, the Read/Write lock would only make the code more complicated and not give you more performance.

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigFactory.java
alvarolobato: done

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigFactory.java
alvarolobato: Not really needed. Removed.

alvarolobato:
@vivek about putting everything on a map, that was my first idea, but then I thought it would be better to have a more strict type safe approach.
I also thought it would be better performant, but on a second thought BO doesn't reload the page on every action like regular Jenkins so, it should be neglectable.

... anyway, I will change it.

michaelneale:
@alvarolobato thanks - when you do that can you also update this to master? was some fixes for test flakyness just in etc.

alvarolobato:
@michaelneale sure, will do.

michaelneale:
Hi @alvarolobato there is a legit looking test failure: 

```
extensionCanOverrideBehaviourTest – io.jenkins.blueocean.config.BlueOceanConfigFactoryTest
a few seconds
Error
Overriden extension point should have been called
Expected: <true>
     but: was <false>
Stacktrace
java.lang.AssertionError: 
Overriden extension point should have been called
Expected: <true>
     but: was <false>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.junit.Assert.assertThat(Assert.java:956)
	at io.jenkins.blueocean.config.BlueOceanConfigFactoryTest.extensionCanOverrideBehaviourTest(BlueOceanConfigFactoryTest.java:47)
```

too BTW (when you get around to this, no hurry)

alvarolobato:
@michaelneale yes, It's not finished, had to go to other stuff. Resuming right now.

alvarolobato:
@vivek can you re-review please?

michaelneale:
ok I will leave this one up to @vivek as to the timing. Public holiday today in USA.

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfig.java
```diff
@@ -0,0 +1,14 @@
+package io.jenkins.blueocean.config;
+
+import org.kohsuke.stapler.export.ExportedBean;
+
+/**
+ * Holds the Blue Ocean configuration
+ *
+ */
+@ExportedBean
+public abstract class BlueOceanConfig {
```
vivek: Please move it to `io.jenkins.blueocean.rest.model` package in blueocean-rest module.

blueocean-config/src/main/java/io/jenkins/blueocean/config/BlueOceanConfigFactory.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.annotation.Nonnull;
+
+import hudson.ExtensionList;
+import hudson.ExtensionPoint;
+
+/**
+ * Blue Ocean configuration factory which allows different extensions to return configuration values.
+ * 
+ */
+public abstract class BlueOceanConfigFactory implements ExtensionPoint {
```
vivek: Please move it to `io.jenkins.blueocean.rest.model.factory` package in blueocean-rest module.

alvarolobato:
@vivek done.

vivek:
LGTM 🐝 Once tests pass, it should be ready for merge.

michaelneale:
@alvarolobato @vivek I took the liberty of updating this to master - as master has good end to end now - which will help with confidence. Will let this run and see how it goes now... 
(I bet you didnt' know you could do that!!)

michaelneale:
thanks @alvarolobato

PR #1083, opened 2017-05-24
update gulp so we can run gulp watch from js-extensions

NicuPascu:
# Description

No ticket related to this

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
review please @cliffmeyers @kzantow @cliffmeyers @sophistifunk (anyone really)

the test failure is: 



so you can ignore it.

sophistifunk:
So this removes js-builder from the build process? If so we probably want to remove it from `package.json` as well.

js-extensions/gulpfile.js
```diff
@@ -1,32 +1,75 @@
...
-     builder.gulp.src('spec/' + filter + '-spec.js')
-     .pipe(mocha({
-         compilers: { js: babel }
-     })).on('error', function(e) {
-         if (builder.isRetest()) {
-             // ignore test failures if we are running retest.
-             return;
-         }
-         throw e;
-     });
```
cliffmeyers: Removing this would break the ability to run tests via `npm run test`?

michaelneale:
@cliffmeyers so - not needed? cc @NicuPascu ?

kzantow:
So, after thinking about this a bit, I think it might be a better idea not to do this here, at least yet. The reason is.. depending on the changes made to js-extensions, it could end up breaking downstream plugins. Sure, upstream changes of any type might... but this one actually contributes to the way the bundles get built - e.g. they have to install it from npm, any changes to that or the way metadata is generated at build time, etc. would be bad. In other words, making changes to js-extensions at this point might be better to make it hard to do. The other facet is, yes, I have some ideas that would make this a lot easier, just a matter of assuring the time is best spent there.

kzantow:
Also, this one isn't actually bundled, it's just transpiled. So using js-builder is not really the right thing to do.

kzantow:
Could the same be accomplished by just doing a `"watch": "babel --watch --presets es2015,react src -d dist"`? If you wanted to move the presets to the `.babelrc` that's also great and simplify the commands.

imeredith:
@kzantow it would probably work, but can't we just make it the same as JDL and core-js?  It bugs me that we do things differently. We use gulp watch in core-js and JDL so it makes sense we would do it for js-extensions too.

kzantow:
@imeredith well, we have `bundle:watch` targets... because they are bundles. `js-extensions` is not a bundle, it's just a random NPM module. Keep in mind JDL and core-js are moving to their own bundles now.

sophistifunk:
Is there a good reason we can't just chuck this functionality into core-js? Do @scherler 's team need it separate?

PR #1091, opened 2017-05-25
JENKINS-43079# Trusted release related changes

vivek:
# Description

See [JENKINS-43079](https://issues.jenkins-ci.org/browse/JENKINS-43079).

How this works:
- Release build is run in a docker container with Java8 and maven 3.3.9
- Works with user scoped credentials
- Create github credentials ( github username and accessToken) and Jenkins credentials (username and password) in **user store**
- Chose which user is doing release from the drop down. At present there are only few, we can add more as needed

For it to work we need changes in pom file where it must be `https://` url under scm developerConnection setting. _This means anyone releasing from their laptop (which we do not want anyways) will no longer be able to use their ssh keys. However there are ways to automate it using GitHub access token._

I have placed release pipeline script under `release` folder for the time being mostly to track changes, we should probably move it out in it's own repo.

Tested doing beta release of another (smaller) plugin on dogfood. It needs to be tested on release environment whenever that gets fixed. 

Script uses credential input step that are not supported yet in BlueOcean, here is how the build screen looks like to do release. 

image

image

TODOs:

[ ] Test it in proper env to release beta version of blueocean

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/release/Jenkinsfile
+++ b/release/Jenkinsfile
@@ -1,0 +1,123 @@
+/**
+ * Pipeline script to be used for releasing BlueOcean
+ */
+node {
+    properties([parameters([
+            choice(choices: 'Vivek Pandey:vivek.pandey@gmail.com\nMichael Neale:mneale@cloudbees.com\nJames Dumay:jdumay@cloudbees.com', description: 'Colon separated Git Username and email.', name: 'GIT_AUTHOR_AND_GIT_EMAIL'),
+            [$class: 'CredentialsParameterDefinition', credentialType: 'com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl', defaultValue: '', description: 'Your GitHub credential', name: 'GITHUB_CREDENTIALS', required: true],
+            [$class: 'CredentialsParameterDefinition', credentialType: 'com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl', defaultValue: '', description: 'Your Jenkins credential', name: 'JENKINS_CREDENTIALS', required: true],
+            string(defaultValue: 'https://github.com/jenkinsci/blueocean-plugin.git', description: 'GIT repo URL', name: 'GIT_REPO_URL'),
+            string(defaultValue: 'master', description: 'GIT branch name to release', name: 'BRANCH')
+    ])])
+
+    deleteDir()
+
+    docker.image('maven:3.3-jdk-8').inside() {
+        def gitNameEmail = (env.GIT_AUTHOR_AND_GIT_EMAIL).split(':')
+        println gitNameEmail[0] + ":" + gitNameEmail[1]
+        String gitName = gitNameEmail[0]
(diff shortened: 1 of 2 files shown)


michaelneale:
sounds like it might be ok @vivek 
Not sure where we will do a 1.1 from though - so up to you when/if this merges.

vivek:
@michaelneale yeah, I guess I would like  to merge it but will have to wait for actually releasing using it till env is ready. 

In any case, we want to do a beta release first from release env (to avoid any unexpected error that might have been missed) and then actually do 1.1 or 1.2 release based on how soon we have the env. Sounds good?

michaelneale:
@vivek yes - not sure we will have the release environment in time though.

release/Jenkinsfile
```diff
@@ -0,0 +1,123 @@
...
+                        sh 'git push -u origin $BRANCH'
+                    }
+
+                    //Sanity build and test
+                    stage('Build and test') {
+                        sh 'echo Java version: `java -version`'
+                        sh 'echo Maven version: `mvn --version`'
+                        sh 'echo Git version: `git --version`'
+
+                        sh 'mvn -T 1C  clean install'
```
michaelneale: I changed my mind - we can drop the test. We want to trust the ATH and previous pipeline runs here anyway.

release/Jenkinsfile
```diff
@@ -0,0 +1,123 @@
...
+    deleteDir()
+
+    docker.image('maven:3.3-jdk-8').inside() {
+        def gitNameEmail = (env.GIT_AUTHOR_AND_GIT_EMAIL).split(':')
+        println gitNameEmail[0] + ":" + gitNameEmail[1]
+        String gitName = gitNameEmail[0]
+        String gitEmail = gitNameEmail[1]
+        withEnv(["GIT_COMMITTER_EMAIL=$gitEmail", "GIT_COMMITTER_NAME=$gitName", "GIT_AUTHOR_NAME=$gitName", "GIT_AUTHOR_EMAIL=$gitEmail"]) {
+            withCredentials([
+                    usernamePassword(credentialsId: '${JENKINS_CREDENTIALS}', passwordVariable: 'JENKINS_PASSWORD', usernameVariable: 'JENKINS_USERNAME'),
```
michaelneale: should this be the cleartext version of it or as encrypted in my desktop settings.xml? 
(probably need to go over this some more)

vivek: Its clear text. If it's running in insecure env, we better keep them encrypted but then we need to tell maven where encryption keys are so anyone who can get to env variable can possibly get to those encryption keys. Something we should run it by Ben?

michaelneale:
my version of the Jenkinsfile: 

https://gist.github.com/michaelneale/821e20c6d417b9ae89eea91599d00183

vivek:
Merging it in order to test a release, main thing is developerConnection url. Once a script is tested to work, we will update it in separate PR.

PR #1093, opened 2017-05-26
lets see if this makes it more robust

michaelneale:
# Description
 cc @imeredith  I tweaked a few assertions to make this less likely to fail on things that seem to be irrelevant


imeredith:
@michaelneale My commitmessagestest is merged into master with the github test. I think its ok but I just had a failure locally, so will keep an eye on it.

PR #1094, opened 2017-05-26
documentation for extension points

michaelneale:
# Description

@kzantow @i386 I have clarified some docs on the current state of extension points. Make sense? 

a little arm wavy on purpose.


i386:
🐝

PR #1098, opened 2017-05-29
this is needed for ATH and LTS

michaelneale:
# Description

Ths makes ATH run against LTS version - for reasons unknown. 
I am not sure we want to add a specific version in like this, but not sure what else to do. Opened for disucssion

cc @vivek @i386 @imeredith

### DIFF
--- a/blueocean/pom.xml
+++ b/blueocean/pom.xml
@@ -78,7 +78,10 @@
             <artifactId>blueocean-git-pipeline</artifactId>
         </dependency>
 
-
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>credentials-binding</artifactId>
+        </dependency>
 
         <!-- Bundled plugins for improved out of the box experience -->
         <dependency>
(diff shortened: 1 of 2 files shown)


michaelneale:
@vivek any idea why https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/fix%2Fcredentials-binding-version/4/pipeline would fail due to jacoco - I removed the stuff from the pom in the previous commit - and it was fine. Somehow that small change tricks jacoco?

vivek:
@michaelneale   Perhaps some t sets failed in that module, the might reflect as jacoco failure. Maybe something else going on there or something spurious. Btw, what failure is requiring addition of credentials-binding plugin? None of our production code needs it, adding it is probably hiding the symptom. If at all ATH needs it, should be added in ATH pom, not blueocean aggregator Pom?

michaelneale:
@vivek the ATH required it (sent a separate email, separate ticket to investigate) when running with latest LTS (as it picks a non sensical combination of versions): 

https://issues.jenkins-ci.org/browse/JENKINS-44533

PR #1100, opened 2017-05-29
parallel clicking: can it really be this trivial

michaelneale:
# Description

See https://issues.jenkins-ci.org/browse/JENKINS-44530

could it be this simple @cliffmeyers @kzantow ?

@i386 do you want to try this one? 

The data seems to have it in it, so why not use it. That calculation never could have worked.


i386:
I will try this out.
whoa

michaelneale:
and @imeredith has terrified with with closure scopes and functions and how this might accientally work. I will now have nighmares.

i386:
I tested with the examples on JENKINS-44530 and JENKINS-44086 - IT WORKS LIKE A CHARM! 

Nice hacking, dude! 💯 

watch-out-we-got-a-badass-over-here-meme

michaelneale:
ok @imeredith is this any better? if so, @i386 this should work the same (worth a look at some point).

PR #1102, opened 2017-05-30
Search API enhancements

i386:
# Description

Ensures that the search patterns detailed in [this document](https://docs.google.com/a/cloudbees.com/document/d/1riU62MGFF7A9cInJx04gIwZQ8PGwN9MkQQORC3dGq40/edit?usp=sharing) can be run against the search API.

See [JENKINS-38982](https://issues.jenkins-ci.org/browse/JENKINS-38982).

@michaelneale @imeredith @NicuPascu this means the search API can accept all the types we discussed in the use case document a few weeks ago

Target of this PR is set to @NicuPascu's branch for https://github.com/jenkinsci/blueocean-plugin/pull/1043

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineSearch.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineSearch.java
@@ -5,6 +5,7 @@ import hudson.Plugin;
 import hudson.model.Item;
 import hudson.model.ItemGroup;
 import io.jenkins.blueocean.commons.ServiceException;
+import io.jenkins.blueocean.commons.ServiceException.UnexpectedErrorException;
 import io.jenkins.blueocean.rest.OmniSearch;
 import io.jenkins.blueocean.rest.Query;
 import io.jenkins.blueocean.rest.factory.organization.OrganizationFactory;
@@ -16,6 +17,7 @@ import jenkins.model.Jenkins;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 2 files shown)


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/PipelineSearch.java
```diff
@@ -107,11 +108,21 @@ public String getType() {
...
             GlobMatcher matcher = pipeline.contains("*") ? new GlobMatcher(pipeline) : null;
-            while (pipelineIterator.hasNext()) {
-                BluePipeline p = pipelineIterator.next();
-                // If using glob syntax try to match using the glob matcher otherwise fall back to equality check
-                if (matcher != null && matcher.matches(p.getFullName())) {
-                    pipelines.add(p);
+            if (matcher != null) {
+                while (pipelineIterator.hasNext()) {
+                    BluePipeline p = pipelineIterator.next();
+                    String decodedName = URLDecoder.decode(p.getFullDisplayName());
```
vivek: p.getFullDisplayName() is individual job/folder's encoded display name separated by '/'.  For example if there is branch with display name `feature/x-1` inside org folder `vivek`. It will appear as `vivek/feature%2Fx-1`. url decoded form will be `vivek/feature/x-1`. Just making sure query is sent accordingly. 

Also, better to use `URLDecoder.decode(param)` is deprecated, use the one that takes encoding in addition, "UTF-8"?

i386: Fixed the deprecated API use.

I'm using `getFullDisplayName` as we are filtering based on what the user sees. Perhaps this should be moved to a different omnisearch implementation that is just used for user search?

vivek: @i386 I think it's alright as its still in context of pipeline search.

i386:
@nicupascu I've merged this into your branch so you'll have to pull updates

NicuPascu:
awesome, thank you

PR #1105, opened 2017-05-31
JENKINS-44571; Fix show more logs

imeredith:
# Description

It seems to work but I don't know what what I deleted was doing. 

See [JENKINS-44571](https://issues.jenkins-ci.org/browse/JENKINS-44571).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
+++ b/blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
@@ -35,14 +35,9 @@ export class LogConsole extends Component {
     componentWillReceiveProps(nextProps) { // eslint-disable-line
         logger.debug('newProps isArray', Array.isArray(nextProps.logArray));
         // We need a shallow copy of the ObservableArray to "cast" it down to normal array
-        const newArray = nextProps.logArray !== undefined && nextProps.logArray.slice();
-        const oldArray = this.props.logArray !== undefined && this.props.logArray.slice();
-        // if have a new logArray, simply add it to the queue and wait for next tick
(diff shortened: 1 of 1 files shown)


cliffmeyers:
IMO there's a bunch of work that needs to be done on this component and we'd be better suited using the bandaid in #1107 and then doing more involved work in separate ticket after 1.1.

Many moons ago, I added some "batching" behavior to `LogConsole` that would sequentially call the `render` method using more and more log data, with the hope that React's DOM diffing could pretty quickly update the DOM by adding just the new rows. This would help reduce browser lock up when you threw it 10K+ log lines at once.

At some point the `isLoading` flag was introduced and it prevents any log elements from being returned from `render` while this log concat process was happening. This seems like an accident, as all it seems to do now is pause 500ms between some array concatenation but not actually render 
anything to the screen. The end result is that React tries to render 10k+ log lines all at once and the browser never seems to recover.

I experimented with removing the `isLoading` guard before the Array.map call for the log lines, so that the "batched rendering" could actually happen. The screen does update, but very very slowly. The browser is still responsive but there's a ton of lag while trying to scroll while new elements are being added. For a 50K line log file it took a couple of minutes or so to finish up. Eventually it does and you're able to scroll, still a bit laggy but much improved over that.

cliffmeyers:
cc @imeredith @i386 @michaelneale @kzantow

michaelneale:
@cliffmeyers that is exactly right - there is a whole epic to focus on log/console stuff. It could be someone full time for a month or two (or spread out over 2 people I think) - but the whole area needs focus (there are known issues with badness and missing features we really need).

PR #1112, opened 2017-06-01
this may solve flaky test

michaelneale:
# Description
cc @vivek  PTAL

The old test was using parallel - which I don't think is needed there (we test that elsewhere actually, including inputs, in the ATH). I think that resulted in things not always being in the right order. Made it simpler and less brittle.


michaelneale:
cc @cliffmeyers not sure if this one flaked out on you - but I saw it several times today. Easy fix.

PR #1114, opened 2017-06-01
Enable the use of multiple Jenkins versions

imeredith:
Installs later deps in ATH.

Other changes in PR make tests nicer to work with.


michaelneale:
as long as passes :bee:

PR #1117, opened 2017-06-02
try this out

michaelneale:
# Description
move editor test case into its own class


michaelneale:
basically just ivan's thing - so merging

PR #1137, opened 2017-06-08
encode search string when pushing it to the url string

NicuPascu:
fixes the search param not being url safe by encoding the value before pushing it to the url

See [JENKINS-44759](https://issues.jenkins-ci.org/browse/JENKINS-44759).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


imeredith:
LGTM 🐝

i386:
@cliffmeyers do we have tests for fast search? How can we improve coverage?

imeredith:
@i386 I believe ATH test for this would be best. We could unit test it specifically too, although probably just that it is using the correctly search term for the pager. The display code is identical for search vs non-search.

PR #1141, opened 2017-06-08
Use basic auth instead of JWT in tests.

imeredith:
# Description
@vivek seeing occasionally flaky tests that I believe are due to JWT not setting the user properly. Will be documenting it in a ticket for you to take a look at

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/ProfileApiTest.java
```diff
@@ -365,7 +365,7 @@ public void badTokenTest1() throws Exception {
     }
 
 
-    @Test
+    //@Test
```
cliffmeyers: Should `@Ignore` be used so we can see in reports that they are actually skipped?

imeredith:
I am trying that but i dont see how that can fix anything given we set that prop in BaseTest
```
  public BaseTest() {
        System.setProperty("BLUEOCEAN_FEATURE_JWT_AUTHENTICATION", "true");
        j = new JenkinsRule();
    }
```

vivek:
@imeredith thats inside constructor, static blocks are loaded during class load time and much before JenkinsRule or anything else gets setup. I think it's worth a shot.

imeredith:
@vivek That static block thing didnt work. Merging this for now because it is blocking other people from getting things done. Can we add them back in if we figure it out.

vivek:
@imeredith 👍

i386:
Cherry picked to release branch

PR #1145, opened 2017-06-12
Story/jenkins 40855 GitHub enterprise credential / domain changes

cliffmeyers:
# Description
After some extended discussion with @vivek we decided to do the following here:
- Leave `/scm/github` endpoint as-is for now, to avoid changing that API.
- Make some changes to "new" `/scm/github-enterprise/validate` API:
  - `apiUrl` becomes a required parameter, whereas it's optional for github
  - All credentials are now stored in single domain named `blueocean-github-enterprise-domain`
- Add "credentials" property to `/scm/github-enterprise` API:
  - Credentials are exposed as a `credentials` property on the enterprise SCM object (see below)
  - The UI can fetch this data and be aware of which API URL's already have a credential registered. I "normalize" the URI when credential is created by always stripping off the trailing slash with the hope we can avoid silly mismatch errors that way.
- [ ] Need to decide if raw URL is okay or whether we really need a 2-way decode. If needed, perhaps Base64 would work?

```
{
    "_class": "...",
    "_links": { ... },
    "credentialId": null,
    "id": "github-enterprise",
    "credentials": [{
        "apiUri": "https://api.github.com",
        "credentialId": "blueocean-github-enterprise-domain:https://api.github.com"
    }]
}
```

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
```diff
@@ -230,7 +227,7 @@ public HttpResponse validateAndCreate(@JsonBody JSONObject request) {
             //Now we know the token is valid. Lets find credential
             StandardUsernamePasswordCredentials githubCredential = CredentialsUtils.findCredential(getId(), StandardUsernamePasswordCredentials.class, new BlueOceanDomainRequirement());
 
-            final StandardUsernamePasswordCredentials credential = new UsernamePasswordCredentialsImpl(CredentialsScope.USER, "github", "Github Access Token", authenticatedUser.getId(), accessToken);
+            final StandardUsernamePasswordCredentials credential = new UsernamePasswordCredentialsImpl(CredentialsScope.USER, getId(), "Github Access Token", authenticatedUser.getId(), accessToken);
```
cliffmeyers: This is intended to avoid two credentials with the same ID of "github" - when I originally tested this code, the cred tied to GitHub Enterprise was actually greyed out in the UI so I assumed there was some kind of conflict happening. This also led to the `credentialId` field in the response being null which is a problem.

i386: I think we need to override `getId` in `GithubEnterpriseScm` to return the `apiUrl`. This would allow you to support multiple Github Enterprise servers that have entirely different credentials.

cliffmeyers: We'll need to discuss how to handle this on the UI side. At present, the UI calls into an endpoint such as `/blue/rest/organizations/${orgname}/scm/${scmId}` to look for a single credential that's tied to the SCM ID. To support multiple SCM's, we probably need some changes that are non-trivial, e.g.

- User selects "GitHub Enterprise"
- UI makes an API call to see if any "GHE" tokens have been registered. Perhaps the endpoint /blue/rest/organizations/${orgname}/scm/github-enterprise` needs to return multiple tokens?
-- If no tokens, user sees "Add Token" step as reflected in design comp
-- If at least one token, user sees list of tokens registered (including the API URL endpoint), as well as an option to "Add New..."
- If user selects existing token, flow proceeds as normal...
- If user selects "Add New...", user sees same "Add Token" step as reflected in design comp

Does this sound correct, or am I making this more complicated that necessary? @i386

i386: Hmm, for `https://github.mycorp.com` there should be 1 credential for that URL. Setting up another Github Enterprise `http://github.example.com` shouldn't use the mycorp credential.

cliffmeyers: One credential per URL makes sense. But the UI needs to make an initial fetch to know which GitHub Enterprise credentials have been saved. Something like...

`/blue/rest/organizations/${orgname}/scm/github-enterprise`
returns
`[ { apiUrl: "https://ghe.mycorp1.com" }, { apiUrl: "https://ghe.somellc2.com" } ]`

And if multiple credentials do exist, the user needs a way to pick one. Unless you're thinking that each endpoint corresponds to an organization? I suppose that's possible but it seems like a mismatch to me.

cliffmeyers:
@vivek I've implemented the changes we discussed yesterday.  @i386 PTAL as well.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
```diff
@@ -30,14 +35,46 @@ public GithubEnterpriseScm(Reachable parent) {
     }
 
     @Override
+    public @Nonnull String getUri() {
+        String apiUri = getCustomApiUri();
+
+        // NOTE: GithubEnterpriseScm requires that the apiUri be specified
+        if (StringUtils.isEmpty(apiUri)) {
+            throw new ServiceException.BadRequestException(new ErrorMessage(400, "URI is required"));
```
vivek: error should be `apiUrl is required parameter`?

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
```diff
@@ -30,14 +35,46 @@ public GithubEnterpriseScm(Reachable parent) {
...
-        java.net.URI uri;
-        try {
-            uri = new URI(getUri());
-        } catch (URISyntaxException e) {
-            throw new ServiceException.UnexpectedErrorException(new ErrorMessage(400, "Invalid Github Enterprise URI: "+getUri()));
+        return DOMAIN_NAME;
+    }
+
+    @Exported(name = CREDENTIALS)
+    public List<GithubEnterpriseScmCredential> getCredentials() {
```
vivek: @cliffmeyers Hmm, I think its not going to fit in here. The thing is if we know apiUrl then it should return at most single element in the list, that is credential for the given apiUrl.  Then that means, `getCredentialId()` would return that same credential anyways. Basically we do not need 'credentials' element in the response in that case.

I think we missed important aspect in our discussion, that is if apiUrl is optional in case of GHE, then Scm abstraction doesn't work as it is, because it returns single scm url. See `Scm.getUri()`.  In that case we need collection of Scm in case of GHE, each on to represent instance of GHE.

I think we are better of calling Credential Api which can give list of credentials for given user (user store) and domain. If there is any issue in that API we can fix such query capability but I am fairly certain its all there. 

My take on this after more thinking is that we should not implement `GithubEnterpriseScm.getCredentials()`.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScmCredential.java
```diff
@@ -0,0 +1,26 @@
+package io.jenkins.blueocean.blueocean_github_pipeline;
+
+import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
+import org.kohsuke.stapler.export.Exported;
+import org.kohsuke.stapler.export.ExportedBean;
+
+
+@ExportedBean(defaultVisibility = 9999)
+public class GithubEnterpriseScmCredential {
```
vivek: Based on my comment above, we do not need this.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
```diff
@@ -185,6 +175,36 @@ public Link getLink() {
...
+        if (!StringUtils.isEmpty(apiUri)) {
+            java.net.URI uri;
+
+            try {
+                uri = new URI(apiUri);
+            } catch (URISyntaxException ex) {
+                throw new ServiceException.BadRequestException(new ErrorMessage(400, "Invalid URI: " + apiUri));
+            }
+
+            customUrl = uri.toString();
```
vivek: You don't really need it. It's not like `new URI()` does any normalization, I guess you are doing it for the reason to fail fast. You can avoid using `customUrl` variable and simply use apiUrl variable for all purposes.

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/CredentialsUtils.java
```diff
@@ -118,6 +119,23 @@ public boolean apply(@Nullable Credentials input) {
...
+            return ImmutableList.of();
+        }
+
+        Domain domain = store.getDomainByName(domainName);
+
+        if (domain == null) {
+            return ImmutableList.of();
+        }
+
+        List<C> filtered = (List<C>) CredentialsMatchers.filter(store.getCredentials(domain), CredentialsMatchers.instanceOf(type));
```
vivek: simply return it?

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
cliffmeyers: @vivek So are you proposing for GHE that we call general-purpose "credential search" API similar to how we do today for Git pipelines? Would we make similar changes to existing GH code too? In principle that sounds fine to me. However, we need to verify we can still correlate the `credentialId` to the `apiUrl` the user specified. In the current credentials model I don't think this is possible?

If I am misunderstanding your suggested approach then perhaps we can jump on a call in a couple hours if you are available.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
cliffmeyers: Fixed here:  https://github.com/jenkinsci/blueocean-plugin/pull/1145/commits/32f33e3ca8c8cec098612dbfd4962a68955b2365#diff-ea94a0b6b0be097866123d104e92d50bR42

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
cliffmeyers: Fixed in 32f33e3ca8c8cec098612dbfd4962a68955b2365

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScmCredential.java
cliffmeyers: Removed in 32f33e3ca8c8cec098612dbfd4962a68955b2365

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
cliffmeyers: Simplified here:  d32a75f1bcb1b9cd58222d0448913895ae9a862a

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/CredentialsUtils.java
cliffmeyers: Fixed c866131a3b2929e8ce578be53895287406b2d30f

cliffmeyers:
@vivek PTAL at the last round of changes. I improved test coverage to account for 200, 400 and 404 responses when performing GET request against `/scm/github-enterprise`.

cliffmeyers:
@vivek PTAL when you have a chance. I've implemented the changes we discussed yesterday. I think we have pretty good test coverage for GET requests against `/scm/github-enterprise` as I check for 200, 400 and 404 responses now.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
```diff
@@ -30,14 +35,49 @@ public GithubEnterpriseScm(Reachable parent) {
...
-        return DOMAIN_NAME + "-" + uri.getHost();
+
+        throw new ServiceException.NotFoundException("Credential not found for apiUrl=" + apiUri);
+    }
+
+
+    @Override
+    protected String createCredentialId(@Nonnull String apiUri) {
+        String domainName = getCredentialDomainName();
+        return domainName + ":" + DigestUtils.sha256Hex(apiUri);
```
vivek: This should be getId()+":"+DigestUtils.sha256Hex(apiUrl);

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
```diff
@@ -30,14 +35,49 @@ public GithubEnterpriseScm(Reachable parent) {
...
+        return DOMAIN_NAME;
+    }
+
+    @WebMethod(name="") @GET @TreeResponse
+    public Object getState() {
+        // will enforce that the apiUrl parameter has been sent
+        String apiUri = getUri();
+
+        User user = getAuthenticatedUser();
+        List<StandardUsernamePasswordCredentials> credentials = CredentialsUtils.findCredentials(StandardUsernamePasswordCredentials.class, user, DOMAIN_NAME);
```
vivek: You can do `CredentialsUtils.findCredential(createCredentialId(apiUrl), StandardUsernamePasswordCredentials.class, new BlueOceanDomainRequirement())`, this will save you from iterating over credentials and finding the right one.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
```diff
@@ -30,14 +35,49 @@ public GithubEnterpriseScm(Reachable parent) {
...
+
+        for (StandardUsernamePasswordCredentials cred : credentials) {
+            if (cred.getId().equals(getCredentialId())) {
+                return this;
+            }
         }
-        return DOMAIN_NAME + "-" + uri.getHost();
+
+        throw new ServiceException.NotFoundException("Credential not found for apiUrl=" + apiUri);
```
vivek: You may like to return array of errors instead of individual ones.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
```diff
@@ -30,14 +35,49 @@ public GithubEnterpriseScm(Reachable parent) {
...
+    }
+
+
+    @Override
+    protected String createCredentialId(@Nonnull String apiUri) {
```
vivek: return should be `@Nonnull` as well?

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
```diff
@@ -29,13 +29,7 @@
...
 import org.apache.commons.lang3.StringUtils;
 import org.jenkinsci.plugins.github_branch_source.GitHubSCMSource;
-import org.kohsuke.github.GHMyself;
-import org.kohsuke.github.GHOrganization;
-import org.kohsuke.github.GHUser;
-import org.kohsuke.github.GitHub;
-import org.kohsuke.github.GitHubBuilder;
-import org.kohsuke.github.HttpException;
-import org.kohsuke.github.RateLimitHandler;
+import org.kohsuke.github.*;
```
vivek: We do not wildcard imports, I set very high limit in idea (under import settings).

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
```diff
@@ -45,16 +39,8 @@
...
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.net.URL;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
+import java.net.*;
```
vivek: Same as above, explicit imports, no wildcards.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
```diff
@@ -45,16 +39,8 @@
...
+import java.util.*;
```
vivek: Please use explicit imports.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
```diff
@@ -228,24 +245,30 @@ public HttpResponse validateAndCreate(@JsonBody JSONObject request) {
             }
 
             //Now we know the token is valid. Lets find credential
-            StandardUsernamePasswordCredentials githubCredential = CredentialsUtils.findCredential(getId(), StandardUsernamePasswordCredentials.class, new BlueOceanDomainRequirement());
-
-            final StandardUsernamePasswordCredentials credential = new UsernamePasswordCredentialsImpl(CredentialsScope.USER, "github", "Github Access Token", authenticatedUser.getId(), accessToken);
-
+            String credentialId = createCredentialId(getUri());
+            StandardUsernamePasswordCredentials githubCredential = CredentialsUtils.findCredential(credentialId, StandardUsernamePasswordCredentials.class, new BlueOceanDomainRequirement());
+            final StandardUsernamePasswordCredentials credential = new UsernamePasswordCredentialsImpl(CredentialsScope.USER, credentialId, "Github Access Token", authenticatedUser.getId(), accessToken);
```
vivek: Should be `Github Enterprise Access Token`.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
cliffmeyers: Fixed in 194545f580c4f95cf72bfdbea81656ec5c5899fc

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
cliffmeyers: Fixed in 769b1e147e06ee61118fb2859f28006bf06f967c

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
cliffmeyers: Fixed in c1857e03da7dec22ac365fbdbf665eb6297f9e23

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
cliffmeyers: Fixed in 44b92f883c0977363faf89a8c54632729991849a

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
cliffmeyers: Fixed in 44b92f883c0977363faf89a8c54632729991849a

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
cliffmeyers: Fixed in 44b92f883c0977363faf89a8c54632729991849a

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScm.java
cliffmeyers: Fixed 242de503c1f593ca5cb7d011430e70b1698a3e5a

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
```diff
@@ -30,14 +35,50 @@ public GithubEnterpriseScm(Reachable parent) {
...
+        if (StringUtils.isEmpty(apiUri)) {
+            throw new ServiceException.BadRequestException(new ErrorMessage(400, "apiUrl is required parameter"));
+        }
+
+        return apiUri;
+    }
+
+    @Override
+    public String getCredentialId() {
+        return createCredentialId(getUri());
```
vivek: As discussed this should be using 
```
String credId = createCredentialId(apiUri)'
Credentials cred = CredentialsUtils.findCredential(credId, StandardUsernamePasswordCredentials.class, new BlueOceanDomainRequirement());
return cred!=null?credId:null;

File: blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java [hunk 16]
```diff
@@ -30,14 +35,50 @@ public GithubEnterpriseScm(Reachable parent) {
...
+    }
+
+    @WebMethod(name="") @GET @TreeResponse
+    public Object getState() {
+        // will enforce that the apiUrl parameter has been sent
+        String apiUri = getUri();
+
+        StandardUsernamePasswordCredentials credential = CredentialsUtils.findCredential(createCredentialId(apiUri), StandardUsernamePasswordCredentials.class, new BlueOceanDomainRequirement());
+
+        if (credential == null) {
```
vivek: As discussed lets not check credential here, as there may not be credential setup in the first place so we should not fail due to this.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
cliffmeyers: Fixed in ef5999fbe389c27ba46865f2a9e133c0dee4fca2

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
cliffmeyers: Fixed in ef5999fbe389c27ba46865f2a9e133c0dee4fca2

vivek:
@cliffmeyers LGTM 🐝 pending CI passes successfully.

PR #1148, opened 2017-06-14
fix official Docker image

ydubreuil:
install-plugins.sh script downloads a too recent version of workflow-job
let's force it to latest compatible version with 2.46 LTS to quickly fix
the Official image.

This is in order to fix the actual 1.1.0 image. 

@i386 @rtyler

### DIFF
--- a/docker/official/Dockerfile
+++ b/docker/official/Dockerfile
@@ -6,6 +6,6 @@ USER root
 # Add the docker binary so running Docker commands from the master works nicely
 RUN apk -U add docker
 
-RUN install-plugins.sh antisamy-markup-formatter matrix-auth pipeline-model-definition blueocean:$BLUEOCEAN_VERSION
+RUN install-plugins.sh workflow-job:2.11 antisamy-markup-formatter matrix-auth blueocean:$BLUEOCEAN_VERSION
 
 USER jenkins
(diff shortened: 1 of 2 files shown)


i386:
LGTM @ydubreuil 👍

rtyler:
This pull request doesn't actually update the Jenkinsfile in any way to allow for the `FORCE_REBUILD` option to be passed in as a parameter. As such, I'm not sure how it's expected that this should be used

ydubreuil:
@rtyler I'll follow up with an update of Jenkinsfile

ydubreuil:
Follow up: https://github.com/jenkinsci/blueocean-plugin/pull/1151

ndeloof:
install-plugins.sh is fixed upstream, but the docker image `jenkinsci/jenkins-lts` used by this image has not be regenerated as the Dockerfile was improved, and will probably only be updated on next LTS release. @rtyler I can't see where this image is built, used to run on https://jenkins.ci.cloudbees.com/job/docker/ but moved to jenkins.io IIUC. We should keep all (recent) tags up to date with repository.

`jenkins` official docker image is up-to-date.

rtyler:
@ndeloof Because publishing the Docker container requires what would be considered "release credentials", we run this in an environment referred to as "trusted.ci.jenkins.io" (not a real name) which is off the public internet.

The next time an LTS is released, the Pipeline from jenkinsci/docker will execute automatically and publish those new images. I think the current solution from @ydubreuil will be sufficient until then.

PR #1157, opened 2017-06-15
bug / jenkins 44522 top nav pipelines link highlight

cliffmeyers:
# Description
- See [JENKINS-44522](https://issues.jenkins-ci.org/browse/JENKINS-44522).
- This changes a little how the main "Pipelines" link works. Previously, in web, it was set up as the default link for the extension point, meaning anyone implementing that extension point would replace it. Now it's contributed by dashboard, so any extension will be additive. I'm not sure if that's actually a problem; I don't think anyone really cares about that use case but I could be wrong.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-dashboard/src/main/js/PipelineRoutes.jsx
```diff
@@ -114,7 +114,7 @@ function onLeaveCheckBackground() {
 }
 
 export default (
-    <Route path="/" component={Dashboard} onChange={persistBackgroundOnNavigationChange}>
```
sophistifunk: Does this mess with  `<link to="/">` within the app?

cliffmeyers: No, since in blueocean-web there's actually an `IndexRedirect` that pushes to `/pipelines`. I wonder if at some point we ought to make the default route an extension point so there isn't a hard-coded relationship between web and dashboard.

blueocean-dashboard/src/main/js/jenkins-js-extension.yaml
```diff
@@ -2,8 +2,8 @@
 # NB: "component" currently maps to modules, not "symbols" so make sure to "export default"
 # WARNING: If you change this you'll have to change io.jenkins.blueocean.jsextensions.JenkinsJSExtensionsTest as well :(
 extensions:
-  - component: AdminNavLink
-    extensionPoint: jenkins.logo.top
```
michaelneale: does this extension point still exist?

cliffmeyers: It does but this was its only usage in the "blueocean-plugin" repo. Perhaps we should remove it?

michaelneale: I think it may be used by others so best to leave it

blueocean/src/test/java/io/jenkins/blueocean/config/JenkinsJSExtensionsTest.java
```diff
@@ -74,7 +74,7 @@ public void test() {
         // Calling JenkinsJSExtensions.getJenkinsJSExtensionData() multiple times should
         // result in the same object instance being returned because the list of plugin
         // has not changed i.e. we have a simple optimization in there where we only scan
-        // the classpath if the active plugin lust has changed.
```
michaelneale: lust for plugins ;)

sophistifunk: Could just be i18n for noozulland.

sophistifunk:
The `PipelinesActive` global is kinda kludgey, we might want some comments in there to explain why it exists.

blueocean/src/test/java/io/jenkins/blueocean/config/JenkinsJSExtensionsTest.java
michaelneale: choice bro

cliffmeyers:
@sophistifunk happy to add comments. But back to one of the hard problems in CS: do you have a suggestion for a better name?

sophistifunk:
OOOoooh no, I'm not falling for that one. My vote is for explanatory comment, and we re-address this when we have multiple top-level tabs and need a better system.

michaelneale:
:bee: from me otherwise

michaelneale:
@cliffmeyers still happy with this?

cliffmeyers:
@michaelneale I'd like to add those comments as Josh suggested. I will circle back on this later today after sorting out some VPN issues and pushing forward w/ GHE work.

michaelneale:
@cliffmeyers no worries 👏

michaelneale:
@cliffmeyers hopefully can get to this this week?

blueocean-dashboard/src/main/js/jenkins-js-extension.yaml
scherler: it is used in our internal project, so good you left the extensionpoint.

michaelneale: @scherler good point

@cliffmeyers reminder: we should not remove/move extension points (add new ones) - think of it as a 18 karat gold rule. Not 24 karat as there may be exceptions, but 18 karat.

PR #1163, opened 2017-06-16
Use official image (with fixed install-script)

ndeloof:
related to https://github.com/jenkinsci/blueocean-plugin/pull/1148
using official `jenkins` docker image we benefit a fixed install-script which uses LTS update center to download plugins, so avoid installing incompatible

### DIFF
--- a/docker/official/Dockerfile
+++ b/docker/official/Dockerfile
@@ -1,11 +1,11 @@
 # This image is based off the latest Jenkins LTS
-FROM jenkinsci/jenkins:lts-alpine
+FROM jenkins:alpine
 
 USER root
 
 # Add the docker binary so running Docker commands from the master works nicely
 RUN apk -U add docker
 
-RUN install-plugins.sh workflow-job:2.11 antisamy-markup-formatter matrix-auth blueocean:$BLUEOCEAN_VERSION
+RUN install-plugins.sh antisamy-markup-formatter matrix-auth blueocean:$BLUEOCEAN_VERSION
 
 USER jenkins


i386:
@ndeloof @rtyler @michaelneale why were we using the `alpine` variant?

ndeloof:
@i386 why not ? :P 
this makes the image smaller for user to download it (even fat Jenkins war just make it big anyway)

michaelneale:
Massively reduced size. 
Builds should run not on master so no need for those other tools. 

@rtyker is best positioned to review this

i386:
shouldn't this change be made upstream so that alpine gets the right install-script ?

ndeloof:
@i386 proposed change IS to rely on upstream which includes the right install-scrip.
The issue here is blueocean relies on jenkinsci/jenkins-lts which isn't rebuilt but when a new LTS is released, i.e. install-script has been fixed since last LTS, waiting for the next release.

i386:
@ndeloof ahh I understand now. I am 👎 for changing this to depend on the non-alpine image type for the time being and the 2.60.x LTS is coming very soon - will be happy to wait until then.

For the record, this image is rebuilt automatically when new LTS releases become available so a new LTS release will fix all the things here.

Given the info above, is there any need for this PR?

ndeloof:
@i386 I didn't proposed to move to a non alpine image, but to use `jenkins:alpine` official image. 

Main goal of this PR is to avoid a fake dependency to workflow-job:2.11, that we will need to later remove, and have to remember why it has been added.

I'm also concerned we have both `jenkins` and `jenkinsci/jenkins-lts` for the same image but with distinct release cycles (cc @rtyler) :-\

i386:
@ndeloof ahh but then everyone would be on 2.65 or something non-lts right? We want our image to always follow LTS because there are problems in the weekly we do not want to deal with (we get bug reports :( )

ndeloof:
@i386 official docker image only provide LTS releases.

i386:
I am pretty confused then. @rtyler is on vacation for two weeks I believe.

ndeloof:
@i386 ok, let's resume :
- blueocean uses `jenkinsci/jenkins-lts:alpine`, which is only build when a new LTS release happens
- official docker image `jenkins:alpine` is built everytime ... one ask it to be updated. So typically when some fixes are provided to install-script, not waiting for a new LTS

both images are built from the exact same Dockerfile, just not with same lifecycle, which IMHO is just confusing for end-users.

- previous fix from Yoann added a fake dependency to `workflow-job:2.11` to enforce we don't get `workflow-job:latest` - which was the issue fixed in install-plugin script. We can wait for next LTS release to automatically update `jenkinsci/jenkins-lts:alpine` but then will need to remember we can/have to remove this dependency, to avoid later confusion

michaelneale:
Yes I think I understand, but Would like to wait for @rtyler to return. I understand the different release cycles of the 2 images, but want to make sure he is on board.

ydubreuil:
I'll let @rtyler chime in. I don't know why there's 2 different images for Jenkins LTS available on DockerHub

rtyler:
@ydubreuil The use of the "official images" `jenkins` library we have been **strongly not recommending** because we do not control its release cycle in any automated form. That image must be updated with a manual pull request which a Docker, Inc. employee must manually merge. 

The `jenkinsci/jenkins` image is automatically updated with new releases, including *critical* security releases.

Plainly stated, the `jenkins` image should never be used for production systems, ever.

ndeloof:
@rtyler ok, then we should just deprecate this official image, ensure no new release is pushed there, and document this on https://hub.docker.com/_/jenkins/ to avoid user confusion

Also, please configure CI so image is rebuilt on changes to github repo, not just new LTS release

ndeloof:
please note: Elastic did the same 
https://hub.docker.com/_/elasticsearch/

ndeloof:
I've created https://github.com/jenkinsci/docker/issues/515 on this topic, @rtyler would appreciate your advice on it.

michaelneale:
@ndeloof @rtyler the official "jenkins" image has 10M+ pulls (FWIW, not sure what counts as a pull). But so does elastic. A LOT of people use it, and few would look at the readme for it (not necessarily related to this PR, but just a point of concern). The "official" docker images get a lot of visibility, so if there is a problem with control of it, perhaps we need to talk to docker inc about it?

ndeloof:
I've been asking Docker to reconsider this official image release process for 2 years without any change, so don't expect a move.
As a new LTS is release and not available as "jenkins" official image, people will for sure learn to RTFM

michaelneale:
@ndeloof you put too much faith in people looking at what they are doing. What will happen is they will keep using it, building on it, and complain that jenkins is broken. The only way to not update the official one is to ask docker to remove it so no one can have it.

ndeloof:
They didn't for Elastic - they consider images as immutable things, so they can't be removed from the hub. Wonder they can implement some "redirect" ?

michaelneale:
@ndeloof right - a new image could just say "don't use me" - but not sure if it would be blessed. 
in any case - is this PR still relevant in this world?

ndeloof:
no, let's follow up on https://github.com/jenkinsci/docker/issues/515

PR #1167, opened 2017-06-19
Hotfix/test flaky stop

michaelneale:
This test seems to only be testing non blue ocean stuff... and it can't guarantee it will abort.


i386:
replaced by https://github.com/jenkinsci/blueocean-plugin/pull/1168

PR #1172, opened 2017-06-19
remove redundant assertions and stabilise abort test

michaelneale:
trying to stabilise build


michaelneale:
I am merging this - as it passed 2x in a row.

PR #1179, opened 2017-06-21
Ensure latest comes from the run container

i386:
# Description

Mic and I wrote this test to try to track down a bug (we thought latestRun might not have shown queued items). Thought it was worth keeping the test and ensuring that the latest run comes from the run container since the data her is synthesised (its a union of what we believe is in the queue and the runlist)

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractPipelineImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractPipelineImpl.java
@@ -84,10 +84,7 @@ public class AbstractPipelineImpl extends BluePipeline {
 
     @Override
     public BlueRun getLatestRun() {
-        if(job.getLastBuild() == null){
-            return null;
-        }
-        return AbstractRunImpl.getBlueRun(job.getLastBuild(), this);
+        return getRuns().iterator().next();
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractPipelineImpl.java
```diff
@@ -84,10 +84,7 @@ public Integer getWeatherScore() {
 
     @Override
     public BlueRun getLatestRun() {
-        if(job.getLastBuild() == null){
-            return null;
-        }
-        return AbstractRunImpl.getBlueRun(job.getLastBuild(), this);
+        return getRuns().iterator().next();
```
vivek: This might fail in case, a job is run for the first time and iterator().next() might result in to `NoSuchElementException` 🐛  I think the existing code is correct, where `job.getLastBuild()` returns null if `runs.isEmpty()` returns null value.

vivek: I think I missed the fact that BluePipeline.getRuns() gives merge of queued and active runs. I still think there is bug where if no job has been run, it might result in to `NoSuchElementException`.

i386: Pushed a fix here.

i386:
@vivek fixed the 🐛  and updated to test to cover null before first run.

PR #1183, opened 2017-06-21
ATH runtime dependencies bump

michaelneale:
# Description
ATH needs to run on newer jenkins versions - will see if this works first with vanilla


michaelneale:
hey @vivek and @imeredith this makes it light up on latest LTS at least - can you take a look and perhaps we can merge - and then try the weekly? I am not sure how to otherwise try it on the weekly job (other than locally)

imeredith:
:bee: if you want to test against weekly just modify the job config to pick up this branch

michaelneale:
running test now...

PR #1184, opened 2017-06-21
[JENKINS-44636] Report queued status for Pipelines correctly

abayer:
# Description

This is work in progress - downstream of https://github.com/jenkinsci/pipeline-graph-analysis-plugin/pull/11, https://github.com/jenkinsci/workflow-api-plugin/pull/42, and https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/42. 

Still needed - tests! UI! Doing something with the `ExecutorTaskInfoAction.getWhyBlocked()` value for queued tasks and the `WorkspaceAction.getNode()` value for running tasks! But we're getting there?

cc @reviewbybees @vivek @svanoort @i386 

See [JENKINS-44981](https://issues.jenkins-ci.org/browse/JENKINS-44981). And see [JENKINS-44636](https://issues.jenkins-ci.org/browse/JENKINS-44636), 'cos this may actually be more that than it is JENKINS-44981. Or they're the same thing in the end. Your call. =)

# Manual test
(from [JENKINS-44636](https://issues.jenkins-ci.org/browse/JENKINS-44636))

1. Setup the app store demo pipeline https://github.com/i386/app-store-demo
2. disable all executors
3. run the master branch
4. The run is reported in Blue Ocean as queued rather than running.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


abayer:
btw, if someone with rights wants to throw a `work-in-progress` label on this PR, I would appreciate it. =)

michaelneale:
Very nice - impressive this passes the gauntlett of all the tests!

I will kick the tyres of this today.

michaelneale:
huh, I was just resolving a conflict (nothing) and it does a merge from master ... I did not intend that (but it is fine...) just surprising ... (there is no way to update with master in cases where there are no conflicts)

michaelneale:
doens't set anything on fire - what am I looking for behavior wise?

abayer:
@michaelneale See [JENKINS-44636](https://issues.jenkins-ci.org/browse/JENKINS-44636), which this is more a fix for than it is for [JENKINS-44981](https://issues.jenkins-ci.org/browse/JENKINS-44981) the more I think about it. The latter will come out of this work, but isn't completely addressed here.

abayer:
Hrm - trying the directions at [JENKINS-44636](https://issues.jenkins-ci.org/browse/JENKINS-44636), once I've run the build once and then removed the executors and tried running again, I'm getting the JS error below - no idea where it's coming from or how to pursue it. =)

```
blueocean.js:6076 Error fetching page TypeError: Cannot read property 'toLowerCase' of undefined
    at TimeHarmonizerUtil.getI18nTitle (blueocean.js:2148)
    at LiveStatusIndicator.render (blueocean.js:1516)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (blueocean.js:85455)
    at ReactCompositeComponentWrapper._renderValidatedComponent (blueocean.js:85478)
    at ReactCompositeComponentWrapper.performInitialMount (blueocean.js:85020)
    at ReactCompositeComponentWrapper.mountComponent (blueocean.js:84916)
    at Object.mountComponent (blueocean.js:91929)
    at ReactCompositeComponentWrapper.performInitialMount (blueocean.js:85029)
    at ReactCompositeComponentWrapper.mountComponent (blueocean.js:84916)
    at Object.mountComponent (blueocean.js:91929)
(anonymous)	@	blueocean.js:6076
```

michaelneale:
@abayer yes maybe we need to leave this for @vivek when he has time next month?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
```diff
@@ -107,6 +108,22 @@ public BlueRunState getStateObj() {
...
+                isQueued = true;
+            } else if (n.getStateObj().equals(BlueRunState.RUNNING)) {
+                isRunning = true;
+            }
+        }
+        if (isQueued && !isRunning) {
+            return BlueRunState.QUEUED;
+
+        }
+
```
i386: @abayer you are seeing a `toLowerCase()` issue because something here is now returning null and this method is `@Nonnull`

i386:
@abayer  I made you a write collaborator on this repository so you can set your own labels etc

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
```diff
@@ -107,6 +108,22 @@ public BlueRunState getStateObj() {
...
         return super.getStateObj();
     }
```
i386: Make this whole method assign a single variable `state` instead of returning immediately (including `super.getStateObj()`) then `return state;` on last line. Set a breakpoint there and you should find the null pretty easily.

michaelneale:
Tested this out - it only shows as queued after I do a reload, so something isn't right with the front end code. It goes to running, and site there indefinitely. If I do a whole page refresh, it loads correctly. I think there is more to this change than meets the eye.

abayer:
@michaelneale Hrm. Any pointers on what could be the problem would be appreciated, since the frontend is all black magic to me. =)

michaelneale:
@abayer a compile error with the latest bump BTW. 

Right in this case, I think it is probably how the data is being pushed back into the client stores, not triggering a refreshing of things that should be refreshed. If you can find @cliffmeyers around this week before he takes off, or otherwise @kzantow they may be able to explain in an appropriate timezone (and actually I don't know the tricks - I am suspecting this is something around mobx). I'll shoot an email to try and set up a time!

So basically it is very very close, but likely need some help for the last bit (it's pretty funky)

abayer:
@michaelneale I'm wondering (from my very little knowledge as I explore the frontend code) if that might be happening because there's no new node or step added, so the frontend doesn't know to refresh?

cliffmeyers:
I should have a chance to look into this tomorrow.

abayer:
@michaelneale I think I nailed the event thing, but verify for me. =)

abayer:
Test failure seems unrelated?

michaelneale:
@abayer yes unrelated test failure - did you update to master recently?

abayer:
@michaelneale Yup.

michaelneale:
I don't know if I am doing it wrong - but when I start it - it still shows up as running until I refresh (I press run from the "branches" tab, on the master branch of the app-store-demo project)

abayer:
@michaelneale Aaaaah, ok, I wasn't testing that. Different event I think. Checking!

abayer:
So from what I can tell so far, the event I updated will fire on the karaoke page, not the branches page. That page, as best as I can tell, is updated based on job events, not Pipeline events, so this needs to fire an event like when the run becomes paused. For the moment, I'm hacking around it by firing a `job_run_started` event, but I'm betting we'll actually want something different - there are queue-related events, but those are for the whole job entering/leaving the queue, not part of a run.

abayer:
ah-ha, that did help. But now it never *leaves* the queued state!

abayer:
And I think I got it! Still needs work for JENKINS-44981 to make a permanent step row for `node` steps, but...we're getting close.

abayer:
@michaelneale Think it was legit, and fixed that one. Updating to master now.

michaelneale:
@abayer so that test is now fixed, but now seeing: 

```
pipelineLatestRunIncludesRunning – io.jenkins.blueocean.rest.impl.pipeline.AbstractRunImplTest
```

which also seems related to this general area - legit failure?

abayer:
@michaelneale Hmm, looks timing based. Will fix tomorrow.

abayer:
Joy, it's not consistent. That's always fun.

michaelneale:
@abayer yeah likely is just noise. If you update to master - Ivan has made some improvements.

abayer:
Trying to run again to see if the failure keeps happening - I can't reproduce locally, which always makes me twitch.

abayer:
Ok, new change up making sure we're not going too fast to properly get the status updated to `QUEUED` - there is a brief period in which the status is legitimately `RUNNING` before it slides over to `QUEUED` and I think the test runs were in some cases getting the status while it was in that window...

abayer:
Woo - the failures now seem to just be the `GitHubEnterpriseCreationTest` ATH tests, which...well, I'm not shocked that they don't pass. =)

cliffmeyers:
@abayer could also be that a change to eventing status has broken the creation tests' which wait for specific SSE events / statuses before redirecting to the activity page. I am catching up still after vacation but might be able to help take a look later today.

michaelneale:
@abayer did you update to master?

abayer:
@michaelneale Yes.

cliffmeyers:
@abayer is this good to go from your perspective with the exception of the `GitHubEnterpriseCreationTest` ATH test? Or is there more work to be done on the UI side? I should have cycles tomorrow to help out so perhaps you can let me know where I can lend a hand. Thanks!

abayer:
@cliffmeyers I *think* it's good to go - it's definitely ready for review. =)

michaelneale:
ah did find an additional problem: favorites - @cliffmeyers personal fav!

This wont' work with them - they just sit in running and never exit it. I am not sure if this would result in a regression or not, but it looks like there are some events or something missing (perhaps @cliffmeyers  can comment?). 

If it doesn't change existing behavior with favourites I would be ok with this, but it does seem like a :ant: for now. Are the upstream libs published now - and can be updated? if this is updated with them - may be ok for now, but there is a gap in functionality with favoutites (if you can find Cliff, he can probably point to where). 

but nice progress!

abayer:
I'll take a look at the favorites stuff today. And no, the upstream dependencies aren't merged/released yet.

abayer:
@cliffmeyers Do you know which events the favorites are listening for?

cliffmeyers:
@abayer so the events that it's listening for aren't exactly specific: it registers a generic "job" listener in `DashboardCards.componentWillMount`. For each event received, the `FavoritesSseListener._filterJobs` method compares the event's `blueocean_job_rest_url` against each favorite item's `_links.self.href` and ignores events where the event doesn't match one of the favorites. If there's a match, then `FavoritesStore` `UPDATE_RUN` action fires and the `latestRun` property of the favorite's item is updated to reflect the new status.

cliffmeyers:
@michaelneale @abayer what is the most effective way to be able to replicate the issue in favorites? Favorite cards update normally when I'm just running a couple pipelines here or there. Do I just need to run enough concurrently to exhaust the number of executors? I believe the default is 2?

cliffmeyers:
@michaelneale so to clarify, favorites cards have always misreported as running when multiple pipelines are run concurrently beyond the executor limit. This is because RUNNING events always fired even when the pipeline was really queued. If you refresh the page, the state of the card reports correctly as queued because the data coming from the favorites fetch is accurate. Right now this PR doesn't appear to change that behavior.

Here's a patch that adds some helpful logging with respect to favorites:
https://gist.github.com/cliffmeyers/fb33d0a198b49b4a2abc1ffa7c2373ad

It shows which events are received that match a user's favorite, and it shows the state of the run object that is fetched via REST API and used to update the data that drives the favorite card. Here's sample output for running a third pipeline when two are running:
```
received event QUEUED for user favorite histos/karaoke-samples-duration-10m-b Object {jenkins_object_type: "org.jenkinsci.plugins.workflow.job.WorkflowJob", jenkins_event_uuid: "9c4a482f-96a6-4de7-82a4-1383b79c5fc5", job_run_status: "QUEUED", job_name: "histos/karaoke-samples/duration-10m-b", jenkins_org: "jenkins"…}
updating run data QUEUED for user favorite histos/karaoke-samples Object {pipeline: "histos/karaoke-samples", _links: Object, state: "QUEUED", result: "UNKNOWN"}
received event ALLOCATED for user favorite histos/karaoke-samples-duration-10m-b Object {jenkins_object_type: "org.jenkinsci.plugins.workflow.job.WorkflowJob", jenkins_event_uuid: "00dd9c8b-f895-42c5-b418-70bfd2becd5b", job_run_status: "ALLOCATED", job_name: "histos/karaoke-samples/duration-10m-b", jenkins_org: "jenkins"…}
received event RUNNING for user favorite histos/karaoke-samples-duration-10m-b Object {jenkins_object_type: "org.jenkinsci.plugins.workflow.job.WorkflowRun", jenkins_event_uuid: "0db3cc22-a3f9-4566-bb24-ece931681942", job_run_status: "RUNNING", job_name: "histos/karaoke-samples/duration-10m-b", jenkins_org: "jenkins"…}
received event RUNNING for user favorite histos/karaoke-samples-duration-10m-b Object {jenkins_object_type: "org.jenkinsci.plugins.workflow.job.WorkflowRun", jenkins_event_uuid: "faa2cfcd-d3ce-490f-8dd1-fc14c1e2d2ca", job_run_status: "RUNNING", job_name: "histos/karaoke-samples/duration-10m-b", jenkins_org: "jenkins"…}
updating run data RUNNING for user favorite duration-10m-b Object {_class: "io.jenkins.blueocean.rest.impl.pipeline.PipelineRunImpl", _links: Object, actions: Array(0), artifactsZipFile: "/job/histos/job/karaoke-samples/job/duration-10m-b/3/artifact/*zip*/archive.zip", causeOfBlockage: "Waiting for next available executor"…}
updating run data RUNNING for user favorite duration-10m-b Object {_class: "io.jenkins.blueocean.rest.impl.pipeline.PipelineRunImpl", _links: Object, actions: Array(0), artifactsZipFile: "/job/histos/job/karaoke-samples/job/duration-10m-b/3/artifact/*zip*/archive.zip", causeOfBlockage: "Waiting for next available executor"…}
```

I might be stating the obvious, but to actually fix this long-standing issue, we need to do either of these:
1. Suppress the running events for the pipeline when it's not really running. This feels the most correct but I think it might be difficult due to how pipeline runs under the hood.
2. Add an extra queued event that fires after the running events. Apart from extra event load, it might cause some flicker in the UI (from initial state, to RUNNING, to QUEUED). We could mitigate that by debouncing the render in the UI, although we'd have to play with it a bit to make sure the app still feels responsive... we don't want the user to press a run button in the card and wait too long to get feedback.

abayer:
@cliffmeyers So, fwiw, this adds logic so that if there's at least one `node` block queued but no other branches running, the run as a whole will report as queued as well. If you disable all the executors on your master (from the traditional UI, click on `Build Executor Status`, then the gear to the right of the master in the table, then set the executor count to 0 and save) and then run a job trying to get an executor, you'll get that result. In the list of branches or runs for a branch in the Blue Ocean UI, the run in question will switch itself to queued, and then to running once you've added an executor again - https://github.com/jenkinsci/blueocean-plugin/pull/1184/commits/f08c78bb25e9ad000b7345528d3e9acef8ea3b06 is where I added the job events to make that happen. 

I'm not sure why they're not being picked up in favorites, but I also haven't had a chance to dig too deep due to other obligations today.

abayer:
Verified that as of now, the favorite won't go from running to queued, but does go from queued to running...

abayer:
So the rendering is getting triggered but for some reason, the state is still `RUNNING` at that point. Not sure why - or why that's not happening on the pipelines page...

michaelneale:
@cliffmeyers in that case - I think I am ok with this as it is if you are - ie "do no harm" - it is no worse than before right?

abayer:
Got further - it's that the `RunMessage` in the event automatically sets the `job_run_status` to `RUNNING`. Sigh. More granularity needed...

michaelneale:
@cliffmeyers I think we can just go ahead with this - ok with you?

cliffmeyers:
@michaelneale as far as I can tell nothing here has regressed. I think we should probably merge it. Perhaps we should file a ticket for the favorites status not be accurate? I think there actually already is one that was filed by me many many moons ago.

abayer:
@michaelneale @cliffmeyers I'm currently churning through the upstream PRs to get them merged and released (and to get the `workflow-cps` release done too, since there's a `SNAPSHOT` dependency here at the moment). Hopefully will have that all done tomorrowish, at which point this can be ready for final testing etc.

michaelneale:
ok thanks @abayer - just mention when you think this is ready for "final". Looking good functionally.

abayer:
Ok, with `pipeline-graph-analysis` 1.5 released, this should be good to go. @michaelneale @cliffmeyers @vivek - could someone do a final review? Thanks!

i386:
🐝 works as advertised! Nice job @abayer 💯

PR #1186, opened 2017-06-21
[JENKINS-43507] new branch source: Plugin version bump only on master

stephenc:
- Still using deprecated APIs

# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean/src/test/java/io/jenkins/blueocean/i18n/BlueI18nTest.java
+++ b/blueocean/src/test/java/io/jenkins/blueocean/i18n/BlueI18nTest.java
@@ -26,6 +26,7 @@ package io.jenkins.blueocean.i18n;
 import hudson.PluginWrapper;
 import io.jenkins.blueocean.service.embedded.BaseTest;
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.Before;
 import org.junit.Test;
 
@@ -130,6 +131,7 @@ public class BlueI18nTest extends BaseTest {
             return;
         }
         String version = plugin.getVersion();
+        Assume.assumeTrue("Test is only valid if git plugin is a suitable probe", version.matches("\"[\\\\d/.]{3,}\""));
         BlueI18n.BundleParams bundleParams = BlueI18n.getBundleParameters(String.format("git/%s/pluginx.bundle", version));
 
         // Should be cacheable because the installed version matches the requested version + the
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     <hpi.dependencyResolution>runtime</hpi.dependencyResolution>
     <jenkins-test-harness.version>2.15</jenkins-test-harness.version>
     <scm-api.version>2.2.0-alpha-1</scm-api.version>
-    <git.version>3.4.0-alpha-1</git.version>
+    <git.version>3.4.0-alpha-4</git.version>
   </properties>
 
   <scm>
@@ -330,7 +330,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-branch-source</artifactId>
-            <version>2.2.0-alpha-3</version>
+            <version>2.2.0-alpha-4</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -330,7 +330,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-branch-source</artifactId>
-            <version>2.2.0-alpha-1</version>
+            <version>2.2.0-alpha-2</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -330,7 +330,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-branch-source</artifactId>
-            <version>2.2.0-alpha-2</version>
+            <version>2.2.0-alpha-3</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -32,8 +32,8 @@
     <jacoco.missedclass.coverage>0.00</jacoco.missedclass.coverage>
     <hpi.dependencyResolution>runtime</hpi.dependencyResolution>
     <jenkins-test-harness.version>2.15</jenkins-test-harness.version>
-    <scm-api.version>2.1.1</scm-api.version>
-    <git.version>3.3.0</git.version>
+    <scm-api.version>2.2.0-alpha-1</scm-api.version>
+    <git.version>3.4.0-alpha-1</git.version>
   </properties>
 
   <scm>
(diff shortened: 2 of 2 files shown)


stephenc:

michaelneale:
thanks @stephenc - I thought this would fail to build due to `@Restricted` though?

michaelneale:
there are some test failures - which look legit. I will run it again just to be sure - and then look locally (as some are github/ATH related, which is a fairly "deep" test).

michaelneale:
@stephenc branch-api 2.0.11-alpha-1 isn't in use here - it is still the prod version. Is this due to `@Restricted`? Would be good to change that to deprecated, then we would know for sure how kosher this is.

michaelneale:
yes failing tests are failing: https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/PR-1186/2/tests 

although without branch-api - not sure how legit the failures are - but this is core things not working.

michaelneale:
tested locally and it breaks hard with github creation with org folders via blue ocean, ie any org folder for github that blue ocean created. 

* Creating a pipeline never triggers the scan, so never discovers
* Saving from editor (if you open an empty pipeline) never triggers 
* Running a scan by hand deleted what projects were already there from a previous version
* When creating for the first time the scan log screen seems hung

I tested by updating the branch-api by hand as well to make sure it was all consistent. 

The main thing blue ocean org folders do that is probably not tested is user scoped creds. If you see the comment above there is a stack trace when you open the config screen to do with credentials. I wonder if that hints the way? 

The Good news is that test coverage picked this up.

michaelneale:
@stephenc ah ok - confused -  thought you said it would break the build once the depdencies were bumped due to `@Restricted` - but seems to almost be working here.

stephenc:
So the test failures should all be resulting from these [6 missing lines](https://github.com/jenkinsci/github-branch-source-plugin/pull/141/commits/394d016d15970d1abf9f085f63dea741137ad4bc#diff-d083863769fe25af396167d27e61e95aR232) in the legacy constructor.

Great that this was caught, but that does point to an issue with how Blue Ocean updates the navigation folder: [Blue Ocean](https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java#L170) blows away the existing source with every change.

The result of that is _if anyone configures traits_ then any changes made by blue ocean will blow those traits away...

So if they want to control Git LFS support, or branch discovery, or ref specs, or submodule options, or clean behaviours... then they will have to stop using the BlueOcean UI.

I will need to adapt #1187 and #1188 to avoid that issue

stephenc:
@michaelneale 
> thought you said it would break the build once the depdencies were bumped due to @Restricted - but seems to almost be working here.

The build _should_ be broken here, but your build is not _enforcing_ the `@Restricted` annotations... which is bad in general - but good here as it means we can test the upgrade

stephenc:
@michaelneale do you have the JIRA ref for the issue where BlueOcean tramples over the branches to build and PR merge vs head options set via the classic UI. I'd like to tag the commits on the other PRs that fix that as a side-effect so that you can track issues correctly

michaelneale:
@stephenc no - I wasn't aware that was a bug with the current version.

michaelneale:
OK ignore that failure (it is due to alpha version cancelling caching) - that is a green build! I will hand verify this shortly.

michaelneale:
this is looking a lot better @stephenc now. No projects are lost. Only remaining thing is the "warning" stacktrace when I go to open the config screen (this may be problematic if the user goes to save it?), but otherwise, looking pretty compatible. I didn't see any rebuilds or anything.

michaelneale:
@stephenc do you mean with the current version it also replaces the branch settings etc (ie the wall of checkboxes) each time it saves?

stephenc:
@michaelneale 
> do you mean with the current version it also replaces the branch settings etc (ie the wall of checkboxes) each time it saves?

yes. right now, without updating the plugins, every time you change an org folder with Blue Ocean it resets the checkboxes to:

screen shot 2017-06-22 at 09 32 48

Irrespective of any configuration options that you set with the classic UI.

Have you a JIRA for that issue?

stephenc:
Oh more fun... you also reset these two:

screen shot 2017-06-22 at 09 34 48

michaelneale:
@stephenc a fresh one: https://issues.jenkins-ci.org/browse/JENKINS-45058 - yes I expect it would change all those. I think I captured it on that ticket correctly? (should it be fixed in this PR? one of the others? Ultimately org folders won't be used by github "real soon now" as none of these settings make sense to set for all projects at once).

stephenc:
@michaelneale so I am not fixing [JENKINS-45058](https://issues.jenkins-ci.org/browse/JENKINS-45058) on this PR as you basically require to use the traits API to fix that issue... (otherwise you are chasing an arms-race with getters and setters)

w.r.t. the configuration options, with traits you could define a new trait that tweaked the discovery options for specific repositories only... *i'm not saying that is what we should do* rather it is a possibility enabled by the traits API... personally I think that would be really confusing to users so I would not recommend it and would not merge it into the core plugins... but it could live in an extension plugin quite happily

michaelneale:
nice one @stephenc - I will do a final sniff check today. Thanks for this!

michaelneale:
ran this again today - still looking good. @stephenc could we do a similar against the 1.1.0-release branch to be doubly sure? (it shouldn't really matter though.. but might be worth it)

michaelneale:
is the warning: 

```
WARNING: Caught exception evaluating: descriptor.calcFillSettings(field,attrs) in /jenkins/job/multibranchorg/configure. Reason: java.lang.IllegalStateException: class org.jenkinsci.plugins.github_branch_source.GitHubSCMSource$DescriptorImpl doesn't have the doFillCredentialsIdItems method for filling a drop-down list
java.lang.IllegalStateException: class org.jenkinsci.plugins.github_branch_source.GitHubSCMSource$DescriptorImpl doesn't have the doFillCredentialsIdItems method for filling a drop-down list
	at hudson.model.Descriptor.calcFillSettings(Descriptor.java:410)
```

anything to worry about? doesn't seem to harm. It is when opening the github org config.

stephenc:
@michaelneale I'll take a peek... may just be a method name that is mis-spelt... oh how I wish for a way to catch those issues at compile time... hmmm

michaelneale:
ok - well 1.1 looks good as well - so binary compat is looking 👍🏻 to me. The warning notwithstanding (doesn't seem to do ill..)

stephenc:
> is the warning:
>
> ```
> WARNING: Caught exception evaluating: descriptor.calcFillSettings(field,attrs) in /jenkins/job/multibranchorg/configure. Reason: java.lang.IllegalStateException: class org.jenkinsci.plugins.github_branch_source.GitHubSCMSource$DescriptorImpl doesn't have the doFillCredentialsIdItems method for filling a drop-down list
> java.lang.IllegalStateException: class org.jenkinsci.plugins.github_branch_source.GitHubSCMSource$DescriptorImpl doesn't have the doFillCredentialsIdItems method for filling a drop-down list
> 	at hudson.model.Descriptor.calcFillSettings(Descriptor.java:410)
> ```
> anything to worry about? doesn't seem to harm. It is when opening the github org config.

Good catch, that's fixed by https://github.com/jenkinsci/github-branch-source-plugin/pull/141/commits/8c5447bd5014844dd27943a7b76fd8bc24a7aeb2 

I'm holding off spinning an -alpha-4 in case there is anything serious

michaelneale:
nice

michaelneale:
@stephenc are there any newer betas to try here?

stephenc:
Beta-1 coming soon

michaelneale:
ok retested as of beta1 manuall - and passes automated tests so - compatibility seems good to me.

michaelneale:
has served its purpose

PR #1187, opened 2017-06-21
[JENKINS-43507] Plugin version and simple API migration

stephenc:
- Only moving to equivalent methods, not taking advantage of the new API

# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


stephenc:

michaelneale:
nice - passes. Can this one be closed now that https://github.com/jenkinsci/blueocean-plugin/pull/1188 exists with more betterer stuff?

stephenc:
@michaelneale depends, if the org folder stuff is being migrated away from, this may be a better short-term... but it is more at risk of JENKINS-45058 as we are hijacking the regex, which can then be edited by the user whereas #1188 explicitly adds a trait to hold the management by BlueOcean

michaelneale:
@stephenc def being migrated away from, but I don't see a downside in #1188 - but will let @vivek decide when he has time which one is most appropriate. The most important is the binary compat - which seems good, so we have some choice here. 

Thanks.

michaelneale:
looking good thanks

michaelneale:
I don't think we want to do it this way right @vivek ?

PR #1188, opened 2017-06-21
[JENKINS-43507] Plugin version and API adoption

stephenc:
- Uses the new API to remove previous regex building and parsing hacks

# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


stephenc:

michaelneale:
@stephenc is this the place to change things so it doesn't wipe out the settings each time? (in the medium term github org folders is going away anyway, as it isn't right for configuration settings, so it may be somewhat academic anyway).

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubPipelineCreateRequest.java
```diff
@@ -141,18 +148,36 @@ public BluePipeline create(Reachable parent) throws IOException {
...
+                        if (t instanceof BlueOceanSCMSourceFilterTrait) {
+                            trait = (BlueOceanSCMSourceFilterTrait) t;
+                        } else if (t instanceof RegexSCMSourceFilterTrait) {
+                            regexTrait = (RegexSCMSourceFilterTrait) t;
+                        } else {
+                            traits.add(t);
                         }
                     }
+                    if (trait == null) {
+                        Matcher matcher = Pattern.compile("\\((.*?)\\\\b\\)\\?").matcher(
```
stephenc: this is migration of existing configuration

stephenc:
@michaelneale so what this PR does is it moves the selection of which repositories out of the user configurable regex trait and into a pure BlueOcean managed trait:

screen shot 2017-06-22 at 10 07 13

With #1187 we also fix JENKINS-45058 but just keep using the regex field (exposing risk that users modify the regex and then tweaks get lost there

stephenc:
Actually here are the defaults that BlueOcean would create with this PR:

screen shot 2017-06-22 at 10 09 55

stephenc:
Changes in dd003a7

screen shot 2017-06-22 at 10 14 31

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubOrganizationFolder.java
```diff
@@ -32,10 +34,12 @@ public boolean isScanAllRepos() {
             SCMNavigator scmNavigator = getFolder().getSCMNavigators().get(0);
             if(scmNavigator instanceof GitHubSCMNavigator){
                 GitHubSCMNavigator gitHubSCMNavigator = (GitHubSCMNavigator) scmNavigator;
-                return (StringUtils.isBlank(gitHubSCMNavigator.getIncludes()) || gitHubSCMNavigator.getIncludes().equals("*"))
-                        && StringUtils.isBlank(gitHubSCMNavigator.getExcludes())
-                        && (StringUtils.isBlank(gitHubSCMNavigator.getPattern())
-                        || gitHubSCMNavigator.getPattern().equals(".*"));
+                for (SCMTrait<?> t: gitHubSCMNavigator.getTraits()) {
+                    if (t.getDescriptor().getClass().getAnnotation(Selection.class) != null) {
```
stephenc: @vivek unclear what the intent of this getter is... are you checking if there are any restrictions at all in place (in which case this is correct)... or are you checking if blueocean is managing the repository selection (in which case we just want 

```
if (t instanceof BlueOceanSCMSourceFilterTrait) {
  return false;
}
```

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubOrganizationFolder.java
```diff
@@ -32,10 +34,12 @@ public boolean isScanAllRepos() {
...
+                        // approximate logic
```
stephenc: I say approximate as *technically* the user could add a useless `new WildcardSCMSourceFilterTrait("*","")` which would not be doing any filtering but we would be returning `false`

michaelneale:
@stephenc great - so we can keep this one open until they are published right, and then continue with this to adapt to it (and fix the issue with writing over traits in one fell swoop?)

stephenc:
@michaelneale well we need @vivek to answer the open question on the `isScanAllRepos()` method and what it is trying to determine.

The name suggests that it returns `false` if not all repos are scanned...

I suspect that what is actually wanted is to return `false` if the repos are being managed by BlueOcean

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubScmContentProvider.java
```diff
@@ -246,15 +246,15 @@ public ScmParamsFromItem(Item item) {
                     if (source.getApiUri() != null) {
                         apiUrl = source.getApiUri();
                     }
-                    credentialId = source.getScanCredentialsId();
+                    credentialId = source.getCredentialsId();
```
i386: No more scan vs checkout credentials? :)

blueocean-github-pipeline/src/test/java/io/jenkins/blueocean/blueocean_github_pipeline/BlueOceanSCMSourceFilterTraitTest.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.*;
+
+public class BlueOceanSCMSourceFilterTraitTest {
+    @ClassRule
+    public static JenkinsRule r = new JenkinsRule();
+
+    @Test
+    public void given__trait__when__decoratingContext__then__prefilterApplied() throws Exception {
```
i386: `givenTrainWhenDecoratingContextThenPrefilterApplied`

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/BlueOceanSCMSourceFilterTrait.java
```diff
@@ -0,0 +1,90 @@
...
+import jenkins.scm.api.trait.SCMNavigatorContext;
+import jenkins.scm.api.trait.SCMNavigatorTrait;
+import jenkins.scm.api.trait.SCMNavigatorTraitDescriptor;
+import jenkins.scm.api.trait.SCMSourcePrefilter;
+import jenkins.scm.impl.trait.Selection;
+import org.apache.commons.lang.StringUtils;
+import org.jenkinsci.plugins.github_branch_source.GitHubSCMNavigator;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+public class BlueOceanSCMSourceFilterTrait extends SCMNavigatorTrait {
```
i386: What is this?

vivek:
>I suspect that what is actually wanted is to return false if the repos are being managed by BlueOcean

Hmm that's not current behavior is though. At present it returns falls if user goes to classic and changed filter settings(include or exclude) or set a  pattern. And it returns true if user goes and sets wildcard in includes for example. 

How's it going to work, if blue ocean has set certain repo based pattern and user upgrades to this version, will user end up getting that selected repos he originally had in previous version? Remember this method would return false for his case. 

Similarly if user chose Scan All Repo, equivalent to .*, will that be honoured with this PR as well?

Sorry, didn't have chance to see all code changes yet, deep down in other stuff I am doing. But will try to look in to it sometime next week.

stephenc:
> At present it returns falls if user goes to classic and changed filter settings(include or exclude) or set a pattern.

@vivek Except JENKINS-45058 means that anything they do set will get trampled over... so effectively you are only returning `true` if BlueOcean has applied a regex.

The code as written in this PR will return `true` if anything is filtering the returned repositories or branches.

If you want `true` based on filtering just the repositories then we would want

```
if (t instanceof SCMNavigatorTrait && t.getDescriptor().getClass().getAnnotation(Selection.class) != null) {
```

If you want `true` based on BlueOcean managing the list of repositories then that would be
 
```
if (t instanceof BlueOceanSCMSourceFilterTrait) {
```

I do not think you want what I have in this PR, but what I have in this PR is technically the correct replacement of what your existing code is doing... but I think checking the `includes` / `excludes` is a bug in the existing code (a bug that thanks to JENKINS-45058 is not important)

I think you actually want 

```
if (t instanceof BlueOceanSCMSourceFilterTrait) {
```

If the user configures their own additional filters on top of BlueOcean using the classic UI, in a sense BlueOcean should not care... those additional filters will just reduce the discovery to the intersection of the criteria... what BlueOcean cares about is whether it is managing the list or whether it is "all" from the BlueOcean perspective... the additional filtering should not IMHO be a concern of BlueOcean.

WDYT?

vivek:
>If the user configures their own additional filters on top of BlueOcean using the classic UI, in a sense BlueOcean should not care... those additional filters will just reduce the discovery to the intersection of the criteria... what BlueOcean cares about is whether it is managing the list or whether it is "all" from the BlueOcean perspective

@stephenc Ok, so the intent of `isScanAllRepo()` is to return false if there is any filter in place, whether blueocean ones or user explicitly set the repo. Return true if repos had been 'auto discovered' or scan all scenario. UI uses this field to render pipeline creation flow differently. 

> ... the additional filtering should not IMHO be a concern of BlueOcean.

It's not but this flag purpose is more to correctly detect if all repos were scanned, thats `true` returned from `isScanAllRepos()`. 

To confirm, lets check with @cliffmeyers. Cliff, how does `isScanAllRepos` field is used by UI.  Here is [UI code](https://github.com/jenkinsci/blueocean-plugin/blob/9fd3ad72262c00b8a2a1c51f234cc4c891c75250/blueocean-dashboard/src/main/js/creation/github/GithubFlowManager.js#L72-L72) that uses this flag? What does UI do using this flag, hide 'Auto Discover button'?

michaelneale:
Looking good

vivek:
@cliffmeyers Need your input on purpose of `isScanAllRepos()`, as how its used by the UI in pipeline creation flow. If its too much info to grasp, may be we can discuss this in our meeting today.

cliffmeyers:
@vivek that flag indicates whether the user chose "auto discover pipelines" when creating the org folder. Originally we were just going to rely on "pipelineFolderNames" for this, but that only tells us what pipelines exist in the org folder, and they could have been created explicitly by the user (adding individual pipelines explicitly) or just via auto-discover.

This flag is important to the UI because we show some warning messages to user the when they have an org folder in "auto discover" mode and switch to "single repo" - we need to tell them that pipelines will no longer be automagically created.

stephenc:
@cliffmeyers @vivek sounds like this flag is really "isManagedByBlueOcean" in which case we are just checking for the BlueOcean filter trait being present.

vivek:
> sounds like this flag is really "isManagedByBlueOcean" in which case we are just checking for the BlueOcean filter trait being present.

@stephenc Not really. Auto-discovery (select all using .*) as well as list of repos as regex - both are managed by blueocean. That is, if `isScanRepo()` true or false return, both cases could mean it's managed by blueocean. cc: @cliffmeyers

cliffmeyers:
@vivek I agree. It's really just a boolean to indicate whether the wild-card regex was set.

michaelneale:
@vivek @cliffmeyers now it has been released - we can probably go ahead with this. Of course we want to move off github org folders, if so - is this still relevant?

michaelneale:
@vivek @stephenc this is currently in a :bug: state - it isn't scanning the git repo correctly when I create it - which the test is picking up. A new test. 

When I create by hand it isn't either - ssh or https. It is something to do with the change for creating git repos here I hope - otherwise the branch-api changes have broken creation of git repos.

stephenc:
@michaelneale fixed

michaelneale:
nice - thanks @stephenc!. 

Unfortunately there is also a new conflict since the refactoring for bitbucket went in. Perhaps @vivek is best positioned to resolve that?

stephenc:
Yes @vivek should probably resolve the conflicts

michaelneale:
thanks @stephenc for keeping this one up to date - has been most helpful. Passing all the tests is 👌🏽
Dogfood is running the latest and greatest BTW!

PR #1189, opened 2017-06-22
Story/jenkins 40855 creation GitHub enterprise

cliffmeyers:
# Description
- See [JENKINS-40855](https://issues.jenkins-ci.org/browse/JENKINS-40855).
- Remaining todos:
   - [x] ATH, single pass that creates a server and completes creation against repo with existing jenkinsfile 
   - [x] Ensure "create access token" link is pointing to correct URL on GHE server
   - [ ] Fix broken GitHub org avatars
   - [ ] Add generic error handling to ensure "error step" displays when unhandled exceptions occur

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


cliffmeyers:
cc @michaelneale as requested

blueocean-dashboard/src/main/js/creation/github/steps/GithubCredentialStep.jsx
```diff
@@ -45,18 +82,61 @@ export default class GithubCredentialsStep extends React.Component {
         return null;
     }
 
+    _getInstructions(stateId) {
```
cliffmeyers: I will probably refactor this method into a composed component regardless of whether I do a more extensive refactor on this step.

i386:
@cliffmeyers I have this a go today and it worked for repositories that have Jenkinsfiles! Nice 👍 However it did not work if my repo did not have a Jenkinsfile (errors on save in the editor). Did you end up getting access to our internal Github Enterprise? Should be pretty clear what is going on when you do.

You were absolutely right about the design of the GH enterprise needing some drop downs to manage what server you are on - typing things out every time in testing sucks :(

Here's the flow I propose:

*User is given a drop down list to pick or add a Github Enterprise server*
github enterprise step 1

*Adding a Github Enterprise server*
add server

*Ask for the Access key in a second step once we've selected the server*
github enterprise step 2

So not to bother Vivek so much I wrote the API to enumerate and create Github servers registrations at https://github.com/jenkinsci/blueocean-plugin/pull/1190. All we need to do is modify `GithubPipelineCreateRequest` so that it validates that the apiUri matches a server we have defined and everything should work (currently it creates it on the fly for the user and we should stop doing that).

What do you think?

michaelneale:
perhaps there is  some  hard coding of URI? could set up a proxy for api.github.com locally for that. 

Otherwise test saving it against github.com (hopefully that breaks then it would be easier to see).

i386:
@michaelneale maybe a quick python bottle we could check in and have proxy the public API endpoint so that it looks like https://developer.github.com/v3/enterprise/#endpoint-urls ?

i386:
The other thing that didn't work was avatars. it looks like you need to have a cookie set to load them. The result is the images are broken because the browser redirects to login. 

I'm going to chase that up with my friends at Github.

cliffmeyers:
@i386 we probably should validate the `apiUrl` parameter for all requests through `GithubEnterpriseScm` including `/validate`, `/organizations` in addition to `GithubPipelineCreateRequest`. Another option would be to assign a unique ID to each server object and actually pass `serverId` as the parameter instead of `apiUrl`. That might be nice as it enforces the semantics a bit.

blueocean-github-pipeline/README.md
```diff
@@ -0,0 +1,9 @@
+# blueocean-github-pipeline
```
i386: I've added a proxy server here you can run to similar GHE 🎉  cc @michaelneale

i386:
@cliffmeyers seems to fail just after I specify a Github Enterprise server https://cl.ly/0I0K0z0y103G

BTW, if you run the proxy you can simulate Github Enterprise!

One small thing:
Github Enterprise endpoint looks like `http://ghe.myserver.com/api/v3/`. To avoid the user typing `/api/v3/` every time could we just append it to the URL?

cliffmeyers:
@i386 I think this change will get you going again. Sorry about that!

i386:
@cliffmeyers 🐛 it looks like styles are not loading correctly 

_On Dogfood_
screencapture-ci-blueocean-io-blue-organizations-jenkins-create-pipeline-1498429319916
_On branch_
screencapture-localhost-8080-jenkins-blue-organizations-jenkins-create-pipeline-1498429316869

blueocean-github-pipeline/nginx.conf
```diff
@@ -0,0 +1,15 @@
...
+  worker_connections  1024;
+}
+http {
+  server {
+    listen 9000;
+    location /api/v3/ {
+      proxy_pass https://api.github.com/;
+      sub_filter_once off;
+      sub_filter_types application/json
+      sub_filter https://api.github.com/ http://localhost:9000/api/v3/;
```
i386: @michaelneale I had a crack at using the `sub_filter` to substitute data in the response body. Couldn't seem to get this to work. Have you had any experience with `proxy_pass` and `sub_filter`?

cliffmeyers:
@i386 styles regression fixed in c9afd30

i386:
@cliffmeyers thanks 👍

cliffmeyers:
Recent builds of this are actually passing but there seems to be another Jenkins build updating the build status (see my msg in Gitter room). Here's a link to the legit builds:
https://ci.blueocean.io/blue/organizations/jenkins/blueocean/activity?branch=story%2FJENKINS-40855-creation-github-enterprise

cliffmeyers:
@i386 could you have a look at this again and let me know how it looks? I have a basic smoke test for GHE creation, although naturally it points to the public github.com API endpoint. It does exercise the server creation dialog pretty well. Please also see my note in 35921c6 and let me know what you think of this change.

Until we have a reliable mechanism for deleting the GitHub servers in automated fashion, we need a little hack like that to allow multiple runs of that ATH test to pass - important for local development.

cliffmeyers:
Note that the server-side Java changes were reviewed in #1190 but that was only merged to this branch, not to master, so this PR appears larger than it really is. Not sure if the branch from #1190 should be merged to master separately to help slim this one down?  @i386

i386:
@cliffmeyers I had to return all the memory in my system so I am unable to test against my GHE server. 

I just merged #1190 to master but now there are conflicts here.

i386:
@cliffmeyers fixed conflicts for you.

michaelneale:
this is looking good - tested it on github.com - needs testing on GHE though. 
Also - master I think is broken since https://github.com/jenkinsci/blueocean-plugin/pull/1190 was merged (it had failing tests) - so not sure what state thigns are at.

i386:
I'll fix master when I get back

michaelneale:
looking good - master was a false alarm - flaky test (I have a WIP to fix it). 

@i386 @cliffmeyers had a chance to test with real GHE? 

One question: 

Is there any name spacing of the org folder with the server? I noticed if I had "multibranchorg" setup via regular github, and then went to set it up with GHE (pointing to api.github.com) - it tried to put things in the same place, although it shouldn't, as one is github, one is GHE. In this case the name was identical, but is that likely to be a problem? As it wouldn't be uncommon to have the same org+repo names in 2 different github endpoints right? - that is my only functional feedback.

i386:
@cliffmeyers @michaelneale I am not in a position to test until next week when my replacement memory arrives.

@cliffmeyers perhaps its time to setup your VPN?

cliffmeyers:
@michaelneale you're correct that if two servers have the same org name that it will be a problem as it stands today. Currently the UI does make a fetch call after the org is selected to see if it already exists... however I am not sure there's enough information in the object returned for us to determine whether it's an org folder from a different API URL or the current one. We could potentially do the following:
- Add a property to `GithubOrganizationFolder` which is the API URL for that SCM. 
- When the above fetch is made...
   - If the URL matches the server the user has selected, then we know we are simply updating the existing org folder.
   - If it doesn't match, then we could throw up a "rename" step to allow them to specify a name. I believe this kind of approach is planned for when we switch from creating org folders to creating multibranch pipelines.

If the above sounds reasonable, do either of you know offhand if there's a good method on `OrganizationFolder` or `SCMSource` that returns the URL? I did a really quick scan and didn't see anything but can keep looking.

@i386 I got access to the VPN a couple days ago... I was just granted access to GHE overnight. I will be testing that this morning.

cliffmeyers:
@i386 @michaelneale did a quick test against GHE 2.9.2 and it worked fine:

(using org "dev1" with repo "stage-example-pipeline":
ghe-test-1

As James noted the org avatars are broken so I added that as a TODO item to this PR so we don't forget.

michaelneale:
@cliffmeyers that is great. No I don't know about getting the URL back from a org folder - perhaps @vivek knows? 

As for the name uniqueness - this is a bit tricky while we are in between org folders and MBP, so what can we do now to get this to a state of merge-able ness? 

Short of adding a prefix to the name or path that includes the url - not sure what we can do. If blue ocean is going to flatten things out between orgs or pipeline names - then we always need to handle renaming or name spacing. This is getting into a design area that @i386 will need to think about with brody (ie do we put server/org name scope in pipeline somehow?). 

@cliffmeyers so on this issue - I would flag it as a follow up ticket, which can be tackled after getting off org folders, so we can work on closing this one down and getting it out as a beta. If you agree - do you mind making a ticket? (I would make one but seems premature until we agree to punt that issue down the road).

i386:
@cliffmeyers @michaelneale @vivek for name uniqueness, handle it in the exact same way that we handle uniqueness of Git based pipelines (we detect the duplicate and ask for an alternative name).

If that becomes hard to do (because org folders) then create another ticket to handle this case and we can schedule it as part of org folder removal.

michaelneale:
@i386 we can't do that with github org folders in the mix, as the name of the pipeline is not controllable - so I think make it as a follow on ticket IMO

i386:
@cliffmeyers would you mind ticketing this up as you understand it? Ill find a place for it to go.

cliffmeyers:
@michaelneale good point, forgot we didn't have control over the org folder name.

@i386 I'll write up a ticket to describe this issue, and the other problem with avatars we discussed.

cliffmeyers:
Duplicate org name problem:
https://issues.jenkins-ci.org/browse/JENKINS-45240

Avatars not loading:
https://issues.jenkins-ci.org/browse/JENKINS-45238

re: avatars, I am implementing a fix now in this PR so that we use a standard placeholder image for GHE.

@i386 re: improved error handling, this probably requires some non-trivial changes and this PR is already becoming somewhat long-running. Would it be okay to split that off as another ticket as well so we can land this? I can try to squeeze it in next week.

michaelneale:
@cliffmeyers ok - in this case - I think this is :bee: from me. Let's get it in!

The follow on work about naming can wait for the github org folder un-doing. Would be good to get this out in beta.

vivek:
@cliffmeyers @michaelneale 👍 Will be great to get it in as some of the improvement done in this PR will come handy for bitbucket work.

PR #1193, opened 2017-06-22
sign images we push to dockerhub

ndeloof:


michaelneale:
review please @ydubreuil - you ok with this? (not sure what this does, but I am sure it is ok...)

ndeloof:
see https://docs.docker.com/engine/security/trust/content_trust/
signed image might ben required for some deployment (typically : linuxkit does)

michaelneale:
@ndeloof so for the image to be signed, the publishing needs to change. that is done by behind the firewall stuff @rtyler runs at jenkins.io - so he would need to be on board with this change. I assume in the meantime - this is a noop until publishing changes right?

ndeloof:
Yes indeed, we might need to apply 
https://docs.docker.com/engine/security/trust/trust_automation/
to get automated (non interactive) signing for images we publish

i.e have `DOCKER_CONTENT_TRUST_ROOT_PASSPHRASE` and `DOCKER_CONTENT_TRUST_REPOSITORY_PASSPHRASE` set on build environment

ydubreuil:
It's fine to me, it'll be great to have signed images. Let's wait for @rtyler

michaelneale:
closing this for now, until jenkins.io has this infra in place (which I don't think it does yet)

ndeloof:
@michaelneale what makes you think we don't have infra ? Signing images is just a question of using your dockerhub ID and set and env variable for docker push.

michaelneale:
@ndeloof I have no visibility into the infra that publishes this docker image, so I can't say if it is setup for that or not? (can reopen it if/when it is)

michaelneale:
@ndeloof oh wait - you mean literally just setting that env var? is there any way this would fail the push? (sorry I misunderstood)

PR #1194, opened 2017-06-22
Publish under "jenkins" dockerhub organization

ndeloof:
see https://github.com/jenkinsci/docker/issues/515

### DIFF
--- a/docker/official/build.sh
+++ b/docker/official/build.sh
@@ -35,6 +35,6 @@ docker tag "jenkinsci/blueocean:$BLUEOCEAN_VERSION" jenkinsci/blueocean:latest
 docker tag "jenkinsci/blueocean:$BLUEOCEAN_VERSION" "jenkinsci/blueocean:$FULL_VERSION"
 
 # push it real good
-docker push "jenkinsci/blueocean:$BLUEOCEAN_VERSION"
-docker push "jenkinsci/blueocean:$FULL_VERSION"
-docker push jenkinsci/blueocean:latest
+docker push "jenkins/blueocean:$BLUEOCEAN_VERSION"
+docker push "jenkins/blueocean:$FULL_VERSION"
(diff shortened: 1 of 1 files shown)


michaelneale:
need a review from @rtyler for this - as shouldn't it publish to both?

ndeloof:
I'd prefer we avoid duplication with two official names for our images organization.
jenkinsci could just be deprecated (like jenkins official image)

michaelneale:
@ndeloof but as mentioned before - there is no real deprecation. People will still expect it to be there. Jenkinsci is the org that I know about so this is -1 from me until @rtyler agrees. Not our call.

ndeloof:
I'm asking docker if they can configure redirects on dockerhub (like github does)
in the meantime we could indeed have dual deployment, at least for a transition period

michaelneale:
thanks @ndeloof !

ndeloof:
@rtyler added (expected to be temporary) dual publishing

michaelneale:
ok, will wait for @rtyler to return from vacay to final bless - but :bee: from me

i386:
I'd also like to see the README.md updated to reflect the documentation in the Jenkins base image. This can be as simple as a link to that container on Docker Hub.

michaelneale:
hey @rtyler still interested?

rtyler:
@michaelneale I'm not not interesting, but this is not something that I can spend any time on.


As far as the Blue Ocean container is concerned, this is fine. If the desire is that all containers move over to the `jenkins` organization, then somebody should champion this and bring it up as a topic in a Governance Meetings or on the dev list.

michaelneale:
can open this back up later when we get time. In time a specific blue ocean image may not permanently be a thing (once blue ocean gets into recommended plugins, for instance). 

Just clearing this out as the number of PRs left open is exploding.

PR #1195, opened 2017-06-22
Bitbucket Cloud and Server

vivek:
# Description

See [JENKINS-44666](https://issues.jenkins-ci.org/browse/JENKINS-44666).

Backend implementation of Bitbucket server complete with all tests. Next tasks:

- [x] Backend implementation for Bitbucket cloud 
- [x] Endpoint implementation for BitBucket
- [x] Add bitbucket cloud user's account in the list of orgs
- [x] Add URL normalization logic and add tests with/without slashes
- [x] Front end development for bitbucket server 
- [x] Adopt content creation on empty repo using BB server 5.2.0-m2
- [x] Enforce version check on bitbucket server, this would be `5.2.0-m2` now. When the fixes get rolled in to later version, we change this value accordingly thru separate PR.
- [ ] Update and test bitbucket pipeline editing with pipeline-editor
version 'content creation on empty repo' feature lands on
- [ ] Update permission repo with `blueocean-bitbucket-pipeline` module

I plan to implement bitbucket cloud first, because its relatively small task, matter of implementing [BitbucketApi](https://github.com/jenkinsci/blueocean-plugin/blob/39cab5430e863e38e4f189a947653ea5ab1d7c3a/blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/BitbucketApi.java#L23-L23). This will also solidify abstractions that encapsulates server vs cloud differences.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


i386:
Wow @vivek 👍

cliffmeyers:
Would it be possible to get a summary of the REST API in a README?

michaelneale:
nice one @vivek ! 

Don't for get the findbugs violations (which is why build failed I think) - also we need to remember to update the permissions repo with this new plugin.

vivek:
@michaelneale yes, its still a WIP, adding cloud specific APIs in this same PR, will get to findbugs eventually. code coverage is at 76% with bitbucket server, will add tests for cloud one as I make progress.

michaelneale:
@vivek 75% is amazing for this! well done. Looking fantastic 👌🏽

vivek:
@cliffmeyers there are slight variations in config section of credential and create API. Will add a README to bitbucket module with what APIs it exposes.

i386:
@vivek @stephenc it appears Bitbucket does not have the concept of stored "endpoints" which means the user must enter their server URL over and over.

Unless we can build stored endpoints (like Github) we cannot reuse the UI we are building for Github enterprise as described in https://github.com/jenkinsci/blueocean-plugin/pull/1189#issuecomment-310311239 

*Enter the Server URL*
bitbucket server step 1

*Accepting credentials*
bitbucket server step 2

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketPipelineCreateRequest.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+public class BitbucketPipelineCreateRequest extends AbstractMultiBranchCreateRequest {
+    @DataBoundConstructor
+    public BitbucketPipelineCreateRequest(String name, BlueScmConfig scmConfig) {
+        super(name, scmConfig);
+    }
+
+    @Override
+    protected SCMSource createSource(@Nonnull MultiBranchProject project, @Nonnull BlueScmConfig scmConfig) {
+        BitbucketSCMSource bitbucketSCMSource =  new BitbucketSCMSource(null, (String)scmConfig.getConfig().get("repoOwner"), (String)scmConfig.getConfig().get("repository"));
```
stephenc: Needs reworking for JENKINS-43507 see #1188 (which will give you a list of endpoints too)

i386:
Confirmed that endpoints are in #1188 - just lovely @stephenc 👍 

configure system jenkins 2017-06-23 16-33-42

i386:
@vivek just chatting with @michaelneale - giving you the A-OK to depend on #1188 so that we can make sure the Bitbucket server and Github Enterprise experience is consistent (we want endpoints).

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerApi.java
```diff
@@ -0,0 +1,340 @@
...
+        }
+    }
+
+
+    @Override
+    public @Nonnull
+    BbPage<BbProject> getProjects(int start, int limit){
+        try {
+            InputStream inputStream = Request.Get(String.format("%s?start=%s&limit=%s",baseUrl+"projects/", start,limit))
+                    .addHeader("Authorization", basicAuthHeaderValue)
```
i386: @vivek actually this one should really be fixed.

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/BitbucketApi.java
```diff
@@ -0,0 +1,63 @@
...
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Map;
+
+/**
+ * @author Vivek Pandey
+ */
+@Restricted(NoExternalUse.class)
+public abstract class BitbucketApi {
```
vivek: It's strictly internal use with only two internal implementation already in place, but I will add some docs.

michaelneale:
looking good - there are errors with github org folder tests I can see in the log

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketPipelineCreateRequest.java
vivek: Thanks @stephenc.

vivek:
@i386 bitbucket server endpoint configuration also includes Manage hooks. Do we want to surface it in blueocean UI. If not then what should we default to, `true`? We cad always attached user's bitbucket server credentials, however in classic it won't be visible as bitbucket server endpoint config UI only lists system credentials. cc: @stephenc  

If we surface 'manage hooks' in blueocean ui, then we can fix it as we can list user (only bitbucket server credentials) as well as system credentials.

image

stephenc:
Whatever about a user adding a server (which should be restricted to admins IMHO) you should not be adding a server with manage hooks. The manage hook credentials need to be in the global store (and depending on the admins security concerns may need to be system scope). The blue ocean credentials store will not be accessible in that context.

If you want to have the hooks managed then add the hook management trait to items you create with ITEM mode.

vivek:
@cliffmeyers https://github.com/jenkinsci/blueocean-plugin/pull/1195/commits/75c4e23ecdef2efbe558a1426ce49cdc44230bf4 has first cut on bitbucket front end. I can create bitbucket cloud pipeline with limited success:) Still lots to do.

Can you scan and let me know your general feedback, am I going in right direction? 

Here is what I think needs to be done for cloud flow (haven't got chance to test server yet):

- Fix bugs related to state management
  - Disabled state handling
  - Completion success state handling to switch over to dashboard (sse event related work)
  - creation failure state handling
  - Cleanup states borrowed from github/github-enterprise
- Test for edge cases
- Test and fix bugs for bitbucket server

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/AbstractBitbucketScm.java
```diff
@@ -0,0 +1,276 @@
...
+        return request;
+    }
+
+    protected String getApiUrlParameter(){
+        return getApiUrlParameter(getStaplerRequest());
+    }
+
+    private String getApiUrlParameter(StaplerRequest request){
+        return request.getParameter("apiUrl");
+    }
```
cliffmeyers: In GHE we had some annoyances with the `apiUrl` containing the trailing slash, or not, which might bypass duplicate detection inappropriately. I think there is some logic downstream to detect this elsewhere in the code, although it might have a bug in it.  Would it be better to add a method here like `String getCanonicalApiUrlParameterValue()` which returns the `apiUrl` with trailing slash always added? Then all downstream logic wouldn't have to worry about it. To be extra safe, we could add the same logic in endpoint handling code in case someone adds an endpoint manually via classic?

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerEndpointContainer.java
```diff
@@ -0,0 +1,120 @@
...
+                String displayName = endpoint.getDisplayName();
+                if (StringUtils.isNotBlank(displayName) && displayName.equals(name)) {
+                    errors.add(new ErrorMessage.Error(ScmServerEndpoint.NAME, ErrorMessage.Error.ErrorCodes.ALREADY_EXISTS.toString(), ScmServerEndpoint.NAME + " already exists"));
+                }
+            }
+            if(StringUtils.isNotBlank(url)) {
+                if (url.startsWith(endpoint.getServerUrl())) {
+                    errors.add(new ErrorMessage.Error(ScmServerEndpoint.API_URL, ErrorMessage.Error.ErrorCodes.ALREADY_EXISTS.toString(), ScmServerEndpoint.API_URL + " already exists"));
+                }
+            }
```
cliffmeyers: It looks like this code would thrown an error if the user tries to add `https://a.com/api/foo` and server `https://a.com/api/` already exists (good). However if user tried to add `https://b.com/api` and `https://b.com/api/` already exists, no error would be thrown (bad?). Please see my other comment regarding URL handling.

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerScm.java
```diff
@@ -0,0 +1,72 @@
...
+        return new BitbucketServerEndpointContainer(this);
+    }
+
+    @Override
+    protected  @Nonnull String createCredentialId(@Nonnull String apiUrl){
+        String url = apiUrl;
+        if(url.charAt(url.length()-1) != '/'){ //ensure trailing slash
+            url = url + "/";
+        }
+        return String.format("%s:%s",getId(),DigestUtils.sha256Hex(url));
```
cliffmeyers: See my other comment about trailing slash logic.

blueocean-bitbucket-pipeline/src/test/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerEndpointTest.java
```diff
@@ -0,0 +1,265 @@
...
+                .post(URL)
+                .build(Map.class);
+
+        List errors = (List) resp.get("errors");
+        assertEquals(1, errors.size());
+
+        Map error1 = (Map) errors.get(0);
+        assertEquals("apiUrl", error1.get("field"));
+        assertEquals("ALREADY_EXISTS", error1.get("code"));
+        assertNotNull(error1.get("message"));
```
cliffmeyers: Might be good to add a test here with a trailing slash mismatch, just because it's very possibly two users may try to add the same server and omit a trailing slash in one but not the other.

blueocean-dashboard/coverage/clover.xml
```diff
@@ -0,0 +1,4452 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cliffmeyers: These files were checked in accidentally. You'll want to delete everything in `blueocean-dashboard` `coverage` and `reports` if they exist. If you merge latest master and re-run build, these files will be instead placed in `./target/jest-coverage` and `./target-jest-reports` respectively which will already be ignored.

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/cloud/BitbucketCloudApi.java
```diff
@@ -0,0 +1,257 @@
...
+                pageNumber = 1;
+            }
+
+            if(pageSize <=0){
+                pageSize = PagedResponse.DEFAULT_LIMIT;
+            }
+            InputStream inputStream = Request.Get(String.format("%s&page=%s&pagelen=%s",baseUrl+"teams/?role=contributor", pageNumber,pageSize))
+                    .addHeader("Authorization", basicAuthHeaderValue)
+                    .execute().returnContent().asStream();
+            return om.readValue(inputStream, new TypeReference<BbCloudPage<BbCloudTeam>>(){});
```
vivek: Yes, you are right. I need to add it there. I have added it to TODO in this PR.

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/AbstractBitbucketScm.java
vivek: I think I am going to write a normalized URL conversion util to be consistent so that we can do thisUrl.equals(thatUrl) comparison instead of startWith, which is a hack.

Endpoint is going to get some changes, what we have is broken.

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerEndpointContainer.java
vivek: Ack.

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerScm.java
vivek: Ack.

blueocean-bitbucket-pipeline/src/test/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerEndpointTest.java
vivek: ack

blueocean-dashboard/coverage/clover.xml
vivek: argh, let me fix that.

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerEndpointContainer.java
stephenc: Please just use the BitbucketEndpointConfiguration.normalizeUrl to fix urls and compare post normalisation

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/AbstractBitbucketScm.java
stephenc: Please don't reinvent the wheel. There should be URL normalisation methods in the endpoint configuration classes (I'll add to GitHub if missing there)

vivek:
@cliffmeyers @i386 I have **bitbucket cloud** working. There are some UI kinks  and there is still  
 to take care of no-jenkinsfile-in-repo case with editor launch, much like github. There are other bugs identified but with this there is more confidence of getting there. And  adding bb server should be pretty easy as it has extra step of listing server (like GHE, builds on top of cloud version).

out

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerEndpointContainer.java
vivek: @stephenc excellent 👍

cliffmeyers:
@vivek nice! I agree that once Bitbucket Cloud is working that Server shouldn't be a ton more code... it's the same kind of additions as I did for GHE which was pretty modest in the end.

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerEndpointContainer.java
vivek: @stephenc  `BitbucketEndpointConfiguration.normalizeServerUrl(apiUrl)`  returns bitbucket cloud URL if apiUrl is null. Why not throw exception? Otherwise it leaves me with doing the blank/null check myself, not too bad but returning bitbucket cloud url is unexpected and stops it being from generic URL normalization utility.

stephenc:
Legacy config used null as a marker for cloud

vivek:
@i386 @michaelneale BB server creation front end work is in place. There are two differences from mock ups:
- Server drop down sows only server URLs (we went over it).
- In add server dialog, I am only asking for server URL. Do not see point of asking for name that doesn't appear in the server drop down list.

Let me know if there are any changes needed. 

bbserver

vivek:
@i386 I was thinking, we could collapse 'BitBucker Cloud' and 'BitBucket Server' in to one button by listing BB cloud api URL in the list of servers. Maybe not obvious but the button space looks bit crowded. wdyt?

i386:
@vivek this is so freaking awesome dude! I'm going to give this a go tonight and send you some review feedback. Nice job 🙌

blueocean-dashboard/src/main/js/creation/bitbucket/steps/BbCredentialStep.jsx
```diff
@@ -0,0 +1,139 @@
...
+        const valid = this._performValidation();
+        if (!valid) {
+            return;
+        }
+        this.props.flowManager.createCredential(this.state.usernameValue, this.state.passwordValue);
+    }
+
+    _getErrorMessage(stateId) {
+        if (stateId === BbCredentialState.INVALID_CREDENTIAL) {
+            return 'Invalid userName and secret';
```
i386: "Invalid username or password"

i386:
@vivek this is really great - thank you! 

Few things:
* According to Atlassian's brand guidelines the name is "Bitbucket" rather than "BitBucket". Can we ensure anywhere where we display the name it is spelled correctly?
* You are using the Github logo instead of the Bitbucket logo. I have asked @brody to follow up with an appropriate SVG.
* Bitbucket doesn't have "Organizations". We should use the specific terminology available for each kind of Bitbucket here instead.
** Cloud has "Teams"
** Server has "Projects
jenkins blue ocean 2017-07-10 18-42-36

blueocean-dashboard/src/main/js/creation/bitbucket/server/BbAddServerDialog.jsx
```diff
@@ -0,0 +1,155 @@
+import React, { PropTypes } from 'react';
+import { observer } from 'mobx-react';
+import debounce from 'lodash.debounce';
+import { Dialog, FormElement, TextInput } from '@jenkins-cd/design-language';
+
+import ServerErrorRenderer from './ServerErrorRenderer';
+
+
+let t = null;
```
i386: What is `t`?

blueocean-dashboard/src/main/js/creation/bitbucket/server/steps/BbChooseServerStep.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+                { this.state.showAddServerDialog &&
+                <BbAddServerDialog
+                    flowManager={flowManager}
+                    onClose={cred => this._onAddServerDialogClosed(cred)}
+                />
+                }
+
+                <button className="button-next-step" disabled={disabledNext} onClick={() => this._onClickNextButton()}>
+                    {t('creation.githubent.choose_server.button_next')}
```
i386: Consider refactoring these i18n strings so that they were a bit more generic? There might be some overlap with the Github ones elsewhere, too.

blueocean-dashboard/src/main/js/creation/bitbucket/steps/BbRepositoryStep.jsx
```diff
@@ -0,0 +1,90 @@
...
+                    >
+                        Create Pipeline
+                    </button>
+                </div>
+                }
+
+                { flowManager.repositories.length === 0 &&
+                <div className="container">
+                    <p className="instructions">
+                        The organization "{orgName}" has no repositories.
```
i386: i18n

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubServerContainer.java
```diff
@@ -9,7 +9,8 @@
 import io.jenkins.blueocean.commons.ServiceException;
```
i386: We should switch all this over to use the URL as the index. Perhaps on a follow up PR?

pom.xml
```diff
@@ -399,6 +405,23 @@
             <version>1.27.0</version>
         </dependency>
 
+        <!-- BitBucket dependencies -->
```
i386: "Bitbucket"

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubServerContainer.java
vivek: @i386 we do not have use case to look for endpoint by URL or name, so not really needed. All thats needed for us is to list all available server endpoints.

blueocean-dashboard/src/main/js/creation/bitbucket/server/BbAddServerDialog.jsx
i386: I think this is a missing import?

vivek: good catch, missing BbServerFlowManager import.

blueocean-dashboard/src/main/js/creation/bitbucket/server/steps/BbChooseServerStep.jsx
vivek: yeah, I am using what was there with GHE, was avoiding making changes all across. Maybe do it in separate PR? Avoiding stepping on too many things:)

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/BitbucketPipelineCreateRequest.java
```diff
@@ -0,0 +1,90 @@
...
+public class BitbucketPipelineCreateRequest extends AbstractMultiBranchCreateRequest {
+    @DataBoundConstructor
+    public BitbucketPipelineCreateRequest(String name, BlueScmConfig scmConfig) {
+        super(name, scmConfig);
+    }
+
+    @Override
+    protected SCMSource createSource(@Nonnull MultiBranchProject project, @Nonnull BlueScmConfig scmConfig) {
+        /* scmConfig.uri presence is already validated in {@link #validate} but lets check just in case */
+        Preconditions.checkNotNull(scmConfig.getUri(), "scmConfig.uri must be present");
```
vivek: Ack.

blueocean-dashboard/src/main/js/creation/bitbucket/server/steps/BbChooseServerStep.jsx
i386: Separate PR is fine. Maybe file another ticket to follow up?

vivek:
> The username and password fields should be full width
ok

> According to Atlassian's brand guidelines the name is "Bitbucket" rather than "BitBucket". Can we ensure anywhere where we display the name it is spelled correctly?

Ok, I will make sure its spelled with right case.

> You are using the Github logo instead of the Bitbucket logo. I have asked @brody to follow up with an appropriate SVG.

Ah, right. I wanted to get back on this and missed. @brody let me know when you have one and I can use that instead.

Bitbucket doesn't have "Organizations". We should use the specific terminology available for each kind of Bitbucket here instead.
** Cloud has "Teams"
** Server has "Projects

Yeah, sounds good.

blueocean-dashboard/src/main/js/creation/bitbucket/server/steps/BbChooseServerStep.jsx
vivek: Yeah, that should be fine. I am going to address these i18n comments, I will see if I can do it myself otherwise separate ticket for doing this is fine as well.

blueocean-dashboard/src/main/js/creation/bitbucket/server/BbAddServerDialog.jsx
vivek: @i386 t is set to [props.flowManager.translate](https://github.com/jenkinsci/blueocean-plugin/blob/e495e01c5e0cba2ce61987174eb931be987202e0/blueocean-dashboard/src/main/js/creation/bitbucket/server/BbAddServerDialog.jsx#L25-L25), so it should be fine. No import needed.

vivek:
@i386 Now I display server URL for the name selected in dropdown box. Maybe some styling needed but looks ok for now. As discussed, if name is null/empty then server URL will appear in the drop down.

image

i386:
@brody so we wanted to disambiguate which Github server they had picked in the dropdown. Idea here was to add some subtext to the dropdown element to show the URL represented by the name (its possible to have the same URL for different names in the dropdown AND there was a security concern that you could setup a fake Bitbucket/Github server to capture credentials). What is appropriate to do here?

stephenc:
@i386 correction:

> (its possible to have the same URL for different names in the dropdown AND there was a security concern that you could setup a fake Bitbucket/Github server to capture credentials). What is appropriate to do here?

No it is the other way around, it is possible to have the same name for different URLs. The URLs will always be unique as we guarantee a dedup on save, but we cannot guarantee a dedup based on names

i386:
>  I was thinking, we could collapse 'BitBucker Cloud' and 'BitBucket Server' in to one button by listing BB cloud api URL in the list of servers. Maybe not obvious but the button space looks bit crowded. wdyt?

@vivek lets leave it for the moment. We may update the design later to reduce the amount of space these buttons take up.

blueocean-dashboard/src/main/js/creation/bitbucket/server/BbAddServerDialog.jsx
i386: @vivek seems weird to put it in the module scope. Better to assign as an instance on `BBAddServerDialog`?

brody:
@vivek You can grab the Bitbucket logo from this Zeplin https://zpl.io/1PPGsB

@i386 I'd need more context around real world data. If possible, it would be ideal to have the URL next to the name, but it all depends on if that would work. Otherwise, we may need to represent the text in a different way below the dropdown as with the screenshot above.

vivek:
@brody Can't access it, says "you are not invited to this project". Maybe send it thru email?

i386:
@brody the URLs in this case could be rather long (in corp environments they can very long). I don't think you would have enough space to fit in the name and URL next to each other.

brody:
@i386 The only problem I can see (which mightn't be a problem) is that you don't know the URL until after selecting the server. Is this an issue? If it isn't then the placement is fine, we could tweak the way its displayed so it doesn't look like the normal labels.

If it is, we could try another way of including is in the dropdown.
image

i386:
@brody right on. @vivek, I think we should clarify this part after this PR lands? I need to go through here and scan for things that need to be converted into JIRA issues.

cliffmeyers:
@vivek a few comments not directly related to the code:
- I fixed a lint issue.
- When I run the build with `mvn -T 1C` or `mvn -T 0.5C` the Java tests seem to hang. In one instance many of them failed with "too many open files" errors. When I push Maven down to use a single thread the build passes but we might want to look into that at some point. These failures were after a system reboot, so pretty confident they were not due to any other zombie processes on my machine.
- There is a Jacoco failure in blueocean-bitbucket-pipeline right now. I checked the docs, it doesn't seem there is a way to dump the error details when it fails. Do you know of a way?

I'll move on now to a review of the UI code itself.

vivek:
@cliffmeyers Please do not check in those changes. The jacoco and lint issue fix is coming.

vivek:
@cliffmeyers 

>When I run the build with mvn -T 1C or mvn -T 0.5C the Java tests seem to hang. In one instance many of them failed with "too many open files" errors. When I push Maven down to use a single thread the build passes but we might want to look into that at some point. These failures were after a system reboot, so pretty confident they were not due to any other zombie processes on my machine.

I get that too. Side effect of wire mock and all these files loaded in parallel in different threads. More of issue on Mac than Linux.

vivek:
@cliffmeyers lint and Jacoco failures are fixed

blueocean-dashboard/src/main/js/creation/bitbucket/cloud/BbCloudFlowManager.js
```diff
@@ -0,0 +1,414 @@
...
+    _saveRepoSuccess(mbp) {
+        LOGGER.debug(`Multi-branch pipeline creation successfully: ${mbp.name}`);
+        this.changeState(STATE.STEP_COMPLETE_SUCCESS);
+        this.pipeline = mbp.name;
+    }
+
+    _saveRepoFailure() {
+        LOGGER.error('Multi-branch pipeline creation failed');
+        this.changeState(STATE.STEP_COMPLETE_SAVING_ERROR);
+    }
```
cliffmeyers: `_saveRepoSuccess` and `_saveRepoFailure` appear to be unused according to IDEA.

blueocean-dashboard/src/main/js/creation/bitbucket/cloud/BbCloudFlowManager.js
```diff
@@ -0,0 +1,414 @@
...
+            });
+            this._showPlaceholder();
+        } else if (result.outcome === CreateMbpOutcome.INVALID_URI || result.outcome === CreateMbpOutcome.INVALID_CREDENTIAL) {
+            this.removeSteps({ afterStateId: STATE.STEP_CREDENTIAL });
+            this._showPlaceholder();
+        } else if (result.outcome === CreateMbpOutcome.ERROR) {
+            this.renderStep({
+                stateId: STATE.ERROR,
+                stepElement: <BbUnknownErrorStep error={result.error} />,
+                afterStateId: STATE.STEP_CONNECT,
```
cliffmeyers: This looks like a copy/paste error from Git creation. You probably need to check if `STATE.STEP_NAME` is added and show it after either that or `STATE.STEP_CHOOSE_REPOSITORY`

blueocean-dashboard/src/main/js/creation/bitbucket/cloud/BbCloudFlowManager.js
```diff
@@ -0,0 +1,414 @@
...
+            this.queuedIndexAllocations++;
+        }
+
+        if (event.job_multibranch_indexing_result) {
+            this.queuedIndexAllocations--;
+        }
+
+        if (event.job_orgfolder_indexing_result === 'FAILURE') {
+            this._finishListening(STATE.STEP_COMPLETE_EVENT_ERROR);
+        }
```
cliffmeyers: I think you can remove this if block?

blueocean-dashboard/src/main/js/creation/bitbucket/cloud/BbCloudFlowManager.js
```diff
@@ -0,0 +1,414 @@
...
+    _findExistingCredentialComplete(success) {
+        if (success) {
+            this.changeState(STATE.PENDING_LOADING_ORGANIZATIONS);
+            this.listOrganizations();
+        } else {
+            this.renderStep({
+                stateId: STATE.STEP_CREDENTIAL,
+                stepElement: <BbCredentialsStep />,
+            });
+        }
```
cliffmeyers: There may be a potential bug in the logic here. If we're in the "server" scenario, then after picking the server, the `findExistingCredential()` method is called which triggers this handler. If no credential is found, then the above `renderStep()` call is made, however it will remove the existing `CHOOSE_SERVER_STEP` because no `afterStateId` is specified here. You probably want to add `afterStateId: this._getCredentialsStepAfterStateId()` - see my comment below in `BbServerFlowmanager`

blueocean-dashboard/src/main/js/creation/bitbucket/server/BbServerFlowManager.js
```diff
@@ -0,0 +1,86 @@
...
+        this.setPlaceholders(translate('creation.core.status.completed'));
+    }
+
+    getApiUrl() {
+        return this.selectedServer ? this.selectedServer.apiUrl : null;
+    }
+
+    _getCredentialsStepAfterStateId() {
+        return STATE.STEP_CHOOSE_SERVER;
+    }
```
cliffmeyers: This method shows as unused in IDEA although it probably should be called from `_findExistingCredentialsComplete` in the "Cloud" manager - see my comment above. You probably need to add this method to the "Cloud" manager and return value of `null` since the step can be added at the beginning every time.

blueocean-dashboard/src/main/js/creation/bitbucket/server/BbAddServerDialog.jsx
cliffmeyers: @i386 yeah, it just makes the code slightly more readable to not preface each translation call with `this.` but could go either way with that.

blueocean-dashboard/src/main/js/creation/bitbucket/server/steps/BbChooseServerStep.jsx
```diff
@@ -0,0 +1,137 @@
...
+        this.dropdown = null;
+    }
+
+    componentWillMount() {
+        t = this.props.flowManager.translate;
+    }
+
+    _onChangeServerDropdown(option) {
+        const { serverManager } = this.props.flowManager;
+        serverManager.validateVersion(option.id)
```
cliffmeyers: Is there some way we could validate the server version _before_ we actually create it? It seems non-ideal to let a user create a server that can't actually be used. Is this handling a scenario where off in the future the minimum version of Bitbucket Server gets raised and we need prevent users from selecting servers that were added a long time ago?

blueocean-dashboard/src/main/js/creation/bitbucket/server/steps/BbChooseServerStep.jsx
```diff
@@ -0,0 +1,137 @@
...
+            .then(
+                success => this._onValidateVersion(success),
+                error => this._onValidateVersion(error),
```
cliffmeyers: Since `BbServerManager` explicitly handles the error from `BbServerApi` and transforms it to a result, I don't think you need the second handler here, just one should do it.

blueocean-dashboard/src/main/js/creation/bitbucket/server/steps/BbChooseServerStep.jsx
vivek: This is fine. This is to take care of cases where user adds server via classic UI, so you could still have servers that are not of supported version.

When user adds new servers, we check the version in backend, see https://github.com/jenkinsci/blueocean-plugin/blob/bbc72facebc13eaa320d85b0f04f23e6b019a240/blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerEndpointContainer.java#L59-L59.

blueocean-dashboard/src/main/js/creation/bitbucket/cloud/steps/BbCloudCompleteStep.jsx
```diff
@@ -0,0 +1,122 @@
...
+                {content}
+            </FlowStep>
+        );
+    }
+}
+
+BbCloudCompleteStep.propTypes = {
+    flowManager: PropTypes.object,
+    status: PropTypes.string,
+};
```
cliffmeyers: Should this step actually go in `/bitbucket/steps` ? I believe it's shared by both cloud and server?

blueocean-dashboard/src/main/js/creation/bitbucket/server/steps/BbChooseServerStep.jsx
vivek: No we need it set correct state, see https://github.com/jenkinsci/blueocean-plugin/blob/d6c42d7573c2cf7a8d1b75922b3a062a3d737680/blueocean-dashboard/src/main/js/creation/bitbucket/server/steps/BbChooseServerStep.jsx#L70-L70. Or maybe there is better way to do it?

blueocean-dashboard/src/main/js/creation/bitbucket/cloud/BbCloudFlowManager.js
vivek: There is no choose server step in cloud. Not sure what you mean.

blueocean-dashboard/src/main/js/creation/bitbucket/server/BbServerFlowManager.js
vivek: I think this should be removed. There is no choose server step in cloud.

blueocean-dashboard/src/main/js/creation/bitbucket/cloud/BbCloudFlowManager.js
vivek: Ack

blueocean-dashboard/src/main/js/creation/bitbucket/cloud/BbCloudFlowManager.js
vivek: Ack

blueocean-dashboard/src/main/js/creation/bitbucket/cloud/BbCloudFlowManager.js
cliffmeyers: Right, but in the `BbServerFlowManager` class, this method will still fire after the credentials are fetched, and if no credentials are available then a `renderStep` method call without an `afterStateId` will remove any steps already registered. If you fire up BB Server, make sure all credentials are removed, then pick a server I think you'll see what I'm talking about, but I could be wrong. I don't have a BB server handy to test against.

blueocean-dashboard/src/main/js/creation/bitbucket/server/steps/BbChooseServerStep.jsx
cliffmeyers: It is really a question of style. If you supply only a single function to `Promise.then`, it will only fire if the Promise is resolved successfully without any error. If you provide two functions then the first fires for success and the second fires for fault. `BbServerManager.validateVersion` explicitly handles both success and failure but they each `return` a value, meaning both will fire the success handler. For consistency, maybe you want to change `BbServerManager._onValidateVersionFailure` to `throw new CreateServerError` rather than return it. That's consistent with how the `_onCreateServerFailure` handler does it. If you make that change, then you want leave this code here as-is to handle both scenarios.

blueocean-dashboard/src/main/js/creation/bitbucket/cloud/BbCloudFlowManager.js
vivek: You are right, let me fix it.

blueocean-dashboard/src/main/js/creation/bitbucket/cloud/BbCloudFlowManager.js
```diff
@@ -0,0 +1,414 @@
...
+        this._saveRepo();
+    }
+
+    @action
+    _saveRepo() {
+        const afterStateId = this.isStateAdded(STATE.STEP_RENAME) ?
+            STATE.STEP_RENAME : STATE.STEP_CHOOSE_REPOSITORY;
+
+        this.renderStep({
+            stateId: STATE.STEP_COMPLETE_SUCCESS,
```
cliffmeyers: This is a nit. Rather than immediately putting this into the `STATE.STEP_COMPLETE_SUCCESS` state, we should probably use `STATE.PENDING_CREATION_SAVING`. In `BbCloudCompleteStep.render` we should check for this state and if so set `loading=true` and then pass that down to the `FlowStep`. You can see the same pattern in `GithubCompleteStep` - it just makes the UI a little smoother and avoids a scenario where we briefly show success state but switch to error state after getting an error from backend.

blueocean-dashboard/src/main/js/creation/bitbucket/server/BbServerFlowManager.js
vivek: Keeping it as per above discussion.

blueocean-dashboard/src/main/js/creation/bitbucket/cloud/BbCloudFlowManager.js
cliffmeyers: Actually this not a nit. In `BbCloudCompleteStep._getContent` it checks for success state and uses `setTimeout` to perform a redirect. If we set the success state immediately this code will fire too soon. It then throws an error in the console:
```
jenkins-js-extension.js:55528 Uncaught TypeError: Cannot read property 'organization' of null
    at BbCloudCompleteStep.navigatePipeline (jenkins-js-extension.js:55528)
    at jenkins-js-extension.js:55601
```

blueocean-dashboard/src/main/js/creation/bitbucket/server/steps/BbChooseServerStep.jsx
vivek: Ack

cliffmeyers:
@vivek after latest push, I can create pipelines from private and public repos in Bitbucket Cloud account (which uses an Atlassian ID). Looks like this is getting close. I don't have a mechanism to test Cloud with "legacy" Bitbucket login though.

I am not sure if anyone else is going to lend a hand a testing w/ Bitbucket Server, as it requires a little setup.

vivek:
@cliffmeyers @i386 has setup.

blueocean-dashboard/src/main/js/creation/bitbucket/cloud/steps/BbCloudCompleteStep.jsx
vivek: Done

vivek:
@cliffmeyers All your comments have been addressed. PTAL.

i386:
@cliffmeyers run Bitbucket server using these instructions:

1. [Install the Atlassian SDK on Mac using Homebrew](https://developer.atlassian.com/docs/getting-started/set-up-the-atlassian-plugin-sdk-and-build-a-project/install-the-atlassian-sdk-on-a-linux-or-mac-system)
2. Run `atlas-run-standalone -u 6.3.0 --product bitbucket --version 5.2.0 --data-version 5.2.0`

i386:
@cliffmeyers if its no bother could you try running server tomorrow?

vivek:
@i386 @cliffmeyers The last two bit from code perspective: all branch source plugins upgraded from beta to release version and feature flag added for bitbucket cloud and server creation. 

To enable bitbucket pipeline creation add ?bitbucket to creation url, e.g. `http://localhost:8080/jenkins/blue/organizations/jenkins/create-pipeline/?bitbucket`.

I will wait for @cliffmeyers's +1 after server testing and then merge it to master based.

i386:
@vivek any extra testing results can be written into a new ticket. Merge away :) 🐝

vivek:
@i386 ok, will do after CI run:)

blueocean-dashboard/src/main/js/creation/bitbucket/server/BbAddServerDialog.jsx
```diff
@@ -0,0 +1,192 @@
...
+
+        if (duplicateName) {
+            newState.nameErrorMsg = t('creation.bbserver.add_server.text_name_error_duplicate');
+        }
+
+        if (duplicateUrl) {
+            newState.urlErrorMsg = t('creation.bbserver.add_server.text_url_error_duplicate');
+        } else if (invalidUrl && error.errors) {
+            error.errors.map(err => (
+                    newState.urlErrorMsg = err.message
```
cliffmeyers: 🐛  Adding a server with name and URL of "foo" results in bad request but the UI does not update with the error message. The `err` object only contains `code` and `field` properties, so `newState.urlErrorMsg` ends up being `undefined`. You might want to just to remove the `map` code and just assign `newState.urlErrorMsg = error.message`

vivek: Fixed, in backend though.

cliffmeyers:
Did some more testing, have a question and found a potential upstream Bitbucket branch source issue, but generally looking pretty good:

- Creating MBP from a repo with a Jenkinsfile seems to work fine. User is navigated to Activity tab and sees the builds.
- Creating MBP from a repo without a Jenkinsfile navigates the user to the Activity tab with the "no Jenkinsfile" message. I assume this is expected since the Editor doesn't yet support editing non Org Folder projects?
- Creating MBP from a repo without a default branch succeeds but the indexing fails with an error message (see below). Perhaps a known issue in Bitbucket branch source? I am not sure if this is a realistic use case but perhaps a bug should be filed. My repro steps are:
   1. Create new repository in BB Server UI
   2. Clone the repo
   3. Switch to a new branch e.g. "inital-jenkinsfile"
   4. Add a Jenkins file, commit, and push
   5. Load Creation UI, select the repo and create
   6. Note that "no Jenkinsfile" message displays in UI. Check classic and find error in indexing log.

```
ERROR: [Tue Jul 18 11:04:30 EDT 2017] Could not update folder level actions from source 24e70d6e-8882-498f-88ed-638cfc5cbf20
java.io.IOException: I/O error when accessing URL: /rest/api/1.0/projects/PROJECT_1/repos/repo-2/branches/default
	at com.cloudbees.jenkins.plugins.bitbucket.server.client.BitbucketServerAPIClient.getDefaultBranch(BitbucketServerAPIClient.java:345)
	at com.cloudbees.jenkins.plugins.bitbucket.BitbucketSCMSource.retrieveActions(BitbucketSCMSource.java:967)
	at jenkins.scm.api.SCMSource.fetchActions(SCMSource.java:746)
	at jenkins.branch.MultiBranchProject.computeChildren(MultiBranchProject.java:587)
	at com.cloudbees.hudson.plugins.folder.computed.ComputedFolder.updateChildren(ComputedFolder.java:266)
	at com.cloudbees.hudson.plugins.folder.computed.FolderComputation.run(FolderComputation.java:162)
	at jenkins.branch.MultiBranchProject$BranchIndexing.run(MultiBranchProject.java:969)
	at hudson.model.ResourceController.execute(ResourceController.java:98)
	at hudson.model.Executor.run(Executor.java:410)
Caused by: java.io.FileNotFoundException: URL: /rest/api/1.0/projects/PROJECT_1/repos/repo-2/branches/default
	at com.cloudbees.jenkins.plugins.bitbucket.server.client.BitbucketServerAPIClient.getRequest(BitbucketServerAPIClient.java:508)
	at com.cloudbees.jenkins.plugins.bitbucket.server.client.BitbucketServerAPIClient.getDefaultBranch(BitbucketServerAPIClient.java:342)
	... 8 more
[Tue Jul 18 11:04:30 EDT 2017] Finished branch indexing. Indexing took 67 ms
FATAL: Failed to recompute children of repo-2
java.io.IOException: I/O error when accessing URL: /rest/api/1.0/projects/PROJECT_1/repos/repo-2/branches/default
	at com.cloudbees.jenkins.plugins.bitbucket.server.client.BitbucketServerAPIClient.getDefaultBranch(BitbucketServerAPIClient.java:345)
	at com.cloudbees.jenkins.plugins.bitbucket.BitbucketSCMSource.retrieveActions(BitbucketSCMSource.java:967)
	at jenkins.scm.api.SCMSource.fetchActions(SCMSource.java:746)
	at jenkins.branch.MultiBranchProject.computeChildren(MultiBranchProject.java:587)
	at com.cloudbees.hudson.plugins.folder.computed.ComputedFolder.updateChildren(ComputedFolder.java:266)
	at com.cloudbees.hudson.plugins.folder.computed.FolderComputation.run(FolderComputation.java:162)
	at jenkins.branch.MultiBranchProject$BranchIndexing.run(MultiBranchProject.java:969)
	at hudson.model.ResourceController.execute(ResourceController.java:98)
	at hudson.model.Executor.run(Executor.java:410)
Caused by: java.io.FileNotFoundException: URL: /rest/api/1.0/projects/PROJECT_1/repos/repo-2/branches/default
	at com.cloudbees.jenkins.plugins.bitbucket.server.client.BitbucketServerAPIClient.getRequest(BitbucketServerAPIClient.java:508)
	at com.cloudbees.jenkins.plugins.bitbucket.server.client.BitbucketServerAPIClient.getDefaultBranch(BitbucketServerAPIClient.java:342)
	... 8 more
Finished: FAILURE
```

@i386 wanted to give you a heads up on this issue so we can decide if it needs to be escalated to another team?

vivek:
@cliffmeyers Looks like bitbucket branch source expects a default branch to be set. If you go to their repo setting and set default branch to the other branch it will work. The same indexing code works in case of GitHub because GitHub automatically sets first branch created as `default` branch.

i386:
@cliffmeyers @vivek https://github.com/jenkinsci/bitbucket-branch-source-plugin/pull/61 this should make this error silent

vivek:
Awesome, thanks @i386! Does this mean the other branch with jenkinsfile gets indexed and pipeline job gets created?

i386:
@vivek if by that you mean scanning doesn't shit itself, then yes :)

BTW Pipeline is part of Blue Ocean internally now. You are free to go and change code there so long as you go through the proper code reviews but no longer need to escalate :)

PR #1197, opened 2017-06-22
Eliminate pointless PipelineNodeUtil.getStatus call

abayer:
# Description

This literally always returned SUCCESS, since it could only *not*
return SUCCESS if errorAction != null, and we already have an if
statement for exactly that. So let's simplify the code a little.
(no ticket - this is such a tiny change I assumed it didn't merit one)

No tests are added - this is a simplification of existing code that should match the existing behavior perfectly, so it just has to pass the existing tests to be valid.

No manual tests are needed either, IMO.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

cc @reviewbybees @vivek


michaelneale:
:bee:

PR #1202, opened 2017-06-26
Anonymous user analytics

i386:
# Description

Use Keen.io to track anonymous user analytics. This would be used to measure the usage of certain features and the success of certain flows within the app. 

For example, we would like to know what the success rate of the pipeline creation process is. We would track events for each stage in the flow and we should be able to use a [funnel analysis](https://mixpanel.com/blog/2009/06/10/introduction-to-analytics-funnel-analysis/) to see what step users drop off (where we have failed)

Will not send data unless `hudson.model.UsageStatistics.DISABLED` is `false`. Does not send any PII back to Keen.io apart from a `sha256(instancePublicKey+username)` to uniquely identify the user.

The data collected here is intended for product development only and is available to contributors under the exact same terms as the [Jenkins Usage Statistics](https://wiki.jenkins.io/display/JENKINS/Usage+Statistics).

See [JENKINS-45144](https://issues.jenkins-ci.org/browse/JENKINS-45144)

**POST /blue/rest/analytics/track**
```
{
  name: 'cool_event',
  properties: {
    prop1: 123,
    prop2: 'hello world'
  }
}
```

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-commons/pom.xml
+++ b/blueocean-commons/pom.xml
@@ -20,6 +20,14 @@
       <artifactId>jackson-databind</artifactId>
       <version>${jackson.version}</version>
     </dependency>
+    <dependency>
+      <groupId>io.keen</groupId>
+      <artifactId>keen-client-api-java</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.modules</groupId>
+      <artifactId>instance-identity</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
--- a/blueocean-commons/pom.xml
+++ b/blueocean-commons/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>io.jenkins.blueocean</groupId>
     <artifactId>blueocean-parent</artifactId>
-    <version>1.2.0-beta-1-SNAPSHOT</version>
+    <version>1.2.0-beta-2-SNAPSHOT</version>
   </parent>
 
   <artifactId>blueocean-commons</artifactId>
--- a/blueocean-commons/pom.xml
+++ b/blueocean-commons/pom.xml
@@ -20,14 +20,6 @@
       <artifactId>jackson-databind</artifactId>
       <version>${jackson.version}</version>
     </dependency>
-    <dependency>
-      <groupId>io.keen</groupId>
-      <artifactId>keen-client-api-java</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.modules</groupId>
-      <artifactId>instance-identity</artifactId>
-    </dependency>
   </dependencies>
 
   <build>
--- a/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/analytics/Analytics.java
+++ b/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/analytics/Analytics.java
@@ -71,7 +71,7 @@ public abstract class Analytics {
     protected abstract void doTrack(String name, Map<String, Object> allProps);
 
     protected final String server() {
-        return Hashing.md5().hashBytes(InstanceIdentity.get().getPublic().getEncoded()).toString();
+        return Hashing.sha256().hashBytes(InstanceIdentity.get().getPublic().getEncoded()).toString();
     }
 
     protected final String identity() {
--- a/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/analytics/Analytics.java
+++ b/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/analytics/Analytics.java
@@ -1,0 +1,70 @@
+package io.jenkins.blueocean.commons.analytics;
+
+import com.google.common.base.Objects;
+import com.google.common.collect.Maps;
+import com.google.common.hash.Hashing;
+import hudson.model.UsageStatistics;
+import hudson.model.User;
+import io.jenkins.blueocean.commons.ServiceException;
+import org.apache.commons.lang.StringUtils;
+import org.jenkinsci.main.modules.instance_identity.InstanceIdentity;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@Restricted(NoExternalUse.class)
+public abstract class Analytics {
+
+    private static final Logger LOGGER = Logger.getLogger(Analytics.class.getName());
+
+    public static class TrackRequest {
+        public final String name;
+        public final Map<String, Object> properties;
+
+        @DataBoundConstructor
+        public TrackRequest(String name, Map<String, Object> properties) {
+            this.name = name;
+            this.properties = properties;
+        }
+    }
+
+    /**
+     * @return analytics instance
+     */
+    public static Analytics get() {
+        return UsageStatistics.DISABLED ? NullAnalytics.INSTANCE : KeenAnalyticsImpl.INSTANCE;
+    }
+
+    /**
+     * @param req to track
+     */
+    public void track(TrackRequest req) {
+        if (req == null) {
+            throw new ServiceException.BadRequestException("missing request");
+        }
+        if (StringUtils.isEmpty(req.name)) {
+            throw new ServiceException.BadRequestException("missing name");
+        }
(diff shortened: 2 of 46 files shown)


blueocean-commons/src/main/java/io/jenkins/blueocean/commons/analytics/Analytics.java
```diff
@@ -0,0 +1,70 @@
...
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+@Restricted(NoExternalUse.class)
+public abstract class Analytics {
```
vivek: This could be extension point as well. Might make it easier to swap analytics backend rather than depending on specific impl.

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/analytics/KeenAnalyticsImpl.java
```diff
@@ -0,0 +1,32 @@
...
+import java.util.logging.Logger;
+
+@Restricted(NoExternalUse.class)
+public class KeenAnalyticsImpl extends Analytics {
+
+    static final Analytics INSTANCE = new KeenAnalyticsImpl();
+
+    private static final Logger LOGGER = Logger.getLogger(KeenAnalyticsImpl.class.getName());
+    private static final String PROJECT_ID = "5950934f3d5e150f5ab9d7be";
+    private static final String WRITE_KEY = "E6C1FA3407AF4DD3115DBC186E40E9183A90069B1D8BBA78DB3EA6B15EA6182C881E8C55B4D7A48F55D5610AD46F36E65093227A7490BF7A56307047903BCCB16D05B9456F18A66849048F100571FDC91888CAD94F2A271A8B9E5342D2B9404E";
```
vivek: We are hard coding this key in product, what if it gets rotated by keen.io?

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/analytics/KeenAnalyticsImpl.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private KeenAnalyticsImpl() {
+        this.client = new JavaKeenClientBuilder().build();
+        KeenClient.initialize(this.client);
+        KeenClient.client().setDefaultProject(new KeenProject(PROJECT_ID, WRITE_KEY, null));
+    }
+
+    @Override
+    protected void doTrack(String name, Map<String, Object> allProps) {
+        client.addEvent(name, allProps);
```
vivek: Probably wrap it in try catch so that even if keen.io is down or there are other networking issue, we do not throw stack trace over to server log.

vivek:
@i386 it's also possible to post analytics directly from UI. Won't it be more efficient to post these directly over to keen.io? I guess, if needed certain opt in info can be provided to UI as well?

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/analytics/KeenAnalyticsImpl.java
i386: Then the installation can no longer send the data. Not much we can do.

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/analytics/KeenAnalyticsImpl.java
i386: added a try catch in `Analytics` and updated this call so it sends the event async

i386:
@vivek we could but then we have to rely on including their JavaScript into all of our pages, make their JS honour `hudson.model.UsageStatistics.disabled` and if we wanted to swap implementation out we'd have to have a frontend wrapper to abstract the events.

With this approach we can also send backend events (which is in my experience something we will likely end up doing).

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/analytics/Analytics.java
i386: I expect it could be but adds very little value right now?

vivek: yeah but sounds like right thing to do without much cost. but I agree no immediate need of seeing it as extension point that is fairly easily replaceable, might defeat the purpose. I guess when this happens it can be big deal anyways. I will leave it up to you.

vivek: Maybe where I was going was that, how could other plugins participate in sending their events as well. Not immediate need but some extensions that could let them do it. For example, what if pipeline editor wants to add it's events, e.g. editor version etc.

i386: Due to costs associated I want to collect this data for Blue Ocean purposes first. If we find it to be useful, we can have a wider discussion about expanding it to other plugins.

vivek: Sounds reasonable.

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/analytics/Analytics.java
```diff
@@ -71,7 +71,7 @@ public void track(TrackRequest req) {
     protected abstract void doTrack(String name, Map<String, Object> allProps);
 
     protected final String server() {
-        return InstanceIdentity.get().getPublic().getPublicExponent().toString(16);
+        return Hashing.md5().hashBytes(InstanceIdentity.get().getPublic().getEncoded()).toString();
```
i386: @vivek this was nonsense and now its not.

vivek: @i386 md5? why not use sha256?

i386: I believe SSH uses md5 hash of the public key as a fingerprint of the public key. So I am using that. We could use sha256 if you want?

vivek: hmm, I thought RSA public key generated using sha256 algorithm in Jenkins, not sure though. regardless I think safer to use sha-256.

i386: @vivek OK - ill switch to sha256.

blueocean-commons/pom.xml
```diff
@@ -20,6 +20,14 @@
       <artifactId>jackson-databind</artifactId>
       <version>${jackson.version}</version>
     </dependency>
+    <dependency>
```
vivek: This should be removed?

rsandell:
Since you are sending the data from the backend you also need to configure the proxy used by Jenkins or your data won't be sent if there is a proxy in the way ;)

i386:
@rsandell good point. Do we get this from `hudson.ProxyConfiguration`? Seems I can only open `URL`s or `InputStream`s

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/analytics/KeenAnalyticsImpl.java
```diff
@@ -0,0 +1,45 @@
...
+    private static final String PROJECT_ID = "5950934f3d5e150f5ab9d7be";
+    private static final String WRITE_KEY = "E6C1FA3407AF4DD3115DBC186E40E9183A90069B1D8BBA78DB3EA6B15EA6182C881E8C55B4D7A48F55D5610AD46F36E65093227A7490BF7A56307047903BCCB16D05B9456F18A66849048F100571FDC91888CAD94F2A271A8B9E5342D2B9404E";
+    private final KeenClient client;
+
+    public KeenAnalyticsImpl() {
+        this.client = new JavaKeenClientBuilder().build();
+        KeenClient.initialize(this.client);
+        ProxyConfiguration proxyConfig = Jenkins.getInstance().proxy;
+        if (Jenkins.getInstance().proxy == null) {
+            KeenClient.client().setProxy(proxyConfig.createProxy(null));
```
i386: @rsandell @vivek does this look right?

i386:
I am going to merge this now to make sure it works on dogfood. Can be disabled at any time.

PR #1205, opened 2017-06-28
trying a retry for flaky test

michaelneale:
This test has been problematic for a long time - this is an attempt to see if a retry and a wait helps (due to consistency of resources and replay - problem is not in blue ocean code)


michaelneale:
@i386 sniffs ok? I am hoping a wait and retry may stop this problem if it is in fact a consistency issue (400 errors happen from time to time).

i386:
@michaelneale code wise it looks fine. Sure this isn't hiding a real problem?

michaelneale:
@i386 it could well be - not sure what though. We have other tests for replay right? 
hrm maybe @vivek may know what a 400 means from rest?

PR #1208, opened 2017-06-29
Properly use the latest Docker credentials in our trusted.ci env for publishing

rtyler:
The assumption that the Jenkins agent will simply have ~/.docker/config.json
around is no longer true.

Related to INFRA-1066


michaelneale:
:bee:

PR #1209, opened 2017-06-30
Sanitise step descriptions

i386:
# Description

@abayer @vivek looks like `ArgumentsAction.getStepArgumentsAsString` might include `\t` and other characters that are bad for JSON. Should this really be updated in workflow-api? I am not sure having a `\t` character in a step name is something good for human eyeballs anyway. Feels some sanitation should occur there rather in blue ocean?

The problem is that when we encounter these issues parts of the UI fail to load as described in the ticket below.

See [JENKINS-45099](https://issues.jenkins-ci.org/browse/JENKINS-45099).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-commons/src/test/java/io/jenkins/blueocean/commons/JSONTest.java
+++ b/blueocean-commons/src/test/java/io/jenkins/blueocean/commons/JSONTest.java
@@ -1,0 +1,11 @@
+package io.jenkins.blueocean.commons;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class JSONTest {
+    @Test
+    public void sanitize() throws Exception {
+        Assert.assertEquals("hello vivek :)", JSON.sanitizeString("hello\r vivek\n\t :)"));
+    }
+}
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
@@ -1,5 +1,6 @@
 package io.jenkins.blueocean.rest.impl.pipeline;
 
+import com.google.common.base.CharMatcher;
 import com.google.common.base.Predicate;
 import hudson.ExtensionList;
 import hudson.FilePath;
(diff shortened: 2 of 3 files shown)


michaelneale:
nice - would like the nod from @vivek on this that this is right place and if there are other problems.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
```diff
@@ -74,7 +75,12 @@ public String getDisplayName() {
 
     @Override
     public String getDisplayDescription() {
-        return ArgumentsAction.getStepArgumentsAsString(node.getNode());
+        String displayDescription = ArgumentsAction.getStepArgumentsAsString(node.getNode());
+        if (displayDescription != null) {
+            // JENKINS-45099 Remove any control characters that may have found their way out of a script
+            displayDescription = CharMatcher.JAVA_ISO_CONTROL.and(CharMatcher.anyOf("\r\n\t").negate()).removeFrom(displayDescription);
```
michaelneale: I wonder if there is a more general sanitise thing that can be done here...

i386: I am not sure myself. Advice needed.

vivek: @i386 Code looks good, if I read it correctly, it filters out ISO controls chars (not json friendly) and line/tab breaks. We should keep it inside commons package, maybe inside Utils.java as generic code, could be used somewhere else if needed, perhaps something along this?

```
public String sanitizeJsonString(@Nonnull String value, @Nullable String additionalFilter){
        
        CharMatcher charMatcher = CharMatcher.JAVA_ISO_CONTROL;
        if(StringUtils.isNotBlank(additionalFilter)){
            charMatcher = charMatcher.and(CharMatcher.anyOf(additionalFilter).negate());
        }
        return charMatcher.removeFrom(value);
    }
```

i386: Done.

blueocean-commons/src/test/java/io/jenkins/blueocean/commons/JSONTest.java
```diff
@@ -0,0 +1,11 @@
+package io.jenkins.blueocean.commons;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class JSONTest {
```
i386: @vivek added a test for the util

PR #1210, opened 2017-06-30
Add browser and operating system stats to all event properties

i386:
# Description

Adds browser and client OS data to each analytics event

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,16 +131,6 @@
       <id>repo.jenkins-ci.org</id>
       <url>http://repo.jenkins-ci.org/public/</url>
     </repository>
-    <repository>
-      <id>twitter</id>
-      <url>http://maven.twttr.com</url>
-      <releases>
-          <enabled>true</enabled>
-      </releases>
-      <snapshots>
-          <enabled>false</enabled>
-      </snapshots>
-    </repository>
   </repositories>
   <pluginRepositories>
     <pluginRepository>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -131,6 +131,16 @@
       <id>repo.jenkins-ci.org</id>
       <url>http://repo.jenkins-ci.org/public/</url>
     </repository>
+    <repository>
+      <id>twitter</id>
+      <url>http://maven.twttr.com</url>
```
michaelneale: hrm - not sure about this if it isn't in one of the other repos - not sure we should be adding another one in - doesn't seem right.

i386: FWIW, I have used this library for the same purpose on a previous project.

michaelneale: the library isn't a problem, but we have jenkins and central repos only  - and a mirror etc, this ads another dep - it really shoudl be in central

i386: No idea how to get it into central...

i386: @michaelneale I think the right thing to do here is make sure this repository is mirrored by the Jenkins project's Artifactory and then we can remove this `<repository>` entry.

I have raised [INFRA-1259](https://issues.jenkins-ci.org/browse/INFRA-1259)

i386: Resolved and removed.

michaelneale: ah ok - in that case - I think this is ok as long as it builds. I think Tyler is ok with it - so probably kosher? I am being a little conservative here ;)

jtnord:
What is this? there is no JIRA and this appears to bring in new libraries that clash with exisiting plugins causing funky runtime dependencies?  (snakeyaml)

//cc @alvarolobato

PR #1212, opened 2017-06-30
JENKINS-45011 make commit id in run result header clickable

NicuPascu:
# Description

See [JENKINS-45011](https://issues.jenkins-ci.org/browse/JENKINS-45011).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


NicuPascu:
@imeredith not sure if this is restable, @i386 asked in comment to this ticket to ask you about maybe adding a test for this, what are your thoughts?

blueocean-dashboard/src/main/js/components/RunDetailsHeader.jsx
```diff
@@ -124,7 +124,13 @@ class RunDetailsHeader extends Component {
             <div className="u-label-value" title={commitLabel + ': ' + commitIdString}>
                 <label className={labelClassName}>{ commitLabel }:</label>
                 <span className="commit">
```
i386: @NicuPascu to test this correctly:

1. Split this code for displaying commit IDs out into something like `CommitIdCell.jsx`
```
export default class CommitIdCell extends Component {
  render() { 
    return (<span className="commit">...</span>);
  }
}
```
2. Write an [enzyme test](http://airbnb.io/enzyme/docs/api/)
import React from 'react';
import { expect } from 'chai';
import { render } from 'enzyme';
import CommitIdCell from '../../main/js/components/CommitIdCell.jsx"';

describe('CommitIdCell', () => {

    it('displays commit ID with link', () => {
        const run =  { ... mocked run object goes here ... };
        const cell = render(<CommitIdCell run={run} t={t} />);
        expect(cell.text()).to.equal(... something ...); // you will have to change this to test the rendered output
    });


There is a good example in `RunMessageCell.jsx` and `run-message-cell-spec.js` on how to write these tests

i386:
🐛 we should remove the custom formatting of the commit ID and use `https://github.com/jenkinsci/jenkins-design-language/blob/master/src/js/components/CommitId.jsx` instead. `CommitId` handles how to shorten a git commit id from `c410e38f46bfb74dfce8947ef4690b2dcdb12e71` to `c410e38`.

For example, you compose like so:
```
<CommitIdCell> // Your new component
   <CommitId commitId={commit}> // Component from JDL
</CommitIdCell>
```

i386:
@NicuPascu actually there looks to be a `<CommitLink>` component that we already use in `RunDetailsChanges` - perhaps we could refactor it to make it our `<CommitIdCell>`?

NicuPascu:
JDL PR - https://github.com/jenkinsci/jenkins-design-language/pull/180

blueocean-dashboard/src/main/js/components/RunDetailsChanges.jsx
```diff
@@ -68,7 +58,7 @@ export default class RunDetailsChanges extends Component {
             <Table headers={headers} className="changeset-table">
                 { changeSet.map(commit => (
                     <tr key={commit.commitId}>
-                        <td><CommitLink {...commit} /></td>
+                        <td><CommitHash commitId={commit.commitId} url={commit.url} /></td>
```
cliffmeyers: Agree, this is much a better approach.

i386: is there a jsspec for RunDetailsChanges? Does it cover CommitHashes? Might be worth creating a test if its not a huge hassle.

i386:
🐝

michaelneale:
Getting there - but there is a test failure @NicuPascu : 

```
[ERROR] 
[ERROR] Summary of all failing tests
[ERROR]  FAIL  src/test/js/run-details-changes-spec.js
[ERROR]   ??? RunDetailsChanges ??? valid changeSet ??? renders a Table with expected data
[ERROR] 
[ERROR]     AssertionError: expected '<CommitId />' to equal '<CommitLink />'
[ERROR]       
[ERROR]       at Object.<anonymous> (src/test/js/run-details-changes-spec.js:70:26)
[ERROR] 
[ERROR] 
```

blueocean-dashboard/src/main/js/components/RunDetailsChanges.jsx
NicuPascu: yes, that was the test that broke, I fixed it

i386: 👍

michaelneale:
nice - @cliffmeyers all look ok? I'll let you confirm you are happy 

LGTM

blueocean-dashboard/src/main/js/components/RunDetailsHeader.jsx
```diff
@@ -119,12 +119,13 @@ class RunDetailsHeader extends Component {
             </div>
         );
 
-        const commitIdString = run.commitId || '—';
+        const commitIdString = changeSet.length && changeSet[changeSet.length - 1].commitId || '—';
```
michaelneale: can changeSet be null? ie if triggered not from a change?

jenkins-design-language/src/js/components/CommitId.jsx
```diff
@@ -9,7 +9,7 @@ const GIT_HASH_REGEX = /\b[0-9a-f]{5,40}\b/;
  */
 export class CommitId extends Component {
     render() {
-        const commitId = this.props.commitId;
+        const {commitId, url, title} = this.props;
```
michaelneale: if commitId is null - still ok with destructuring like this? (my js foo is weak)

michaelneale:
nice - @cliffmeyers all look ok? I'll let you confirm you are happy 

I had a few Q's about the likelihood of nulls I put above

jenkins-design-language/src/js/components/CommitId.jsx
NicuPascu: yes, I just double checked to be sure :)

blueocean-dashboard/src/main/js/components/RunDetailsHeader.jsx
NicuPascu: yes, in that case commitIdString becomes '-'

i386: @NicuPascu check but I think the `CommitId` component will handle converting a null commitId to `–` for you.

blueocean-dashboard/src/main/js/components/RunDetailsHeader.jsx
```diff
@@ -119,12 +119,13 @@ class RunDetailsHeader extends Component {
             </div>
         );
 
-        const commitIdString = run.commitId || '—';
+        const commitIdString = changeSet.length && changeSet[changeSet.length - 1].commitId;
```
NicuPascu: I removed the '-' attribution because it is indeed handled by the CommitId component, we do still need to check for changeSet.length to not go out of bounds. (I tried to do it intentionally to see what would happen and mobx threw an out of bounds error)

michaelneale: so what about null changeSet?

michaelneale:
Nice @NicuPascu. Still concerned about nulls - ie the case where a pipeline is not multibranch, and has no repos at all, just runs - surely that may be a problem? 

Also needs updating due to package.json conflict/blah blah etc. 

@cliffmeyers is out today but may get to review this tomorrow

sophistifunk:
Looks OK to me 🐝

michaelneale:
I tried this - the commitId didn't show up at all for me for a multibranch project. Just showed up as "-". 

When I switched back to master I saw the commit hash show up - but it was not linkable (obviously) - so clearly some :bug: here hanging around... 

To see: 
* add a project like: https://github.com/multibranchorg/app-store-demo from say the master branch of blue ocean. 
* Let the builds run
* switch to this branch and rebuild
* open the project again and note that there are no commit ids in the results page for finished builds

NicuPascu:
Yeah, I can see the problem, the cangeset is empty in this case, the run.commitID that was used before is present in the run obj but there is no url. The only solution I can see is to make a change in the backend to send the url of the commit in the run obj so we don't depend on the changeset

michaelneale:
@NicuPascu right - otherwise lots of speial case code to handle changeset or not... can park this until you get time to get back to it if you like as you get ready to travel

michaelneale:
@NicuPascu actually this looks like it has picked up a problem with the ATH - it really should be looking for the commit hash. Perhaps as part of this we should enhance it to look for this?

kzantow:
Just a note with [a comment on the JIRA issue that has a potential solution](https://issues.jenkins-ci.org/browse/JENKINS-45011?focusedCommentId=307383&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-307383)

NicuPascu:
Thanks @kzantow for the help, the code is now ready for review

michaelneale:
I tried this with a multibranch one - it wasn't clickable - am I missing somthing? @NicuPascu can you make a gif movie and some screenshots showing this in action?

NicuPascu:
@michaelneale 
recording_2017-07-20_185939

michaelneale:
@NicuPascu for some reason I didn't see that - if @kzantow agrees - he can approve this.

PR #1215, opened 2017-07-03
Merge JDL to blueocean-plugin

kzantow:
# Description

Just moving JDL into the blueocean-plugin

See [JENKINS-42602](https://issues.jenkins-ci.org/browse/JENKINS-42602).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


.npmignore
```diff
@@ -0,0 +1,13 @@
...
+node_modules/
+npm-debug.log
+.DS_Store
+*.iml
+.idea/
+/.project
+reports
+/website/storybook
+/website/component-docs/doc*
+/website-build
```
cliffmeyers: I believe this was added to the JDL before but I'm not really familiar with how it works. Will npm check a parent directory for this file?

kzantow: According to the docs:
> Like git, npm looks for .npmignore and .gitignore files in all subdirectories of your package, not only the root directory.
Besides, this commit just changes where we are actually releasing from, which won't hit a parent `.npmignore`, as far as I understand it. Good question, though -- had to look it up :)

PR #1218, opened 2017-07-03
Add ability to post to the analytics track endpoint

i386:
# Description

Adds `AnalyticsService` to core-js that allows events to be tracked:
```
import analyticsService from '@jenkins-cd/blueocean-core-js';
analyticsService.track('cool_event', { prop1: 123, prop2: 'hello world' });
```

**POST /blue/rest/analytics/track**
```
{
  name: 'cool_event',
  properties: {
    prop1: 123,
    prop2: 'hello world'
  }
}
```

See [JENKINS-45144](https://issues.jenkins-ci.org/browse/JENKINS-45144).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-core-js/src/js/analytics/AnalyticsService.js
+++ b/blueocean-core-js/src/js/analytics/AnalyticsService.js
@@ -4,19 +4,18 @@ import utils from '../utils';
 
 export class AnalyticsService {
     track(eventName, properties) {
-        const path = config.getJenkinsRootURL();;
+        const path = config.getJenkinsRootURL();
         const url = utils.cleanSlashes(`${path}/blue/rest/analytics/track`);
-        const body = {
-            name: eventName,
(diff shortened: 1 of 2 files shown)


imeredith:
@i386 You can use nock to mock an http call https://www.npmjs.com/package/nock

sophistifunk:
I think the last thing we need is more global const singletons that make everything that uses them harder to test. These things are contextual, and belong in a context of some sort.

michaelneale:
I may lack imagination here, but 90% of this PR is calling IO - so I am not sure what virtue testing that isn't testing it for real is - I think this is a case of signing off as ok with out it (unless an ATH test can be done that has it post somewhere it can capture). Seriously almost all the code in this is just a request being posted somewhere...

PR #1227, opened 2017-07-05
JENKINS-45016 Branch filter clear behaviour

NicuPascu:
Set the original value back if blured with an empty value, force the branch filter to lose focus after change

the setTimeout hack is used to get the functions out of the call stack and only run it when the stack is empty

as stated in the comment already in the ColumnFilter.jsx, the Autocomplete component should be replaced at some point with our own custom dropdown 

# Description

See [JENKINS-45016](https://issues.jenkins-ci.org/browse/JENKINS-45016).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
is there a way to write a test for this? 

Given it is timing based - possibly an ATH test would be best if any (filter is tested by the ATH so isn't lack of coverage, just wanted to verify this).

blueocean-dashboard/src/main/js/components/ColumnFilter.jsx
```diff
@@ -39,8 +43,14 @@ export class ColumnFilter extends Component {
         e.target.select();
     }
 
-    blur() {
+    blur = e => {
+        const targetElem = e.target;
```
michaelneale: Is e actually set? in the code above it is called with no param?

michaelneale:
@sophistifunk review of code please

blueocean-dashboard/src/main/js/components/ColumnFilter.jsx
sophistifunk: @michaelneale the blur handler here is not being called from the code in onChange (directly)

blueocean-dashboard/src/main/js/components/ColumnFilter.jsx
```diff
@@ -39,8 +43,14 @@ export class ColumnFilter extends Component {
         e.target.select();
     }
 
-    blur() {
+    blur = e => {
```
sophistifunk: Using arrow funcs for event handlers is the right way to go, but then we need to also update `render()` to set listers from `onEvent={e=>this.eventHandler(e)}` to `onEvent={this.eventHandler}` otherwise we don't get the benefits.

sophistifunk:
@NicuPascu could you please go into the timeouts a bit more? I'm having trouble understanding the point of it. What is it you're trying to wait for? What goes wrong if you don't use them?

i386:
@NicuPascu does this component need more time for a refactor?

blueocean-dashboard/src/main/js/components/ColumnFilter.jsx
```diff
@@ -26,7 +26,11 @@ export class ColumnFilter extends Component {
         if (event.type === 'select'
             || event.type === 'blur'
             || (event.type === 'keypress' && event.key === 'Enter')) {
+            this.setState({ originalValue: value });
             onChange(value);
+            setTimeout(() => {
+                this.refs.autocomplete.refs.input.blur();
```
kzantow: `this.refs` is going away, no?

sophistifunk: Yes, but this line can't be refactored to use callback refs without also changing the autocomplete component, so we should probably leave it until we can replace it with our dropdown.

michaelneale:
feedback has been left above...

sophistifunk:
I think this is probably as nice as we're going to get, using this particular auto-complete component. Nice work 🐝

michaelneale:
ok then :bee:

michaelneale:
LGTM - that failure should be resolved on master soon when @imeredith gets retry in - so after that, update it and check it is ok. then it can go in! nice one @NicuPascu !

michaelneale:
if this gets a green build you can merge

michaelneale:
I am not sure this has latest master - pull and try again? that failure should be spurious.

PR #1228, opened 2017-07-05
Use GraphListener and FlowExecutionListener to receive events

jglick:
# Description

Uses more recently added listener APIs to avoid tricks with `RunListener`.

No special test coverage; presumably would be covered by existing tests.

@reviewbybees esp. @abayer

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
+++ b/blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
@@ -1,10 +1,8 @@
 package io.jenkins.blueocean.events;
 
-import com.google.common.util.concurrent.ListenableFuture;
 import hudson.Extension;
+import hudson.model.Queue;
 import hudson.model.Run;
-import hudson.model.TaskListener;
-import hudson.model.listeners.RunListener;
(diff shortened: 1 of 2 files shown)


blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
```diff
@@ -45,135 +44,146 @@
...
+    private final Map<FlowExecution, String> currentStageId = new WeakHashMap<>();
 
-        private final Run run;
-        private final PubsubBus pubSubBus;
-        private String currentStageName;
-        private String currentStageId;
+    @Override
+    public void onNewHead(FlowNode flowNode) {
+        // test whether we have a stage node
+        if (PipelineNodeUtil.isStage(flowNode)) {
```
jglick: [for reviewers](https://github.com/jenkinsci/blueocean-plugin/pull/1228/files?w=1)

michaelneale:
cc @vivek for your review

michaelneale:
this is ok with me if we can get one other review. Passes the ATH etc. 

thanks @jglick - yeah tests cover this (should anyway)

i386:
Looks reasonable to me 🐝

i386:
Thanks for this one @jglick 💯

PR #1229, opened 2017-07-05
[JENKINS-45058] update 1.1.0 branch to not stop github settings

stephenc:
@reviewbybees


stephenc:
@michaelneale can you tag this DO NOT MERGE

michaelneale:
@stephenc this works with older and to be released branch source plugin? (compatibility is most short term priority)

michaelneale:
talked this over - and seems reasonable. @vivek @cliffmeyers @i386 if one of you bless it - we can make it happen. 

This makes things better for the upcoming changes. Master branch will have a slightly different set of fixes.

i386:
Seems reasonable to me. The only thing I'd love to see here is some tests!

kzantow:
👀

michaelneale:
dont' need to merge this until 1.1.5 is out I think... 
Or will this be irrelevant as we move away from org folders? 

WDYT @vivek ?

PR #1230, opened 2017-07-05
[JENKINS-45058] Fix that does not rely on JENKINS-43507

stephenc:
@reviewbybees


stephenc:
@michaelneale can you tag this DO NOT MERGE

michaelneale:
this also works with old and "new" branch source right @stephenc ? (I am trying to make sure we don't need to do a co-ordinated release)

michaelneale:
@stephenc I am not sure why this is failing - is this fork of a very recent master or possibly a very old one?

stephenc:
@michaelneale Should be ok now

@vivek you may need to review (and perhaps revise) my attempt to fix the use of setEndpoints

michaelneale:
Right - this kicked the hornets nest - a whole lot of errors due to duplicates - seems odd this would be due to this change - https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/PR-1230/4/pipeline might need to drop this in @vivek's queue to look at.

michaelneale:
closing this as I think @vivek is going to take the other approach

PR #1237, opened 2017-07-10
verify pubsub-light 1.10 compatibility

tfennelly:
No need to merge this (unless ye want it). Just verifying BO compatibility with pubsub-light 1.10.

If ye want to upgrade, we can create a ticket etc and relabel this PR.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -449,7 +449,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>pubsub-light</artifactId>
-            <version>1.8</version>
+            <version>1.10</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>


i386:
Looks fine Tom. We probably should merge - what say you @vivek?

michaelneale:
is this failure anything to do with this? 

```
parameterizedBranchTest(io.jenkins.blueocean.rest.impl.pipeline.MultiBranchTest)  Time elapsed: 18.664 sec  <<< FAILURE!
org.junit.ComparisonFailure: expected:<[SUCCESS]> but was:<[UNKNOWN]>
```

?

vivek:
@michaelneale Probably flaky test. That should not fail because of pubsub plugin changes. I replayed it.

tfennelly:
Definitely seems like this has made the ATH more unstable. The `master` branch seems to be building ok, but this one is hanging.

The only thing I can think of is that it's causing a bit of latency that's screwing with things coz the only diff here should be the fact that there's a `jenkins-instance-url` property in events.

michaelneale:
@tfennelly is your branch well up to date with master? 

If so, does seem related as that hanging is when it misses events right?

michaelneale:
I'll run it again to confirm...

michaelneale:
yes this is very much not working with the ATH

tfennelly:
Prob related to https://issues.jenkins-ci.org/browse/JENKINS-45416 ... obviously more than some latency thing ... I'll look into it first thing in the morning.

michaelneale:
yes that would be it exactly - no events are being picked up, so not just ATH.

i386:
Running a rollback https://github.com/jenkinsci/blueocean-plugin/pull/1248

michaelneale:
will close this - can re-open when we have a new version of pubsub lite to try that has the desired changes (latest one is essentially a rolled back version of bits)

PR #1242, opened 2017-07-11
Topic/use add endpoint

michaelneale:
Stephen tells me that addEndpoint() will survive the API changes he
has planned for Github Branch Source and endpoint unification.

@reviewbybees


michaelneale:
@i386 gosh I am a bit of a dill - I merged master back in to this branch which was the "undo" 

... so I guess I should undo the undoing and see it it passes... again.

stephenc:
Ultimately you'll need to move off addEndpoint, but that is very long term. Once we get unification, use of setEndpoints may trigger loss of configuration for admins, so the *critical* thing is to stop using setEndpoints and switch to addEndpoint

michaelneale:
I know @vivek has previously blessed this one (just chasing down some test issues...)

PR #1244, opened 2017-07-12
Lets see if this makes it clearer

michaelneale:
# Description

api.github.com already in the list messes up the GHE test, as it wants to use it to test GHE (as github cloud is same API).


i386:
🐝

PR #1249, opened 2017-07-13
Upgrade to pubsub-light 1.12-beta-1

i386:
Just for testing

Related to JENKINS-45416


michaelneale:
this one is kaput - just like expected (hangs, as no events are picked up)

PR #1252, opened 2017-07-13
[JENKINS-45442] [JENKINS-45482] [JENKINS-45360] 1.1 branch: Pipeline creation compatibility with custom orgs

jtnord:
# Description

Fixes compatabilty with pipeline (GitHubOrg/MBP) creation and custom organisations.

See [JENKINS-45442](https://issues.jenkins-ci.org/browse/JENKINS-45442) / [JENKINS-45482](https://issues.jenkins-ci.org/browse/JENKINS-45482) / [JENKINS-45360](https://issues.jenkins-ci.org/browse/JENKINS-45260)

To Manually test, build and deploy inside a setup that uses a custom organisation.

- Create a GitHubOrg limited to one repo.
- Update the GitHUb Org (add a different repo)
- Create a MBP from a git repo
- try and create a MBP from the same repo

# Submitter checklist
- [x] Link to JIRA ticket in description, if appropriate.
- [x] Change is code complete and matches issue description
- [x] Appropriate unit or acceptance tests or explanation to why this change has no tests
- [x] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below.

# Reviewer checklist
- [ ] Run the changes and verified the change matches the issue description
- [ ] Reviewed the code
- [ ] Verified that the appropriate tests have been written or valid explanation given

@reviewbybees

PR #1262, opened 2017-07-19
[FIX JENKINS-43148] Add API for creating/reading a user's private SSH key

kzantow:
# Description

Adds a get-or-create API for a user's Jenkins-managed generated public key. Can test this by logging in and accessing (a similar URL):

http://localhost:8080/jenkins/blue/rest/organizations/jenkins/user/publickey

See [JENKINS-43148](https://issues.jenkins-ci.org/browse/JENKINS-43148).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


i386:
@daniel-beck security related PR that would be deserving of your attention

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/CredentialApi.java
```diff
@@ -50,6 +55,20 @@ public String getStore(){
         return credentialStoreAction.getUrlName();
     }
 
+    @GET
```
i386: Make this Get more friendly (remove the caps):
```
@Get('publickey')
public void getPublicKey()
```

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/SSHCredentialManager.java
```diff
@@ -0,0 +1,188 @@
...
+        return key;
+    }
+    
+    /**
+     * Gets a readable SSH-compatible public key a user could paste somewhere
+     * @param user
+     * @param key 
+     * @return 
+     */
+    public String getReadablePublicKey(@Nonnull User user, BasicSSHUserPrivateKey key) {
```
i386: `getPublicKey` and method `@Notnull`

michaelneale:
LGTM - any feedback from hat wearing people like @daniel-beck appreciated. 

Nice tests too 👌🏽

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/SSHCredentialManager.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    /**
+     * Encodes the public key according to some spec somewhere
+     * @param key public key to use
+     * @return the ssh-rsa bytes
+     * @throws IOException 
+     */
+    private static byte[] encodePublicKey(RSAPublicKey key) throws IOException {
```
i386: perhaps `getReadablePublicKey` moves to a static on the key generation class? I think all the crypto usage shouldn't be in this class.

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/CredentialApiTest.java
```diff
@@ -252,5 +252,18 @@ public void createSshCredentialUsingDefaultSshOnMasterInUserStore() throws IOExc
...
     }
+    
+    @Test
+    public void createPersonalSSHKey() throws IOException, UnirestException {
+        User user = login();
+
+        Map resp = new RequestBuilder(baseUrl)
+                .status(200)
+                .jwtToken(getJwtToken(j.jenkins,user.getId(), user.getId()))
+                .get("/organizations/jenkins/credentials/user/personalSSHKey").build(Map.class);
```
jtnord: should surface test multiple organisations to prevent regressions

kzantow: @jtnord is there a way to do that in tests here? just modify the url?

jtnord: @kzantow see the approach taken in #1252

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/SSHCredentialManager.java
```diff
@@ -0,0 +1,188 @@
...
+import javax.annotation.Nonnull;
+import jenkins.model.Jenkins;
+
+/**
+ * Helper to deal with managing a Jenkins-created per-user SSH key
+ * @author kzantow
+ */
+public class SSHCredentialManager {
+    public static final int KEY_SIZE = 2048;
+    public static final String BLUEOCEAN_GENERATED_SSH_KEY_ID = "jenkins-generated-ssh-key";
```
kzantow: nope, hidden.

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/SSHCredentialManager.java
```diff
@@ -0,0 +1,188 @@
...
+import java.util.List;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import jenkins.model.Jenkins;
+
+/**
+ * Helper to deal with managing a Jenkins-created per-user SSH key
+ * @author kzantow
+ */
+public class SSHCredentialManager {
```
kzantow: `UserSSHKeyManager` (renamed) is used by a different plugin already, can't mark it as `@Restricted(NoExternalUse)`... adding a `reset` method - done.

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/SSHCredentialManager.java
```diff
@@ -0,0 +1,188 @@
...
+    public static final BlueOceanDomainSpecification BLUEOCEAN_DOMAIN_SPECIFICATION = new BlueOceanDomainSpecification();
+    public static final BlueOceanDomainRequirement BLUEOCEAN_DOMAIN_REQUIREMENT = new BlueOceanDomainRequirement();
+    
+    /**
+     * Gets the existing generated SSH key for the user or creates one and
+     * returns it in the user's credential store
+     * @param user owner of the key
+     * @return the user's personal private key
+     */
+    public BasicSSHUserPrivateKey getOrCreatePersonalCredential(@Nonnull User user) {
```
kzantow: done.

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/SSHCredentialManager.java
```diff
@@ -0,0 +1,188 @@
...
+        List<BasicSSHUserPrivateKey> credentials = CredentialsProvider.lookupCredentials(BasicSSHUserPrivateKey.class, Jenkins.getInstance().getItemGroup(), user.impersonate(), BLUEOCEAN_DOMAIN_REQUIREMENT);
```
kzantow: done.

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/SSHCredentialManager.java
kzantow: done.

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/SSHCredentialManager.java
```diff
@@ -0,0 +1,188 @@
...
+        return ((userId == null ? "user" : userId) + "@"
+            + Jenkins.getInstance().getDisplayName())
+                .replaceAll("[^@._a-zA-Z0-9]", "");
+    }
+    
+    /**
+     * Generates a new RSA key with specified keySize
+     * @return a public/private key pair
+     */
+    private static KeyPair generateRSAKey(int keySize) {
```
kzantow: done, though hackish test for it in a different plugin :-/

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/SSHCredentialManager.java
kzantow: belay that... i added the annotation and it seemed to be fine, checking and will keep it if it builds fine

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/SSHCredentialManager.java
kzantow: done.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/CredentialApi.java
kzantow: done. but named `userkey` with a corresponding DELETE for resetting

i386:
@kzantow did you forget to push? ;)

kzantow:
@i386 no, ran into an issue with the tests... doesn't seem to persist credentials the same way it does when running; hoping @vivek can shed some light on that

kzantow:
@i386 also, after actually looking at the url again, renamed the method `publickey` as you requested, made the most sense

kzantow:
e.g. http://localhost:8080/jenkins/blue/rest/organizations/jenkins/credentials/user/publickey

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/UserSSHKeyManager.java
```diff
@@ -0,0 +1,169 @@
...
+        // if none found, create one
+        BasicSSHUserPrivateKey key;
+        try {
+            // create one!
+            KeyPair keyPair = SSHKeyUtils.generateRSAKey(KEY_SIZE);
+            RSAPrivateKey privateKey = (RSAPrivateKey)keyPair.getPrivate();
+            String id_rsa = Base64.encode(privateKey.getEncoded());
+            BasicSSHUserPrivateKey.DirectEntryPrivateKeySource keySource = new BasicSSHUserPrivateKey.DirectEntryPrivateKeySource(id_rsa);
+            key = new BasicSSHUserPrivateKey(CredentialsScope.USER, BLUEOCEAN_GENERATED_SSH_KEY_ID, user.getId(), keySource, null, BLUEOCEAN_GENERATED_SSH_KEY_ID);
+            store.addCredentials(BLUEOCEAN_DOMAIN, key);
```
vivek: Where is domain BLUEOCEAN_DOMAIN getting created? Look at [CredentialApi.create()](https://github.com/jenkinsci/blueocean-plugin/blob/a986f24023c4192f94ce6fc5b28b2f4b1e7b1626/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/CredentialApi.java#L55). but we should not create these APIs when we already have one.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/CredentialApi.java
```diff
@@ -88,6 +94,32 @@ public CreateResponse create(@JsonBody JSONObject body, StaplerRequest request)
         throw new ServiceException.UnexpectedErrorException("Unexpected error, failed to create credential");
     }
 
+    @GET
+    @WebMethod(name="publickey")
+    public void getPersonalKey(StaplerResponse rsp) throws IOException {
```
kzantow: @vivek This is a different API, it automatically generates a special private key for the user that only Jenkins has access to.

blueocean-pipeline-scm-api/src/main/java/io/jenkins/blueocean/credential/UserSSHKeyManager.java
```diff
@@ -0,0 +1,169 @@
...
+    private static final String BLUEOCEAN_DOMAIN_NAME = "blueocean-git-domain";
+    private static final Domain BLUEOCEAN_DOMAIN = new Domain(BLUEOCEAN_DOMAIN_NAME, null, null);
+    
+    /**
+     * Gets the existing generated SSH key for the user or creates one and
+     * returns it in the user's credential store
+     * @param user owner of the key
+     * @return the user's personal private key
+     */
+    public static @Nonnull BasicSSHUserPrivateKey getOrCreate(@Nonnull User user) {
```
kzantow: @vivek I do *not* want to push this responsibility to the client

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/credential/CredentialApi.java
```diff
@@ -88,6 +94,32 @@ public CreateResponse create(@JsonBody JSONObject body, StaplerRequest request)
...
+        String publicKey = UserSSHKeyManager.getReadablePublicKey(authenticatedUser, 
+            UserSSHKeyManager.getOrCreate(authenticatedUser));
+        
+        rsp.setStatus(200);
+        rsp.getWriter().print(JsonConverter.toJson(ImmutableMap.of("key", publicKey)));
+    }
+
+    @DELETE
+    @WebMethod(name="publickey")
+    public void resetPersonalKey(StaplerResponse rsp) throws IOException {
```
kzantow: Again, this is a special key and I'd rather have it with a special API @vivek

vivek:
@kzantow Ignore my comments made regarding using existing API. I misunderstood its purpose. However I still think its placed at wrong place. This key is user specific and we should have these keys API as part of user api. Please see BlueUser and UserImpl.

kzantow:
@vivek I'm not sure that this is best suited off the user object, this still being a specific credential API, though definitely not a bad suggestion. Also, you hit the nail why the tests were failing, strange it worked from a running / restarted instance: no saving the domain object...

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/util/UserSSHKeyManager.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+    
+    private static Domain getDomain(CredentialsStore store) {
+        Domain domain = store.getDomainByName(BLUEOCEAN_DOMAIN_NAME);
+        if (domain == null) {
+            try {
+                //create new one
+                boolean result = store.addDomain(new Domain(BLUEOCEAN_DOMAIN_NAME, null, null));
+                if (!result) {
+                    throw new RuntimeException(String.format("Failed to create credential domain: %s", BLUEOCEAN_DOMAIN_NAME));
```
vivek: better to throw `UnexpectedErrorException`. It gets proper json response rendering.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/util/UserSSHKeyManager.java
```diff
@@ -0,0 +1,214 @@
...
+                }
+                domain = store.getDomainByName(BLUEOCEAN_DOMAIN_NAME);
+                if (domain == null) {
+                    throw new RuntimeException(String.format("Domain %s created but not found", BLUEOCEAN_DOMAIN_NAME));
```
vivek: better to throw `UnexpectedErrorException` for proper json rendering.

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/UserSSHKeyTest.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+    
+    @Test
+    public void createPersonalSSHKey() throws IOException, UnirestException {
+        User user = login();
+
+        Map resp = new RequestBuilder(baseUrl)
+                .status(200)
+                .jwtToken(getJwtToken(j.jenkins, user.getId(), user.getId()))
+                .get("/organizations/jenkins/user/publickey").build(Map.class);
```
vivek: Does it work? I would expect it to be `/organizations/jenkins/users/{user.getId()}/publickey`.

vivek: ah, never mind both will work, there is `BlueOrganization.getUser()` and `BlueOrganization.getUsers()`:)

vivek: Maybe add a test with `/organizations/jenkins/users/{user.getId()}/publickey`  as well, up to you. /organizations/user API is more as a workaround to find logged in user, we might stop using from client once we have JWT enabled as token will tell who the logged in user is.

vivek:
Strange looking test failure https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/PR-1262/7/tests. I replayed it, lets see if it was a flaky one.

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/UserSSHKeyTest.java
kzantow: This is something the user can only get their own public key, not someone else's. Maybe we can add that in the future, but I don't think it needs to be part of this PR

kzantow:
@vivek I made a different change that should fix the tests

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/UserSSHKeyTest.java
kzantow: Oh, misread what you wrote at first... right... the `/organizations/jenkins/users/{user.getId()}/publickey` will NOT currently work. It uses the logged in user.

vivek: @kzantow Fact that `/organizations/jenkins/users/{user.getId()}/publickey`, better to authorize if logged in user UserImpl.user.getId()  is same as User.current.getId(), if not same then throw `UnauthorizedException`. Something like below for both GET and DELETE implementation. I think this can pose as security hole 🐛 

```
User authenticatedUser =  User.current();
if(authenticatedUser == null) {
    throw new ServiceException.UnauthorizedException("No authenticated user found");
}
if(!user.getId.equals(authenticatedUser.getId()){
    throw new ServiceException.ForbiddenException("Not authorized");
}
```

vivek: You should also add tests to ensure Bob can't see Alice's key.

kzantow: @vivek done

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/UserImpl.java
```diff
@@ -6,6 +6,8 @@
 import hudson.model.User;
 import hudson.tasks.Mailer;
 import hudson.tasks.UserAvatarResolver;
+import hudson.util.HttpResponses;
```
kzantow: @vivek this was the key... *not* the stapler version, but the `hudson.util` version has all those JSON utility methods...

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/UserImpl.java
```diff
@@ -143,6 +151,43 @@ public Link getLink() {
...
+     * Deletes the user's private Jenkins-managed key
+     * @return
+     * @throws IOException 
+     */
+    @DELETE
+    @WebMethod(name="publickey")
+    public HttpResponse resetPublicKey() throws IOException {
+        User authenticatedUser =  User.current();
+        if(authenticatedUser == null || !StringUtils.equals(getId(), authenticatedUser.getId())) {
+            throw new ServiceException.UnauthorizedException("Not authorized");
```
vivek: Hmm, I think it should be split to throw UnauthorizedException if authenticatedUser is null, if user ids do not match then Forbidden. Basically, Unauthorized or 401, which tells client to ask user to login/authenticate as authentication info was not found. Forbidden is 403, which is to tell client its unauthorized access and user is not allowed to access the resource.

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/UserSSHKeyTest.java
vivek: Looks great. Another nit pick about throwing correct status code in case of error.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/UserImpl.java
vivek: @kzantow it extends from `org.kohsuke.stapler.HttpResponses` so it could return this type as well in method signature. 

You may like to add method signature to BlueUser with stapler one (as we do not add jenkins model classes in blueocean-rest module. This could come in later PR though.

kzantow: Right, I'm a little skeptical about that; the BlueUser seems like a data holder, I'm not a fan of mixing service methods with data classes. But this is what we have today, so; I'll defer this until it's needed for some reason, if that's ok.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/UserImpl.java
kzantow: done

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/UserSSHKeyTest.java
kzantow: right, added a check for that, too

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/CredentialApiTest.java
kzantow: @jtnord i see, this needs to be genericized at the test base class level, I think and applied here, as well as other places

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/UserImpl.java
vivek: `BlueUser` is a rest resource (extends `Resource`), so yes it exposes data as well as could expose other resources. This is fine we can add later, for this PR purpose its fine.

vivek:
@kzantow Looks good 🐝 merge ahead after successful CI run.

kzantow:
@vivek thanks for the feedback / review / etc.! :)

i386:
woohoo!

PR #1265, opened 2017-07-20
[JENKINS-45703] 1.1 branch - Check the organization from the url and the factory instead of jenkins root

alvarolobato:
# Description

See [JENKINS-45703](https://issues.jenkins-ci.org/browse/JENKINS-45703).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

cc. @vivek

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/preload/OrganizationStatePreloader.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/preload/OrganizationStatePreloader.java
@@ -67,7 +67,7 @@ public class OrganizationStatePreloader extends PageStatePreloader {
         }
     }
 
-    private static final Pattern pattern = Pattern.compile("\\/organizations\\/([^\\/]*)\\/");
+    private static final Pattern pattern = Pattern.compile("/blue/organizations/([^/]*)/");
 
     private String getOrganizationFromURL() {
         StaplerRequest currentRequest = Stapler.getCurrentRequest();
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/preload/OrganizationStatePreloader.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/preload/OrganizationStatePreloader.java
@@ -1,21 +1,28 @@
 package io.jenkins.blueocean.preload;
 
+import java.io.StringWriter;
+import java.util.Iterator;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import javax.annotation.Nonnull;
+
+import org.kohsuke.stapler.Stapler;
+import org.kohsuke.stapler.StaplerRequest;
+
 import hudson.Extension;
 import io.jenkins.blueocean.commons.PageStatePreloader;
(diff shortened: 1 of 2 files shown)


alvarolobato:
Ups I have one, it didn't go in the push, let me fix that.

michaelneale:
@alvarolobato is this on master too? 1.1 is frozen right now so this probably wont' make it in until next week some time.

alvarolobato:
@vivek  added the test. I ended up writing a better new one and changing some of the code.

@michaelneale yes, it should be applied to master too, but I'm sending it to 1.1 first as is the one we are testing right now.

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/preload/BlueOceanWebURLBuilderOrgOnFolderTest.java
```diff
@@ -0,0 +1,296 @@
...
+        request = mock(StaplerRequest.class);
+
+        mockStatic(OrganizationFactory.class);
+        when(OrganizationFactory.getInstance()).thenReturn(orgFactory);
+
+        mockStatic(Stapler.class);
+        when(Stapler.getCurrentRequest()).thenReturn(request);
+    }
+
+    //private static final String MATCHER_ORG = "/setState.*'organization'.*\"name\".*\"%s\"/g"; //Matches setState('organization', {"name":"ORG_NAME"
```
vivek: Remove the commented code 🐛

vivek:
@alvarolobato code looks good, minor issue with leftover commented code.

vivek:
@alvarolobato thanks for the tests, looks great!

alvarolobato:
@vivek comment addressed

michaelneale:
@vivek ok - when this passes this looks ok? If we approve we can probably cherry pick this to master along with https://github.com/jenkinsci/blueocean-plugin/pull/1252 at some point

michaelneale:
nice @vivek can you check this looks ok to you?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/preload/OrganizationStatePreloader.java
```diff
@@ -44,4 +66,26 @@ public String getStateJson() {
             return "{}";// if will happen only when there is no implementation of BlueOrganization found.
         }
     }
+
+    private static final Pattern pattern = Pattern.compile("\\/organizations\\/([^\\/]*)\\/");
```
jtnord: this pattern looks wrong - the `/` is not special (unless I missed something) so does not need escaping with `\`?

jtnord: should this not match `/blue/organisations/([^/]+)/`  otherwise if someone deployed their jenkins to "orgainizations" in the context it would fail (just narrows the window of things going wrong, does not completely eliminate it.

alvarolobato: @jtnord addressing both comments.

michaelneale:
has been included in the rollup

PR #1269, opened 2017-07-21
Update Dockerfile

ndeloof:
jenkinsci/jenkins is deprecated


michaelneale:
@rtyler can I get an amen?

michaelneale:
awaiting @rtyler to bless this

PR #1274, opened 2017-07-25
JENKINS-45589 check for null when using OrganizationFactory

i386:
# Description

It's possible not to get back a valid `BlueOrganization` when using methods on `OrganizationFactory`. There still feels like there is a bug somewhere here but without paths for the items that we are resolving there is no way to tell.

This should add some better exception messages and setup expectations that `OrganizationFactory` may not work in some circumstances.

See [JENKINS-45589](https://issues.jenkins-ci.org/browse/JENKINS-45589).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineImpl.java
@@ -6,6 +6,8 @@ import hudson.model.Job;
 import io.jenkins.blueocean.rest.Reachable;
 import io.jenkins.blueocean.rest.annotation.Capability;
 import io.jenkins.blueocean.rest.factory.BluePipelineFactory;
+import io.jenkins.blueocean.rest.factory.organization.OrganizationFactory;
+import io.jenkins.blueocean.rest.model.BlueOrganization;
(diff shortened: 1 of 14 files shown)


i386:
Expecting we may get some findbugs violations

i386:
@vivek you may want to take a look at this again

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineImpl.java
```diff
@@ -18,17 +20,18 @@
...
     }
 
     @Extension(ordinal = 1)
     public static class PipelineFactoryImpl extends BluePipelineFactory {
 
         @Override
         public BluePipeline getPipeline(Item item, Reachable parent) {
-            if (item instanceof WorkflowJob) {
-                return new PipelineImpl((Job) item);
+            BlueOrganization org = OrganizationFactory.getInstance().getContainingOrg(item);
```
i386: We should never be doing this lookup inside of the CTOR for the Pipeline Impls...

vivek: Yeah, better to pass them thru constructor param.

michaelneale:
:bee:

PR #1275, opened 2017-07-25
JENKINS-45099 escape step description so that it is ECMA script safe

i386:
# Description

See [JENKINS-45099](https://issues.jenkins-ci.org/browse/JENKINS-45099).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-commons/pom.xml
+++ b/blueocean-commons/pom.xml
@@ -20,6 +20,11 @@
       <artifactId>jackson-databind</artifactId>
       <version>${jackson.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <version>3.5</version>
+    </dependency>
   </dependencies>
 
   <build>
--- a/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/JSON.java
+++ b/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/JSON.java
@@ -1,6 +1,7 @@
 package io.jenkins.blueocean.commons;
 
 import com.google.common.base.CharMatcher;
+import org.apache.commons.lang3.StringEscapeUtils;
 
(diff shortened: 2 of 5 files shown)


sophistifunk:
I can't see a reason for this to be yellow, so I've restarted it.

michaelneale:
nah failure looks legit: 

```
BlockStageStepsWithDesc – io.jenkins.blueocean.rest.impl.pipeline.PipelineNodeTest
Error
expected:<echo ["Building]"> but was:<echo [\"Building\]">
```

So @i386, is there some unfortunate side effect of this or is that escaping correct now and the test should be adjusted? As it looks funny...

sophistifunk:
I didn't mean that I found the error and thought it was bogus, I just couldn't for the life of me find the error in the logs :(

sophistifunk:
We get a "whole pipeline is yellow" alarm and then every individual step is green :(

sophistifunk:
I'm not sure I understand the original problem well enough- why would escape characters in the log break UI stuff? We have them in ours and while it's not being shown in colour the way we'd like, it doesn't break.

i386:
@michaelneale thinking about this fix - its not at the right level. We should be making sure stapler encodes JSON properly (hint: It doesn't).

michaelneale:
@i386 agree - but that may be a battle. Should we open a ticket anyway for it?

sophistifunk:
Sounds like fun. Surely we don't have code that relies on Stapler being able to generate bad JSON?

michaelneale:
@sophistifunk stapler says "hold my beer..."

i386:
@michaelneale I've submitted a PR for this in Stapler https://github.com/stapler/stapler/pull/122. When its approved, we can back port it to our own copy of stapler in master and close the ticket out.

michaelneale:
@i386 nice! that will be great and I bet help a lot of people.

PR #1277, opened 2017-07-26
Handle skipped stages

sophistifunk:
# Description

See [JENKINS-39628](https://issues.jenkins-ci.org/browse/JENKINS-39628).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


sophistifunk:
screen shot 2017-07-26 at 8 33 15 am

sophistifunk:
screen shot 2017-07-24 at 7 25 07 pm

i386:
giphy-1

i386:
@brody would you mind reviewing? Do you still feel strongly about these rules? (from the JIRA ticket)
- If the first stage is skipped then the pipeline should start from a "Start" node.
- If the last stage is skipped then the pipeline should end with a "End" node.

sophistifunk:
I didn't see those conditions, because hulk read güt. But it's no big change (now) either way :D

sophistifunk:
I will fight against the flat-bottomed single-node-skip curve tho :D

image

sophistifunk:
Not sure if relative unpredictability of wether or not users will see start/end nodes may be an issue tho, if they don't intuit the logic of it. Also, if nobody's skipping first/last nodes there's a chance we might break that functionality in the future and not notice ;-)

sophistifunk:
ATH failure doesn't *seem* related, so re-running.

brody:
@i386 The alternative is to randomly have them start/stop, which doesn't feel right. So yeah, start/end nodes give you more context around whats happening so lets stick with that. 👍

sophistifunk:
New ATH issue might be valid, investigating :)

sophistifunk:
I'd like to get @cliffmeyers and/or @kzantow to look at the layout code changes for this, because after working with the editor they'd have a perspective I'm missing.

michaelneale:
nice one josh - code looks lovely!

sophistifunk:
Updated with a little cleanup, and shortening the connections for start/end. Got a strange non-error from npm on core-js publish so ATH can check if everything went OK

michaelneale:
:bee: from me - unless @cliffmeyers disagrees overnight get it in.

acceptance-tests/src/test/java/io/blueocean/ath/offline/multibranch/ParallelNavigationTest.java
```diff
@@ -89,7 +89,7 @@ public void parallelNavigationTestInput () throws IOException, GitAPIException,
         wait.until(By.cssSelector(".btn.inputStepSubmit")).click();
 
         // and clicking on the unselected node will yield us the second branch
-        wait.until(By.cssSelector(".pipeline-node")).click();
+        wait.until(By.xpath("//*[contains(@class, 'pipeline-node')][3]")).click();
```
cliffmeyers: I feel like `.pipeline-node:nth-child(3)` reads a bit better, but probably just personal taste :)

blueocean-dashboard/src/main/js/components/Activity.jsx
```diff
@@ -107,18 +107,18 @@ export class Activity extends Component {
...
-        const runButton = !isMultiBranchPipeline && (
-            <RunButton
-                buttonType="run-only"
-                innerButtonClasses="btn-secondary"
-                runnable={pipeline}
-                latestRun={latestRun}
-                onNavigation={onNavigation}
-            />
-        );
+        const runButton = isMultiBranchPipeline ? null : (
```
cliffmeyers: Was this change actually fixing an issue? I assumed a false expression doesn't render anything at all. Just curious.

jenkins-design-language/src/js/components/status/StatusIndicator.jsx
```diff
@@ -6,15 +6,17 @@ import SvgStatus from './SvgStatus';
 
 const validResultValues = {
     // Mostly from io/jenkins/blueocean/rest/model/BlueRun.java
+    // TODO: Do we want to move this? I have NFI
```
cliffmeyers: It would be great if this was enum defined in blue ocean somewhere, but we don't have a great way of sharing code from there down to the JDL. I think we should leave this TODO and revisit after XP 2.0 lands.

acceptance-tests/src/test/java/io/blueocean/ath/offline/multibranch/ParallelNavigationTest.java
sophistifunk: I'm still hoping that a non-awful solution to this sort of fragility will occur to me like some sort of Newtonian apple-falling moment.

blueocean-dashboard/src/main/js/components/Activity.jsx
sophistifunk: No I was just rearranging it so it wasn't checking for (NOT foo AND) because it seemed awkward.

jenkins-design-language/src/js/components/status/StatusIndicator.jsx
sophistifunk: Yeah I put that comment in when I was planning to move the component, but I decided against it because I didn't want to give up Flow. I'll remove it.

michaelneale:
:bee: - just keep in mind how up to date this may or may not be with master.

PR #1280, opened 2017-07-27
[JENKINS-27395] Per-stage test results

abayer:
# Description

See [JENKINS-27395](https://issues.jenkins-ci.org/browse/JENKINS-27395).

Downstream of https://github.com/jenkinsci/junit-plugin/pull/76.

Very preliminary work here - nothing on visualization, obviously. More tests in `blueocean-pipeline-api-impl` needed. Everything could change radically depending on what happens with https://github.com/jenkinsci/junit-plugin/pull/76 going forward. And we probably need to be storing info on block-scoped steps somehow so that we can get any tests they may have recorded as well. Buuuuuut hey, it's a start?

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

cc @reviewbybees esp @svanoort


abayer:
So latest change here is that `BlueJUnitTestResult#getName` is now taking advantage of the latest changes to `CaseResult#getDisplayName` - that, plus the `junit` step changes, now results in `stage / branch / test name` being returned. So...this may actually bare-bones work now on top of the junit PR...

abayer:
FYI, unit tests are all fine, ATH seems to be flaky, but I could be wrong.

kzantow:
Could this have conflicts resolved so I can review?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
```diff
@@ -169,4 +194,35 @@ FlowNodeWrapper getFlowNodeWrapper(){
         return node;
     }
 
+    @Extension
+    public static class NodeFactoryImpl extends BlueTestResultFactory {
```
i386: This class looks specific to JUnit - perhaps it should move into `BlueJUnitTestResult`?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
```diff
@@ -169,4 +194,35 @@ FlowNodeWrapper getFlowNodeWrapper(){
...
+        @Override
+        public Result getBlueTestResults(Run<?, ?> run, final Reachable parent) {
+            Iterable<BlueTestResult> results = null;
+            TestResultAction action = run.getAction(TestResultAction.class);
+            if (action != null && parent instanceof BluePipelineNode) {
```
i386: Other plugins are going to want to know how to do this kind of processing - why not encode this in the `BlueTestResultFactory`? 

The intent might be easier if we added something like: 
`public Result getBlueTestResults(Run<?,?> run, BluePipelineNode node, final Reachable parent)`

Then the code in this factory can move to that method on `io.jenkins.blueocean.service.embedded.rest.junit.BlueJUnitTestResult.FactoryImpl`

i386: Something like https://github.com/abayer/blueocean-plugin/pull/1

michaelneale:
@abayer what state is this one in? reviewable? feature semi-complete?

abayer:
Reviewable for sure.

michaelneale:
ok, given it is mostly backend - @vivek mind casting eye over?

pom.xml
```diff
@@ -693,7 +693,13 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>junit</artifactId>
-            <version>1.21</version>
+            <version>1.22-beta-2-20171019.094026-1</version> <!-- TODO: Switch to 1.22 post-beta -->
```
michaelneale: is this out of beta now?

abayer: No, not yet.

abayer:
ping - the junit changes are very close to actually releasing, so I'd like to make sure this is ready to go soon after.

michaelneale:
@vivek able to do a code review? 

Thanks Andrew 🎉

i386:
@abayer I tried this PR but the described UI (JENKINS-46166) doesn't seem to be done? FWIW, its all data driven and there isn't any frontend work to do.

imeredith:
Looking forward to this for bluecoean ci job!

abayer:
@i386 I thought it was? If it's not, then it probably needs someone who understands the Blue Ocean side of things better than myself to finish it off.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
```diff
@@ -72,6 +89,16 @@ public String getDisplayDescription() {
     }
 
     @Override
+    public BlueTestResultContainer getTests() {
+        return new BlueTestResultContainerImpl(this, run);
+    }
+
+    @Override
+    public BlueTestSummary getTestSummary() {
```
vivek: This can result in to performance issues. test results are read off disk. So far flow graph returned thru nodes API is pretty light weight, embedding test summary could be expensive for large number of nodes. 

I would like @svanoort to weigh in here too, specially keeping in mind perf issues discussed at: https://github.com/jenkinsci/blueocean-plugin/pull/1517#issuecomment-341268422.

michaelneale:
What is best way to test it? Archiving same results 2x may not be right way.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BluePipelineNode.java
```diff
@@ -69,6 +70,18 @@
...
+     * @return Gives tests in this node.
+     */
+    @Navigable
+    public abstract BlueTestResultContainer getTests();
+
+    /**
+     * @return Gives the test summary for this node.
+     */
+    @Exported(name = TEST_SUMMARY, inline = true, skipNull = true)
+    public abstract BlueTestSummary getTestSummary();
```
vivek: See my comments at implementation. Returning test summary as part of each node could be expensive.

abayer:
@michaelneale Well, that'd actually be a useful test, since the ability to run the same named set of tests in different stages/parallel branches and have them recorded distinctly is the driving reason for this whole thing.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
abayer: I'm going to defer to you and @svanoort on performance matters.

abayer: If I'm not mistaken, we're only reading from disk once - once the action's in memory, we shouldn't be reading from disk again, should we?

abayer:
Ah, the always fun random creation errors...rebuilding.

i386:
@abayer think you could post a few screenshots here?

abayer:
I have one screenshot handy: 
Test results

i386:
@abayer what browser do you use? Shouldn't see the Blue bar (not a problem for this PR).

visually the tests LGTM 💯  👍

abayer:
@i386 That was Chrome, and may be my screenshot plugin being crappy. =)

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BluePipelineNode.java
i386: I don't think this is necessary for how it is implemented today so this method could be removed before merging. WDYT @abayer @vivek ?

abayer: I honestly can't remember what the reason for this was. =)

i386:
Left a minor nit https://github.com/jenkinsci/blueocean-plugin/pull/1280#discussion_r149784727

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BluePipelineNode.java
vivek: 👍 yeah better to remove it from here.

abayer:
I'm getting rid of the summary - but we definitely want the ability to get the tests per-node, even if we don't yet have a UI usage of that.

abayer:
i.e., we could defer the node-level test stuff, but we're going to want to come back to it eventually when we have a UI design for "show me the tests for THIS_STAGE", etc. So if there's an inherent problem with that, we should address it now.

svanoort:
@abayer My suggestion is to attach the tests to the run API response object, but have the UI take that mapping of test result to nodes and use it to decorate the appropriate node using the ID/stage.  It should have the ID info it needs to know where to use that information. 

Attaching that information on the nodes themselves makes them kind of bloated and doesn't really fit with the way the data is stored internally. 

It'll LOOK to the user like the test result is on the node, but the only we'll be the wiser. ;)

abayer:
@svanoort Ah, ok. Let me see if I can wrap my head around that and figure out how to implement it.

abayer:
...ok, I have no idea how to implement that. 

I guess I should remove the test stuff from nodes completely and just let the node/test relationship logic be determined down the road?

abayer:
@svanoort @vivek Removed the per-node test stuff. I'll work on a separate PR on the side for that in the future.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BluePipelineNode.java
```diff
@@ -7,6 +7,7 @@
 
 import java.util.List;
 
+import static io.jenkins.blueocean.rest.model.BlueRun.TEST_SUMMARY;
```
vivek: Remove unused import.

blueocean-rest-impl/pom.xml
```diff
@@ -130,6 +130,16 @@
             <artifactId>workflow-basic-steps</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
```
vivek: Are these dependencies still needed after removing node/test related code?

abayer: Probably not. Removing.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BluePipelineNode.java
abayer: Will do.

michaelneale:
@i386 able to manually review this for functionality? to bless it?

i386:
It looks like this https://github.com/i386/app-store-demo/blob/jw/boothdemo1/Jenkinsfile will cause the parallel names to be doubled up like:

screencapture-localhost-8080-jenkins-blue-organizations-jenkins-app-store-demo-detail-jw-2fboothdemo1-10-tests-1511397565183

abayer:
@i386 ah, yeah, we should probably ignore branch names with one identically named stage within them. That needs to be addressed over in junit, though, not here.

michaelneale:
I don’t get the doubling up - can someone explain it to me?

i386:
@michaelneale it repeats the parallel name "Browser Tests / Chrome / Chrome" rather than "Browser Tests / Chrome"

michaelneale:
The test results aren’t doubled up but the “safari / safari” is right? What is that other coming from?

i386:
@michaelneale I had the same tests run in each parallel - this is what its supposed to look like - however there is an erroneous addition of the stage to the test name, so the parallel stage appears twice in the name of the test.

michaelneale:
@i386 still confused. Will have to talk this one over. This PR needs to get in, been dragging on

i386:
@michaelneale 

```
stage('Browser Tests') {
      parallel {
        stage('Firefox') {
          steps {
            sh 'echo \'setting up selenium environment\''
            sh 'ping -c 5 localhost'
            sh 'mvn test'
          }
          post {
            always {
              junit '**/target/surefire-reports/TEST-*.xml'
            }
          }
        }
}
```

Then:
image 2017-11-23 12-17-12

abayer:
I’ll open a JIRA against junit for this and get a PR up. Declarative has an actual `stage` named “Chrome” inside a `parallel` branch also named `Chrome` for its parallel stages syntax behind the scenes, and junit needs to discard one of those names in that scenario.

i386:
OK, I think this one is good to go from my PoV. Will still need the follow up on JUnit :)

michaelneale:
@abayer ah thanks - that explains it. Sorry I am really slow today... couldn't make sense of it, but it makes sense. 

In that case, if there is a PR/Ticket for the junit plugin to remove the duplication, I think this is :bee: from me - @svanoort had some feedback, I know viveks was addressed, but @svanoort ?

abayer:
@i386 @michaelneale https://issues.jenkins-ci.org/browse/JENKINS-48196 opened for the `junit` redundant branch/stage names, fyi.

michaelneale:
thanks @abayer very nice

abayer:
FYI, JENKINS-48196 (the "foo / foo" duplicate naming) is fixed in junit 1.23, which I released today. And I have updated this PR to use junit 1.23. So tada?

i386:
@abayer merge-away!

abayer:
@i386 Mind merging for me when the build finishes? I'm heading offline a little early today due to head cold. =)

i386:
@abayer sure thing!

i386:
So great to have this one done @abayer - great job 💯

michaelneale:
yes thanks again, it was a long road, but this will help people out with clarity.

PR #1290, opened 2017-08-01
JENKINS-45920 Backend for JIRA integration

i386:
# Description

When a JIRA issue is quoted in a comment message it should link to that ticket on JIRA.

Adds the following structure to every runs changesets:

```
{
  "changeSet": [
    {
      "_class": "io.jenkins.blueocean.service.embedded.rest.ChangeSetResource",
      "affectedPaths": [
        "bug/asdsada"
      ],
      "author": {
        "_class": "io.jenkins.blueocean.service.embedded.rest.UserImpl",
        "avatar": null,
        "email": "noreply@github.com",
        "fullName": "noreply",
        "id": "noreply",
        "permission": null
      },
      "commitId": "1e96c711d6e57f763e99eeb5b0843c7cf1585965",
      "issues": [
        {
          "_class": "io.jenkins.blueocean.service.embedded.jira.BlueJiraIssue",
          "id": "JENKINS-35808",
          "url": "https://issues.jenkins-ci.org/browse/JENKINS-35808"
        }
      ],
      "msg": "JENKINS-35808 hello world",
      "timestamp": "2017-08-02T08:53:45.000+1000",
      "url": "https://github.com/i386/hellonode/commit/1e96c711d6e57f763e99eeb5b0843c7cf1585965"
    }
  ]
```

See [JENKINS-45920](https://issues.jenkins-ci.org/browse/JENKINS-45920).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/jira/JiraSCMListener.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/jira/JiraSCMListener.java
@@ -1,0 +1,71 @@
+package io.jenkins.blueocean.service.embedded.jira;
+
+import com.atlassian.jira.rest.client.api.domain.Issue;
+import com.google.common.base.Function;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Lists;
+import com.google.common.collect.Sets;
+import hudson.Extension;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.model.listeners.SCMListener;
+import hudson.plugins.jira.JiraBuildAction;
+import hudson.plugins.jira.JiraSession;
+import hudson.plugins.jira.JiraSite;
+import hudson.plugins.jira.model.JiraIssue;
+import hudson.scm.ChangeLogSet;
+import hudson.scm.SCM;
+
+import java.util.List;
+import java.util.Set;
+
+@Extension
+public class JiraSCMListener extends SCMListener {
+    @Override
+    public void onChangeLogParsed(Run<?, ?> run, SCM scm, TaskListener listener, ChangeLogSet<?> changelog) throws Exception {
+        JiraSite jiraSite = JiraSite.get(run.getParent());
+        if (jiraSite == null) {
+            return;
+        }
+        List<String> issueKeys = Lists.newArrayList();
+        for (ChangeLogSet.Entry entry : changelog) {
+            issueKeys.addAll(BlueJiraIssue.findIssueKeys(entry.getMsg(), jiraSite.getIssuePattern()));
+        }
+        String jql = constructJQLQuery(issueKeys);
+        JiraSession session = jiraSite.getSession();
+        // Query for JIRA issues
+        Set<JiraIssue> issuesFromJqlSearch = Sets.newHashSet(Iterables.transform(session.getIssuesFromJqlSearch(jql), new Function<Issue, JiraIssue>() {
+            @Override
+            public JiraIssue apply(Issue input) {
+                return new JiraIssue(input);
+            }
+        }));
+        // If there are no JIRA issues, do not update the actions
+        if (issuesFromJqlSearch.isEmpty()) {
+            return;
+        }
+        // Create or update the JiraBuildAction
+        JiraBuildAction action = run.getAction(JiraBuildAction.class);
+        if (action == null) {
+            run.addAction(new JiraBuildAction(run, issuesFromJqlSearch));
+        } else {
+            action.addIssues(issuesFromJqlSearch);
+        }
+        run.save();
+    }
+
+    static String constructJQLQuery(List<String> issueKeys) {
+        StringBuilder jql = new StringBuilder();
+        jql.append("key in (");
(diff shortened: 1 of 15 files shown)


vivek:
@i386 needs more test coverage and there is conflict.

```
[WARNING] Rule violated for bundle blueocean-rest-impl: lines covered ratio is 0.67, but expected minimum is 0.70
[WARNING] Rule violated for bundle blueocean-rest-impl: classes missed count is 13.00, but expected maximum is 0.00
```

pom.xml
```diff
@@ -440,6 +441,51 @@
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>jira</artifactId>
+            <version>2.3</version>
+            <exclusions>
```
vivek: why was it needed to explicitly depend on jura-plugin dependencies?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/jira/JiraSCMListener.java
```diff
@@ -0,0 +1,74 @@
...
+public class JiraSCMListener extends SCMListener {
+    @Override
+    public void onChangeLogParsed(Run<?, ?> run, SCM scm, TaskListener listener, ChangeLogSet<?> changelog) throws Exception {
+        JiraSite jiraSite = JiraSite.get(run.getParent());
+        if (jiraSite == null) {
+            return;
+        }
+        List<String> issueKeys = Lists.newArrayList();
+        for (ChangeLogSet.Entry entry : changelog) {
+            issueKeys.addAll(BlueJiraIssue.findIssueKeys(entry.getMsg(), jiraSite.getIssuePattern()));
```
vivek: @i386 Its possible change set has multiple duplicate JIRA issues, in that case issueKeys will have duplicates? would it mean later on when jql is executed it would return duplicates as well?

i386: Possible. We could change the list to a hashset above to deduplicate?

pom.xml
i386: @vivek JIRA client 3.0.0 actually included `org.apache.httpcomponents` classes inside of it which caused a conflict with the jar from the same package. 4.0.0 does not have this problem. 

There is a fix upstream as part of https://github.com/jenkinsci/jira-plugin/pull/126

i386: @vivek that change was merged. When I get https://github.com/jenkinsci/jira-plugin/pull/127 merged I will ask the maintainers to release a new version.

pom.xml
```diff
@@ -440,6 +441,51 @@
...
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>com.atlassian.jira</groupId>
+            <artifactId>jira-rest-java-client-core</artifactId>
+            <version>${jira.client}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.google.guava</groupId>
+                    <artifactId>guava</artifactId>
```
michaelneale: here be dragons?

pom.xml
```diff
@@ -440,6 +441,17 @@
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>jira</artifactId>
+            <version>2.4</version>
```
i386: @vivek this version straightened out all the dependency issues.

pom.xml
```diff
@@ -500,6 +512,16 @@
             <version>${jackson.version}</version>
         </dependency>
         <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
```
i386: @vivek noticed that we didn't have these in the right spot

pom.xml
i386: seems to work just fine. Needed to do this so it picks our guava rather than Atlassian's version via the jira-rest-client

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/jira/JiraSCMListener.java
i386: Switched to using a LinkedHashSet. BlueJiraIssue has toEquals and hashCode implemented

i386:
@vivek ready for your review.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/jira/JiraSCMListener.java
vivek: Great, thanks.

pom.xml
```diff
@@ -34,6 +34,7 @@
     <jenkins-test-harness.version>2.23</jenkins-test-harness.version>
     <scm-api.version>2.2.0</scm-api.version>
     <git.version>3.4.0</git.version>
+    <jira.client>4.0.0</jira.client>
```
vivek: Unused?

vivek:
@i386 please hold off merging as we have master locked for now.

PR #1297, opened 2017-08-03
[JENKINS-45804] - Resolve user permissions against organization base - Master Branch

alvarolobato:
# Description

See [JENKINS-45804](https://issues.jenkins-ci.org/browse/JENKINS-45804).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
for @i386 to review at some point

alvarolobato:
@vivek @i386  please take a look.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
```diff
@@ -56,7 +57,7 @@ protected FetchData getFetchData(@Nonnull BlueUrlTokenizer blueUrl) {
         User jenkinsUser = User.current();
 
         if (jenkinsUser != null) {
-            UserImpl blueUser = new UserImpl(jenkinsUser);
+            UserImpl blueUser = new UserImpl(OrganizationUtil.getOrganization(OrganizationUtil.getOrganizationNameFromURL(), true), jenkinsUser);
```
i386: @alvarolobato isn't this the same as:
```
OrganizationFactory.getInstance().getContainingOrg(Jenkins.getInstance());
```

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
```diff
@@ -95,7 +96,7 @@ public PullRequest getPullRequest() {
                     cnt++;
                     String id = e.getCommitId();
                     if (id == null) id = String.valueOf(cnt);
-                    m.put(id, new ChangeSetResource(e, this));
+                    m.put(id, new ChangeSetResource(OrganizationUtil.getOrganization(this.getOrganization(), true), e, this));
```
i386: 🐛 Organisation is already in context - just use `this.organization`

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/util/OrganizationUtil.java
```diff
@@ -0,0 +1,87 @@
...
+     * Returns the organization for a given name, defaulting to the first organization if not found or if the
+     * organization name parameter was null
+     * 
+     * @param orgName name of the organization to lookup
+     * @param defaultToFirst if true the result will default to the first organization when the name comes as null or no
+     * organization with that name could be found. When false it will return null if the organization couldn't be found
+     * @return the organization or null if it couldn't be found.
+     */
+    @CheckForNull
+    public static BlueOrganization getOrganization(@CheckForNull String orgName, boolean defaultToFirst) {
```
i386: Delete this method and use the snippet below in the call sites in `OrganizationStatePreloader` and `UserStatePreloader`

```
Objects.firstNonNull(OrganizationFactory.getInstance().getOrg(orgName), Iterables.getFirst(OrganizationFactory.getInstance.list(), null))
```

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
alvarolobato: @i386 if the organization is in a folder that won't work. In fact the whole point of this PR is to support permissions when the organization is not in Jenkins root.

The reason why I think that in the preloaders we should get it from the url is because the whole API supports multiple organizations, given the OrgFactory with the containingOrg and the of methods.

If we don't take into account which org the user is accessing through the url the only solution we have left is accessing to the one on Jenkins root (which in my case the is none) or get the first one, which makes pointless supporting organizations in folders.

I agree with you that we had to remove every place where the url was accessed to get the organization and the Jenkins.getInstance() use, but the case of the preloaders is different because is where the front-end get the information about the org the user is accessing.
WDYT?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
alvarolobato: You are right, this had to be this way in 1.1 branch, not for master. Changing it.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/util/OrganizationUtil.java
alvarolobato: Ok, changing it.

alvarolobato:
@i386 please, see my comment above about the url. I'll change the rest of the thing in my morning.
Let me know if you want to discuss the URL thing in the morning. Thanks

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
i386: Right - because the permissions are at the folder (org) level. That makes sense.

Could `getOrganizationNameFromURL` be moved to `OrganizationImpl` and be marked as `@Restricted`? 

Then delete OrganizationUtils completely

alvarolobato: Sure, I'll do that.

alvarolobato:
Changes done as agreed.

alvarolobato:
Same as https://github.com/jenkinsci/blueocean-plugin/pull/1279

@vivek as agreed I've removed all the places where the url is used to get the organization (preloaders) and defaulted to the first one.
We will change it once BO team has decided a better approach.

cc. @i386

alvarolobato:
Rebased to master. Good to merge again

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ChangeSetResource.java
```diff
@@ -32,15 +35,19 @@
     private final ChangeLogSet.Entry changeSet;
     private final Reachable parent;
 
-    public ChangeSetResource(Entry changeSet, Reachable parent) {
+    @Nullable
+    private final BlueOrganization organization;
+
+    public ChangeSetResource(@Nullable BlueOrganization organization, Entry changeSet, Reachable parent) {
```
vivek: Why is this null?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/UserImpl.java
```diff
@@ -44,15 +56,20 @@
 
     protected final User user;
 
+    private final BlueOrganization organization;
+    private final AccessControlled organizationBase;
+
     private final Reachable parent;
-    public UserImpl(User user, Reachable parent) {
+
+    public UserImpl(BlueOrganization organization, User user, Reachable parent) {
```
vivek: These params should be `@Nonnull`, to help findbugs report any error elsewhere.

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/ChangeSetResourceTest.java
```diff
@@ -30,9 +31,10 @@ public void testChangeSet(){
         when(entry.getCommitId()).thenReturn("12345");
         when(entry.getMsg()).thenReturn("test changeset");
         when(entry.getAffectedPaths()).thenReturn(Collections.singleton("/foo/bar"));
-        ChangeSetResource changeSetResource = new ChangeSetResource(entry,reachable);
+        ChangeSetResource changeSetResource = new ChangeSetResource(null, entry, reachable);
```
vivek: null org?

michaelneale:
@i386 looks like comments addressed. @vivek looks similar to the 1.1 branch - still want changes?

PR #1298, opened 2017-08-03
Translation files for Chinese

meiboyu:
I don't know how to reopen...so i create a new pull request....


michaelneale:
thanks @meiboyu - will probably get this in next week! (yes it was very confusing how the other PR got in that state)

michaelneale:
merged via https://github.com/jenkinsci/blueocean-plugin/pull/1370

PR #1306, opened 2017-08-08
Improvement/jenkins 45743 remove material icons

NicuPascu:
# Description
This PR removes our reliance on the react-material-icons package, it adds the material-ui svg icons in the JDL and creates an "Icon" component in the JDL that is a drop-in replacement for the react-material-icons Icon component with a few changes that makes it more flexible.

I removed the "Icon" story from the blueocean-dashboad and created a new "Icon" story in the JDL that replaces it.
 
See [JENKINS-45743](https://issues.jenkins-ci.org/browse/JENKINS-45743).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


i386:
I gave this a spin and its looking very good. What you might want to look at is building this branch locally then changing https://github.com/jenkinsci/blueocean-pipeline-editor-plugin/blob/master/pom.xml#L28 to match the version of blue ocean you have locally, build it and test the editor

blueocean-core-js/src/js/parameter/ParametersRunButton.jsx
```diff
@@ -157,7 +157,7 @@ export class ParametersRunButton extends Component {
             </Dialog>);
         }
         return (<div className="ParametersRunButton">
-            <RunButton {...runButtonProps} />
+            <RunButton {...runButtonProps} iconColor={this.props.iconColor ? this.props.iconColor : '#ffffff'} hoverIconColor={this.props.hoverIconColor ? this.props.hoverIconColor : ''} />
```
cliffmeyers: Would this be better as a `defaultProp`?

blueocean-core-js/src/js/components/ReplayButton.jsx
```diff
@@ -83,7 +83,7 @@ export class ReplayButton extends Component {
         return (
             <div className={`replay-button-component ${outerClass}`} onClick={(event => stopProp(event))}>
                 <a className={`replay-button ${innerButtonClass}`} title={replayLabel} onClick={() => this._onReplayClick()}>
-                    <Icon size={24} icon="replay" />
+                    <Icon size={24} icon="AvReplay" />
```
cliffmeyers: What's the meaning of "Av" at the beginning?

cliffmeyers: Ok, looks like maybe it's the "category" from https://material.io/icons/   But aren't the icon names unique across all categories? Some quick searches seemed to indicate so. If that's the case, I would just use the icon name, probably all lowercase with hyphen or underscore as a delimiter.

jenkins-design-language/src/js/components/material-ui/svg-icons/action/accessibility.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import pure from 'recompose/pure';
```
cliffmeyers: I am curious what exact performance benefit this gives for a stateless functional component, and am always hesitant to take on any React-based library that might inhibit us from upgrading React later.

cliffmeyers:
cc @sophistifunk as his feedback for UI components is always A+

jenkins-design-language/src/js/components/material-ui/SvgIcon/SvgIcon.js
```diff
@@ -0,0 +1,102 @@
...
+  };
+
+  static defaultProps = {
+    onMouseEnter: () => {},
+    onMouseLeave: () => {},
+    viewBox: '0 0 24 24',
+  }; 
+
+  state = {
+    hovered: false,
```
cliffmeyers: Would it be feasible to apply the hover color via CSS alone? If so, this could be reduced down to a purely functional component.

sophistifunk: If it's just things like `strokeColor` and `fillOpacity` and such we should be able to do it by CSS `:hover`

blueocean-core-js/src/js/components/ReplayButton.jsx
sophistifunk: FWIW I don't care what the names are, so long as there's an example of all of them in the storybooks so we know which names are available :)

jenkins-design-language/src/js/components/material-ui/svg-icons/action/accessibility.js
sophistifunk: I believe this functionality is already in React for function components, and possibly also if you explicitly `class MyComponent extends PureComponent`

sophistifunk: But yes, we definitely don't want to do anything to hurt upgrading React - v16 (IIRC) with the fiber engine allows components to return multiple children from `render()`, which will allow us to clean up a bunch of our HOCs that cause layout issues due to wrapping divs that are only there to satisfy the "single div" rule.

jenkins-design-language/src/js/components/material-ui/SvgIcon/SvgIcon.js
sophistifunk: We might need per-icon styles though, depending on if we've got a mix of visual strokes, rather than relying on fills. Would need a visual inspection.

jenkins-design-language/less/components/button.less
```diff
@@ -129,6 +129,10 @@ button, .btn, .btn-default, .btn-primary, .btn-secondary, .btn-danger, a.btn-lin
 button, .btn, .btn-default, .btn-primary, .btn-secondary {
     &.monochrome {
         .btn-styles(transparent, @gray-light, @text-color, @gray-lighter, @gray-lighter);
+
+        svg {
+            fill: @text-color !important;
```
sophistifunk: What was overriding this that we needed `!important`? It might be better to increase specificity on whatever it was conflicting with.

sophistifunk:
Having a boatload of individual Icon components with inline SVG in separate files is definitely the way to go IMO, nice work @NicuPascu. 

We might want to do some experiments to make sure we're only including the ones referenced during bundle creation, or the cost may be a significant # of ms in both loading and (more importantly from a responsiveness perspective) parsing the bundle files.

i386:
We would need another PR for the pipeline editor, correct?

NicuPascu:
@i386 yes, I am working on that as well

jenkins-design-language/less/components/button.less
NicuPascu: I did this because of the monochrome icons I talked about during our meeting, it is just to make sure that all svgs that are children of elements with ".monochrome" have the color of the text, I can't increase specificity in this case because the fill rule that needs to be beaten is inline, only important can trump that. This could be removed if all the components that have monochrome icons correctly sends the fill color, I added it because the way some components are made right now make this really difficult right now, can't remember which one right now

jenkins-design-language/src/js/components/material-ui/SvgIcon/SvgIcon.js
NicuPascu: I kept it like this because if we send the fill as a prop, shouldn't we be able to send the hover color as well? I think it gives us an extra layer of flexibility so we can make the icon look specifically how we want without any extra css

jenkins-design-language/src/js/components/material-ui/svg-icons/action/accessibility.js
NicuPascu: I left all the icons as they are declared in "https://github.com/callemall/material-ui" because it would make it really easy to upgrade all the icons in the future, I don't really understand why they decided to use the recompose/pure component but if we want to remove it, we have to remove it from every svg and every time we want to upgrade the icons, remember to remove it again from the upgraded icons

blueocean-core-js/src/js/components/ReplayButton.jsx
NicuPascu: The names are declared like this in "material-ui/svg-icons/index.js", I didn't want to change them because I wanted the update process to be as easy as possible in the future, I also created a story in the JDL storybook called Icon->all icons that renders all the icons with their respective names

jenkins-design-language/less/components/button.less
sophistifunk: But this wouldn't be a problem if we weren't inlining the colours, would it?

jenkins-design-language/src/js/components/material-ui/svg-icons/action/accessibility.js
sophistifunk: If we're using their classes verbatim, why clog up our repo with them?

jenkins-design-language/less/components/button.less
NicuPascu: Yup, it wouldn't

jenkins-design-language/src/js/components/material-ui/svg-icons/action/accessibility.js
NicuPascu: Well we don't need everything in the material-ui, the icons are just a small part of it, secondly, if we have them in our repo, we can control if/when we update the icons and maybe add our own. That doesn't mean we can't change anything, I just explained my rationale for what I did, if you think it is better to remove the recompose/pure dep and/or change the name of the icons to not include the category name, it is totally doable and the benefit of hosting the icons ourselves, the only downside will be when we will want to update the icons. What do you think?

sophistifunk: No need to change the names I think, unless it really bothers Cliff :) But we could probably just write a script to copy them and strip the recompose.pure stuff mechanically. I'm happy to help with that if you like, since I'm the one suggesting it.

NicuPascu: Yup, would really appreciate the help, thanks

michaelneale:
hrm - a couple of failures in a row, ma need to look into ATH live tests...

NicuPascu:
I think there might be some legitimate tests failing, I'm trying to run the ATH locally to see if I can do anything about them.

michaelneale:
@NicuPascu good idea - yes it seems consistent - so something isn't happy. It likely will show up if you follow it along while it runs via vnc

michaelneale:
Yes is a :bug: somewhere @NicuPascu. The dreaded invariant error (see below). 

To see this - run it up - then go through the new github pipeline flow - when you pick a github org and then the "new pipeline button" it fails. 

```
Error: Minified React error #130; visit http://facebook.github.io/react/docs/error-decoder.html?invariant=130&args[]=undefined&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at reactProdInvariant (blueocean.js:133964)
    at ReactCompositeComponentWrapper.instantiateReactComponent [as _instantiateReactComponent] (blueocean.js:133758)
    at ReactCompositeComponentWrapper.performInitialMount (blueocean.js:122556)
    at ReactCompositeComponentWrapper.mountComponent (blueocean.js:122447)
    at Object.mountComponent (blueocean.js:129460)
    at ReactDOMComponent.mountChildren (blueocean.js:128338)
    at ReactDOMComponent._createInitialChildren (blueocean.js:123914)
    at ReactDOMComponent.mountComponent (blueocean.js:123739)
    at Object.mountComponent (blueocean.js:129460)
    at ReactCompositeComponentWrapper.performInitialMount (blueocean.js:122560)
```

NicuPascu:
Thanks for spotting it, I will take a look

jenkins-design-language/src/js/components/material-ui/svg-icons/action/accessibility.js
cliffmeyers: I'm okay with the names as-is.

blueocean-core-js/src/js/components/ReplayButton.jsx
cliffmeyers: That's okay by me, now that I understand the change.

i386:
@NicuPascu are you blocked on something?

NicuPascu:
I am having problems running the ATH tests locally, they all time out for some reason, will ask Ivan for help because I spent too much time on this without success

michaelneale:
@NicuPascu you don't need to run ATH to see this problem  just try to create a pipeline even when running hpi:run - ATH is just picking up same issue

NicuPascu:
@michaelneale Ah, thanks, I missed that info in your previous comment, just saw the error log for some reason, thanks for the tip

michaelneale:
@NicuPascu ok well if this passes - I expect it should be ok right? Is there any area to look at? Or just some screen shots?

NicuPascu:
@michaelneale Basically everything should look exactly the same as before removing the react-material-icons. I changed all the components that had an icon so they use the new ones instead of the old ones, I checked them all manually to be sure they look fine and I think that is the only way to check them, just one screenshot would not guarantee that all the changes are fine, the icons are peppered all over the project

jenkins-design-language/less/components/button.less
NicuPascu: I just realised we can fix this easily by setting the fill as an attribute instead of inline style... will try to see if my hunch is correct and if it is... it we can use css AND props to set the colour and hover state

michaelneale:
:bee: nice one @NicuPascu !

jenkins-design-language/less/components/button.less
NicuPascu: I think I just opened a can of worms, a lot of icons have a bad inherited colour if we do it as I proposed. I don't think the benefit is worth it right now, if we have a different reason to remove this important rule or to add the fill colour by attr instead of inline style I will come back to it in a different PR, as it is, everything works as intended and I think we should merge as is.

michaelneale: ok

i386:
Nice work mate!!!

PR #1310, opened 2017-08-08
JENKINS-40616 Show publishHTML() reports on Blue Ocean

i386:
# Description

*Example Jenkinsfile*
```
node {
    sh 'mkdir coverage'
    sh 'touch coverage/foo.html'
    sh 'echo "hello world" > coverage/index.html'
    
    publishHTML (target: [
      allowMissing: false,
      alwaysLinkToLastBuild: false,
      keepAll: true,
      reportDir: 'coverage',
      reportFiles: 'index.html',
      reportName: "My Cool report"
    ])
}
```

*Showing on artifact screen*
publishhtml

See [JENKINS-40616](https://issues.jenkins-ci.org/browse/JENKINS-40616).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
+++ b/blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
@@ -100,6 +100,22 @@ export default class RunDetailsArtifacts extends Component {
             const urlArray = artifact.url.split('/');
             const fileName = urlArray[urlArray.length - 1];
             logger.debug('artifact - url:', artifact.url, 'artifact - fileName:', fileName);
+
+            let downloadLink = null;
+            if (artifact.downloadable) {
+                downloadLink = (
+                    <a target="_blank"
+                       className="action-button-colors"
+                       download={fileName} title={downloadLabel}
+                       href={`${rootURL}${artifact.url}`}
+                    >
+                        <Icon icon="file_download" />
+                    </a>
+                );
+            }
+
+            const artifactSize = artifact.size >= 0 ? (<FileSize bytes={artifact.size} />) : (<span>–</span>);
+
             return (
                 <TableRow key={artifact.url}>
                     <TableCell>
@@ -108,16 +124,10 @@ export default class RunDetailsArtifacts extends Component {
                         </a>
                     </TableCell>
                     <TableCell>
-                        <FileSize bytes={artifact.size} />
+                        {artifactSize}
                     </TableCell>
                     <TableCell className="TableCell--actions">
-                        <a target="_blank"
-                           className="action-button-colors"
-                           download={fileName} title={downloadLabel}
-                           href={`${rootURL}${artifact.url}`}
-                        >
-                            <Icon icon="file_download" />
-                        </a>
+                        {downloadLink}
                     </TableCell>
                 </TableRow>
             );
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/htmlpublisher/HTMLArtifact.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/htmlpublisher/HTMLArtifact.java
@@ -1,0 +1,75 @@
+package io.jenkins.blueocean.htmlpublisher;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
+import htmlpublisher.HtmlPublisherTarget;
+import hudson.Extension;
+import hudson.Util;
+import hudson.model.Run;
+import io.jenkins.blueocean.rest.Reachable;
+import io.jenkins.blueocean.rest.factory.BlueArtifactFactory;
+import io.jenkins.blueocean.rest.hal.Link;
+import io.jenkins.blueocean.rest.model.BlueArtifact;
+
+import java.util.Collection;
+import java.util.List;
+
+public class HTMLArtifact extends BlueArtifact {
+
+    private final HtmlPublisherTarget.HTMLBuildAction action;
+    private final Reachable parent;
+
+    public HTMLArtifact(HtmlPublisherTarget.HTMLBuildAction action, Reachable parent) {
+        this.action = action;
+        this.parent = parent;
+    }
+
+    @Override
+    public Link getLink() {
+        return parent.getLink().rel(Util.rawEncode(action.getHTMLTarget().getReportName()));
+    }
+
+    @Override
+    public String getName() {
+        return action.getHTMLTarget().getReportName();
+    }
+
+    @Override
+    public String getPath() {
+        return action.getHTMLTarget().getReportName();
+    }
+
+    @Override
+    public String getUrl() {
+        return String.format("/%s%s", action.getOwner().getUrl(), action.getUrlName());
+    }
+
+    @Override
+    public long getSize() {
+        return -1;
+    }
+
+    @Override
+    public boolean isDownloadable() {
+        return false;
+    }
+
+    @Extension
+    public static class FactoryImpl extends BlueArtifactFactory {
+        @Override
+        public Collection<BlueArtifact> getArtifacts(Run<?, ?> run, Reachable parent) {
+            List<HtmlPublisherTarget.HTMLBuildAction> actions = run.getActions(HtmlPublisherTarget.HTMLBuildAction.class);
+            if (actions.isEmpty()) {
+                return null;
+            }
+            List<BlueArtifact> foundArtifacts = Lists.newArrayList();
+            for (HtmlPublisherTarget.HTMLBuildAction action : actions) {
+
+
+
+                foundArtifacts.add(new HTMLArtifact(action, parent));
+            }
+            return foundArtifacts;
+        }
+    }
+}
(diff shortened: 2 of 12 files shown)


blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
```diff
@@ -100,6 +100,22 @@ export default class RunDetailsArtifacts extends Component {
             const urlArray = artifact.url.split('/');
             const fileName = urlArray[urlArray.length - 1];
             logger.debug('artifact - url:', artifact.url, 'artifact - fileName:', fileName);
+
+            let downloadLink = null;
+            if (artifact.downloadable) {
```
i386: If not downloadable, don't show the link.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/htmlpublisher/HTMLArtifact.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public long getSize() {
+        return -1;
+    }
+
+    @Override
+    public boolean isDownloadable() {
+        return false;
```
vivek: its false, so how does it get rendered in blueocean?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/htmlpublisher/HTMLArtifact.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public String getPath() {
+        return action.getHTMLTarget().getReportName();
+    }
+
+    @Override
+    public String getUrl() {
+        return String.format("/%s%s", action.getOwner().getUrl(), action.getUrlName());
```
vivek: So the url is relative to blueocean artifact page?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ArtifactContainerImpl.java
```diff
@@ -1,122 +1,47 @@
...
-        return Iterators.transform(artifacts.iterator(), new Function<Run.Artifact, BlueArtifact>() {
+        return Iterators.find(iterator(), new Predicate<BlueArtifact>() {
             @Override
-            public BlueArtifact apply(@Nullable Run.Artifact artifact) {
-                if(artifact == null) {
-                    return null;
-                }
-
-                return new ArtifactImpl(run, artifact, ArtifactContainerImpl.this);
+            public boolean apply(@Nullable BlueArtifact input) {
```
vivek: Artifact permission check `if(Functions.isArtifactsPermissionEnabled() && !run.hasPermission(Run.ARTIFACTS)) {}` is removed. 🐛  

Also, now that it returns ArtifactImpl or any implementation of BlueArtifact, `BlueArtifact.getSize()` is not implemented any more. Current functionality shows the  size of download.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ArtifactContainerImpl.java
```diff
@@ -1,122 +1,47 @@
...
+                return input != null && input.getName().equals(name);
             }
-        });
+        }, null);
     }
 
     @Override
+    @Nonnull
     public Iterator<BlueArtifact> iterator() {
```
vivek: Artifact permission check `if(Functions.isArtifactsPermissionEnabled() && !run.hasPermission(Run.ARTIFACTS)) {}` is removed. 🐛

blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
```diff
@@ -100,6 +100,22 @@ export default class RunDetailsArtifacts extends Component {
...
+                       className="action-button-colors"
+                       download={fileName} title={downloadLabel}
+                       href={`${rootURL}${artifact.url}`}
+                    >
+                        <Icon icon="file_download" />
+                    </a>
+                );
+            }
+
+            const artifactSize = artifact.size >= 0 ? (<FileSize bytes={artifact.size} />) : (<span>–</span>);
```
i386: If the artefact does not have a valid size, show –

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ArtifactImpl.java
```diff
@@ -41,7 +54,24 @@ public long getSize() {
...
+    public boolean isDownloadable() {
+        return true;
+    }
+
+    @Extension(ordinal = -1)
+    public static class FactoryImpl extends BlueArtifactFactory {
+        @Override
+        public Collection<BlueArtifact> getArtifacts(final Run<?, ?> run, final Reachable parent) {
+            // Check security for artifacts
+            if(Functions.isArtifactsPermissionEnabled() && !run.hasPermission(Run.ARTIFACTS)) {
```
vivek: Ah, this is moved over here, but then we are leaving this check up to BlueArtifactFactory implementors. e.g. HtmlArtifact's factory is not doing this check. We should move it up to `ArtifactContainerImpl.get()` and `ArtifactContainerImpl.iterator()`

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/htmlpublisher/HTMLArtifact.java
i386: without the download link. The name is still clickable

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/htmlpublisher/HTMLArtifact.java
i386: Thats the URL for the action as constructed by the HTMLPublish plugin. Out of my control.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ArtifactImpl.java
i386: Done.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ArtifactContainerImpl.java
```diff
@@ -1,122 +1,45 @@
...
-    }
-
-    @Override
-    public Iterator<BlueArtifact> iterator(int start, int limit) {
-        // Check security for artifacts
-        if(Functions.isArtifactsPermissionEnabled() && !run.hasPermission(Run.ARTIFACTS)) {
-           return Iterators.emptyIterator();
-        }
-
-        List<Run.Artifact> artifactsUpTo = run.getArtifactsUpTo(limit);
```
i386: @vivek I am not sure if I have produced a performance issue here by removing the pagination like this - WDYT?

vivek: Ah, I missed it. I think we should revert to `iterator(start, limit)` and then `BlueArtifactFactory.resolve()` should also take start/limit in to account, otherwise we will end up reading all artifacts. Even probably BlueArtifactFactory.getArtifacts() could be expensive. I think we should fix it before merging, so I am marking it as 🐛

abishek3876:
Wouldn't it be better if instead of including every requested feature from other plugins in here, blueocean made it easier for other plugins to plug content into it. Right now, my jenkins install has a huge number of totally useless plugins just because I want to use blueocean. And I don't think it's ever going to be possible to integrate all the charts/UI views from all the plugins into blueocean.

i386:
@abishek3876 this PR includes an extension point to do just that. Now, we could move it to the publishHTML plugin but many maintainers have pushed back on us including Blue Ocean extensions in their plugins. They either do not want to rely on the newer version of Jenkins we have a hard requirement on or they are uncooperative on depending on Blue Ocean's API modules at all. 

We've tried to do the right thing but there is deadlock in the community. If there are particular pieces of functionality you think should be moved from Blue Ocean to a different plugin, please take that up with the maintainer. 

As for the "useless plugins", it is better to think of Blue Ocean as a distribution rather than a bunch of plugins. We provide a developer experience that makes Jenkins easier to use. Removing any of the plugins the aggregator could result in undefined behaviour and we won't be offering any support in that situation.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/factory/BlueArtifactFactory.java
```diff
@@ -0,0 +1,37 @@
+package io.jenkins.blueocean.rest.factory;
```
i386: @abishek3876 this is the package that contains our primary extension points. We are also working on a way for developers to extend the user interface directly but it is immature at this time. If there are specific places in the UI that you think could or should be extensible, please open a ticket.

amuniz:
> Now, we could move it to the publishHTML plugin but many maintainers have pushed back on us including Blue Ocean extensions in their plugins

Then put the extension into a new plugin, not in BO core.

PR #1313, opened 2017-08-09
JENKINS-46070 PR page broken on dogfood

sophistifunk:
# Description

* Fix an issue passing translate function to child components

See [JENKINS-46070](https://issues.jenkins-ci.org/browse/JENKINS-46070).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/PullRequestRow.jsx
+++ b/blueocean-dashboard/src/main/js/components/PullRequestRow.jsx
@@ -29,7 +29,7 @@ function noRun(pr, openRunDetails, t, columns) {
 }
 
 export class PullRequestRowRenderer extends Component {
-    
+
     render() {
         const {
             columns,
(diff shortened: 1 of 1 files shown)


sophistifunk:
I'm still confused as to why this didn't fail ATH :-/

i386:
@imeredith do we not have ATH coverage for this page?

NicuPascu:
LGTM, I totally agree with this way of passing the props along compared with the "..." way, this way is a lot clearer. And regarding the ATH, is it possible to fail the build if there is any JS error in the console? That would have caught this, don't know if it is possible though. :bee:

sophistifunk:
It's not just that using the `...` is ugly for a single property, but it was also incorrect, which is the main reason I changed it when I noticed. It was `{...t}` which will explode `t` and use its enumerables as props. When you just want the same meaning as `t={t}` the syntax is `{..{t}}`

sophistifunk:
Oh sure, ATH fails when I *fix* stuff :)

sophistifunk:
@i386 @imeredith there's definitely coverage for PR tab, there must be something peculiar to the environment or the limited repos used for the tests that causes the `RunHistoryButton` to not render. Maybe it's only ever exercised for "virgin" pipelines that have never run, or the user does not have the permission.

PR #1316, opened 2017-08-10
Import script for icons

sophistifunk:
# Description

Just put it in its own PR so you can see the diffs easily. Always good to get a sanity check on mass mechanical changes to files 😁


jenkins-design-language/bin/import-material-icons.js
```diff
@@ -0,0 +1,248 @@
...
+
+/**
+ * Load an icon source file, remove the references to recompose/pure, then write into the JDL
+ *
+ * Very simplistic transform, and won't write anything if the file's not recognised / transformed correctly.
+ */
+function processIcon({ filePath, relativePath }, destPathRoot) {
+
+    // The patterns we're looking for. Only write the new file if all are met.
+    const pureImportRegex = /^import pure from 'recompose\/pure';\s*\n/gm;
```
kzantow: minor suggestion: might use `\s+` / `\s*` instead of spaces, just in case they randomly change upstream for some unknown reason...

kzantow:
Looks cool! 👍

jenkins-design-language/bin/import-material-icons.js
sophistifunk: I had that originally, but then I changed it to be less flexible so we get an early warning if things are changed. I think those files are mechanically generated anyway so it seems unlikely.

michaelneale:
@NicuPascu @sophistifunk I guess this is failing in the same way the other one was? (at least the test failures look similar) - ie when you press "new pipeline" things break? (or may just be flaking test failures?)

sophistifunk:
Failures here don't matter, it's against Nicu's branch not master.

PR #1321, opened 2017-08-11
JENKINS-46156 Only attempt to stop a run once when the user clicks the stop link

philsttr:
# Description

See [JENKINS-46156](https://issues.jenkins-ci.org/browse/JENKINS-46156).

Previously, Blue Ocean would send the stop signal to a job multiple times until the job completed.
This did not allow jobs to handle aborts properly.

For example, if a job caught the abort exception, and needed to execute some additional steps to cleanup, Blue Ocean would send more stop signals while the job was attempting to clean up.

Now, Blue Ocean will send the stop signal once, and wait for the job to complete.

No existing unit test for the stop workflow to modify.  Just trying to speed up getting this fixed. 

### Manual Test

1. Create a job with the scripted pipeline from [JENKINS-46156](https://issues.jenkins-ci.org/browse/JENKINS-46156).
2. Run it
3. Abort it via the Blue Ocean UI
4. Ensure that the sh step in the cleanup stage succeeds

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
@@ -216,14 +216,11 @@ public abstract class AbstractRunImpl<T extends Run> extends BlueRun {
                     throw new ServiceException.BadRequestException("timeOutInSecs must be >= 0");
                 }
 
-                long timeOutInMillis = timeOutInSecs*1000;
+                stoppableRun.stop();
 
+                long timeOutInMillis = timeOutInSecs*1000;
                 long sleepingInterval = timeOutInMillis/10; //one tenth of timeout
                 do{
-                    if(isCompletedOrAborted()){
-                        return this;
-                    }
-                    stoppableRun.stop();
                     if(isCompletedOrAborted()){
                         return this;
                     }


i386:
I am cool with this change so long as this doesn't change the behaviour for stuck runs. In Classic, it is almost impossible to stop hung runs. Admittedly this was a bit of a hack to route around this behaviour.

kzantow:
@i386 : just a note - I'm not convinced it was working to stop stuck runs, anyway. This just happened on CI a few days ago, where I was unable to stop a hung build and I had to go in to the pipeline log and click an alternate link to abort.

michaelneale:
I was under the impression that there was a grace period (10 seconds or so) where it could clean up. If a cleanup takes longer... 

basically do we want the stop to really stop (mostly blue ocean does a good job here, perhaps too good?). 

Thoughts @vivek ?

philsttr:
Under the current code (prior to this PR), the request would [wait up to 10 seconds](https://github.com/jenkinsci/blueocean-plugin/blob/4cefa8d4a94bb513d252e436ae8228f809368209/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java#L212-L214) for the job to stop.  During those 10 seconds, it would send the stop signal [every 1 second](https://github.com/jenkinsci/blueocean-plugin/blob/4cefa8d4a94bb513d252e436ae8228f809368209/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java#L221). Therefore, the job would have had to cleanup within 1 second to avoid getting stopped again during cleanup.  Hardly enough time for any job.

In my case, cleanup can possibly take multiple minutes.  The job where I first noticed this problem is a CD pipeline, so "cleanup" involves rolling back an environment to a previous state, which is not instantaneous.

i386:
@abayer @svanoort @michaelneale we should discuss this at the Pipeline catchup this week. Certainly we don't want a "Stop (maybe)" and a "No, really, Stop NOW" button.

michaelneale:
@philsttr right - the preferred thing would be to perhaps give it 10 seconds (that can be negoiatble) and then kill it until it stops - need @vivek to weigh in

i386:
@philsttr I understand and value your use case. However the problem that I have with this change is that most users, when they click stop, want an immediate stop. This is tricky and the team will need to have a think about this before we can merge.

philsttr:
Yup.  Understood.  I just need a way to let the scripted pipeline cleanup after an abort with a long/configurable timeout.  (Perhaps a job property to set the abort timeout?)

But having said that, as @kzantow alluded to, I doubt sending the stop signal 9 more times after the first time is going to actually stop a "stuck" job.  If a job is truly "stuck", and it doesn't respond to the first stop signal, it's probably not going to respond to the next 9 either.

michaelneale:
@philsttr based on anecdotal experience, it does seem to work (so perhaps a better behavior is to try it once politely, timeout, and then hit it hard to make it stop as best as possible).

philsttr:
Ok.  That sounds great to me, as long as I can control the timeout, preferably on a per-job basis (CD-related jobs will have long timeouts.  CI-related jobs won't).

If that is going to take a while to implement, please consider this PR as a short-term solution.  From a consistency standpoint, I think it would be better for Blue Ocean's stop button to behave the same as the old UI's abort button.  Even if that means I have to click the abort button 10 times instead of once in both UIs to mimic the current behavior of the Blue Ocean's stop button.

Thanks for the attention to detail.  I'm really digging Blue Ocean.

vivek:
@philsttr Thanks for the PR but backend code already supports non-blocking one time fire and forget abort behavior - thats default. UI sends `?blocking=true` to try force abort in 10 sec, see https://github.com/jenkinsci/blueocean-plugin/blob/41936a8b5a1a33c8ea5ab69483f946397b5bc130/blueocean-core-js/src/js/rest/RunApi.js#L27. Its also documented at https://github.com/jenkinsci/blueocean-plugin/blob/0e5974efa7bb8330e9d4610b4d9283f6331fa6a1/blueocean-rest/README.md#stop-a-build-as-blocking-call. Given this, we do not need any changes in backend code.

If we want to have both behavior 'Stop' and 'Really Really stop', it needs to surface somehow in UI. More of question for @i386?

kzantow:
Maybe we could try to stop nicely if the user clicks again try to stop it forcefully, at least it wouldn't clutter up the UI. We could change the icon to a "stopping, click to terminate now" type of icon. I can see a desire to have some time to clean up with some process when a user stops the job, maybe this actually needs to be exposed as a pipeline property or something for people who really need it. Anyway, just spitballin'. Carry on here, nothing to see.

michaelneale:
@kzantow @philsttr interesting. 

I would be a bit more sympathetic but the classic approach appeared to work - as it is it seems completely broken any time I have used it (perhaps this is more a surface thing, there is no feedback that I pressed it). I think few people would stop and clear out builds as much as I do, so I am wary of changing behavior to the classic broken approach without knowing what it means. 

@philsttr can you provide some mode details/examples of the type of cleaning up you do, perhaps even a Jenkinsfile sample, or anything really to explain what sort of cleanup is expected?

philsttr:
Sure thing @michaelneale 

My use case can be simplified down to any run that takes a non-trivial amount of time to cleanup after an abort is requested by the user.

My specific use case is for a CD pipeline that uses [terraform](https://www.terraform.io).
The pipeline will deploy a new environment via [`terraform apply`](https://www.terraform.io/docs/commands/apply.html), and then wait for it to become "healthy".  If the user aborts the run while the pipeline is deploying or while waiting for the new environment to become healthy, then the pipeline will destroy the new enviroment via a [`terraform destroy`](https://www.terraform.io/docs/commands/destroy.html) command, which takes a non-trivial amount of time.  The Blue Ocean UI is currently aborting the cleanup one second after it starts cleaning up.  Therefore the environment is not destroyed properly, and it requires one of our engineers to go clean up the environment.

Unfortunately I cannot provide the Jenkinsfile for this pipeline, since our pipeline is pretty complex, and uses some custom-built shared libraries, etc.  However, the simple pipeline in [JENKINS-46156]( https://issues.jenkins-ci.org/browse/JENKINS-46156) captures the essence of the problem.  Just image that the `sh 'sleep 10'` step is executing a `terraform destroy`.  The `sh` step will be interrupted by the Blue Ocean UI.

michaelneale:
@philsttr thanks (excuse use of PR for discussion, but why not, github have made a nice UI). 

In that case - would 10 seconds be enough? 30? or talking minutes in that case? Are you using declarative pipeline or scripted?

philsttr:
Minutes in some cases.  It depends on the number of resources being destroyed, and the speed of the cloud provider.  Both of which are variable, which can cause the cleanup time to vary.

Truthfully, if I was given an option to set the "cleanup timeout" for these CD pipelines, I would probably set it to an hour or something large.  Mainly because I want to avoid Blue Ocean killing the cleanup in all but the most extreme situations.  Any time Blue Ocean kills a cleanup, it means somebody has to manually clean up an environment, which is lost time, effort, and focus.

This is a scripted pipeline.  (We can't switch to a declarative at this point.  It's just doesn't offer the flexibility we need.)


Since this is a discussion, I'm going to propose a few options with varying complexity that would allow a job to properly cleanup without interruption.  This is probably not an exhaustive list, so feel free to propose more.

1. When the user requests the run to stop, only send the stop signal to the job once .  For properly behaved runs, this should be enough.  If the run is not behaving as the user expects, the user can press the stop button any more number of times to try to further interrupt the job.  This is similar to the old Jenkins UI, and similar to sending a SIGTERM to a unix process.
2. When the user requests the run to stop, send the stop signal once, then wait a timeout, then send it a bunch more times on behalf of the user.  Allow the user to configure the length of the timeout as a job properly, since it will vary based on the type of job (CD jobs require long timeout, CI jobs do not)
3. Give a run the ability to indicate that it received the abort signal, and is in the process of cleanup.  When the user requests the run to stop, send the stop signal repeatedly (with delays between attempts) until the run acknowledges that it received the abort signal, or finishes.   The abort request and acknowledgement would need to be correlated with some sort of id unique to a user request, so that if a user sends multiple request, each of them can be handled individually. 
4. Provide two options to the user, analogous to SIGTERM and SIGKILL for unix processes.  Call them what you want, but SIGTERM will give the process a chance to catch the signal and cleanup.  SIGKILL is a 'force kill' that allows no cleanup.  SIGTERM would just send the stop signal once per user request to stop.  SIGKILL would send the stop signal repeatedly (since that appears to be the only way to stop a run).

svanoort:
We do have the primitives in pipeline core to support variants of this -- in the WorkflowJob doTerm (trigger failure in running steps, which may lead to cleanup triggering), doStop (first try to stop the pipeline's base task, then do a hard kill if that's already dead and invoked), and doKill (kill the job and put a stake through its heart). 

This is analogous to what @philsttr is describing (yup, Sigterm == doTerm, Sigkill ~= doKill). 

> We could change the icon to a "stopping, click to terminate now" type of icon.

Yup, this would be a good approach.  Provides a more graceful approach to how vanilla pipeline offers a hard-kill link.

michaelneale:
Hi @philsttr thanks for the comprehensive reply. That makes sense to me. I have done similar things with setup and tear down infra. In my case I have used both out of band "garbage collectors" to clean up things as long as idempotent jobs I would fire off. I never would quite trust pipeline to have enough time to clean up every time (eg what if agent gets killed?) so I would need to have some out of band garbage collection anyway. So I automated that. 

What I think you have uncovered is a new feature. The pipeline needs to have a state of "shutting down" so that can be reflected in the UI etc. This can then allow a lifecycle similar to posix processes like @svanoort says: press a button - it is ALWAYS in shutting down state. Press again later when frustrated - SIGKILL. This is a fairly familiar experience. However, this does require some changes to pipeline and is a bit of a bigger story to make it not feel broken. I will write that up as a feature - it may not be a huge amount of work but it cuts into a few areas. 

As for this PR - I think if there was a way to make this a feature flag to solve your immediate problem (via a system property?) perhaps it would be ok as a short term measure, as it would solve your problem for now, until we get the proper statuses and so on in place.

philsttr:
A system property for making Blue Ocean's abort button work the same way as the classic UI's abort button is a compromise that will work for me.  I'll be happy to use it while waiting on a more thorough feature.

Having said that, it would be easier to just make Blue Ocean's abort button work the same way as the classic UI's abort button _by default_.  I don't think this would be a bad thing™.  As a long-time jenkins user, most runs that I've ever encountered will respond to a single abort request appropriately.  However, maybe I just haven't aborted enough jobs, or my jobs are particularly well behaved.  In the rare case (that I'm not too familiar with) where the run doesn't respond to the abort request, I think the user would instinctively try a few more times (mimicking the current Blue Ocean behavior).  So, they still have a way to abort those runs.  As a side note, I really would like to understand the cases where a run does not respond to the first abort request.  Perhaps some focus should be placed on making those runs more receptive to abort requests.

michaelneale:
@philsttr no - as the classic abort button doesn't "work", it happens to work nicely in your case though! but for the vast majority - no it doesn't work. I press it all the time and 1 time out of 20 it will stop. Having said that- let me try this out to see what the perception is...

michaelneale:
hrm - @philsttr this certainly works better than classic, so maybe this is ok as a starting point. There may be screams once it gets in the wild, but in that case we could undo it etc... worth considering I think 

@vivek  ?

philsttr:
Ok.  Thanks for giving it a second consideration!

BTW, I did some [JIRA searching](https://issues.jenkins-ci.org/browse/JENKINS-22776?jql=status%20not%20in%20%20(Resolved%2C%20Closed)%20AND%20text%20~%20%22abort%20work%22) trying to understand when other users are encountering runs that will not abort when requested.  I found these:

* [JENKINS-37790](https://issues.jenkins-ci.org/browse/JENKINS-37790) - seems to indicate that there is already a force terminate option after failing to stop a build.  Although, it looks like it's prompting multiple times, which would be bad. 
* [JENKINS-42049](https://issues.jenkins-ci.org/browse/JENKINS-42049) - also indicates a force kill feature is available.
* [JENKINS-45772](https://issues.jenkins-ci.org/browse/JENKINS-45772) - also indicates a force kill feature is available (although it also mentions [JENKINS-32986](https://issues.jenkins-ci.org/browse/JENKINS-32986) which seems to be that a force kill doesn't clean up threads properly)
* [JENKINS-33164](https://issues.jenkins-ci.org/browse/JENKINS-33164) - also indicates a force kill feature is available, but it's not working in this case.
* [JENKINS-44838](https://issues.jenkins-ci.org/browse/JENKINS-44838) - seems to be a bug in `SCMStep.StepExecutionImpl.stop`.  

A couple thoughts on those issues:

1. They seem to confirm what @svanoort mentioned, i.e. that term/kill already exists (perhaps that is what was introduced in [JENKINS-25550](https://issues.jenkins-ci.org/browse/JENKINS-25550)).  Maybe Blue Ocean can take advantage of this when it implements the term-then-kill functionality
2. I don't get the impression that Blue Ocean's current approach of repeatedly calling stop would solve any of those problems. It would be interesting to test these cases out with Blue Ocean's current stop behavior to see if it solves them.


Do you see a common pattern for the runs that you have that won't abort after the first request?  I'm wondering if it's specific to certain jobs or conditions.  (Also, are there other jira issues that I didn't find that more accurately describe the "can't-be-aborted-easily" behavior you are seeing?)  Even when the newly proposed term-then-kill logic is implemented in Blue Ocean, it would be nice to fix those bugs so that term has a better chance of success.  i.e. so the kill is not required very often.

michaelneale:
@philsttr not a common pattern - I suspect the UI experience in classic is just so broken it seems like you have to press things lots of times when you won't. 

I think we may be able to move forward with this, and add on a feature request to have a "stopping" status visible - WDYT?

philsttr:
Sounds perfect to me!

I guess the only question would be if the fix is in `AbstractRunImpl` as I have done here, or in `RunApi.js` (to make it so it doesn't pass blocking=true).  Either would work.

vivek:
@philsttr @michaelneale I am ok with controlling blocking stop behavior with env variable as temporary solution, that is we might remove it in future versions when we have proper solution. Such env variable are more of technical debt and do not want them to live forever:)

On the change needed, all we need is to control blocking query parameter behavior, that is something like this is all thats needed:

```patch
diff --git a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
index 3c13558d7..edaa0957f 100644
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
@@ -29,7 +29,6 @@ import io.jenkins.blueocean.rest.model.GenericResource;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
 import org.kohsuke.stapler.QueryParameter;
-import org.kohsuke.stapler.export.Exported;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
@@ -263,6 +262,9 @@ public abstract class AbstractRunImpl<T extends Run> extends BlueRun {
             if(blocking == null){
                 blocking = false;
             }
+            if(Boolean.getBoolean("DISABLE_STOP_BLOCKING_BEHAVIOR")){
+                blocking = false;
+            }
             try {
                 long start = System.currentTimeMillis();
                 if(timeOutInSecs == null){
```

philsttr:
@vivek , I think @michaelneale  and I had come to the conclusion that an environment variable would not be used, and the default behavior would just be changed to only call stop once.  @michaelneale , please correct me if I misunderstood.

vivek:
Thanks @philsttr, LGTM 🐝

vivek:
I am waiting for master CI build to look healthy then will merge it.

PR #1323, opened 2017-08-14
Optimizations to REST API

i386:
# Description

Serving this request took 75ms. It now takes 30ms.

before

after

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


i386:
@vivek I am not sure how to interpret the failure here - is it just saying I need more tests?

vivek:
@i386 look in `${project.build.directory}/code-coverage/report`, there should be index.html, that should give you details.

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/ModelBuilder.java
```diff
@@ -44,12 +44,26 @@
     /*package*/ final Map<Class, Model> models = new ConcurrentHashMap<Class, Model>();
 
     public <T> Model<T> get(Class<T> type) throws NotExportableException {
+        if (type.getAnnotation(ExportedBean.class) == null) {
```
vivek: Directly updating stapler imported code? This can leave us with pain later on, better to submit PR on stapler then we import them over here.

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/Property.java
```diff
@@ -141,15 +141,14 @@ public void writeTo(Object object, TreePruner pruner, DataWriter writer) throws
         if (merge) {
             // merged property will get all its properties written here
             if (d != null) {
-                Model model;
-                try {
-                    model = owner.get(d.getClass(), parent.type, name);
-                } catch (NotExportableException e) {
+                if (d.getClass().getAnnotation(ExportedBean.class) == null) {
```
vivek: Same as above, lets not update stapler code directly here. Please open a PR on stapler.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
```diff
@@ -53,7 +56,7 @@
 
     /** Date String format */
     public static final String DATE_FORMAT_STRING = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
-
+    public static final DateTimeFormatter DATE_FORMAT = DateTimeFormat.forPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
```
vivek: Can we move joda specific code to AbstractRunImpl? This will also result in to `blue-rest` module not to depend on specific data time library.

blueocean-rest/pom.xml
```diff
@@ -23,6 +23,10 @@
           <groupId>io.jenkins.blueocean.rest.annotation</groupId>
           <artifactId>capability-annotation</artifactId>
       </dependency>
+      <dependency>
```
vivek: See my other comment, we should avoid depending on specific date time library in models.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/OrganizationFolderRunImpl.java
```diff
@@ -97,6 +96,21 @@ public Date getEndTime() {
     }
 
     @Override
+    public String getStartTimeString() {
```
vivek: Maybe declare these as abstract? default implementation returning null can be misleading.

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/ModelBuilder.java
i386: Done https://github.com/stapler/stapler/pull/127

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/Property.java
i386: See above

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/OrganizationFolderRunImpl.java
i386: WDYM?

michaelneale:
Not sure about the build failure, is this a test coverage failure? 

```
[WARNING] Rule violated for bundle blueocean-rest-impl: classes missed count is 3.00, but expected maximum is 0.00
```

?? (jacoco seems annoying in every single way, except that it does enforce coverage levels)

vivek:
@michaelneale it means there are 3 classes without test coverage. Reports are in that modules `target/code-coverage/report` directory. Guess time to write Jacoco HTML publisher:)

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/Property.java
vivek: Changes in Property.java in this PR is different from the ones in PR. I think its cleaner to remove changes from here and merge stapler changes in separate PR, after stapler PR is merged. This way we can merge this PR without waiting for stapler changes and get immediate perf benefits.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/OrganizationFolderRunImpl.java
vivek: Ignore that comment, for some reason I mistook it as one of the model class. All good here:)

i386:
@vivek the only class I can see that does not have coverage is `BlueJUnitTestResult`. Would you mind taking a look? I am stumped :(

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
```diff
@@ -186,7 +224,21 @@ public BlueTestResultContainer getTests() {
...
+            try {
+                return TEST_SUMMARY.get(run.getExternalizableId(), new Callable<Optional<BlueTestSummary>>() {
+                    @Override
+                    public Optional<BlueTestSummary> call() throws Exception {
+                        BlueTestSummary summary = BlueTestResultFactory.resolve(run, parent).summary;
+                        return summary == null ? Optional.<BlueTestSummary>absent() : Optional.of(summary);
+                    }
+                }).orNull();
+            } catch (ExecutionException e) {
+                return null;
```
vivek: should log error message or else the checked exception thrown by value loaded callable will be silently ignored.

vivek:
@james since we don't have coverage report archived, if you build that module locally (Where its reporting coverage error) you should be able to see the index.html (target/code-coverage/report/index.html) and see which classes lack coverage (it includes anonymous inner classes as well).

i386:
@vivek I have removed the stapler code.

PR #1329, opened 2017-08-16
reorder source control buttons

NicuPascu:
# Description

See [JENKINS-46218](https://issues.jenkins-ci.org/browse/JENKINS-46218).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/creation/CreatePipelineScmListRenderer.jsx
+++ b/blueocean-dashboard/src/main/js/creation/CreatePipelineScmListRenderer.jsx
@@ -23,6 +23,19 @@ export class CreatePipelineScmListRenderer extends React.Component {
         this._initialize();
     }
 
+    customSortProviders(providers, sortedProviders, providerToSearch) {
+        const providerKeys = Object.keys(providers);
+
+        for (let i = 0; i <= providerKeys.length; i++) {
+            if (providers[i] && ((providers[i].constructor.name == providerToSearch) || !providerToSearch)) {
+                sortedProviders.push(providers[i]);
+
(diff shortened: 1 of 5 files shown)


blueocean-dashboard/src/main/js/creation/CreatePipelineScmListRenderer.jsx
```diff
@@ -37,8 +51,18 @@ export class CreatePipelineScmListRenderer extends React.Component {
...
 
+            // eslint-disable-next-line
+            let sortedProviders = [];
+
+            this.customSortProviders(providers, sortedProviders, 'BbCloudScmProvider');
+            this.customSortProviders(providers, sortedProviders, 'BbServerScmProvider');
+            this.customSortProviders(providers, sortedProviders, 'GithubScmProvider');
+            this.customSortProviders(providers, sortedProviders, 'GithubEnterpriseScmProvider');
+            this.customSortProviders(providers, sortedProviders, 'GitScmProvider');
+            this.customSortProviders(providers, sortedProviders); // add all other providers
```
cliffmeyers: I think perhaps we should use the `ordinal` property in `jenkins-js-extension.yaml` as it's the standard way to sort extensions for example in `Extensions.Renderer`. See https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-personalization/src/main/js/jenkins-js-extension.yaml for an example how to specify it. Then we should just be able to add a `providers.sort` call below the `providers.filter` call.

michaelneale:
@NicuPascu feedback above make sense (build failure is just noise...)

PR #1339, opened 2017-08-20
Replace Stapler JSON with Jackson

i386:
# Description

To do
- [x] Honor `ExportedBean` and `Export`
- [ ] Error handling
- [x] `Exported#skipNull`
- [ ] `Exported#visibility` / `Exported#inline` 
- [ ] `Exported#merge`
- [ ] `Exported#verboseMap`
- [ ]  Port Stapler tests to Jackson `JSON.toJson()`


i386:
@michaelneale @vivek most of the UI loads. Seems there is some problem with one of the responses (invalid characters) but the responses are valid JSON. Needs some more work.

Alternatively we can just switch to Jackson and forget the whole @ExportedBean shenanigans. 

app-store-demo 2017-08-21 11-11-29

vivek:
@i386 whats the motivation behind replacing stapler serialization with Jackson alone? There is always going to be this maintenance cost to keep it working with Stapler annotated beans. 

> Alternatively we can just switch to Jackson and forget the whole @exportedbean shenanigans.

Only for serialization? Or replace Stapler completely with jackson while preserving tree/graphQL like functionality. Having this mixed thing is always going to be a pain. Blueocean is closer to core and if we diverge from that, then we are going to be living with different ways of writing plugins where one set of annotation to make the plugin work with classic Jenkins and another set for blueocean. This is something we need to think bit carefully to keep in mind plugin authors and how they can extend blueocean APIs.

PR #1349, opened 2017-08-22
Hotfix/js extensions revert

cliffmeyers:
# Description
- Current state of js-extensions in master did not match what was last published as 0.0.36. The revert of js-extensions in #1132 was incomplete. See df42fdf which rolls back the remaining code that was added to js-extensions in #1043.
- Also a small tweak to an ATH test case that was frequently flaky.
- Discovered while implementing code cleanup for [JENKINS-46218](https://issues.jenkins-ci.org/browse/JENKINS-46218).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/GithubAddServerDialogPage.java
+++ b/acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/GithubAddServerDialogPage.java
@@ -44,6 +44,7 @@ public class GithubAddServerDialogPage {
 
 
     public void enterServerName(String name) {
+        logger.info(String.format("enter server name '%s", name));
         WebElementUtils.setText(
             wait.until(ExpectedConditions.visibilityOf(textName)),
             name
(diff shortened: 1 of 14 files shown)


cliffmeyers:
cc @NicuPascu to see if this makes sense to you, and cc @sophistifunk since you had made some test tweaks in js-extensions that I think were related to the original changes.

sophistifunk:
I don't get this, why are we reverting these changes? I put things in to make the tests pass, but the original changes must have been there for a reason?

cliffmeyers:
The code in master as it exists is completely broken and causes Blue Ocean to not load at all. Code was added in #1043 and then partially removed in #1132. From what I can tell it never made it into a prod release of js-extensions. Perhaps @NicuPascu can shed some light?

cliffmeyers:
To be clear, I downloaded the latest prod release (0.0.36) and diff'd the "dist" directory against current master build of js-extensions. There are differences in the code, and this removes them since they are not being used in any prod release.

sophistifunk:
Fair enough, sounds like something went awry in a merge and we don't know why, do the tests pass on the new reverted state?

cliffmeyers:
@sophistifunk yep :)  And my follow-on PR will add js-extensions to the CI build so we don't have failures going unnoticed.

sophistifunk:
Nicccce

cliffmeyers:
Thanks gents.

PR #1354, opened 2017-08-23
add the pipeline editor to the cleanInstall.js and checkshrinkwrap.js

NicuPascu:
# Description

Attempt to add the pipeline editor to the cleanInstall and checkshrinkwrap scripts

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


sophistifunk:
LGTM 🐝

michaelneale:
this one will need some updating if still relevant

cliffmeyers:
Per request of @NicuPascu will be taking a look at this one later today or tomorrow.

NicuPascu:
@cliffmeyers I pushed my latest attempt, fixed some things but still not working, there is a different error now, really appreciate the help

cliffmeyers:
@NicuPascu I'm going to paste in the process I took to resolve this with the hope it shares some knowledge. It will also shed some light on how painful this entire process is...

Initially the script is complaining about imprecise numbers in the devDeps: we avoid caret operator because it can allow new deps to come in when we don't explicitly ask for them, which is bad for build reproducibility:

```
npm install -D -E chai@3.5.0
npm install -D -E gulp@3.9.1
npm install -D -E gulp-mocha@2.2.0
npm install -D -E mocha@2.4.5
npm install -D -E nock@8.0.0
npm install -D -E eslint-plugin-react@4.3.0
```

(Note that I downgraded the last dep to align with the other modules: I seem to recall an issue with version 5 causing more of our rules to fail.)

After that a new set of errors is displayed by the script:

```
./bin/checkshrinkwrap.js
...
@jenkins-cd/eslint-config-jenkins@0.0.2 missing in shrinkwrap`
```
... (~20 more)

I'd wager that when you created shrinkwrap you did not include devDeps as that's an explicit flag you have to pass: `--dev`

So I ran `npm shrinkwrap --dev` and was greeted with this error:

```
npm shrinkwrap --dev
peer invalid: @jenkins-cd/js-builder@^0.0.36, required by @jenkins-cd/js-test@1.2.3
```

As to why js-test is complaining about js-builder I am not sure, since we have js-builder@0.0.59 installed. peerDeps always seemed flaky to me and I avoid them if I can. But we don't actually use js-test anyways, so...

```
npm uninstall -D @jenkins-cd/js-test
npm shrinkwrap --dev
wrote npm-shrinkwrap.json
```

Now the shrinkwrap file is significantly larger because it includes our dev dependencies as well. This is again important so that each time we run a build, it uses the same the dev deps to build our JS files. We don't want a new version of babel or something else changing the output of our build a month from now.

```
./bin/checkshrinkwrap.js
explicit version checks in shrinkwrap failed
create-hmac:1.1.4: create-hmac cannot exceed 1.1.4 due to an incompatibility with safe-buffer and buffer. the version may have changed if you recently upgraded js-builder. discard the updates in npm-shrinkwrap.json manually. see https://github.com/crypto-browserify/createHmac/issues/20
```

This was a check I added months ago. I released a new version of js-builder, but since js-builder doesn't use shrinkwrap, each time we install a new version, new dependencies come along for the ride, and can break our build. So we can also fix this with by including an explicit dep:

`npm install -D -E create-hmac@1.1.4`

```
 ./bin/checkshrinkwrap.js 
validating dependencies in /Users/cmeyers/Development/code/blueocean/blueocean-dashboard
success!
validating dependencies in /Users/cmeyers/Development/code/blueocean/blueocean-personalization
success!
validating dependencies in /Users/cmeyers/Development/code/blueocean/blueocean-web
success!
validating dependencies in /Users/cmeyers/Development/code/blueocean/blueocean-config
success!
validating dependencies in /Users/cmeyers/Development/code/blueocean/blueocean-core-js
success!
validating dependencies in /Users/cmeyers/Development/code/blueocean/blueocean-pipeline-editor
success!
validating dependencies in /Users/cmeyers/Development/code/blueocean/jenkins-design-language
success!
all dependencies look good! took 78ms
```

Ah but wait...

```
npm run test
Message:
    Cannot find module 'jsdom'
```

Seems that 7.2.2 is a popular version in other shrinkwraps, so let's try it.

```
npm install -D -E jsdom@7.2.2
npm run test
No window.localStorage. Creating a mock in-memory localStorage, assuming this is running in a test environment.
Browser configuration of @jenkins-cd/logging is explained at https://tfennelly.github.io/jenkins-js-logging/index.html#browser-config
Unexpected state. Blue Ocean preload state not on page as expected. This is okay if running in a test.
Jenskins SSE Gateway Client: This browser does not support EventSource. Where's the polyfill?


  Pipeline Syntax Converter
    ✓ converts from JSON: agent any
    ✓ converts from JSON: agent docker
    ✓ converts from JSON: single stage
    ✓ converts from JSON: parallel stage
    ✓ converts from JSON: named parameter values properly
    ✓ converts from JSON: unnamed parameter values properly
    ✓ converts from JSON: nested steps
    ✓ converts to JSON: basic
    ✓ restores unknown sections from JSON
    ✓ handles unknown steps


  10 passing (11ms)
```

I'm going to check in these changes now and say a prayer. :)

cliffmeyers:
@NicuPascu ack, I forgot to pull latest before making these changes. I'm going to use "my" version of package.json and shrink and then sort out the other issues after...

NicuPascu:
@cliffmeyers Thanks for the help and detailed explanation of the whole process, it looks like it is now ready for merge right?

cliffmeyers:
@NicuPascu yep, I will just go ahead and merge it now.

PR #1355, opened 2017-08-23
test change

michaelneale:
test change
This is just to test if jacoco failures are flaky - as master passed - this is the same as master.


michaelneale:
well, passes now so deleteing

PR #1357, opened 2017-08-23
Remove the pipeline-editor dependency

rtyler:
I heard this lives in the repo now


rtyler:
522b6c04e691b209f948b81f_736

i386:
Nah still needed in the POM root

rtyler:
ORLY, I thought because it was a module  that wasn't necessary.

3qfzt0

i386:
Nope :( you would have to repeat the version everywhere if you were to use it. Mavens :(

PR #1363, opened 2017-08-26
Task/jenkins 42283 fetch promise error handlers

cliffmeyers:
# Description
- See [JENKINS-42283](https://issues.jenkins-ci.org/browse/JENKINS-42283).
- Add `ErrorUtils` which will log unhandled JS errors, and log rejected promises that were not handled. In my opinion, this is the most important part of the PR, so we can identify errors more easily during testing or during bug reports. We can also enhance this down the road to pump the error info into support bundle or similar if we so choose.
- Added some unit tests for Fetch

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


kzantow:
... so excited about this. But is there a way to _not_ import "bluebird" all over the place?

cliffmeyers:
@kzantow yeah, that's a good question. The first time you import bluebird, I believe it overrides the global promise. Unfortunately if you are running say a unit test it may not necessarily get imported. I need to put a little more thought on whether there's a clean way to ensure it's always loaded first in every scenario.

cliffmeyers:
Actually we might be able to handle this cleanly with Jest, and in normal deployment this will get handled automagically when core-js is loaded.. I'll do a little more experimentation with it.

michaelneale:
@cliffmeyers if still relevant - want to update this? review from @NicuPascu or @sophistifunk ?

cliffmeyers:
Yes, it's relevant, and I'll update the npm deps once it gets a thumbs up.

cliffmeyers:
@sophistifunk your point is taken. The one thing we really need to evaluate in addition to basic Promise support is whether the "unhandledrejection" event is supported. Right now it doesn't look so good:

https://developer.mozilla.org/en-US/docs/Web/Events/unhandledrejection

PR #1364, opened 2017-08-28
[JENKINS-46452] Don't reference null Jira session

MarkEWaite:
[JENKINS-46452](https://issues.jenkins-ci.org/browse/JENKINS-46452) reports null pointer exceptions which cause build failures if the Jira plugin is configured without credentials.

I use the Jira plugin to rewrite bug report URL's, but do not want it to submit updates to the bugs.  This change should allow my use model to work again.

My apologies that I have not written a test for the bug.  I'm OK if you reject this change in favor of a change which includes a test that shows the failure.

My apologies that I don't have a reliable set of steps to duplicate the problem.  I suspect that the steps would include something like:

- Build the docker image from [my lts-with-plugins branch](https://github.com/MarkEWaite/docker-lfs/tree/lts-with-plugins)
- Execute the jobs in the "Bugs - Individual Checks" folder
- Review failures, checking for a null pointer exception as described in the bug report

# Submitter checklist

- [x] Appropriate unit or acceptance tests or **explanation to why this change has no tests**

# Reviewer checklist

- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/jira/JiraSCMListener.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/jira/JiraSCMListener.java
@@ -36,6 +36,9 @@ public class JiraSCMListener extends SCMListener {
         }
         String jql = constructJQLQuery(issueKeys);
         JiraSession session = jiraSite.getSession();
+        if (session == null) {
+            return;
+        }
         // Query for JIRA issues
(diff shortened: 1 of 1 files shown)


rtyler:
FWIW, calls to `getSession()` should [always check for null return values](https://github.com/jenkinsci/jira-plugin/blob/master/src/main/java/hudson/plugins/jira/JiraSite.java#L256) from my understanding. 

JiraSite itself [does many null checks](https://github.com/jenkinsci/jira-plugin/blob/master/src/main/java/hudson/plugins/jira/JiraSite.java#L452-L455) when it calls getSession()

i386:
This is very reasonable - thanks for the PR 👍

PR #1366, opened 2017-08-29
JENKINS-38442 Extract layout logic from PipelineGraph into separate function + Unit Tests

sophistifunk:
Squashed commit of the following:

    hackweek/complex-graphs * Add a comprehensive test-case for the layout algorithm to ensure we don't break it during modifications
    hackweek/complex-graphs * Update existing test to use new interface to layout
    hackweek/complex-graphs * Flow and eslint fixes
    hackweek/complex-graphs * Move ypStart into the Layout object and remove the old busted constructor usecase from storybooks
    hackweek/complex-graphs * Extract layout into a single-entry function in separate file for reuse
    hackweek/complex-graphs * Move pipeline graph components into their own subdir, extract types into a model file

# Description

See [JENKINS-38442](https://issues.jenkins-ci.org/browse/JENKINS-38442).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/jenkins-design-language/src/js/components/pipeline/PipelineGraphLayout.jsx
+++ b/jenkins-design-language/src/js/components/pipeline/PipelineGraphLayout.jsx
@@ -1,0 +1,230 @@
+// @flow
+
+import React from 'react';
+
+import type {
+    NodeColumn,
+    LabelInfo,
+    LayoutInfo,
+    StageInfo,
+} from './PipelineGraphModel';
+
+// TODO: Docs
+export function layoutGraph(newStages: Array<StageInfo>, layout: LayoutInfo) {
+
+    const stageNodeColumns = createNodeColumns(newStages);
+    const { nodeSpacingH, ypStart } = layout;
+
+    const startNode = {
+        x: 0,
+        y: 0,
+        name: 'Start',
+        id: -1,
+        isPlaceholder: true,
+        key: 'start-node',
+        type: 'start',
+    };
+
+    const endNode = {
+        x: 0,
+        y: 0,
+        name: 'End',
+        id: -2,
+        isPlaceholder: true,
+        key: 'end-node',
+        type: 'end',
+    };
+
+    const allNodeColumns = [
+        { nodes: [startNode] },
+        ...stageNodeColumns,
+        { nodes: [endNode] },
+    ];
+
+    positionNodes(allNodeColumns, layout);
+
+    const bigLabels = createBigLabels(allNodeColumns);
+    const smallLabels = createSmallLabels(allNodeColumns);
+    const connections = createConnections(allNodeColumns);
+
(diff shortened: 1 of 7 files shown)


sophistifunk:
Extracted the layout code and removed any references to the component, to make modifications and testing easier. Should also make life easier when updating the editor to use any future improvements to the layout code.

Added a thorough test that checks pixel positions and connections to keep it robust while I screw around with the algorithm to support more complex graphs.

Should not have any visible effect, and doesn't need a re-publish of JDL just for this patch.

sophistifunk:
Also: Moved ypStart from a const into the layout object

jenkins-design-language/src/js/components/pipeline/PipelineGraphLayout.jsx
```diff
@@ -0,0 +1,230 @@
...
+
+        previousTopNode = topNode;
+    }
+}
+
+
+/**
+ * Generate label descriptions for big labels at the top of each column
+ */
+function createBigLabels(columns: Array<NodeColumn>) {
```
cliffmeyers: Would `createColumnLabels` be better name?

jenkins-design-language/src/js/components/pipeline/PipelineGraphLayout.jsx
```diff
@@ -0,0 +1,230 @@
...
+        });
+    }
+
+    return labels;
+}
+
+/**
+ * Generate label descriptions for small labels under the nodes
+ */
+function createSmallLabels(columns: Array<NodeColumn>) {
```
cliffmeyers: Perhaps `createNodeLabels`?

jenkins-design-language/src/js/components/pipeline/PipelineGraphLayout.jsx
```diff
@@ -0,0 +1,230 @@
...
+
+    const labels = [];
+
+    for (const column of columns) {
+        if (column.nodes.length === 1) {
```
cliffmeyers: Should we null check `column.nodes` first?

jenkins-design-language/src/js/components/pipeline/PipelineGraphLayout.jsx
sophistifunk: I'm not sure about these two, as single-node columns don't have a small label, just a big label... I'll mull it over while I'm in here screwing around tho :)

jenkins-design-language/src/js/components/pipeline/PipelineGraphLayout.jsx
sophistifunk: That's `flow`'s job :) All the columns are generated within this module, they're not inputs so if it's bolloxed I want errors.

PR #1368, opened 2017-08-31
Stage trend impl

i386:
# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/StageRuntimeTrend.java
```diff
@@ -0,0 +1,101 @@
...
+        @Override
+        public String getId() {
+            return run.getId();
+        }
+
+        @Override
+        public Object toExportedObject() {
+            ImmutableMap.Builder<String, Object> builder = ImmutableMap.builder();
+            builder.put("id", getId());
+            for (BluePipelineNode node : run.getNodes()) {
```
i386: @vivek is this the right way to list out the stages of the pipeline?

cliffmeyers:
@i386 Made a few changes here:
- Tweak to `StageRuntimeTrend` to do a no-op impl of new `BlueTable.getColumns` method. Since the column names (really the names of the stages) can vary from run to run, it's left empty.
- Add some UI logic to handle a Trend with no `getColumns` impl: it scans all the rows to find all unique column names. 
- Add a UI extension mechanism so that totally custom React charts can be built.

cliffmeyers:
I think we can merge this back into topic/trends whenever... just waiting for CI to pass.

PR #1369, opened 2017-09-01
Bug/jenkins 46348 46417 46374 change the way material icons fill color is passed along

NicuPascu:
# Description

This PR removes the inline css fill rule for the material icons, uses the fill attr instead making css specificity rules work again

See [JENKINS-46348](https://issues.jenkins-ci.org/browse/JENKINS-46348) [JENKINS-46417](https://issues.jenkins-ci.org/browse/JENKINS-46417) [JENKINS-46374](https://issues.jenkins-ci.org/browse/JENKINS-46374).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-core-js/src/less/run-button.less
```diff
@@ -11,6 +11,18 @@
...
+            svg {
+                fill: #4A90E2;
+            }
+        }
+
+        svg {
+            fill: #ffffff;
+        }
+    }
+
```
sophistifunk: Can we integrate this with the `.action-button-colors` mixin? It's used just below this. Ideally we can re-use that mixin with the more-specific selector, or if there's a problem with the mixin we might want to update it? It's also defined in dashboard.

Ideally we want to get away from using hex codes in individual-component `.less` files, and reuse (or define new) colour/style names in `variables.less`

sophistifunk: This stuff is going to need some overhaul in the next couple of months with the need for theming coming from Blue Steel, so the fewer places we need to look for codes, the easier that process will go.

blueocean-dashboard/src/main/js/components/karaoke/components/LogToolbar.jsx
```diff
@@ -24,15 +24,15 @@ export default class LogToolbar extends Component {
                     href: logUrl,
                 }}
                 >
-                    <Icon size={24} {...{ color: '#4a4a4a', icon: 'ActionLaunch' }} />
```
kzantow: so happy to get rid of any of this type of coding!

jenkins-design-language/src/js/components/material-ui/SvgIcon/SvgIcon.js
```diff
@@ -90,6 +78,7 @@ class SvgIcon extends Component {
         {...other}
         onMouseEnter={this.handleMouseEnter}
         onMouseLeave={this.handleMouseLeave}
+        fill={color ? color : 'white'}
```
kzantow: is this necessary?

blueocean-core-js/src/less/run-button.less
kzantow: In general, it would be simple to just override styles with a downstream CSS file, or build a custom one from the original. Don't think it's necessary to allow custom colors and styles etc. to be passed in. Not sure exactly what you're saying, we might be talking about exactly the same thing :)

jenkins-design-language/src/js/components/material-ui/SvgIcon/SvgIcon.js
NicuPascu: that fill attr is kind if necessary because white was the default color in the react-material-icons, if I remove that, a lot of icons become black and we need to write css to make them white again, I think white is a good default color because in a lot of places that we use the icons they are white, there is no place where they are black so I would leave this for convenience

kzantow: Can't we just make a low-specificity CSS rule that defaults them to white? e.g. add a `className="icon"` to `SvgIcon` and then use `svg.icon { fill: #fff }` ?

NicuPascu: yup, will do that, I will add the other default css icon styles in that class as well

michaelneale:
:bee:

jenkins-design-language/src/js/components/material-ui/SvgIcon/SvgIcon.js
```diff
@@ -70,24 +56,18 @@ class SvgIcon extends Component {
...
-      userSelect: 'none',
-      transition: 'all 450ms cubic-bezier(0.23, 1, 0.32, 1) 0ms', 
     }, style);
 
     delete other.iconName;
 
     return (
       <svg
         {...other}
+        className="material-icon"
```
kzantow: Not sure if there are any instances, but `className` might get swallowed if it's passed in `other`

sophistifunk:
bbm

PR #1372, opened 2017-09-05
More java ATH

imeredith:
A few tests ported from JS. But more importantly support for Pipeline jobs which JS ath uses a lot. As well as some refactoring


acceptance-tests/src/main/java/io/blueocean/ath/model/MultiBranchPipeline.java
```diff
@@ -12,7 +12,7 @@
 import java.io.IOException;
 import java.util.List;
 
-public class MultiBranchPipeline extends Pipeline{
+public class MultiBranchPipeline extends AbstractPipeline {
```
cliffmeyers: Should this override `isMultibranch()` to return true?

acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/RunDetailsArtifactsPage.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Inject
+    public RunDetailsArtifactsPage(WebDriver driver, @Assisted AbstractPipeline pipeline) {
+        this.driver = driver;
+        this.pipeline = pipeline;
+        PageFactory.initElements(driver, this);
+    }
+
+    public void open(String pipeline, Integer runNumber) {
+        driver.get(base+"/blue/organizations/jenkins/"+ pipeline + "/detail/master/" + runNumber +"/artifacts");
```
cliffmeyers: `pipeline` probably needs URI encoding to support jobs nested in folders?

acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/RunDetailsArtifactsPage.java
```diff
@@ -0,0 +1,75 @@
...
+    public void checkUrl(String branch, int runNumber) {
+        wait.until(ExpectedConditions.urlContains(getUrl(branch, runNumber)), 30000);
+    }
+    public String getUrl(int runNumber) {
+        return getUrl(null, runNumber);
+    }
+
+    public String getUrl(String branch, int runNumber) {
+        if(pipeline.isMultiBranch()) {
+            String tempBranch = branch == null ? "master" : branch;
```
cliffmeyers: May also need URI encoding for branch name.

acceptance-tests/src/main/java/io/blueocean/ath/model/Pipeline.java
```diff
@@ -1,81 +1,44 @@
...
+import java.io.BufferedInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.nio.file.Files;
 
-    public Folder getFolder() {
-        return folder;
+public class Pipeline extends AbstractPipeline {
```
cliffmeyers: `ClassicPipeline` might be a better name, just to differentiate it from our more generic use of "pipeline" throughout.

acceptance-tests/src/main/java/io/blueocean/ath/model/MultiBranchPipeline.java
imeredith: yes

acceptance-tests/src/main/java/io/blueocean/ath/model/Pipeline.java
imeredith: Ah yeah, i was trying to figure our job vs pipeline. Sounds lke a good idea

michaelneale:
@imeredith you picking this up?

michaelneale:
LGTM - if @cliffmeyers is satisfied with his objections above

PR #1257, opened 2017-07-17
[FIX JENKINS-42602] Convert JDL and core-js to HPI & auto-watch all plugins to bundle

kzantow:
# Description

Convert JDL and core-js into plugins to avoid NPM publishing dance.

[No NPM required!](https://www.youtube.com/watch?v=dOtN5uol1Tc)

NOTE: there is a beta version of this published: https://github.com/jenkinsci/js-builder/pull/19 and it'll need to be merged for future releases.

See [JENKINS-42602](https://issues.jenkins-ci.org/browse/JENKINS-42602).

# Submitter checklist
- [x] Link to JIRA ticket in description, if appropriate.
- [x] Change is code complete and matches issue description
- [ ] Appropriate unit or acceptance tests or explanation to why this change has no tests
- [x] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below.
- [ ] infra permissions repo updated with any new plugins added here

# Reviewer checklist
- [ ] Run the changes and verified the change matches the issue description
- [ ] Reviewed the code
- [ ] Verified that the appropriate tests have been written or valid explanation given

PR #1380, opened 2017-09-08
JENKINS-45920 - add link to Jira tickets in the commit table

NicuPascu:
Adds link to Jira tickets in branch, activity and run details changes

[JENKINS-45920](https://issues.jenkins-ci.org/browse/JENKINS-45920) requests the JIRA tickets to also be in the run details header, for this to happen we need changes in the backend to add the issue related to the cause to the run object

See [JENKINS-45920](https://issues.jenkins-ci.org/browse/JENKINS-45920).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
@NicuPascu update this to master, and then the test failure should go away

can you possibly add some screen shots of this in action for review? (I am sure it is fine functionally) - code LGTM just a lot of changes

perhaps @cliffmeyers code review?

NicuPascu:
Sure, here they are:

jira_link
jira_link2

michaelneale:
nice @NicuPascu what happens when there is a `#THING-123` type link in a commit that isn't really to a valid jira issue? just a random hash entry of text?

NicuPascu:
@michaelneale I use the "issues" obj that was added in this PR: https://github.com/jenkinsci/blueocean-plugin/pull/1290

I search for the id's in the message texts and if an exact match is found I add a link to it, so it has to be a valid jira issue.

michaelneale:
seems ok with me - if @cliffmeyers could bless it, coudl merge it?

blueocean-dashboard/src/main/js/components/RunMessageCell.jsx
```diff
@@ -25,22 +28,73 @@ export default class RunMessageCell extends Component {
...
+            if (run.changeSet[run.changeSet.length - 1].issues) {
+                let issuesIdString = '';
+                const issuesObj = {};
+
+                for (const issue of run.changeSet[run.changeSet.length - 1].issues) {
+                    issuesIdString += `${issue.id}|`;
+                    issuesObj[issue.id] = issue.url;
+                }
+
+                if (issuesIdString) {
```
sophistifunk: Is this block of code duplicating the functionality of the `AddIssuesLinksToMsg` component?

blueocean-dashboard/src/main/js/components/ActivityDetailsRow.jsx
```diff
@@ -87,7 +87,7 @@ class ActivityDetailsRow extends Component {
                 <TableCell linkTo={runDetailsUrl}>{run.id}</TableCell>
                 <TableCell linkTo={runDetailsUrl}><CommitId commitId={run.commitId} /></TableCell>
                 { isMultibranch && <TableCell linkTo={runDetailsUrl}>{branchName}</TableCell> }
-                <TableCell linkTo={runDetailsUrl}><RunMessageCell run={run} t={t} /></TableCell>
+                <TableCell><RunMessageCell linkTo={runDetailsUrl} run={run} t={t} /></TableCell>
```
imeredith: Why did you change where the linkTo happens? from what I remeber one of the reasons we need to have the link in the TableCell is to properly cover the whole cell. This may not be an issue here, just wondering

blueocean-dashboard/src/main/js/components/RunMessageCell.jsx
```diff
@@ -1,15 +1,18 @@
 import React, { Component, PropTypes } from 'react';
 import Lozenge from './Lozenge';
+import { Link } from 'react-router';
 
 export default class RunMessageCell extends Component {
     propTypes = {
         run: PropTypes.object,
         t: PropTypes.func,
+        linkTo: PropTypes.string,
```
imeredith: Why linkTo in here vs TableCell?

blueocean-dashboard/src/main/js/components/RunMessageCell.jsx
```diff
@@ -1,15 +1,18 @@
...
     };
 
     render() {
         const run = this.props.run;
         const t = this.props.t;
+        const linkTo = this.props.linkTo || '';
```
imeredith: Seems like if we needed to link in here we could just generate it from the run data. We cant do that with TableCell so we have to pass it in. Seems redundant here and just something else to remember to do.

blueocean-dashboard/src/main/js/components/RunMessageCell.jsx
NicuPascu: @sophistifunk I just did a project wide search of AddIssuesLinksToMsg and got no results, can you point me in the right direction to find this component? I have no idea what it does.

blueocean-dashboard/src/main/js/components/ActivityDetailsRow.jsx
NicuPascu: I had to remove the whole link over the cell because a tags cannot contain other a tags, so to have links in the message, I had to remove the a tag over the whole cell and only add it to the text that is not linked to any jira issue

blueocean-dashboard/src/main/js/components/RunMessageCell.jsx
sophistifunk: Unless Github is having a stroke, you added it to RunDetailsChanges in this PR?

blueocean-dashboard/src/main/js/components/RunMessageCell.jsx
NicuPascu: Given the fact that I removed the link from the TableCell,I thought that passing it along where the table is constructed made most sense as it was already generated. I can definitely generate the link from the run but I will also have to pass along the pipeline  because buildRunDetailsUrl needs it to generate the url. I am still not sure that would make it better in the long run. What do you think?

blueocean-dashboard/src/main/js/components/RunMessageCell.jsx
NicuPascu: argh, my bad, I was on the wrong branch and didn't realise, there is a difference between the two codes, the one in RunMessageCell adds a link to the text that is not a jira issue, in RunDetailsChanges the AddIssuesLinksToMsg component only adds the jira links to a text

sophistifunk: Righto, just wanted to check as it seemed pretty similar :)

NicuPascu: I preferred to have them separate because the code is pretty complicated as is and if I make the component take another optional link prop for the text that is not a jira issue it would add another bit of complexity that would make it even harder to maintain, what do you think, should I make AddIssuesLinksToMsg a separate component that takes in that optional prop and use it in both use cases?

cliffmeyers: I am wondering if we could solve this by having a more general purpose component called `LinkifiedText` which would take in two props:
1. `text:String` as the string of text (commit message, or whatever)
2. `links:Map` where the key is the text that should be linked, and the value is the absolute/relative URL. 

If there are no links, the component could just return a `span` with the `text` as its child. That component wouldn't be too hard to impl and could have a unit test or two.

Then each part of the UI that needs linkified text just implements a simple pure function that takes in the data of interest (commit, changeSet, etc) and builds the `links` Map. If `links` is empty, then we know there are no links and we can fall back to the normal linking behavior.

Note: we don't need to use an actual ES Map impl, just an object literal would be fine.

I may have missed a subtle difference but if so, please clarify and we can find the best way to resolve it.

NicuPascu:
@cliffmeyers I have refactored the code so now there is a separate component that accepts the following props: text, textLink (the optional link for the text ), issues obj (the issues obj as it is passed along from the backend in the changeset). I have also written some tests for the new component.

blueocean-dashboard/src/main/js/components/LinkifiedText.jsx
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        return (<span>{textWithIssues}</span>);
+    }
+
+    return (textLink ? (<Link to={textLink} className="unstyled-link" >{text}</Link>) : (<span>{text}</span>));
+};
+
+LinkifiedText.propTypes = {
+    issues: PropTypes.object,
```
cliffmeyers: This is an improvement, although it's still coupled to the `issues` structure which makes it difficult to reuse in other contexts - and linking to external systems is a pretty realistic future use case. Would you like me to take a quick stab at this or perhaps @michaelneale can weigh in on whether I'm being too picky :)

NicuPascu: No I can change the obj structure if needed, I just thought that it would be better to keep it as it comes from the backend because it was pretty clear and it was made especially for this use case (https://github.com/jenkinsci/blueocean-plugin/pull/1290), it just seems redundant to me to format it one way in the backend and then reformat it in the frontend before using it, maybe change the way it is sent from the backend?

cliffmeyers: I am doing a poor job of explaining this, so let me try again:  suppose we have another place in the UI where we display a string of text and we want to replace certain portions of the text with hyperlinks. This could be Github PR's or anything else we dream up. The idea behind a generic `LinkifyText` component was that it would be generic and not coupled to JIRA, Github, or any other 3rd party service. So rather than having an `issues` prop, it would have a more generic prop of key/value pairs which indicate what a text fragment should be linked to. This component could be used in any context, JIRA, Github, whatever. 

In the specific case of JIRA, it would require some code outside of `LinkifyText` to transform the `issues` structure into the key/value pairs that `LinkifyText` expects. You could even create a `JiraIssueLinks` component which handles that data transform then uses the `LinkifyText`component internally. This kind of component composition is generally a favorable thing to strive for.

Please let me know if that makes sense. If not, I may create a quick branch tomorrow to demonstrate the idea.

michaelneale:
hey @NicuPascu why not build this one off the imminent: https://github.com/jenkinsci/blueocean-plugin/pull/1257 change? would be so much easier ;)

cliffmeyers:
@NicuPascu LGTM! 🐝

michaelneale:
nice  has tests. @cliffmeyers is happy

:bee: from me.

PR #1383, opened 2017-09-11
JENKINS-45333# PCT fix: upgrade plugin pom

vivek:
# Description

See [JENKINS-45333](https://issues.jenkins-ci.org/browse/JENKINS-45333).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-bitbucket-pipeline/pom.xml
+++ b/blueocean-bitbucket-pipeline/pom.xml
@@ -37,17 +37,8 @@
     </dependency>
 
     <dependency>
-      <groupId>org.apache.httpcomponents</groupId>
-      <artifactId>fluent-hc</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.httpcomponents</groupId>
-      <artifactId>httpclient</artifactId>
-    </dependency>
-
-    <dependency>
-      <groupId>org.apache.httpcomponents</groupId>
-      <artifactId>httpmime</artifactId>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>apache-httpcomponents-client-4-api</artifactId>
     </dependency>
 
     <dependency>
--- a/blueocean-bitbucket-pipeline/pom.xml
+++ b/blueocean-bitbucket-pipeline/pom.xml
@@ -20,6 +20,15 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>blueocean-pipeline-api-impl</artifactId>
+      <exclusions>
+        <!-- Upper bound fix: credentials plugin is bringing it's own version of anlr4-runtime,
+             wiremock uses later version
+        -->
+        <exclusion>
+          <groupId>org.antlr</groupId>
+          <artifactId>antlr4-runtime</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
     <dependency>
--- a/blueocean-bitbucket-pipeline/pom.xml
+++ b/blueocean-bitbucket-pipeline/pom.xml
@@ -21,7 +21,7 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>blueocean-pipeline-api-impl</artifactId>
       <exclusions>
-        <!-- Upper bound fix: credentials plugin is bringing it's own version of anlr4-runtime,
+        <!-- Upper bound fix: credentials plugin is bringing its own version of antlr4-runtime,
              wiremock uses later version
         -->
         <exclusion>
@@ -39,6 +39,10 @@
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>fluent-hc</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
     </dependency>
 
     <dependency>
--- a/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/BlueOceanConfigProperties.java
+++ b/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/BlueOceanConfigProperties.java
@@ -6,8 +6,10 @@ package io.jenkins.blueocean.commons;
  * @author Ivan Meredith
  */
 public class BlueOceanConfigProperties {
-    public static final boolean ROLLBAR_ENABLED = Boolean.getBoolean("BLUEOCEAN_ROLLBAR_ENABLED");
+    public static final String BLUEOCEAN_FEATURE_JWT_AUTHENTICATION_PROPERTY = "BLUEOCEAN_FEATURE_JWT_AUTHENTICATION";
+    public static final String BLUEOCEAN_ROLLBAR_ENABLED_PROPERTY = "BLUEOCEAN_ROLLBAR_ENABLED";
 
-    public static final boolean BLUEOCEAN_FEATURE_JWT_AUTHENTICATION = Boolean.getBoolean("BLUEOCEAN_FEATURE_JWT_AUTHENTICATION");
+    public static final boolean ROLLBAR_ENABLED = Boolean.getBoolean(BLUEOCEAN_ROLLBAR_ENABLED_PROPERTY);
+    public static final boolean BLUEOCEAN_FEATURE_JWT_AUTHENTICATION = Boolean.getBoolean(BLUEOCEAN_FEATURE_JWT_AUTHENTICATION_PROPERTY);
 
 }
--- a/blueocean-github-pipeline/pom.xml
+++ b/blueocean-github-pipeline/pom.xml
@@ -26,6 +26,13 @@
     <dependency>
       <groupId>org.kohsuke</groupId>
       <artifactId>github-api</artifactId>
+      <exclusions>
+        <!-- Upper bound fix: Pick the one from jenkins core, it has later version 2.4-->
+        <exclusion>
+          <groupId>commons-io</groupId>
+          <artifactId>commons-io</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
     <dependency>
@@ -62,6 +69,13 @@
       <groupId>com.github.tomakehurst</groupId>
       <artifactId>wiremock</artifactId>
       <scope>test</scope>
+      <exclusions>
+        <!-- Upper bound fix: httpclient comes from blueocean-rest-impl with later version than the one in wiremock -->
+        <exclusion>
+          <groupId>org.apache.httpcomponents</groupId>
+          <artifactId>httpclient</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
   </dependencies>
 
--- a/blueocean-github-pipeline/pom.xml
+++ b/blueocean-github-pipeline/pom.xml
@@ -52,12 +52,8 @@
     </dependency>
 
     <dependency>
-      <groupId>org.apache.httpcomponents</groupId>
-      <artifactId>fluent-hc</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.httpcomponents</groupId>
-      <artifactId>httpclient</artifactId>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>apache-httpcomponents-client-4-api</artifactId>
     </dependency>
 
     <!-- test dependencies -->
--- a/blueocean-github-pipeline/pom.xml
+++ b/blueocean-github-pipeline/pom.xml
@@ -21,6 +21,14 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>blueocean-pipeline-api-impl</artifactId>
+      <exclusions>
+        <!-- Upper bound fix: credentials plugin is bringing it's own version of anlr4-runtime,
+        wiremock uses later version -->
+        <exclusion>
+          <groupId>org.antlr</groupId>
+          <artifactId>antlr4-runtime</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
     <dependency>
--- a/blueocean-github-pipeline/pom.xml
+++ b/blueocean-github-pipeline/pom.xml
@@ -22,7 +22,7 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>blueocean-pipeline-api-impl</artifactId>
       <exclusions>
-        <!-- Upper bound fix: credentials plugin is bringing it's own version of anlr4-runtime,
+        <!-- Upper bound fix: credentials plugin is bringing its own version of antlr4-runtime,
         wiremock uses later version -->
         <exclusion>
           <groupId>org.antlr</groupId>
@@ -54,6 +54,10 @@
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>fluent-hc</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
     </dependency>
 
     <!-- test dependencies -->
--- a/blueocean-github-pipeline/pom.xml
+++ b/blueocean-github-pipeline/pom.xml
@@ -24,15 +24,8 @@
     </dependency>
 
     <dependency>
-      <groupId>org.kohsuke</groupId>
+      <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>github-api</artifactId>
-      <exclusions>
-        <!-- Upper bound fix: Pick the one from jenkins core, it has later version 2.4-->
-        <exclusion>
-          <groupId>commons-io</groupId>
-          <artifactId>commons-io</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
 
     <dependency>
@@ -48,6 +41,11 @@
     <dependency>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>pubsub-light</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>fluent-hc</artifactId>
     </dependency>
 
     <!-- test dependencies -->
--- a/blueocean-jwt/src/main/java/io/jenkins/blueocean/auth/jwt/impl/JwtAuthenticationFilter.java
+++ b/blueocean-jwt/src/main/java/io/jenkins/blueocean/auth/jwt/impl/JwtAuthenticationFilter.java
@@ -4,7 +4,6 @@ import hudson.Extension;
 import hudson.init.Initializer;
 import hudson.util.PluginServletFilter;
 import io.jenkins.blueocean.auth.jwt.JwtTokenVerifier;
-import io.jenkins.blueocean.commons.BlueOceanConfigProperties;
 import org.acegisecurity.Authentication;
 import org.acegisecurity.context.SecurityContext;
 import org.acegisecurity.context.SecurityContextHolder;
@@ -21,6 +20,8 @@ import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import java.io.IOException;
 
+import static io.jenkins.blueocean.commons.BlueOceanConfigProperties.BLUEOCEAN_FEATURE_JWT_AUTHENTICATION_PROPERTY;
+
 /**
  * {@link Filter} that processes JWT token
  *
@@ -32,6 +33,7 @@ public class JwtAuthenticationFilter implements Filter {
      * Used to mark requests that had a valid JWT token.
      */
     private static final String JWT_TOKEN_VALIDATED = JwtAuthenticationFilter.class.getName()+".validated";
+    private boolean isJwtEnabled;
 
     @Initializer(fatal=false)
     public static void init() throws ServletException {
@@ -40,7 +42,15 @@ public class JwtAuthenticationFilter implements Filter {
 
     @Override
     public void init(FilterConfig filterConfig) throws ServletException {
-        // noop
+        /**
+         * Initialize jwt enabled flag by reading BLUEOCEAN_FEATURE_JWT_AUTHENTICATION_PROPERTY jvm property
+         *
+         * {@link io.jenkins.blueocean.commons.BlueOceanConfigProperties.BLUEOCEAN_FEATURE_JWT_AUTHENTICATION} doesn't
+         * work in certain test scenario - specially when test sets this JVM property to enable JWT but this class has
+         * already been loaded setting it to false.
+         *
+         */
+        this.isJwtEnabled = Boolean.getBoolean(BLUEOCEAN_FEATURE_JWT_AUTHENTICATION_PROPERTY);
     }
 
     @Override
@@ -94,7 +104,7 @@ public class JwtAuthenticationFilter implements Filter {
      * Returns true for requests that JWT token processing should apply.
      */
     protected boolean shouldApply(HttpServletRequest req) {
-        if (!BlueOceanConfigProperties.BLUEOCEAN_FEATURE_JWT_AUTHENTICATION)
+        if (!isJwtEnabled)
             return false;
 
         String path = req.getRequestURI().substring(req.getContextPath().length());
--- a/blueocean-pipeline-editor/src/main/java/io/blueocean/rest/pipeline/editor/PipelineMetadataService.java
+++ b/blueocean-pipeline-editor/src/main/java/io/blueocean/rest/pipeline/editor/PipelineMetadataService.java
@@ -1,24 +1,24 @@
 package io.blueocean.rest.pipeline.editor;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.List;
-import java.util.Set;
-
+import hudson.Extension;
+import hudson.ExtensionList;
 import hudson.FilePath;
 import hudson.Launcher;
 import hudson.model.Describable;
+import hudson.model.Descriptor;
+import hudson.security.csrf.CrumbIssuer;
 import hudson.tasks.Builder;
 import hudson.tasks.Publisher;
 import hudson.tools.ToolDescriptor;
 import hudson.tools.ToolInstallation;
+import io.jenkins.blueocean.commons.stapler.TreeResponse;
+import io.jenkins.blueocean.rest.ApiRoutable;
+import jenkins.model.Jenkins;
 import jenkins.tasks.SimpleBuildStep;
 import org.jenkinsci.plugins.pipeline.modeldefinition.agent.DeclarativeAgent;
 import org.jenkinsci.plugins.pipeline.modeldefinition.agent.DeclarativeAgentDescriptor;
-import org.jenkinsci.plugins.pipeline.modeldefinition.ast.ModelASTStep;
 import org.jenkinsci.plugins.pipeline.modeldefinition.model.BuildCondition;
+import org.jenkinsci.plugins.pipeline.modeldefinition.validator.BlockedStepsAndMethodCalls;
 import org.jenkinsci.plugins.structs.SymbolLookup;
 import org.jenkinsci.plugins.structs.describable.DescribableModel;
 import org.jenkinsci.plugins.workflow.steps.Step;
@@ -26,15 +26,13 @@ import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 import org.kohsuke.stapler.NoStaplerConstructorException;
 import org.kohsuke.stapler.verb.GET;
 
-import hudson.Extension;
-import hudson.ExtensionList;
-import hudson.model.Descriptor;
-import hudson.security.csrf.CrumbIssuer;
-import io.jenkins.blueocean.commons.stapler.TreeResponse;
-import io.jenkins.blueocean.rest.ApiRoutable;
-import jenkins.model.Jenkins;
-
 import javax.annotation.CheckForNull;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.List;
+import java.util.Set;
 
 /**
  * This provides and Blueocean REST API endpoint to obtain pipeline step metadata.
@@ -186,7 +184,7 @@ public class PipelineMetadataService implements ApiRoutable {
 
     private boolean includeStep(StepDescriptor d) {
         boolean include = true;
-        if (ModelASTStep.getBlockedSteps().containsKey(d.getFunctionName())) {
+        if (BlockedStepsAndMethodCalls.blockedInSteps().containsKey(d.getFunctionName())) {
             include = false;
         } else if (d.isAdvanced()
                 && !INCLUDED_ADVANCED_STEPS.contains(d.getFunctionName())) {
--- a/blueocean-rest-impl/pom.xml
+++ b/blueocean-rest-impl/pom.xml
@@ -83,6 +83,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>fluent-hc</artifactId>
+            <scope>test</scope>
+        </dependency>
+        
+        <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>junit</artifactId>
         </dependency>
@@ -122,7 +128,6 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>matrix-auth</artifactId>
-            <version>1.7</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/blueocean-rest-impl/pom.xml
+++ b/blueocean-rest-impl/pom.xml
@@ -72,22 +72,9 @@
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
-            <artifactId>token-macro</artifactId>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>matrix-project</artifactId>
             <scope>test</scope>
         </dependency>
-
-        <dependency>
-            <groupId>org.apache.httpcomponents</groupId>
-            <artifactId>fluent-hc</artifactId>
-            <scope>test</scope>
-        </dependency>
-        
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>junit</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -145,11 +145,6 @@
 
   <dependencies>
       <dependency>
-          <groupId>junit</groupId>
-          <artifactId>junit</artifactId>
-          <scope>test</scope>
-      </dependency>
-      <dependency>
           <groupId>org.mockito</groupId>
           <artifactId>mockito-core</artifactId>
           <scope>test</scope>
@@ -167,6 +162,7 @@
       <dependency>
           <groupId>org.apache.httpcomponents</groupId>
           <artifactId>httpclient</artifactId>
+          <scope>test</scope>
       </dependency>
       <dependency>
           <groupId>org.powermock</groupId>
@@ -483,14 +479,6 @@
             <version>3.5.0</version>
         </dependency>
 
-        <!-- Needed by blueocean-rest-impl for ssh unititlies -->
-        <dependency>
-            <groupId>org.jenkins-ci.plugins</groupId>
-            <artifactId>jsch</artifactId>
-            <version>0.1.54.1</version>
-        </dependency>
-
-
         <!-- Other -->
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
@@ -529,7 +517,9 @@
                     <artifactId>guava</artifactId>
                 </exclusion>
 
-                <!-- Upper bound dependency fix -->
+                <!-- Upper bound dependency fix
+                    TODO: Remove it after https://github.com/jenkinsci/jira-plugin/pull/130 is merged and released.
+                -->
                 <exclusion>
                     <groupId>org.springframework</groupId>
                     <artifactId>spring-beans</artifactId>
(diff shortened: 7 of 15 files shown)


michaelneale:
hrm 

```
[ERROR] Failed to execute goal on project blueocean: Could not resolve dependencies for project io.jenkins.blueocean:blueocean:hpi:1.3.0-beta-3-SNAPSHOT: Could not find artifact org.jenkins-ci.main:jenkins-war:executable-war:2.46.3 in cache.blueocean.io (http://172.18.128.166/mirrors/jenkins) -> [Help 1]
```

that is an odd error - it should be a pull through cache @ydubreuil ?

jglick:
The error comes from the improper override of the version of `maven-hpi-plugin` in `blueocean/pom.xml`. You should pick up the version specified in the parent. (Probably the same for `docker/official/pom.xml`.)

jglick:
By the way the `groovy-maven-plugin` execution in `blueocean/pom.xml` should be deleted. This was rendered obsolete long ago by https://github.com/jenkinsci/maven-hpi-plugin/pull/25.

pom.xml
```diff
@@ -739,6 +739,127 @@
             <scope>test</scope>
         </dependency>
 
+        <!-- Upper bound fix: list of transitive dependency to enforce highest version library -->
+        <dependency>
+            <groupId>com.jcraft</groupId>
+            <artifactId>jsch</artifactId>
```
jglick: This is not a plugin so it should not be here. Whatever plugin added this dep improperly, either update to a version of that plugin which has itself been fixed to use 2.31+ POMs, or add an exclusion for this dep so we pick up the version from core.

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>0.1.54</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
```
jglick: Ditto.

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>4.5.3</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpcore</artifactId>
```
jglick: Ditto.

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>4.4.6</version>
+        </dependency>
+        <dependency>
+            <groupId>org.objenesis</groupId>
+            <artifactId>objenesis</artifactId>
```
jglick: IIRC this is due to a stupid packaging of Powermock + Mockito. We should not be using `*-all` deps.

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>2.4</version>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-beans</artifactId>
```
jglick: Wrong.

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>2.5.6.SEC03</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.ant</groupId>
+            <artifactId>ant</artifactId>
```
jglick: Wrong.

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>1.9.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-embedder</artifactId>
```
jglick: Best to just exclude all of `maven-plugin`, at least pending https://github.com/jenkinsci/maven-plugin/pull/102, and make sure none of your transitive dependencies consider it mandatory (if they do that is an architectural bug we can solve).

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>3.5.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.sisu</groupId>
+            <artifactId>org.eclipse.sisu.plexus</artifactId>
```
jglick: ditto

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>0.3.3</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-core</artifactId>
```
jglick: ditto

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>3.5.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-aether-provider</artifactId>
```
jglick: ditto

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>3.3.1</version>
+        </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
```
jglick: Should not be here. This is from core.

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>2.5</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.icon-shim</groupId>
+            <artifactId>icon-set</artifactId>
```
jglick: Ask @tfennelly to revisit https://github.com/jenkinsci/plugin-pom/pull/72.

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>2.0.3</version>
+        </dependency>
+        <dependency>
+            <groupId>org.codehaus.plexus</groupId>
+            <artifactId>plexus-classworlds</artifactId>
```
jglick: `maven-plugin`, ditto

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>2.5.2</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-compat</artifactId>
```
jglick: ditto

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>3.5.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.ow2.asm</groupId>
+            <artifactId>asm</artifactId>
```
jglick: IIRC this is from core, should not be here.

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>5.0.3</version>
+        </dependency>
+        <dependency>
+            <groupId>org.ow2.asm</groupId>
+            <artifactId>asm-tree</artifactId>
```
jglick: Ditto.

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>5.0.3</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci</groupId>
+            <artifactId>symbol-annotation</artifactId>
```
jglick: No, you should never add this dep directly. Depend on `structs` where you actually need `@Symbol`.

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>1.10</version>
+        </dependency>
+        <dependency>
+            <groupId>com.infradna.tool</groupId>
+            <artifactId>bridge-method-annotation</artifactId>
```
jglick: Part of core, should not be here.

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>1.15</version>
+        </dependency>
+        <dependency>
+            <groupId>org.antlr</groupId>
+            <artifactId>antlr4-runtime</artifactId>
```
jglick: Ditto I think.

pom.xml
```diff
@@ -739,6 +739,127 @@
...
+            <version>4.5.1-1</version>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-server</artifactId>
```
jglick: ???

pom.xml
vivek: Ok, certain cases we can definitely do that, for others I will list them individually.

pom.xml
vivek: right, jira-rest-java-client-core bring older version, should be excluded over there.

pom.xml
vivek: ok.

vivek:
@jglick Thanks for your review. I am going to work at plugin level to exclude those individual jar or upgrade to those plugins if they are 2.31+.

pom.xml
vivek: @jglick This case is different. core is bring asm-tree 4.0, and then favorite plugin has 5.0.3. Clearly we can't exclude it in favorite plugin. How do we deal with such cases where core has older version and a plugin dependency is bring later version of a jar dependency( in this case transitive)?

pom.xml
vivek: @jglick github-branch-source has (transitive) dependency on symbol-annotation 1.5 where as core and struts have direct dependency on ver 1.1. Can't depend on struts for this. Either we define dependency on it directly or somehow exclude it from core and struts?

```
Require upper bound dependencies error for org.jenkins-ci:symbol-annotation:1.1 paths to dependency are:
+-io.jenkins.blueocean:blueocean-bitbucket-pipeline:1.3.0-beta-3-SNAPSHOT
  +-org.jenkins-ci.main:jenkins-core:2.46.3
    +-org.jenkins-ci:symbol-annotation:1.1
and
+-io.jenkins.blueocean:blueocean-bitbucket-pipeline:1.3.0-beta-3-SNAPSHOT
  +-org.jenkins-ci.plugins:cloudbees-bitbucket-branch-source:2.2.3
    +-org.jenkins-ci.plugins:structs:1.10
      +-org.jenkins-ci:symbol-annotation:1.1
and
+-io.jenkins.blueocean:blueocean-bitbucket-pipeline:1.3.0-beta-3-SNAPSHOT
  +-io.jenkins.blueocean:blueocean-pipeline-api-impl:1.3.0-beta-3-SNAPSHOT
    +-org.jenkins-ci.plugins:github-branch-source:2.2.3
      +-com.coravy.hudson.plugins.github:github:1.28.0 (managed) <-- com.coravy.hudson.plugins.github:github:1.26.0
        +-org.jenkins-ci:symbol-annotation:1.5
```

pom.xml
vivek: Hmm, in this case test dependency wire mock depends on antlr4-runtime 4.5.1-1 where as credentials on 4.5. So we need one version for test dependency and another for runtime, excluding antlr4-runtime from credentials won't work (I think) since wire mock is test dependency. In this case better to define antlr4-runtime explicit dependency from corresponding blueocean module blueocean-butbucket-pipeline?

```
Require upper bound dependencies error for org.antlr:antlr4-runtime:4.5 paths to dependency are:
+-io.jenkins.blueocean:blueocean-bitbucket-pipeline:1.3.0-beta-3-SNAPSHOT
  +-io.jenkins.blueocean:blueocean-pipeline-api-impl:1.3.0-beta-3-SNAPSHOT
    +-org.jenkins-ci.plugins:credentials:2.1.14
      +-org.antlr:antlr4-runtime:4.5
and
+-io.jenkins.blueocean:blueocean-bitbucket-pipeline:1.3.0-beta-3-SNAPSHOT
  +-com.github.tomakehurst:wiremock:2.6.0
    +-com.github.jknack:handlebars:4.0.6
      +-org.antlr:antlr4-runtime:4.5.1-1
```

pom.xml
vivek: Fixed, excluded from wire mock as jenkins test harness has later version.

pom.xml
vivek: Thats right but core has older version than the one from github-api plugin. How do we do it?

```
Require upper bound dependencies error for com.infradna.tool:bridge-method-annotation:1.13 paths to dependency are:
+-io.jenkins.blueocean:blueocean-bitbucket-pipeline:1.3.0-beta-3-SNAPSHOT
  +-org.jenkins-ci.main:jenkins-core:2.46.3
    +-com.infradna.tool:bridge-method-annotation:1.13
and
+-io.jenkins.blueocean:blueocean-bitbucket-pipeline:1.3.0-beta-3-SNAPSHOT
  +-io.jenkins.blueocean:blueocean-pipeline-api-impl:1.3.0-beta-3-SNAPSHOT
    +-org.jenkins-ci.plugins:git:3.5.1
      +-com.infradna.tool:bridge-method-annotation:1.13
and
+-io.jenkins.blueocean:blueocean-bitbucket-pipeline:1.3.0-beta-3-SNAPSHOT
  +-io.jenkins.blueocean:blueocean-pipeline-api-impl:1.3.0-beta-3-SNAPSHOT
    +-org.jenkins-ci.plugins:github-branch-source:2.2.3
      +-org.jenkins-ci.plugins:github-api:1.86
        +-org.kohsuke:github-api:1.86
          +-com.infradna.tool:bridge-method-annotation:1.14
```

pom.xml
vivek: Pretty much same failure, matrix-auth has 2.0.3. It getting fixed in core won't help because blueocean is not going to upgrade to latest core. I think its easier for us to define this dependency to make things work. When we get to later version of core we can clean it up ourselves.

```
Require upper bound dependencies error for org.jenkins-ci.plugins.icon-shim:icon-set:1.0.5 paths to dependency are:
+-io.jenkins.blueocean:blueocean-rest-impl:1.3.0-beta-3-SNAPSHOT
  +-org.jenkins-ci.main:jenkins-core:2.46.3
    +-org.jenkins-ci.plugins.icon-shim:icon-set:1.0.5
and
+-io.jenkins.blueocean:blueocean-rest-impl:1.3.0-beta-3-SNAPSHOT
  +-org.jenkins-ci.plugins:matrix-auth:1.7
    +-org.jenkins-ci.plugins.icon-shim:icon-shim:2.0.3
      +-org.jenkins-ci.plugins.icon-shim:icon-set:2.0.3
```

vivek:
@jglick I have tried to exclude from individual plugs - where ever I could. Excluding for each plugin is complicated depending on how combination of plugins are used. We are probably better off explicitly listing these transitive dependency with the upper bound version. Anyways PTAL, there are some of dependencies explicitly listed, these are there because core has lower version than the ones brought in thru other plugins.

raul-arabaolaza:
@vivek FYI This PR has got a tremendous positive impact on PCT results so in that sense (not able to judge technically) you have my :bee:

blueocean-github-pipeline/pom.xml
```diff
@@ -26,6 +26,13 @@
     <dependency>
       <groupId>org.kohsuke</groupId>
       <artifactId>github-api</artifactId>
+      <exclusions>
+        <!-- Upper bound fix: Pick the one from jenkins core, it has later version 2.4-->
```
jglick: This _was_ a bug in `github-api-plugin/pom.xml` but it should be fixed in 1.76. You should *not* have a dependency on the library itself, only on the plugin `org.jenkins-ci.plugins:github-api`. And then you need to make sure that you are consuming Jackson only via `org.jenkins-ci.plugins:jackson2-api`.

blueocean-rest-impl/pom.xml
```diff
@@ -51,14 +51,14 @@
...
-            <artifactId>git-client</artifactId>
-        </dependency>
         <dependency>
             <groupId>oro</groupId>
             <artifactId>oro</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.jcraft</groupId>
+            <artifactId>jsch</artifactId>
+        </dependency>
```
jglick: OK, you are bundling this in your `*.hpi` and not sharing it or anything.

blueocean-rest-impl/pom.xml
```diff
@@ -76,12 +76,10 @@
         <dependency>
```
jglick: Check the `jackson-*` deps above ^^^ which you perhaps want to pick up from the Jenkins plugin, but TBD. Ideally you would only be using the plugin dep (after upgrading it).

blueocean-rest-impl/pom.xml
```diff
@@ -136,11 +134,6 @@
 
         <dependency>
             <groupId>org.jenkins-ci.main</groupId>
-            <artifactId>maven-plugin</artifactId>
-            <scope>test</scope>
```
jglick: :+1: no need to wait for https://github.com/jenkinsci/maven-plugin/pull/102

pom.xml
```diff
@@ -476,6 +483,24 @@
                     <groupId>com.google.guava</groupId>
                     <artifactId>guava</artifactId>
                 </exclusion>
+
+                <!-- Upper bound dependency fix -->
+                <exclusion>
+                    <groupId>org.springframework</groupId>
+                    <artifactId>spring-beans</artifactId>
```
jglick: Very likely a bug in `jira-plugin/pom.xml`.

pom.xml
```diff
@@ -639,6 +659,13 @@
             <version>1.21</version>
         </dependency>
 
+        <!-- Needed by blueocean-rest-impl for ssh unititlies -->
+        <dependency>
+            <groupId>com.jcraft</groupId>
+            <artifactId>jsch</artifactId>
```
jglick: Preferably would be shared with other copies via a library wrapper plugin.

pom.xml
```diff
@@ -739,6 +805,43 @@
...
+        <!-- Upper bound fix:
+                These dependencies are defined in core but other plugins brings in higher version of dependencies
+                resulting in upper bound check failure.
+
+                After core is updated to depend on later version of these dependencies AND blueocean depends on that core,
+                 we are going to list these dependencies here.
+        -->
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.icon-shim</groupId>
+            <artifactId>icon-set</artifactId>
```
jglick: Ask @tfennelly for advice on https://github.com/jenkinsci/plugin-pom/pull/72.

pom.xml
jglick: Here we have a bug in `token-macro-plugin/pom.xml` because it bundles ASM 5 but in fact loads ASM 4 from `jenkins.war`. That should be fixed and released and you pick it up; or in the short term, exclude it as a workaround.

pom.xml
jglick: This is a bug in `github-plugin/pom.xml` which I will file a PR for. You must never add a dep on `symbol-annotation` from a plugin. Depend on `structs`, or exclude this dep from `github` for now.

pom.xml
jglick: This is a bug in `github-api-plugin/pom.xml`. I will file a PR to fix it. In the short term you can exclude this dep, so as to pick it up from core.

pom.xml
jglick: You could just exclude the dep from `credentials` or `blueocean-pipeline-api-impl` or `blueocean-bitbucket-pipeline`.

pom.xml
jglick: And in fact there is a `jsch` plugin! So use it.

blueocean-rest-impl/pom.xml
jglick: Use the plugin.

jglick: Cf. https://github.com/jenkinsci/ssh-credentials-plugin/pull/19

pom.xml
jglick: https://github.com/jenkinsci/github-plugin/pull/176

pom.xml
jglick: https://github.com/jenkinsci/github-api-plugin/pull/15

pom.xml
ljader: Yes, the blueocean-rest-impl should use jsch-plugin (library wrapper plugin)

blueocean-github-pipeline/pom.xml
vivek: Fixed, we depend on github-api-plugin.

blueocean-rest-impl/pom.xml
vivek: Fixed, depends on jsch plugin.

pom.xml
vivek: Done

pom.xml
vivek: This error is because we have **test** dependency on matrix-auth, that brings in different version of icon-set. Looks like this is a bug in matrix-auth, icon-set dependency comes from core and should not be overridden in matrix-auth. I am going to exclude icon-set from matrix-auth dependency for now.

```
Require upper bound dependencies error for org.jenkins-ci.plugins.icon-shim:icon-set:1.0.5 paths to dependency are:
+-io.jenkins.blueocean:blueocean-rest-impl:1.3.0-beta-3-SNAPSHOT
  +-org.jenkins-ci.main:jenkins-core:2.46.3
    +-org.jenkins-ci.plugins.icon-shim:icon-set:1.0.5
and
+-io.jenkins.blueocean:blueocean-rest-impl:1.3.0-beta-3-SNAPSHOT
  +-org.jenkins-ci.plugins:matrix-auth:1.7
    +-org.jenkins-ci.plugins.icon-shim:icon-shim:2.0.3
      +-org.jenkins-ci.plugins.icon-shim:icon-set:2.0.3
]
```

pom.xml
jglick: https://github.com/jenkinsci/token-macro-plugin/pull/31

pom.xml
```diff
@@ -429,6 +429,25 @@
...
+            <exclusions>
+                <!-- Upper bound fix.
+                     TODO: Remove it once github plugin commit
+                     https://github.com/jenkinsci/github-plugin/commit/a9308f124eeb5e599b2564d6355c5dd5cdafee34
+                     is released -->
+                <exclusion>
+                    <groupId>org.jenkins-ci</groupId>
+                    <artifactId>symbol-annotation</artifactId>
+                </exclusion>
+                <!-- Upper bound fix: Remove it once token-micro plugin is fixed to remove dependency on asm-tree -->
```
jglick: Link to https://github.com/jenkinsci/token-macro-plugin/pull/31.

michaelneale:
probably worth updating this to master too @vivek

pom.xml
vivek: Done

vivek:
@jglick I think I have taken care of all your feedback, its ready for final review. PTAL.

pom.xml
jglick: @daniel-beck may know.

blueocean-bitbucket-pipeline/pom.xml
```diff
@@ -20,6 +20,15 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>blueocean-pipeline-api-impl</artifactId>
+      <exclusions>
+        <!-- Upper bound fix: credentials plugin is bringing it's own version of anlr4-runtime,
```
jglick: `antlr4-runtime`

`its` BTW

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/BlueOceanConfigProperties.java
```diff
@@ -6,8 +6,10 @@
  * @author Ivan Meredith
  */
 public class BlueOceanConfigProperties {
-    public static final boolean ROLLBAR_ENABLED = Boolean.getBoolean("BLUEOCEAN_ROLLBAR_ENABLED");
+    public static final String BLUEOCEAN_FEATURE_JWT_AUTHENTICATION_PROPERTY = "BLUEOCEAN_FEATURE_JWT_AUTHENTICATION";
+    public static final String BLUEOCEAN_ROLLBAR_ENABLED_PROPERTY = "BLUEOCEAN_ROLLBAR_ENABLED";
```
jglick: Related somehow??

blueocean-github-pipeline/pom.xml
```diff
@@ -21,10 +21,18 @@
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>blueocean-pipeline-api-impl</artifactId>
+      <exclusions>
+        <!-- Upper bound fix: credentials plugin is bringing it's own version of anlr4-runtime,
```
jglick: ditto

blueocean-github-pipeline/pom.xml
```diff
@@ -43,6 +51,11 @@
         <artifactId>pubsub-light</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>fluent-hc</artifactId>
```
jglick: So, you are deliberately bundling this JAR, whatever it is?

blueocean-jwt/src/main/java/io/jenkins/blueocean/auth/jwt/impl/JwtAuthenticationFilter.java
```diff
@@ -40,7 +42,15 @@ public static void init() throws ServletException {
...
-        // noop
+        /**
+         * Initialize jwt enabled flag by reading BLUEOCEAN_FEATURE_JWT_AUTHENTICATION_PROPERTY jvm property
+         *
+         * {@link io.jenkins.blueocean.commons.BlueOceanConfigProperties.BLUEOCEAN_FEATURE_JWT_AUTHENTICATION} doesn't
+         * work in certain test scenario - specially when test sets this JVM property to enable JWT but this class has
+         * already been loaded setting it to false.
+         *
+         */
+        this.isJwtEnabled = Boolean.getBoolean(BLUEOCEAN_FEATURE_JWT_AUTHENTICATION_PROPERTY);
```
jglick: Unclear how this related to the PR.

blueocean-pipeline-editor/src/main/java/io/blueocean/rest/pipeline/editor/PipelineMetadataService.java
```diff
@@ -186,7 +184,7 @@ private boolean isWrapper(StepDescriptor d) {
 
     private boolean includeStep(StepDescriptor d) {
         boolean include = true;
-        if (ModelASTStep.getBlockedSteps().containsKey(d.getFunctionName())) {
+        if (BlockedStepsAndMethodCalls.blockedInSteps().containsKey(d.getFunctionName())) {
```
jglick: Forced by some dependency update?

pom.xml
```diff
@@ -165,6 +165,10 @@
           <scope>test</scope>
       </dependency>
       <dependency>
+          <groupId>org.apache.httpcomponents</groupId>
+          <artifactId>httpclient</artifactId>
```
jglick: This will force that library to be bundled with _every plugin_ in your reactor. Sounds wrong.

(Generally speaking, you should avoid `dependencies` in a POM parent; only use `dependencyManagement` and have each component declare those `dependencies` it actually needs. I guess it does not harm much to add `test`-scoped dependencies, though even those are probably undesirable. `junit:junit` in particular, @ line 148 above, is redundant.)

pom.xml
```diff
@@ -429,6 +433,27 @@
...
+                <!-- Upper bound fix.
+                     TODO: Remove it once github plugin commit
+                     https://github.com/jenkinsci/github-plugin/commit/a9308f124eeb5e599b2564d6355c5dd5cdafee34
+                     is released -->
+                <exclusion>
+                    <groupId>org.jenkins-ci</groupId>
+                    <artifactId>symbol-annotation</artifactId>
+                </exclusion>
+                <!-- Upper bound fix: Remove it once token-micro plugin is fixed to remove dependency on asm-tree
+                     PR: https://github.com/jenkinsci/token-macro-plugin/pull/31
```
jglick: https://github.com/jenkinsci/token-macro-plugin/commit/66ede22a9fefe9f3a459472a60f8230bc3c50851 now (unreleased)

pom.xml
jglick: https://github.com/jenkinsci/jira-plugin/pull/130

pom.xml
```diff
@@ -572,9 +622,18 @@
             <version>${jackson.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.kohsuke</groupId>
+            <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-api</artifactId>
             <version>1.86</version>
+            <exclusions>
+                <!-- Upper bound fix
+                     TODO: Remove when PR https://github.com/jenkinsci/github-api-plugin/pull/15 is merged and released
```
jglick: merged toward 1.87

pom.xml
```diff
@@ -620,6 +679,17 @@
             <groupId>org.jvnet.hudson.plugins</groupId>
             <artifactId>favorite</artifactId>
             <version>2.3.0</version>
+            <exclusions>
+                <!-- Upper bound fix: Remove it once token-micro plugin is fixed to remove dependency on asm-tree -->
```
jglick: merged toward 2.4

pom.xml
jglick: So why is this still here?

pom.xml
```diff
@@ -654,6 +724,28 @@
...
+        </dependency>
+
+        <!-- used in blueocean-rest-impl test -->
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>matrix-auth</artifactId>
+            <version>1.7</version>
+            <scope>test</scope>
+            <exclusions>
+                <!-- upper bound fix: icon-set comes from core, lets ignore the one coming from matrix-auth-->
```
jglick: cf. https://github.com/jenkinsci/plugin-pom/pull/72

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/BlueOceanConfigProperties.java
vivek: Not related, this had to do with fixing test flakiness.

blueocean-jwt/src/main/java/io/jenkins/blueocean/auth/jwt/impl/JwtAuthenticationFilter.java
vivek: Not related, fixes test flakiness.

blueocean-github-pipeline/pom.xml
vivek: yes we are using this library.

blueocean-pipeline-editor/src/main/java/io/blueocean/rest/pipeline/editor/PipelineMetadataService.java
vivek: Yeah upgrading to plugin 2.33 resulted in `@Restricted` violation coming from `ModelASTStep.getBlockedSteps()` required this change.

blueocean-bitbucket-pipeline/pom.xml
vivek: Done

blueocean-github-pipeline/pom.xml
vivek: Done

pom.xml
vivek: ack, added to comment

pom.xml
vivek: Removed junit.  I think declaring test is something makes it easy for sub-modules. Changed httpclient scope to test.

vivek:
@jglick jsch dependency is removed, there was already plugin version of it. Other comments addressed as well. PTAL.

blueocean-github-pipeline/pom.xml
```diff
@@ -55,6 +55,10 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>fluent-hc</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpclient</artifactId>
```
jglick: These plugins should not both independently bundle this library. Talk to @oleg-nenashev about it.

vivek: @jglick  httpclient is a dependency is fluent-hc. For some reason it's not visible resulting in explicitly adding this dependency. Why this would be a problem? Specially since parent pom makes sure both are same version. If it's a trouble I could exclude it from fluent-hc. 

@oleg-nenashev you know what could be causing this? Or how to get around? I could of course exclude it from fluent-hc.

vivek:
@raul-arabaolaza it's pretty much ready to be merged. Can you give it a try with that pct branch? I think we are getting closer.

michaelneale:
great to see this moving ahead !

blueocean-github-pipeline/pom.xml
vivek: I think we are good. Don't see double inclusion of httpclient. This is what's printed with `mvn dependency:tree`:

```
[INFO] +- org.apache.httpcomponents:fluent-hc:jar:4.5.3:compile
[INFO] |  \- commons-logging:commons-logging:jar:1.2:provided
[INFO] +- org.apache.httpcomponents:httpclient:jar:4.5.3:compile
```

vivek:
@michaelneale  This PR upgrade plugin-pom to 2:33 (required for PCT), there is side effect of it though. Once merged we won't be able to run unit tests from within IDEA till https://youtrack.jetbrains.com/issue/IDEA-175538 or https://issues.jenkins-ci.org/browse/JENKINS-45245 are fixed. 
See discussion on dev alias: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/jenkinsci-dev/4hfmOCvlmko/EZj2nGZrAwAJ.

Its a blocker issue for developer productivity but getting PCT to run is a priority as well.

raul-arabaolaza:
@vivek Running PCT

blueocean-github-pipeline/pom.xml
oleg-nenashev: Use [Apache HttpComponents Client 4.x API Plugin for Jenkins](https://github.com/jenkinsci/apache-httpcomponents-client-4-api-plugin), Luke

vivek:
@oleg-nenashev thanks, this should work, trying now.

blueocean-github-pipeline/pom.xml
vivek: Done

raul-arabaolaza:
I have run the PCT for this branch of BO, the result compared with current master is a great improvement.

Without this PR many plugins were not able to be exercised by the PCT due to the lack of `war-for-test` artifact, with this PR all plugins are properly exercised by the PCT and there is just one test failing in the entire BO suite. So this PR solves 99% of the BO PCT problems!

Great work here @vivek and thanks a lot to @jglick and @oleg-nenashev for reviews

vivek:
@oleg-nenashev I have replaced httpcomponents libraries (httpclient, httpmime, fluent-hc) by apache-httpcomponets-client-4-api plugin. If you could take a look great, otherwise I plan to merge it sometime today.

michaelneale:
@raul-arabaolaza that linked PR was merged - so perhaps @vivek can update this to master and we can try again?

vivek:
@michaelneale rebased with master, @raul-arabaolaza is running PCT against this branch.

michaelneale:
ok, well if it passes and @oleg-nenashev is around - I guess this can go in

raul-arabaolaza:
I have successfully run the PCT without errors, you have a :bee: and a :beer: from my side

Package | Duration | Fail | (diff) | Skip | (diff) | Pass | (diff) | Total | (diff)
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
io.blueocean.rest.pipeline.editor | 37 sec | 0 |   | 0 |   | 10 |   | 10 |  
io.jenkins.blueocean | 49 sec | 0 |   | 0 |   | 92 |   | 92 |  
io.jenkins.blueocean.auth.jwt.impl | 1 min 27 sec | 0 |   | 0 |   | 12 |   | 12 |  
io.jenkins.blueocean.blueocean_bitbucket_pipeline.cloud | 3 min 16 sec | 0 |   | 0 |   | 63 |   | 63 |  
io.jenkins.blueocean.blueocean_bitbucket_pipeline.server | 6 min 14 sec | 0 |   | 0 |   | 111 |   | 111 |  
io.jenkins.blueocean.blueocean_git_pipeline | 6 min 46 sec | 0 |   | 8 |   | 108 |   | 116 |  
io.jenkins.blueocean.blueocean_github_pipeline | 19 min | 0 |   | 0 |   | 315 |   | 315 |  
io.jenkins.blueocean.commons | 2.4 sec | 0 |   | 0 |   | 72 |   | 72 |  
io.jenkins.blueocean.commons.stapler.export | 0.7 sec | 0 |   | 0 |   | 342 |   | 342 |  
io.jenkins.blueocean.config | 2 min 8 sec | 0 |   | 0 |   | 23 |   | 23 |  
io.jenkins.blueocean.events | 3 min 37 sec | 0 |   | 0 |   | 55 |   | 55 |  
io.jenkins.blueocean.htmlpublisher | 56 sec | 0 |   | 0 |   | 12 |   | 12 |  
io.jenkins.blueocean.i18n | 1 min 37 sec | 0 |   | 7 |   | 144 |   | 151 |  
io.jenkins.blueocean.preload | 36 sec | 0 |   | 0 |   | 12 |   | 12 |  
io.jenkins.blueocean.rest.impl.pipeline | 1 hr 28 min | 0 | -12 | 96 |   | 1272 | +12 | 1368 |  
io.jenkins.blueocean.service.embedded | 35 min | 0 |   | 91 |   | 1040 |   | 1131 |  
io.jenkins.blueocean.service.embedded.analytics | 1 min 24 sec | 0 |   | 0 |   | 130 |   | 130 |  
io.jenkins.blueocean.service.embedded.jira | 28 sec | 0 |   | 0 |   | 13 |   | 13 |  
io.jenkins.blueocean.service.embedded.rest | 2 min 7 sec | 0 |   | 0 |   | 52 |   | 52 |  
io.jenkins.blueocean.service.embedded.util | 90 ms | 0 |   | 0 |   | 78 |   | 78

oleg-nenashev:
Just for the future, IMHO it's better to explicitly set dependency instead of adding exclusions in some cases. In such case the dependency check won't be suppressed when updating dependencies.

vivek:
@oleg-nenashev yeah, exclusions looks fragile, I had the mixed approach at some point in this PR but @jglick thinks otherwise, he prescribed not to explicitly depend on library dependency, that is let the dependency get resolved from one of the plugin while excluding such library resolving transitively. I am sure there can be cases where explicit dependency could be better.

vivek:
@oleg-nenashev and icon-shim hit us due to test dependency on matrix-auth, exclusion in this case was better solution as it comes from core and that supersedes everything else anyways.

raul-arabaolaza:
@vivek Reluanched the PCT after your changes, still successfull, you still have my :bee:

PR #1384, opened 2017-09-11
defensive code - for issue trackers that like to blow up with NPEs

michaelneale:
# Description

See https://issues.jenkins-ci.org/browse/JENKINS-46778

There have been issues since JIRA has been bundled. This is an attempt to be generically defensive against issue tracking plugins that may poop on things. 

please consider @vivek @kzantow


michaelneale:
@oleg-nenashev oh really? I actually thought even doing it once was a nasty hack, but that makes sense (I did think of it). Let me try that.

michaelneale:
trying per iteration as @oleg-nenashev suggested is common practice with extensions.

PR #1386, opened 2017-09-11
Bug/jenkins 46723 bitbucket empty repo

cliffmeyers:
# Description
- See [JENKINS-46723](https://issues.jenkins-ci.org/browse/JENKINS-46723).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-jira/pom.xml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
michaelneale: huh - I wonder if it would be possible to get this upstreamed in the jira plugin itself? it would need to depend on blue ocean rest right?

kzantow: @michaelneale yeah, that's the right thing to do. same with some of the git and github stuff, but we can tombstone these small plugins when we get around to properly splitting this stuff up. I was asked this very question at JW and think the right answer is to implement a transitional plugin until we can get the functionality merged to the upstream one, they can use an optional dependency on blue ocean.

michaelneale: @kzantow ok - fair enough. Wil need to open the permission PR for a new plugin (and I will get chewed out a bit). 

Also makes it easier to get upstream when the blue ocean extension points are published and stable.

michaelneale:
this seems to work for me - tried it with bitbucket cloud so :bee: from me. 

however, I would like Karl to retest this if possible, and @vivek to give the ok that having this as a module is the right thing to do, but I believe this is the right way. 

https://github.com/jenkins-infra/repository-permissions-updater/ will need to be updated so a new plugin can be deployed for this if we think this is the right approach (which I can do). 

Nice one @cliffmeyers

vivek:
I am not sure if splitting it in to new plugin is necessary. Did you try excluding the dependency from rest-impl that brought this HttpException? You would do his exclusion in bitbucket modules pom where rest-impl or pipeline-impl dependency is declared. I may get to try in few hours and report back.

michaelneale:
@vivek good point. I do think getting it in JIRA plugin would be perfect but I think we can't really do that as a first pass until blue ocean 1.3 is out and then it can depend on the right module/plugin from the jira side.

vivek:
I gave a quick try: https://github.com/jenkinsci/blueocean-plugin/commit/848a6e02cefd0780e888dec695de99713d81d277.

Basically, this change ensures bitbucket module is picking up its own httpclient(HttpResponseException). I also ended up cleaning rest-impl. It was incorrectly depending on maven-plugin and git-client.

michaelneale:
@vivek do you want to make that an alt-PR?

cliffmeyers:
@vivek when I discussed this with @kzantow he was not confident the Maven exclude would actually have the desired impact when actually deployed in Jenkins and that it could give a false positive in the `hpi:run` scenario.

vivek:
@michaelneale @cliffmeyers @kzantow  exclusions work fine with hpi:run. I opened PR on that branch PTAL, best if you can try it locally as well: https://github.com/jenkinsci/blueocean-plugin/pull/1388.

cliffmeyers:
@vivek yes, assumed it would work fine with hpi:run but Keith was unsure it would be work ok in real Jenkins deployment scenario. That said, would it be better to not include jira as a dependency in one of our core modules.

vivek:
@cliffmeyers 
> That said, would it be better to not include jira as a dependency in one of our core modules.

Yes absolutely, thats how it should have been done from the beginning. I was saying from perspective of the issue that needed to be fixed didn't require splitting it in to a new module. There is extra cost of splitting in to new module, that is can we do it outside of blueocean repo and include in the aggregator? We need this clean up for other modules as well so the question is do we do this for Jira now - as an extra maven module. Lets chat when you are back online and decide which way to go.

michaelneale:
My preference is to get to the end game where this is all upstream in JIRA - just a plugin outside of blue ocean. to get there though we need blue ocean 1.3. 

I am ok with whatever approach you agree on that works (it would be slightly nicer to not make another plugin we have to tombstone, if possible...) - but we should make a follow on ticket in either case to get it upstreamed to the JIRA plugin after blue ocean 1.3 is out.

vivek:
@michaelneale +1. As I said, I do not see necessity of splitting this in to separate plugin to fix bitbucket empty repo regression. So that the work on this is not wasted, we can use it post 1.3 to split this in to an upstream plugin via separate ticket/PR. I plan to sync up with @cliffmeyers this evening.

cliffmeyers:
Closing this in favor of #1388. Since it needs to split off to a separate repo long term no real point in leaving this open, although I'll make sure the code doesn't get lost.

michaelneale:
hrm in retrospect, we may have to break this out @vivek @cliffmeyers so lets keep this branch around...

PR #1387, opened 2017-09-11
Experiment/ath wiremock

cliffmeyers:
# Description
- Experimenting with using Wiremock in ATH so we can more easily test edge cases

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


cliffmeyers:
cc @kschultzcb

cliffmeyers:
Need to track down a test failure first...

cliffmeyers:
Fixed up now.

michaelneale:
nice @cliffmeyers - could this help test things like bitbucket server as well? that may be a great use case as that has a gap right now.

cliffmeyers:
@michaelneale absolutely. By default wiremock will proxy a HTTP/S endpoint, pass through requests to the endpoint and record all of the responses. This is how @vivek set up the unit tests for Github and Bitbucket so I'm really just building off his great work there.

I think for ATH which involves these third party services, we should strive to write the tests so they can be run in "live" or "mock" mode and perhaps we run mocks for our regular CI builds and then run live tests on a nightly or weekly basis.

cliffmeyers:
cc @imeredith perhaps you can provide some feedback and a bee if you like what see here?

imeredith:
I like what you have done but I would like to talk about this more. For example, i don't think we can use this method for end to end tests I think? :bee: for what is here, but will talk more next week. This ties in to what im looking at atm.

michaelneale:
@cliffmeyers LGTM - I think we should just go with it, this covers things we don't otherwise have covered right?

cliffmeyers:
@michaelneale strictly speaking it covers the same scenarios that the Github Enterprise test always did, although hopefully a bit more reliably without having to reach out to Github.

PR #1393, opened 2017-09-14
[JENKINS-46879] Check permissions against the item where the pipeline  is in

alvarolobato:
# Description

See [JENKINS-46879](https://issues.jenkins-ci.org/browse/JENKINS-46879).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/ScmResourceImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/ScmResourceImpl.java
@@ -59,7 +59,7 @@ public class ScmResourceImpl extends BluePipelineScm {
     }
 
     private @Nonnull User checkPermission(){
-        ACL acl = Jenkins.getInstance().getACL();
+        ACL acl = item.getACL();
         Authentication a = Jenkins.getAuthentication();
         User user = User.get(a);
         if(user == null){
--- a/blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/OrganizationFolderTest.java
+++ b/blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/OrganizationFolderTest.java
@@ -121,6 +121,14 @@ public class OrganizationFolderTest{
         assertNotNull(folderPipeline.getQueue());
         assertNotNull(folderPipeline.getQueue().iterator());
 
+        //Make sure the user does has permissions to that folder
+        PowerMockito.when(orgFolder.getACL()).thenReturn(new ACL() {
+            @Override
+            public boolean hasPermission(Authentication arg0, Permission arg1) {
+                return true;
+            }
+        });
+
         ScmResourceImpl scmResource = new ScmResourceImpl(orgFolder, folderPipeline);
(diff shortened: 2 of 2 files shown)


alvarolobato:
@reviewbybees

michaelneale:
:bug: seems to be a test error: 

```
testOrganizationFolderFactory(io.jenkins.blueocean.rest.impl.pipeline.OrganizationFolderTest)  Time elapsed: 26.284 sec  <<< ERROR!
java.lang.NullPointerException
	at io.jenkins.blueocean.rest.impl.pipeline.ScmResourceImpl.checkPermission(ScmResourceImpl.java:68)
	at io.jenkins.blueocean.rest.impl.pipeline.ScmResourceImpl.<init>(ScmResourceImpl.java:33)
	at io.jenkins.blueocean.rest.impl.pipeline.ScmResourceImpl.<init>(ScmResourceImpl.java:26)
```

michaelneale:
Ah, a NPE on a Nonnull element (more evidence that nonnull annotation stuff is often just junk - you really can't trust it can you). 

One way: 

```
        if (acl == null) acl = Jenkins.getInstance().getACL();
```
(but no idea of the intention here, and if local to test).

alvarolobato:
@vivek @michaelneale fix pushed. Let wait for CI.

michaelneale:
:bee: once green - I think @vivek is happy with this. Should be ok to get on release/1.2 as well 👍

vivek:
@michaelneale yes, once green, will merge and port it to release/1.2 wait for release/1.2 branch go green then do the release.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/ScmResourceImpl.java
```diff
@@ -59,7 +59,7 @@ public Link getLink() {
     }
 
     private @Nonnull User checkPermission(){
-        ACL acl = Jenkins.getInstance().getACL();
+        ACL acl = item.getACL();
```
jtnord: I'm not sure about any ACL test here.  The permission IIUC is controlled by the underlying SCM (e.g. github) using a user based ket/token.
Thus an admin user may be in chanrge of creating the multibranch pipelines - but if a user can push to github why not allow the user to use BO to do it even if they can not configure the job?

This is certainly better than it was so it gets an approval from me - but seems questionable in the bigger scheme of things.
/cc @vivek @michaelneale

vivek: @jtnord This was added as extra check as editing a Jenkinsfile on existing job was considered job configuration and that user is operating in terms of existing job and editing pipeline was attached to configure permission. Main idea behind this was to reject edit of pipeline for users without configure permission. We can relax this permission if needed, but thats anyways is orthogonal to this PR.

michaelneale: yeah its all a bit odd to me - but then again, a "root" user that can be not allowed to access a folder (which is what alvaro was catering for) kind of also does my head in, so I am not sure I really understand what is going on. 

As you can configure a lot with the Jenkinsfile, it does make sense to trust the SCM as the source of access truth...

PR #1394, opened 2017-09-14
[FIX JENKINS-46854] [FIX JENKINS-46862] Fix error display, credentials for http/https, editor roundtrip

kzantow:
# Description

See [JENKINS-46854](https://issues.jenkins-ci.org/browse/JENKINS-46854).

Test by: using the pipeline editor in many scenarios...

Not really tests at the moment as these are integration issues with external services...

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


kzantow:
Fixed the error display issue for global errors, e.g. an invalid script block:
image

michaelneale:
findbugs is complaining: 

```
[INFO] Possible null pointer dereference in io.jenkins.blueocean.blueocean_git_pipeline.GitReadSaveRequest.getCredential() due to return value of called method [io.jenkins.blueocean.blueocean_git_pipeline.GitReadSaveRequest, io.jenkins.blueocean.blueocean_git_pipeline.GitReadSaveRequest] Method invoked at GitReadSaveRequest.java:[line 73]Known null at GitReadSaveRequest.java:[line 73] NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE
```

michaelneale:
FINDBUGS STILL NOT HAPPY FEED IT

michaelneale:
tried it with git - looking good @kzantow !

michaelneale:
I tried adding a repo via git with existing Jenkinsfile, it still showed the "no pipelines" page (but I didn't wait to see if it refreshed...) just FYI (no different than before in that simple case).

kzantow:
Actually the hate was totally misguided: it's really Jacoco that I hate. Not just for this issue.

kzantow:
these 3 tests...

Jenkinsfile
```diff
@@ -30,7 +30,12 @@ node() {
         }
 
         stage('Building BlueOcean') {
-          sh "mvn clean install -B -DcleanNode -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn -Dmaven.test.failure.ignore -s settings.xml -Dmaven.artifact.threads=30"
+          try {
+            sh "mvn clean install -B -DcleanNode -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn -Dmaven.test.failure.ignore -s settings.xml -Dmaven.artifact.threads=30"
+          } catch(e) {
+            archive '*/target/code-coverage/**/*.html'
+            throw e;
+          }
```
cliffmeyers: Would `post always` for the stage also work?

blueocean-core-js/src/js/services/DefaultSSEHandler.js
```diff
@@ -76,6 +76,16 @@ export class DefaultSSEHandler {
         if (event.job_ismultibranch && !event.blueocean_job_branch_name) {
             return;
         }
+        // don't care about indexing events
+        if (event.job_multibranch_indexing_status === 'INDEXING') {
+            return;
+        }
+        // If we have a queued item but the branch isn't present, we need to refresh the pager
+        // this happens, for example, when you create a new pipeline in a repo that did not have one
+        const pipeline = this.pipelineService.getPipeline(`/blue/rest/organizations/${event.jenkins_org}/pipelines/${event.blueocean_job_pipeline_name}/`);
```
cliffmeyers: Is `event.blueocean_job_pipeline_name` the fully-qualified name of the pipeline, e.g. the `fullName`? If not, this will fail for nested pipelines. That might not matter today in creation use case, but if there's an easy way to ensure we're using the full name we should.

blueocean-pipeline-editor/src/main/js/services/PipelineValidator.js
```diff
@@ -20,6 +20,10 @@ export function isValidEnvironmentKey(key: string): boolean {
     return false;
 }
 
+function _isArray(o) {
+    return o instanceof Array || typeof o === 'array';
```
cliffmeyers: If this data ever becomes managed by MobX just be aware that `ObservableArray` will fail these tests. May be fine to ignore for now.

Jenkinsfile
kzantow: This one isn't `declarative`, does post still work?

blueocean-core-js/src/js/services/DefaultSSEHandler.js
kzantow: hm, good question, I'll check and escape it or not

Jenkinsfile
kzantow: ... magic 8-ball says "no".

blueocean-core-js/src/js/services/DefaultSSEHandler.js
kzantow: good call, it is _not_ escaped

pom.xml
```diff
@@ -319,7 +319,7 @@
         <dependency>
             <groupId>org.jenkinsci.plugins</groupId>
             <artifactId>pipeline-model-definition</artifactId>
-            <version>1.2-beta-4</version>
+            <version>1.2-beta-5</version>
```
michaelneale: should be finaly any second now...

michaelneale:
this is looking good to me - I think my complaints are addressed .

:bee: from me in that case. My main concern is how well we have test coverage for this

michaelneale:
@kzantow if this is clean it can go in any time BTW.. def improved things.

blueocean-rest-impl/pom.xml
```diff
@@ -51,9 +51,9 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>ssh-credentials</artifactId>
         </dependency>
-        <dependency>
-            <groupId>com.jcraft</groupId>
-            <artifactId>jsch</artifactId>
+        <dependency> <!-- need this to solve a JSch linkage problem -->
```
michaelneale: hey @vivek is this ok? should this be test scoped?

michaelneale:
@kzantow looking good, jacoco violations again (last error I saw), not sure what that is about, maybe @vivek has an idea?

blueocean-rest-impl/pom.xml
kzantow: this is required, at least for hpi:run. otherwise, it fails with a linkage error due to the jsch libraries between the plugins somehow @michaelneale

blueocean-git-pipeline/src/test/java/io/jenkins/blueocean/blueocean_git_pipeline/GitUtilsTest.java
```diff
@@ -94,4 +97,25 @@ public void testOperations() throws Exception {
...
+        Map resp = new RequestBuilder(baseUrl)
+            .status(200)
+            .get("/organizations/jenkins/user/publickey/").build(Map.class);
+
+        String id = (String)resp.get("id");
+        Assert.assertTrue(id != null);
+
+        resp = put("/organizations/jenkins/scm/git/validate/",
+            ImmutableMap.of(
+                "repositoryUrl", "git@github.com:vivek/test-no-jenkins-file.git",
```
michaelneale: if this repo disappears or changes... will it break the test?

kzantow: yes, but it's used in [another test already](https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-git-pipeline/src/test/java/io/jenkins/blueocean/blueocean_git_pipeline/GitScmTest.java#L134) @michaelneale

kzantow: err, mebby not. no idea where i got that repo from, copied it from some test somewhere :)

blueocean-rest-impl/pom.xml
kzantow: maybe we can get away with it test scoped? but will it get jsch from the upstream plugin or no? probably @vivek knows for sure :)

kzantow: ... tried it, doesn't work.

Jenkinsfile
michaelneale: yeah only for declarative

blueocean-git-pipeline/src/main/java/io/jenkins/blueocean/blueocean_git_pipeline/GitUtils.java
```diff
@@ -125,6 +128,46 @@
...
+     * @param repo local repository
+     * @param remoteUrl git repo url
+     * @param credential credential to use when accessing git
+     * @return list of Errors. Empty list means success.
+     */
+    public static void validatePushAccess(@Nonnull Repository repo, @Nonnull String remoteUrl, @Nullable StandardCredentials credential) throws GitException {
+        try (org.eclipse.jgit.api.Git git = new org.eclipse.jgit.api.Git(repo)) {
+            // we need to perform an actual push, so we try a deletion of a very-unlikely-to-exist branch
+            // which needs to have push permissions in order to get a 'branch not found' message
+            String pushSpec = ":this-branch-is-only-to-test-if-jenkins-has-push-access";
```
michaelneale: one day I will name a branch this and troll myself...

michaelneale:
I think this is ok - would like @vivek to bless the jsch dependency

blueocean-rest-impl/pom.xml
vivek: @kzantow Dependence on git-client sounds wrong. can you paste the linkage error you are getting? I am going to try your branch locally anyways to see whats going on.

michaelneale: yes this will probably upset things for people downstream... adding this dependency here, so need to find another way.

kzantow: @vivek when the dependency is `jsch`, it gives a linkage error due to both `git-client` and this plugin loading their own copies. You only see it when you try to load a pipeline in the editor using a git pipeline.

michaelneale:
yes I think the git client dep on rest-impl won't play nice..

blueocean-rest-impl/pom.xml
```diff
@@ -51,9 +51,9 @@
...
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>git-client</artifactId>
```
vivek: Dependence of git-client is 🐛 Do not see master failing in CI run or locally with any linkage error. Perhaps something local in your env? Need to get more details before we know whats the right dependency we need to add if/any.

kzantow: Well, we need `jsch`. It doesn't work to include it directly. Master isn't failing because there are no `git` tests in ATH. Again, it only fails when you try to load something in the editor using a Git pipeline.

blueocean-rest-impl/pom.xml
michaelneale: @kzantow that isn't new functoinality though is it? confused as to why it is picked up now

kzantow: FYI, this is the exception:

```
Caused by: java.lang.LinkageError: loader constraint violation: loader (instance of hudson/ClassicPluginStrategy$AntClassLoader2) previously initiated loading for a different type with name "com/jcraft/jsch/JSch"
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at jenkins.util.AntClassLoader.defineClassFromData(AntClassLoader.java:1139)
	at hudson.ClassicPluginStrategy$AntClassLoader2.defineClassFromData(ClassicPluginStrategy.java:876)
	at jenkins.util.AntClassLoader.getClassFromStream(AntClassLoader.java:1310)
	at jenkins.util.AntClassLoader.findClassInComponents(AntClassLoader.java:1363)
	at jenkins.util.AntClassLoader.findClass(AntClassLoader.java:1326)
	at jenkins.util.AntClassLoader.loadClass(AntClassLoader.java:1079)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at org.eclipse.jgit.transport.JschConfigSessionFactory.createSession(JschConfigSessionFactory.java:220)
	at org.eclipse.jgit.transport.JschConfigSessionFactory.createSession(JschConfigSessionFactory.java:176)
	at org.eclipse.jgit.transport.JschConfigSessionFactory.getSession(JschConfigSessionFactory.java:110)
	at org.eclipse.jgit.transport.SshTransport.getSession(SshTransport.java:137)
	at org.eclipse.jgit.transport.TransportGitSsh$SshPushConnection.<init>(TransportGitSsh.java:322)
	at org.eclipse.jgit.transport.TransportGitSsh.openPush(TransportGitSsh.java:167)
	at org.eclipse.jgit.transport.PushProcess.execute(PushProcess.java:160)
	at org.eclipse.jgit.transport.Transport.push(Transport.java:1272)
	at org.eclipse.jgit.api.PushCommand.call(PushCommand.java:161)
	at io.jenkins.blueocean.blueocean_git_pipeline.GitUtils.validatePushAccess(GitUtils.java:152)
	at io.jenkins.blueocean.blueocean_git_pipeline.GitBareRepoReadSaveRequest$1.invoke(GitBareRepoReadSaveRequest.java:61)
	at io.jenkins.blueocean.blueocean_git_pipeline.GitBareRepoReadSaveRequest$1.invoke(GitBareRepoReadSaveRequest.java:57)
	at jenkins.plugins.git.GitSCMFileSystem$3.invoke(GitSCMFileSystem.java:193)
	at org.jenkinsci.plugins.gitclient.AbstractGitAPIImpl.withRepository(AbstractGitAPIImpl.java:29)
	at org.jenkinsci.plugins.gitclient.CliGitAPIImpl.withRepository(CliGitAPIImpl.java:71)
	at jenkins.plugins.git.GitSCMFileSystem.invoke(GitSCMFileSystem.java:189)
	at io.jenkins.blueocean.blueocean_git_pipeline.GitCacheCloneReadSaveRequest.invokeOnScm(GitCacheCloneReadSaveRequest.java:149)
	at io.jenkins.blueocean.blueocean_git_pipeline.GitBareRepoReadSaveRequest.read(GitBareRepoReadSaveRequest.java:57)
	at io.jenkins.blueocean.blueocean_git_pipeline.GitReadSaveService.getContent(GitReadSaveService.java:165)
	at io.jenkins.blueocean.rest.impl.pipeline.ScmResourceImpl.getContent(ScmResourceImpl.java:42)
	at java.lang.invoke.MethodHandle.invokeWithArguments(MethodHandle.java:627)
	at org.kohsuke.stapler.Function$MethodFunction.invoke(Function.java:343)
	... 103 more

```

blueocean-rest-impl/pom.xml
vivek: @kzantow ah ok, this is good info, I can try to reproduce it. will get back.

michaelneale:
@vivek is this up to date with master and the PCT branch changes?

vivek:
@michaelneale yes it includes PCT fixes.

michaelneale:
@kzantow @vivek could this be solved by moving jsch into aggregator something like: 

https://gist.github.com/michaelneale/798a1145f5a4b04c599834ea48b07f18
(and removing git client)?

michaelneale:
another :bug: is that I wasn't able to load it up on a totally empty repo (ie no branches):

```
fetch failed: 500 for http://localhost:8080/jenkins/blue/rest/organizations/jenkins/pipelines/plaingit/scm/content?path=Jenkinsfile
```

https://github.com/multibranchorg/plaingit
should I be able to?

kzantow:
@michaelneale @vivek the problem isn't the version of `jsch`, it's the fact that `git-client` loads `jsch` through its own classloader. If we directly include `jsch` in `blueocean-rest-impl`, it gets loaded another time from _that_ classloader, then down the line in `blueocean-git-pipeline`, when it tries to pass `jsch` objects created from `blueocean-rest-impl` to the `git-client`, the failure happens.

vivek:
@kzantow @michaelneale Took bit time but I have a fix. Its on branch `fix-jsch-linkage-error`
and commit https://github.com/jenkinsci/blueocean-plugin/commit/bc9c56c0e7f001eadd0856a04b370747b124354f. 

Lets wait for CI run https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/fix-jsch-linkage-error/1/pipeline, if it looks good feel free to merge to your branch.

kzantow:
@vivek thanks for looking into this, I'll give this a try locally too right now

vivek:
@kzantow right. Its fixed to ensure rest-impl jsch doesn't get in the way. Eventually it needs fixing in git-client plugin where it should exclude jsch from jgit dependency and then declare it's own explicit dependency on jsch-plugin (not library), at that point we can start depending on jsch-plugin as well.

kzantow:
@vivek right, `jsch` would need to be a plugin to be properly shared. I see there is one: https://wiki.jenkins.io/display/JENKINS/JSch+Plugin but this seems like something 'modular core components' would solve, just being able to share specific libraries. anyway, 'nuff discussion on that here

vivek:
Yeah jsch plugin is already there and most of my time went in to make it work but till it gets picked up in git-client while excluding jsch lib from jgit we can’t use it. 

Other options I tried were shading but it doesn’t work well with hpi. Provided scope also fixes it but optional is more suited in this case.

michaelneale:
@kzantow the test failure for `createPersonalSSHKey – io.jenkins.blueocean.ssh.UserSSHKeyTest`
is the legit? re-running again

kzantow:
@michaelneale could be legit, investigating now

michaelneale:
ok, once I rm ~/.jenkins-hpl... file thing (that is BAAAD)

I was able to save a Jenkinsfile into an empty repo ok. However saving it - never went to activity screen (but it did build) - that is one :ant:. I had to click on pipelines to see it

May be my slow connection though to github, and htings being dropped... wil test again shortly

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/UserRoute.java
```diff
@@ -0,0 +1,17 @@
+package io.jenkins.blueocean.rest;
+
+import hudson.ExtensionPoint;
+import io.jenkins.blueocean.Routable;
+import io.jenkins.blueocean.rest.model.BlueUser;
+
+/**
+ * Route contributing to {@link io.jenkins.blueocean.rest.model.BlueUser}: url path /organization/:id/users/:user/:userRoute.urlName()
+ */
+public interface UserRoute extends Routable, ExtensionPoint {
```
vivek: Looks good @kzantow!

michaelneale:
ok tested with BB - it went back to activity screen... it just seems overly latency sensitive it seems... 

I think this moves a bunch of things forward so :bee:

I am not sure how useful git editing is on hi-latency from master <-> git repo though, at least via ssh it is problematic...

PR #1398, opened 2017-09-15
JENKINS-44033 hide the favorite button in the header of empty multibranch pipelines

NicuPascu:
# Description
just added a check for multibranch pipelines that if they do not have any branches the favorite button should not appear

See [JENKINS-44033](https://issues.jenkins-ci.org/browse/JENKINS-44033).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-personalization/src/main/js/components/FavoritePipeline.jsx
+++ b/blueocean-personalization/src/main/js/components/FavoritePipeline.jsx
@@ -79,13 +79,13 @@ export class FavoritePipeline extends Component {
             return null;
         }
 
-        return (
+        return (!(this.props.pipeline.branchNames && !this.props.pipeline.branchNames.length) ? (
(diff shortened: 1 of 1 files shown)


blueocean-personalization/src/main/js/components/FavoritePipeline.jsx
```diff
@@ -79,13 +79,13 @@ export class FavoritePipeline extends Component {
             return null;
         }
 
-        return (
+        return (!(this.props.pipeline.branchNames && !this.props.pipeline.branchNames.length) ? (
```
cliffmeyers: Alternatively you should be able to use `cond1 && cond2 && <ReactComponent>` instead of the ternary form, might read slightly better.

PR #1399, opened 2017-09-15
JENKINS-46879# Pipeline could be an org folder, handle it accordingly

vivek:
# Description

Ported handling of permission check for org folder during loading content. 1.3 doesn't support creation of org folder pipeline but we  need this fix for backward compatibility.

See [JENKINS-46879](https://issues.jenkins-ci.org/browse/JENKINS-46879).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/ScmResourceImpl.java
```diff
@@ -59,7 +60,12 @@ public Link getLink() {
     }
 
     private @Nonnull User checkPermission(){
-        ACL acl = item.getACL();
+        ACL acl;
+        if(item.getParent() != null && item.getParent() instanceof OrganizationFolder){
```
kzantow: Why would the `item` `ACL` not allow something that the `item.parent` would?

alvarolobato: Because in this case we are inside an OrganizationFolder which have computed folders that you can't configure.

PR #1400, opened 2017-09-18
feature/JENKINS-46913-result-item-links * Linkify URLs in ResultItem headers

sophistifunk:
Linkify URLs in ResultItem headers

# Description

See [JENKINS-46913](https://issues.jenkins-ci.org/browse/JENKINS-46913).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/jenkins-design-language/package.json
+++ b/jenkins-design-language/package.json
@@ -41,6 +41,7 @@
     "url": "https://github.com/jenkinsci/jenkins-design-language.git"
   },
   "dependencies": {
+    "linkifyjs": "2.1.4",
     "lodash.debounce": "4.0.8",
     "moment": "2.15.1",
     "moment-duration-format": "1.3.0",
--- a/jenkins-design-language/src/js/components/ResultItem.jsx
+++ b/jenkins-design-language/src/js/components/ResultItem.jsx
@@ -80,7 +80,6 @@ export class ResultItem extends Component {
     };
 
     urlClicked = (e: Event) => {
-        console.log('urlClicked');
         e.stopPropagation();
     };
 
--- a/jenkins-design-language/src/js/components/ResultItem.jsx
+++ b/jenkins-design-language/src/js/components/ResultItem.jsx
@@ -2,6 +2,7 @@
 
 import React, { Component, PropTypes } from 'react';
 import ReactCSSTransitionGroup from 'react-addons-css-transition-group';
+import Linkify from 'linkifyjs/react';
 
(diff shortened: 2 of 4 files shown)


sophistifunk:
screen shot 2017-09-18 at 6 41 33 pm

jenkins-design-language/package.json
```diff
@@ -41,6 +41,7 @@
     "url": "https://github.com/jenkinsci/jenkins-design-language.git"
   },
   "dependencies": {
+    "linkifyjs": "2.1.4",
```
michaelneale: @sophistifunk this one? https://github.com/SoapBox/linkifyjs/blob/master/package.json so doesn't bring anything along, which is nice, right? (ie fairly small)

sophistifunk: Yeah there's not a lot of code in there. It has a dependency on JQ, but only for functions we're not using so we can just ignore it.

sophistifunk:
I can't seem to see what's actually going wrong here :-/

jenkins-design-language/src/js/components/ResultItem.jsx
```diff
@@ -78,6 +79,11 @@ export class ResultItem extends Component {
         }
     };
 
+    urlClicked = (e: Event) => {
+        console.log('urlClicked');
```
cliffmeyers: Probably just make sure we remove that before PR merge?

jenkins-design-language/package.json
cliffmeyers: I eyeballed the deps too and looks fine. I think as long as we verify that the bundles aren't bloating up after this change we are good. They do strange things sometimes :/

jenkins-design-language/src/js/components/ResultItem.jsx
sophistifunk: Nice catch, cheers

michaelneale:
Nice - this works great. Wire it in and should go for it. Very nice.

michaelneale:
@sophistifunk doesn't this need to be published/incorporated to take effect?

sophistifunk:
@NicuPascu it's just to stop the expando from opening / closing when you click a link. It's the easiest way to do it.

cliffmeyers:
Actually I think in the future we should be sure to publish every package change as part of the PR so it can through ATH. Otherwise we are just pushing that risk onto the next developer who actually needs to publish. Am quite confident this change won't be an issue, but.. famous last words :)

PR #1402, opened 2017-09-18
Bug/icon invalid icon prop

cliffmeyers:
# Description
- Found during downstream upgrade from Blue Ocean 1.1 -> 1.2. Since material icons impl changed, the icon names being used did not align with the new component. e.g. "search" instead of "ActionSearch". In particular this was a problem in `TextInput.iconLeft/iconRight`.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-core-js/npm-shrinkwrap.json
+++ b/blueocean-core-js/npm-shrinkwrap.json
@@ -1,11 +1,11 @@
 {
   "name": "@jenkins-cd/blueocean-core-js",
-  "version": "0.0.158-SNAPSHOT-nicu",
+  "version": "0.0.159-SNAPSHOT-1",
   "dependencies": {
     "@jenkins-cd/design-language": {
-      "version": "0.0.149",
-      "from": "@jenkins-cd/design-language@0.0.149",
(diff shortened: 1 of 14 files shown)


michaelneale:
@cliffmeyers does this need to be backported/cherry picked to release/1.2? 

cc @scherler @vivek

michaelneale:
nice one @cliffmeyers 

If this goes in, can you `git cherry-pick` the commit for a PR on release/1.2 (or vivek or I can do it, should be clean) as I expect this may be wanted there one way or another.

vivek:
@cliffmeyers Its change in JDL, does it not require JDL release to NPM and have that version used by rest of blueocean frontend?

cliffmeyers:
@michaelneale just FYI, this fix isn't critical as 1. all of our downstream code in Blue Ocean was already fixed to use appropriate values for `icon` prop, and 2. Blue Steel has done the same thing on their end.

For that reason, it's perhaps debatable whether we need to cherry pick to 1.2 if it will cause much extra work.

To @vivek point, yes absolutely I need to publish this. Will do now.

PR #1404, opened 2017-09-19
[FIX JENKINS-46959] Remove unnecessary "no branch"

batmat:
Removing all this logic since I don't think there's much value anyway.

With Git, the sha is enough to identify the exact code base associated to the thing in use. So I don't think it's worth spending any time making this work better/again.

I am still building this change locally to test it since I'm still mostly a React/jsx newbie, but pushing this for review as this seems relatively straightforward. I guess a code review is enough for testing, but again I'm building it locally to assess this anyway. 

# Description

See [JENKINS-46959](https://issues.jenkins-ci.org/browse/JENKINS-46959).

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

### DIFF
--- a/blueocean-web/src/main/js/DevelopmentFooter.jsx
+++ b/blueocean-web/src/main/js/DevelopmentFooter.jsx
@@ -14,17 +14,11 @@ export class DevelopmentFooter extends Component {
             return null;
         }
 
-        let includeBranch = true;
-
-        if(!revisionInfo.branch.includes('(no branch)')) {
-            includeBranch = false;
-        }
         return (
           <div className="development-footer">
               <span> {AppConfig.getConfig().version}&nbsp;</span>
               <span> &#183; Core {AppConfig.getJenkinsConfig().version}&nbsp;</span>
               <span> &#183; {revisionInfo.sha.substring(0,7)}&nbsp; </span>
-              { includeBranch && <span> &#183; {revisionInfo.branch}&nbsp;</span> }
               <span> &#183; {moment(revisionInfo.timestamp).format('Do MMMM YYYY hh:mm A')}</span>
               
           </div>


michaelneale:
thanks @batmat 
Congrats you are a 1337 react hacker now! A hoodie is in the mail ;)

Yes I think this is fine - will check with others but I think this makes sense. the branch name is occasionally helpful, but TBH I only ever look at the commit, it is the truth so I thin this is a good change. 

And any change that is deleted good is a good change. Deleted code has no bugs.

sophistifunk:
When it's master/trunk, do we have a branch name, or does it give "no branch"? Coz it seems nice to be able to see quickly if you're looking at master, or at a release/beta-tagged commit

michaelneale:
@sophistifunk nah - it never works properly when you need to. Commit is fine IMO

daniel-beck:
> the branch name is occasionally helpful

Given that apparently the removed condition has inverted logic, it looks like this never worked in the first place.

batmat:
FWIW, tested it yesterday locally:
image

scherler:
@daniel-beck actually that condition was added after the creation, so it worked before and I liked it for the reason @sophistifunk liked it "seems nice to be able to see quickly if you're looking at master, or at a release/beta-tagged commit"

blueocean-web/src/main/js/DevelopmentFooter.jsx
```diff
@@ -14,17 +14,11 @@ export class DevelopmentFooter extends Component {
             return null;
         }
 
-        let includeBranch = true;
-
-        if(!revisionInfo.branch.includes('(no branch)')) {
```
scherler:
```
if(revisionInfo.branch.includes('(no branch)')) {
```

would have been enough to fix the issue.

batmat:
@scherler though I'm still slightly unconvinced of the interest of that information, since a sha does help you find the *exact* state of the code. So that's why I chose getting rid of it.
Also, the reality is that it's nice but it was not working since a (very?) long time, so overall people didn't seem to miss it much already.
Less maintenance FTW.

PR #1406, opened 2017-09-19
Improvement/jenkins 43211 redux markdown cleanup

cliffmeyers:
# Description
- Still a work in process but wanted to get this open so I can more easily review CI / ATH results.
- See [JENKINS-43211](https://issues.jenkins-ci.org/browse/JENKINS-43211).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
yes yes yes! 

lets do this

michaelneale:
some conflicts, but when it is green, get it in.

cliffmeyers:
All right, hopefully this is ready to go after the latest series of commits. Getting `react-remarkable` out of `blueocean-dashboard` was ridiculous... maybe a good rant session for the next time we are in a room together with beer.

michaelneale:
@cliffmeyers I love rooms with beer!

Runnign again as that made it really angry... will take a look today...

cliffmeyers:
@michaelneale don't sweat it too much, if there's that many failures I probably made a mistake in the publishing or something. I had to really fight to get `react-remarkable` out of there due to npm bugs with shrinkwrap. I will have another look in the morning most likely

cliffmeyers:
@michaelneale yeah maybe I will take a look tonight once I get a little time. I may have mispublished core-js... resolving massive conflicts and shrink is hard to get right the first time.

michaelneale:
@cliffmeyers the discussion came up yesterday about yarn - I guess at some point may have to revisit that (as shrinkwrap pain wont go away any time soon I guess.. may be a little easier with publishing though).

michaelneale:
sorry I am really out of ideas... 

Running specific tests that look like they failed on CI work locally. hpi:run works locally. If things are busted, they are in a very very very subtle way... I really have no clue at all.

michaelneale:
I even ran the full ATH locally - and it ran to completion. All I can think is there some new race condition which is vulnerable in some environments, but I really can't see what is wrong. Perhaps a republish/hail mary is worth it. Other builds are ok.

cliffmeyers:
I've pushed two commits to see how the build behaves in CI.
- 1f83426 is just deletion of all redux-related code, and elimination of markdown usage in blueocean dashboard. markdown usage also removed in blueocean-core-js, but not published, so it's really still in the app.
- 1d4b676 includes publish and usage of new core-js which will remove markdown usage from parameterized run dialog as well.

If these builds both behave, then the two remaining subtasks are
- [ ] remove react-remarkable and redux + friends from deps
- [ ] remove redux + friends from package.json's  "jenkinscd/import" list. the more I think about it, maybe this change here is having the impact in CI, although hard to imagine why it's not repro locally.

cliffmeyers:
The second commit actually breaks everything in the same way as prior versions of the PR, so at least I'm getting a little bit closer to isolating it.

cliffmeyers:
Possible that breakage is being caused by i18n bugs being worked in #1257

cliffmeyers:
I merged #1412 into this branch temporarily to see if we can get the browser logs from ATH to see why this is failing.

michaelneale:
@cliffmeyers a comment on #1257 is that this failure is specific to that image... will comment there (good you are getting closer).

cliffmeyers:
I'm putting this PR on hold until either (A) we get #1412 working after migration off of LXD in CI environment, or (B) #1257 is merged with i18n fixes that may fix the underlying issue here.

IMO, the ROI on this PR is just not there given the time invested and the total opacity of errors in CI. If fixed by #1257, great, and if not then #1412 should give us some insight. I spent a couple hours trying to diff the blueocean.js file and my eyes started to bleed, so no more for now.

michaelneale:
@cliffmeyers good call I think. I think this is worth doing but not beating head bloody over yet. 
If you rebase this on #1257 might be interesting (but depends if that is a hassle), but yeah, can park this and revisit when things are in a better shape. Other fish to fry for now

cliffmeyers:
@michaelneale the last CI build definitely shows that the bundle is completely failing to load. See logs linked below, search for "GithubEditorTest":

https://ci.blueocean.io/blue/rest/organizations/jenkins/pipelines/blueocean/branches/improvement%252FJENKINS-43211-redux-markdown-cleanup/runs/29/nodes/57/steps/58/log/?start=0

There is an unhandled JS exception prior to the overall bundle death messag. Unfortunately the console logs from Chrome + Selenium don't seem to render objects, e.g.

`console.log('I am a string param and I render okay',  butComplexObjectWontDisplay);`

... which is a shame because this method of logging is very convenient and renders nicely in the browser.

I may fiddle with this a bit more to get better output. This also probably points to a need for some improved error handling in js-modules.

michaelneale:
thanks for update @cliffmeyers - continue the research!

Note that if you lock the build to the label "bare-docker4" it will likely get further. 
Good point on highlighting the need for error handling, can use this as a driver for that.

cliffmeyers:
@michaelneale I know I said I was going to put this on the shelf for a bit but my curiosity keeps getting the best of me and pulling me back in.

Yesterday I managed to get the runtime error handling slightly improved and it led me to find that the dashboard JS bundle was failing with a totally opaque error. The hpi's built in CI vs. my local are **2.3 MB vs. 1.6 MB**. Digging into the contents, I found that the "jenkins-js-extension.js" for dashboard is **~650 KB when built in CI** (the broken version) versus **3.5 MB when built locally** (the working version).

dashboard-ci-vs-local

There is a lot of library code missing from the bundle built in CI and that results in a broken bundle. I have to assume it's some subtle js-builder bug - perhaps some difference in browserify behavior between operating systems? The specific breakage occurs when code references to react-remarkable are removed from core-js, a new version is published, and then that dependency is installed into our modules.

I'd like to try and get a rough POC of drag and drop done this week, so going to really try to back burner this for a few days. Hopefully I can keep my curiosity at bay for now :)

michaelneale:
@cliffmeyers good update, I am sure this will yield something useful

BTW the new agents are online without LXD - so that removes one layer of confusion and reproducibility at least...

michaelneale:
@cliffmeyers might be worth looking at https://github.com/jenkinsci/blueocean-plugin/pull/1435 too - seems to be same deal.. this is fucked up.

cliffmeyers:
I've cherry-picked commits to fix the mfep setup and get those GitHub ATH fixes. If this succeeds then I just need to resolve shrink conflicts in master and do publish dance. Fingers crossed!

michaelneale:
Sweet. Wonder what impact on bundles/loading may be. In any case, deletes code makes no bugs.

kzantow:
Bundles? how dare you @michaelneale

michaelneale:
Too soon?

michaelneale:
OMG it passed @cliffmeyers ! 

So much deleted code... *luxuriates*

Did you end up using selenium with chrome on this one? (didn't look like a major change)

blueocean-dashboard/src/main/js/components/karaoke/components/InputStep.jsx
```diff
@@ -9,7 +9,6 @@ import {
     logging,
 } from '@jenkins-cd/blueocean-core-js';
 import { Alerts } from '@jenkins-cd/design-language';
-import Markdown from 'react-remarkable';
```
sophistifunk: /me *swoons*

pom.xml
```diff
@@ -20,6 +20,7 @@
     <jackson.version>2.8.9</jackson.version>
     <jenkins.version>2.46.3</jenkins.version> <!-- Should be kept in sync with Dockerfile FROM statement-->
     <javadoc.exec.goal>javadoc-no-fork</javadoc.exec.goal> <!-- stop initialize phase plugins executing twice -->
+    <frontend-version>1.6</frontend-version>
```
sophistifunk: Vas ist das?

michaelneale: that fixes the maven front end plugin to the version of 1.6 which has more sensible failure mode when running npm/scripts that return failure (to avoid that stupid thing of last week you found). 

This is already on master BTW (so we don't get a new front end plugin by surprise in future).

sophistifunk:
dovescry

cliffmeyers:
I just need to bring this up to date with master and it should ship now we found the root cause of the bundling issues.

michaelneale:
hey @cliffmeyers - we have selenium too so this could be brought up to date, and get it in!

cliffmeyers:
Probably going to focus on editor drag and drop before coming back to this one, but it should be much easier to finish up in light of other things having landed. 👍

michaelneale:
@cliffmeyers no worries at all. Given the impact downstream on steel - would like @scherler etc to sign off on this anyway (may not impact them at all, but just a heads up)

acceptance-tests/src/main/java/io/blueocean/ath/ATHJUnitRunner.java
```diff
@@ -75,8 +77,10 @@ public void evaluate() throws Throwable {
 
                 try {
                     next.evaluate();
+                    outputConsoleLogs();
```
scherler: How is this and the other changes in ATH related to redux?

blueocean-core-js/src/js/parameter/ParametersRunButton.jsx
```diff
@@ -10,6 +9,8 @@ import { ParameterApi as parameterApi, ParameterService, ParametersRender, suppo
 
 import logging from '../logging';
 const logger = logging.logger('io.jenkins.blueocean.core.ParametersRunButton');
+// eslint-disable-next-line no-unused-vars
+import Markdown from 'react-remarkable';
```
scherler: Why?

blueocean-dashboard/package.json
```diff
@@ -14,8 +14,8 @@
     "test:watch": "gulp test:debug",
     "bundle": "gulp bundle",
     "bundle:watch": "gulp bundle:watch",
-    "mvnbuild": "gulp bundle",
-    "mvntest": "gulp test lint"
+    "mvnbuild": "gulp lint bundle",
```
scherler: why lint in build and not in build?

Further did I understood wrong that this PR had been to drop redux? If so why it is still a dep?

blueocean-dashboard/src/main/less/forms/index.less
```diff
@@ -9,15 +9,17 @@
     justify-content: center;
 
     .Alert {
-        min-width: 400px;
```
scherler: I made the experience that for xbrowser you should have both. IE being especially special.

blueocean-dashboard/src/main/resources/jenkins/plugins/blueocean/dashboard/Messages.properties
```diff
@@ -248,6 +248,7 @@ servererror.errors.message=Message: {0}
 servererror.message=Message: {0}
 servererror.title=An unknown error has occurred. You may try again.
 Show.complete.logs=Show complete log
-inputStep.error.message=This pipeline uses input types that are unsupported. \nUse [Jenkins Classic]({0}) to resolve this input step.
+inputStep.error.linktext=Resolve Input
```
scherler: That will lead to `inputStep.error.linktext` being `Resolve Input` in all language other then `ru`

blueocean-dashboard/src/test/js/log-spec.js
```diff
@@ -101,44 +92,3 @@ describe("React component test of different runs", () => {
         assert.equal(wrapper.find('Step').length, runNodesFail.length)
     });
 });
-/*
```
scherler: Maybe it would be a good moment to add some unit tests?

blueocean-personalization/package.json
```diff
@@ -12,8 +12,8 @@
     "test:watch": "gulp test:debug",
     "bundle": "gulp bundle",
     "bundle:watch": "gulp bundle:watch",
-    "mvnbuild": "gulp bundle",
-    "mvntest": "gulp test lint"
+    "mvnbuild": "gulp lint bundle",
```
scherler: why?

blueocean-web/src/main/js/ErrorUtils.js
```diff
@@ -22,18 +22,26 @@ function isFirefox() {
 
 function logApplicationError(messageOrEvent) {
     const message = messageOrEvent.error || messageOrEvent;
-    console.error('Hnhandled Error: ', message);
 
-    if (messageOrEvent.preventDefault) {
-        messageOrEvent.preventDefault();
+    console.log('Unhandled Error 1: ' + JSON.stringify(messageOrEvent, null, 4));
```
scherler: why not use the logger but `console`?

jenkins-design-language/package.json
```diff
@@ -38,7 +38,7 @@
   ],
   "repository": {
     "type": "git",
-    "url": "https://github.com/jenkinsci/jenkins-design-language.git"
+    "url": "https://github.com/jenkinsci/blueocean-plugin.git"
```
scherler: Not really related to the PR, right?

pom.xml
scherler: ...but why here?

cliffmeyers:
@scherler appreciate your feedback. TL;DR: this branch is badly in need of a merge/rebase with master.

If you read the lengthy comment stream, you'll see that while trying to remove these deps, it resulted in a completely broke js-bundle and a lot of time spent trying to get better insight into why the application was broken in CI.

I will respond to your feedback once I'm able to do that. Cheers

michaelneale:
@scherler yeah like @cliffmeyers said, there was a bit of a mess here for a while. Let me explain. 



Things failed in strange ways where we noted that the bundle sizes made by js-builder were oddly small. On looking closer, entire bundles were missing. 
It took a while but we realised that there were bundle failures due to errant space characters and eslint rules, but the failures were getting ignored, and bundling continued. 

I believe this turned out to be due to upgrading the parent pom recently, which brought in a newer maven front end plugin which ignored failures like this due to the maven flag to ignore test failures (linting was run as part of the test phase). it is idiomatic for maven to do this as you collect the results and mark the build as unstable, but there were no results so it thought everything was fine. 

This was both a problem with the maven front end plugin (which was corrected in a later release I believe) and also the fact that we didn't "fix" the version so we wouldn't be surprised by it changing. The other thing we did was move lint checks out of test in package.json to part of build (which probably makes more sense) - and THEN the lint stuff started failing correctly. 

Clear?

sophistifunk:
This is why I drink.

cliffmeyers:
Closing this PR in favor of #1627

PR #1411, opened 2017-09-21
 [FIX JENKINS-47032] IE will give focus to StatusIndicator in creation flow

scherler:
# Description 

Can be seen in any creation flow. When using the keybaord to navigate in the create flow if you tab from an input field you get nothing and then tab again and you get the next element in the focus.

Basically you need to shift-tab or tab twice to move between the control elements when a statusIndicator is in between since it will "steal" the focus due the the svg not having focusable="false".

Happens in IE only, Edge is correct as is Chrome.

Testing needs to be done manually in an IE.

See [JENKINS-47032](https://issues.jenkins-ci.org/browse/JENKINS-47032).

# Note on IE

All SVG need to have focusable="false" otherwise IE may give them focus even if they are not anymore in the DOM and your document.activeElement becomes null and TAB_INDEX starts at top level (either document or browser(!)).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/jenkins-design-language/src/js/components/status/StatusIndicator.jsx
+++ b/jenkins-design-language/src/js/components/status/StatusIndicator.jsx
@@ -66,7 +66,6 @@ class StatusIndicator extends Component {
             width = '24px',
             height = '24px',
             noBackground,
-            focusable = false,
         } = this.props;
 
         const groupClasses = [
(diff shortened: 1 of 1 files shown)


michaelneale:
huh, very nice @scherler !

@sophistifunk quick review? should be fine.

sophistifunk:
Making it an option for the status indicator (which doesn't serve as input) seems overkill, but code looks good 🐝

sophistifunk:
It's good to know that this behaviour exists though, and a bummer that it's browser-specific. Basically tab-navigation is awful across the entirety of BO, particularly WRT popups, dropdowns, and other modals. All of which are outside the scope of this issue!

scherler:
@sophistifunk yeah I agree regarding the prop, I guess that will never be set true, so dropping it.

Cheers for the quick review as always!

PR #1413, opened 2017-09-22
Enhancement/jenkins 47041 check plugin compatibility

michaelneale:
# Description

See [JENKINS-47041](https://issues.jenkins-ci.org/browse/JENKINS-47041).

This ads a test that installs @kzantow's executor view plugin, and then checks it loads. Rudimentary test of plugin backwards compatibility. 

This will not build yet, as the plugin needs to be published to the real UC to be used in this way. 
If you want to try this locally - checkout https://github.com/kzantow/blueocean-executor-info-plugin and `mvn install` it locally, you can then run the ATH and see it light up. 

PTAL @cliffmeyers @kzantow @vivek 

And @kzantow - if we publish this plugin, can make this real. 

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/ExecutorPluginPage.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public void checkComputers() {
+            wait.until(ExpectedConditions.visibilityOfElementLocated(getSelectorExecutor()));
+
+
+    }
+
+    public By getSelectorExecutor() {
+        return By.cssSelector(".executor-info-cell");
```
michaelneale: are there other things we can do to check that things like JDL still work? what should appear? can we look for errors of some kind? must be other things we can do

michaelneale:
dependent on: https://issues.jenkins-ci.org/browse/HOSTING-427

michaelneale:
now waiting on https://github.com/jenkins-infra/repository-permissions-updater/pull/452 to be merged so I can deploy it

michaelneale:
PTAL @kzantow @vivek @alvarolobato @i386 

this checks that an external plugin actually loads in blue ocean (ie backwards compat)

michaelneale:
@kzantow not sure if circular, but the way the ATH runs (with maven) means it messes up with the ones from the blue ocean under test, so need to explicitly say not to use them (and thus it will use the blue ocean being tested).

PR #1416, opened 2017-09-24
upgrade declarative to the released version 1.2

michaelneale:
# Description

A simple version bump. see 

https://issues.jenkins-ci.org/browse/JENKINS-47083

@abayer @kzantow anything more than this needed?

### DIFF
--- a/blueocean-rest-impl/pom.xml
+++ b/blueocean-rest-impl/pom.xml
@@ -36,10 +36,6 @@
             <artifactId>display-url-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.jenkins-ci.plugins</groupId>
-            <artifactId>scm-api</artifactId>
-        </dependency>
-        <dependency>
             <groupId>${project.groupId}</groupId>
             <artifactId>blueocean-jwt</artifactId>
         </dependency>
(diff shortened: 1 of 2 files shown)


michaelneale:
@vivek will need help resolving this one with all the bounds checking... (which is new, and probably good, I hope)

vivek:
@michaelneale upper bound check is enforced from parent plugin-pom, and yeah its supposed to do good to ensure one version of plugin/library version in class path. I am going to take a look at it.

vivek:
@michaelneale pipeline-model-definition 1.2 was bringing in scm-api 2.2.2, we needed to bump it up from 2.2.1 to 2.2.2 as well.

michaelneale:
this one is likely made redundant by https://github.com/jenkinsci/blueocean-plugin/pull/1394 - so will give that one a chance to land first...

vivek:
@michaelneale This PR includes cleanup in rest-impl too (removes scm-api dependency). Can you get that change in #1394 as well? Or maybe upgrade declarative with this other change in this PR?

michaelneale:
@vivek #1394 could be updated to have the rest-impl cleanup of removing scm api I think if we wanted...

PR #1421, opened 2017-09-26
JENKINS-47114# Upgrade plugin-pom and jenkins-test-harness

vivek:
Fixes running unit tests from inside IDEA.

# Description

See [JENKINS-47114](https://issues.jenkins-ci.org/browse/JENKINS-47114).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.33</version>
+    <version>2.35</version>
   </parent>
 
   <groupId>io.jenkins.blueocean</groupId>
@@ -31,7 +31,7 @@
     <jacoco.line.coverage>0.70</jacoco.line.coverage>
     <jacoco.missedclass.coverage>0.00</jacoco.missedclass.coverage>
     <hpi.dependencyResolution>runtime</hpi.dependencyResolution>
-    <jenkins-test-harness.version>2.23</jenkins-test-harness.version>
+    <jenkins-test-harness.version>2.27</jenkins-test-harness.version>
     <scm-api.version>2.2.2</scm-api.version>
     <git.version>3.5.1</git.version>
   </properties>
@@ -143,6 +143,18 @@
   </pluginRepositories>
 
   <dependencies>
+      <!-- Needed for jenkins-test-harness -->
+      <dependency>
+          <groupId>org.eclipse.jetty</groupId>
+          <artifactId>jetty-webapp</artifactId>
+          <scope>test</scope>
+      </dependency>
+      <dependency>
+          <groupId>org.eclipse.jetty</groupId>
+          <artifactId>jetty-security</artifactId>
+          <scope>test</scope>
+      </dependency>
+
       <dependency>
           <groupId>org.mockito</groupId>
           <artifactId>mockito-core</artifactId>
@@ -556,6 +568,27 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.jenkins-ci.main</groupId>
+            <artifactId>jenkins-test-harness</artifactId>
+            <version>${jenkins-test-harness.version}</version>
+            <scope>test</scope>
+            <!-- jenkins-test-harness depends on jetty 9.4.5 that doesn't play well with wiremock (uses jetty 9.2.22) -->
+            <exclusions>
+                <exclusion>
+                    <groupId>org.eclipse.jetty</groupId>
+                    <artifactId>jetty-webapp</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.eclipse.jetty</groupId>
+                    <artifactId>jetty-server</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.eclipse.jetty</groupId>
+                    <artifactId>jetty-security</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
             <version>1.10</version>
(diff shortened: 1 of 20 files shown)


vivek:
Multiple test failures due to probably issue with build agent. Running it again.
 
```
getOrganizationsAndRepositories(io.jenkins.blueocean.blueocean_github_pipeline.GithubApiTest)  Time elapsed: 1.043 sec  <<< ERROR!
com.github.tomakehurst.wiremock.common.FatalStartupException: java.lang.RuntimeException: java.lang.IllegalStateException: Insufficient threads: max=10 < needed(acceptors=4 + selectors=8 + request=1)
	at com.github.tomakehurst.wiremock.WireMockServer.start(WireMockServer.java:146)
	at com.github.tomakehurst.wiremock.junit.WireMockRule$1.evaluate(WireMockRule.java:68)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:533)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.RuntimeException: java.lang.IllegalStateException: Insufficient threads: max=10 < needed(acceptors=4 + selectors=8 + request=1)
	at com.github.tomakehurst.wiremock.jetty9.JettyHttpServer.start(JettyHttpServer.java:132)
	at com.github.tomakehurst.wiremock.WireMockServer.start(WireMockServer.java:144)
	... 6 more
Caused by: java.lang.IllegalStateException: Insufficient threads: max=10 < needed(acceptors=4 + selectors=8 + request=1)
	at org.eclipse.jetty.server.Server.doStart(Server.java:414)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68)
	at com.github.tomakehurst.wiremock.jetty9.JettyHttpServer.start(JettyHttpServer.java:130)
	... 7 more
```

michaelneale:
@vivek build failing due to code coverage violations - any idea why when nothing there has changed?

vivek:
@michaelneale no, code coverage is fine. Scroll up to see the error, its `Caused by: java.lang.IllegalStateException: Insufficient threads: max=10 < needed(acceptors=4 + selectors=8 + request=1)
`

The failure is legit, first I thought its build agent issue, but running locally I could reproduce as well. Looking in to it.

vivek:
@michaelneale looks like caused by https://github.com/jenkinsci/jenkins-test-harness/pull/69/files. jenkins-test-harness limits acceptor and selector to hard coded 1, wire mock uses default jetty config where it requires 4 and 8 respectively. Checking if there is way to configure jetty in wire mock.

michaelneale:
@vivek ok

michaelneale:
@vivek well I can certainly run this test individually locally, so that is good... now if we can get the thread thing sorted - should get this in.

michaelneale:
@vivek if you get this passing (and will need to fix the JWT test, which should be fixed on master shortly) then get it in. Locally worked well.

vivek:
@michaelneale its bit complicated that error (thread one) I could fix but there is mess around jetty version in use by jenkins-test-harness(9.4.5) and wiremock(9.2.22). There is API incompatibility between them and we are picking one of the higher version (test-harness) but then that causes runtime failure:

```
java.lang.NoSuchMethodError: org.eclipse.jetty.util.ssl.SslContextFactory.selectCipherSuites([Ljava/lang/String;[Ljava/lang/String;)[Ljava/lang/String;
```

Well picking one vs another version won't work because both wire mock and test-harness use these version specific API. So fix is really, either test-harness downgrades jetty version to 9.2.22 or wire mock upgrades jetty to 9.4.5.

I have sent details to @jglick and trying to figure best course of action/workaround(if any).

michaelneale:
@vivek if no quick solution could you post to jenkins-ux mailing list with instructions on how to run individual tests so others can reference it?

vivek:
@michaelneale sure will do

michaelneale:
getting closer. A few errors now seem to be related to eclipse/jetty I guess: 

```
createPersonalSSHKey – io.jenkins.blueocean.ssh.UserSSHKeyTest
a few seconds
anonymousUserToken – io.jenkins.blueocean.auth.jwt.impl.JwtAuthenticationServiceImplTest
a few seconds
getToken – io.jenkins.blueocean.auth.jwt.impl.JwtAuthenticationServiceImplTest
a few seconds
testOrg – io.jenkins.blueocean.service.embedded.BlueOceanWebURLBuilderTest
a few seconds
test_freestyle – io.jenkins.blueocean.service.embedded.BlueOceanWebURLBuilderTest
a few seconds
testCustomerUrlMapper – io.jenkins.blueocean.service.embedded.BlueOceanWebURLBuilderTest
a few seconds
```

eg:

```
org/eclipse/jetty/security/UserStore
Stacktrace
java.lang.NoClassDefFoundError: org/eclipse/jetty/security/UserStore
	at org.jvnet.hudson.test.JenkinsRule.configureUserRealm(JenkinsRule.java:685)
	at org.jvnet.hudson.test.JenkinsRule.createWebServer(JenkinsRule.java:659)
```

vivek:
Almost there, rebased with master to pull ssh user key test fix and other latest stuff.

michaelneale:
:bee: works lovely @vivek

pom.xml
```diff
@@ -790,57 +816,74 @@
...
                     <artifactId>asm</artifactId>
                 </exclusion>
             </exclusions>
         </dependency>
+        <!--
+            wiremock jetty, depends on 9.2.22.v20170606, jenkins-test-harness on jetty 9.4.5.
+
+            jetty-util 9.4.5 breaks backward compatibility, where
+
+            SslContextFactory.selectCipherSuites([Ljava/lang/String;[Ljava/lang/String;)[Ljava/lang/String;
```
jtnord: https://github.com/tomakehurst/wiremock/issues/551#issuecomment-268647572

use wiremock-standalone that shades jetty...  did you try this?

vivek: @jtnord thanks for the pointer, no I haven't. Happy to try as it will clean all that work around.

PR #1422, opened 2017-09-26
temporarily Ignore flaky userSSH test

michaelneale:
# Description

@kzantow just trying this - this test seems problematic


michaelneale:
This passes - the test passes when run individually too. The problem is the setting of a naughty static that turns JWT on or off.

PR #1423, opened 2017-09-26
restore-scm api dep

michaelneale:
this was passing when on keiths fork. Branches are easier to see what was the difference before merging so this is just a guess.

### DIFF
--- a/blueocean-rest-impl/pom.xml
+++ b/blueocean-rest-impl/pom.xml
@@ -35,6 +35,10 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>display-url-api</artifactId>
         </dependency>
+         <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>scm-api</artifactId>
+        </dependency>        
         <dependency>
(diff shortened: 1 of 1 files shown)


blueocean-rest-impl/pom.xml
```diff
@@ -35,6 +35,10 @@
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>display-url-api</artifactId>
         </dependency>
+         <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>scm-api</artifactId>
```
vivek: 🐛  We can't add it here. scm api shouldn't be in this module. Some other class path issue (sigh)... will get to it. Lets do the right way.

michaelneale: @vivek this is just to see if it changes the result of the failing test on master... which seems a brittle test

vivek: ah ok 👍

michaelneale: so - as you can see @vivek this made it pass by probably changing how the test class is loaded :(

PR #1424, opened 2017-09-26
Set jwt disable const for test

michaelneale:
Not really sure how this test is meant to work when parent sets the jwt mode to on in a static... so trying this...

### DIFF
--- a/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/BlueOceanConfigProperties.java
+++ b/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/BlueOceanConfigProperties.java
@@ -10,6 +10,6 @@ public class BlueOceanConfigProperties {
     public static final String BLUEOCEAN_ROLLBAR_ENABLED_PROPERTY = "BLUEOCEAN_ROLLBAR_ENABLED";
 
     public static final boolean ROLLBAR_ENABLED = Boolean.getBoolean(BLUEOCEAN_ROLLBAR_ENABLED_PROPERTY);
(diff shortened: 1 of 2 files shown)


blueocean-commons/src/main/java/io/jenkins/blueocean/commons/BlueOceanConfigProperties.java
```diff
@@ -10,6 +10,6 @@
     public static final String BLUEOCEAN_ROLLBAR_ENABLED_PROPERTY = "BLUEOCEAN_ROLLBAR_ENABLED";
 
     public static final boolean ROLLBAR_ENABLED = Boolean.getBoolean(BLUEOCEAN_ROLLBAR_ENABLED_PROPERTY);
-    public static final boolean BLUEOCEAN_FEATURE_JWT_AUTHENTICATION = Boolean.getBoolean(BLUEOCEAN_FEATURE_JWT_AUTHENTICATION_PROPERTY);
+    public static boolean BLUEOCEAN_FEATURE_JWT_AUTHENTICATION = Boolean.getBoolean(BLUEOCEAN_FEATURE_JWT_AUTHENTICATION_PROPERTY);
```
vivek: Oh, making mutable global variable in production code is bad 🐛

michaelneale: @vivek do you want to back it out if you like? I am not fussed if needs be (but the test is broken right now, it can't set the state of a final thing reliably)

PR #1428, opened 2017-09-26
[FIX JENKINS-47148] Add a11y support for textInput

scherler:
# Description

See [JENKINS-47148](https://issues.jenkins-ci.org/browse/JENKINS-47148).

# Submitter checklist
- [x] Link to ticket in description, if appropriate.
- [x] Change is code complete and matches issue description.
- [x] Commit messages are self contained and descriptive.
- [x] Testing notes provided and reviewed in the JIRA.
- [ ] Link to any acceptance tests not part of this PR.

# Reviewer checklist
- [ ] Run the changes and verified the change matches the issue description.
- [ ] Reviewed the code.
- [ ] Verified that the tests match the testing notes.
- [ ] Verified the commit messages are self contained and descriptive.

@reviewbybees @cloudbees/team-astro

### DIFF
--- a/jenkins-design-language/src/js/components/forms/TextInput.jsx
+++ b/jenkins-design-language/src/js/components/forms/TextInput.jsx
@@ -24,7 +24,7 @@ export class TextInput extends React.Component {
     render() {
         const classLeft = this.props.iconLeft ? 'u-icon-left' : '';
         const classRight = this.props.iconRight ? 'u-icon-right': '';
-        const { ariaLabel = this.props.placeholder } = this.props;
+        const ariaLabel = this.props['aria-label'] || this.props.placeholder;
 
         return (
             <TextControl {...this.props} className={`TextInput ${this.props.className} ${classLeft} ${classRight}`}>
@@ -47,7 +47,7 @@ TextInput.propTypes = {
     iconRight: PropTypes.string,
     onChange: PropTypes.func,
     onBlur: PropTypes.func,
-    ariaLabel: PropTypes.string
+    'aria-label': PropTypes.string
 };
 
 TextInput.defaultProps = {
--- a/jenkins-design-language/src/js/components/forms/TextInput.jsx
+++ b/jenkins-design-language/src/js/components/forms/TextInput.jsx
@@ -24,11 +24,12 @@ export class TextInput extends React.Component {
     render() {
         const classLeft = this.props.iconLeft ? 'u-icon-left' : '';
         const classRight = this.props.iconRight ? 'u-icon-right': '';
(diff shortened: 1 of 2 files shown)


michaelneale:
very nice @scherler !

tests and aria support 👌🏽

jenkins-design-language/src/js/components/forms/TextInput.jsx
```diff
@@ -46,6 +47,7 @@ TextInput.propTypes = {
     iconRight: PropTypes.string,
     onChange: PropTypes.func,
     onBlur: PropTypes.func,
+    ariaLabel: PropTypes.string
```
sophistifunk: Why not stick with `aria-label`? Doing it this way means we're using two different prop names depending on whose component we're using.

michaelneale: other ones aren't snake-case tho?

sophistifunk: What I mean is in your calling render you'd have `aria-label` on html elements or other-people's components, and `ariaLabel` on our TextInput component.

scherler: @sophistifunk that is more or less the "react-way" to solve the `-` problem. 

You cannot do `const { ariaLabel = this.props.placeholder } = this.props;` anymore because this.props now contains `aria-label` and you cannot do ` const {aria-label}` so that needs to become `const ariaLabel = this.props['aria-label'] || this.props.placeholder;`

I changed it now 
thingdoforlove

scherler: @sophistifunk Actually the preferred way is to have something like

TextInput.propTypes = {
    'aria-*: PropTypes.string
}

...and allow arbitrary `aria` props, but I do not know how to do that.

sophistifunk:
bbm-bueno

jenkins-design-language/src/js/components/forms/TextInput.jsx
sophistifunk: Yes a general solution would be better, next time we see this kind of thing we should cook up a little helper func. There is definitely documentation recommending not renaming `data-foo` to `dataFoo` though, and I'm assuming it holds for `aria-foo`.

sophistifunk: I think it's just something we're going to have eat, since W3C likes hyphens in names and ECMA does not :)

scherler: @sophistifunk +1

PR #1445, opened 2017-10-03
[FIX JENKINS-47249] Remove name parallel group dialog (1.3 branch)

kzantow:
# Description

Removes "Name Parallel Group" dialog. To test: create a parallel group. Needs more tests in this area.

See [JENKINS-47249](https://issues.jenkins-ci.org/browse/JENKINS-47249).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-editor/src/main/js/components/editor/EditorMain.jsx
+++ b/blueocean-pipeline-editor/src/main/js/components/editor/EditorMain.jsx
@@ -17,7 +17,6 @@ import pipelineValidator from '../../services/PipelineValidator';
 import { ValidationMessageList } from './ValidationMessageList';
 import focusOnElement from './focusOnElement';
 import debounce from 'lodash.debounce';
-import { Dialog, TextInput, FormElement } from '@jenkins-cd/design-language';
 
 type Props = {
 };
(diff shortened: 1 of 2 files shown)


michaelneale:
so this is against release/1.3 and presumably will then go back to master after that?

kzantow:
@michaelneale yep, will cherry pick back after verified & merged here.

michaelneale:
this is also bitten by the flaky create test, which master has cc @vivek @imeredith 

something happened with github to affect the timing that is making this break, or environment (it will work locally).

michaelneale:
@vivek WDYT about this for 1.3?

michaelneale:
maybe it does end up easier to do on master and then cherry pick across (so we aren't cherry picking the "bad" way...)

vivek:
@michaelneale isn't it late for 1.3 if we plan to release early this week? I see the associated ticket is blocker so maybe we should add it to 1.3 but then need to release new beta and check with @alvarolobato if he can take this new beta in.

michaelneale:
@vivek it is - but maybe for 1.3.1 if we do it. - so hold fire on this just yet... (perhaps move this to master).

vivek:
@michaelneale sounds good. I can open a PR for it on release/1.3 with DO NOT MERGE so that its tracked and merged post 1.3.

PR #1452, opened 2017-10-03
1.3 branch: Skip creation test

michaelneale:
This was passing, and with no change isn't, so may need to skip it for this branch


michaelneale:
it does yes - just checking a few other things before we resort to this...

PR #1453, opened 2017-10-03
Redirect on timeout

michaelneale:
# Description

If unable to auto redirect, load the activity page (which is future functionality)


michaelneale:
ok looking better now - when green lets do it

PR #1454, opened 2017-10-04
1.3 branch: Redirect on timeout (#1453)

michaelneale:
from master... to correct timing out test


michaelneale:
please review @vivek @imeredith

imeredith:
lgtm, although i did remove that retry, but maybe we wont need it now.

PR #1461, opened 2017-10-05
do lint at earlier stage

michaelneale:
# Description

see: https://issues.jenkins-ci.org/browse/JENKINS-47304 

lint should be done as mvnbuild not test phase

### DIFF
--- a/blueocean-dashboard/package.json
+++ b/blueocean-dashboard/package.json
@@ -14,8 +14,8 @@
     "test:watch": "gulp test:debug",
     "bundle": "gulp bundle",
     "bundle:watch": "gulp bundle:watch",
-    "mvnbuild": "gulp bundle",
-    "mvntest": "gulp test lint"
+    "mvnbuild": "gulp lint bundle",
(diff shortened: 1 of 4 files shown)


cliffmeyers:
Should we intentionally introduce a lint error in dashboard to validate it works as we expect? Then remove lint error, let build pass, and merge?

michaelneale:
@cliffmeyers will do once I know this is green

michaelneale:
ok I made a branch off of this: https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/testing%2Flint-fail/1/pipeline which confirms it doe fail \o/ so I will wait for this to green, and get it in

michaelneale:
This is due to https://github.com/eirslett/frontend-maven-plugin/commit/39bcb48ea6e248776bdd9fc189406ba690739257#diff-f28d4bf058f4adf52359a52fb6841edfR118 being changed - ie maven front end will NEVER fail on things in the test phase, which is where lint was running.

PR #1462, opened 2017-10-06
Do not allow the new branch input in the editor to accept space chars

NicuPascu:
# Description
Couldn't use the TextInput component for this because of the way it manages its state value

See [JENKINS-46104](https://issues.jenkins-ci.org/browse/JENKINS-46104).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/acceptance-tests/src/test/java/io/blueocean/ath/live/GithubEditorTest.java
+++ b/acceptance-tests/src/test/java/io/blueocean/ath/live/GithubEditorTest.java
@@ -129,7 +129,7 @@ public class GithubEditorTest {
         sseClient.clear();
         BranchPage branchPage = activityPage.clickBranchTab();
         branchPage.openEditor("master");
-        editorPage.saveBranch("new-branch");
+        editorPage.saveBranch("new - branch");
         activityPage.checkUrl();
         activityPage.getRunRowForBranch("new-branch");
 
--- a/blueocean-pipeline-editor/src/main/js/EditorPage.jsx
+++ b/blueocean-pipeline-editor/src/main/js/EditorPage.jsx
@@ -36,6 +36,8 @@ class SaveDialog extends React.Component {
            { branch: branch, toString: () => ['Commit to ', <i>{branch}</i>]},
            { branch: '', toString: () => `Commit to new branch`},
        ];
+
+       this.processBranchName = this.processBranchName.bind(this);
     }
 
     save() {
@@ -82,6 +84,12 @@ class SaveDialog extends React.Component {
         this.setState({ saving: false, errorMessage });
     }
 
+    processBranchName(event) {
+        const cleanBranchName = event.target.value.replace(/\s/g, "");
+
+        this.setState({branch: this.branchOptions[1].branch = cleanBranchName});
+    }
+
     render() {
         const { branch } = this.props;
         const { errorMessage } = this.state;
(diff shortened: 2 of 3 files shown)


blueocean-pipeline-editor/src/main/js/EditorPage.jsx
```diff
@@ -103,8 +111,15 @@ class SaveDialog extends React.Component {
...
                 <div className="indent-form" style={{marginBottom: '-6px'}}>
                 <FormElement className="customBranch" errorMessage={this.state.branchError}>
-                    <TextInput placeholder="my-new-branch" onChange={value => this.setState({branch: this.branchOptions[1].branch = value})}
-                        disabled={this.state.branch !== this.branchOptions[1].branch || this.state.saving} />
+                    <div className="TextInput">
+                        <input
+                            className="TextInput-control"
+                            type="text"
+                            value={this.state.branch}
+                            onChange={this.processBranchName}
```
kzantow: only issue is, i would do `onChange={event => this.processBranchName(event)}` instead of the `.bind(this)` earlier

michaelneale:
might want to update to master for react and the fix for lint

michaelneale:
the failure may be legit, running it again to check

michaelneale:
@NicuPascu this seems to break the editor somehow: 

https://ci.blueocean.io/job/blueocean/job/bug%252FJENKINS-46104_branch_name_shouldnt_allow_spaces/4/artifact/acceptance-tests/target/screenshots/io.blueocean.ath.live.GithubEditorTest_testEditor.png

https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/bug%2FJENKINS-46104_branch_name_shouldnt_allow_spaces/4/tests

michaelneale:
@i386 are we ok with not allowing spaces (I am sure there are some SCMs that may allow it) in the editor save dialog? (I think I am ok with it, as long as this passes)

i386:
@michaelneale I am OK with that.

michaelneale:
ok @NicuPascu is there any chance we can get a test for this - unit or ATH would be fine?

NicuPascu:
I will try to improve the GithubEditorTest ATH to test this behaviour in that input

michaelneale:
It may be something that can easily be added here: 

https://github.com/jenkinsci/blueocean-plugin/blob/master/acceptance-tests/src/test/java/io/blueocean/ath/live/GithubEditorTest.java#L132

add something that sets it to something with a space, and check it is in error state, before trying the real one.

blueocean-pipeline-editor/src/main/js/EditorPage.jsx
```diff
@@ -82,6 +82,12 @@ class SaveDialog extends React.Component {
         this.setState({ saving: false, errorMessage });
     }
 
+    processBranchName(event) {
+        const cleanBranchName = event.target.value.replace(/\s/g, "");
+
+        this.setState({branch: this.branchOptions[1].branch = cleanBranchName});
```
cliffmeyers: This code looks susceptible to a NPE.

cliffmeyers: Also assignment inside of a `setState` like that is not something I've seen us do before, should probably be pulled out to its own line with null / index check.

cliffmeyers: Sorry, I see the original code was actually written that way, but I would split it up anyways.

acceptance-tests/src/test/java/io/blueocean/ath/live/GithubEditorTest.java
```diff
@@ -129,7 +129,7 @@ public void testEditor() throws IOException {
         sseClient.clear();
         BranchPage branchPage = activityPage.clickBranchTab();
         branchPage.openEditor("master");
-        editorPage.saveBranch("new-branch");
+        editorPage.saveBranch("new - branch");
```
michaelneale: Hrm - I think that the test for this should be explicit in this file - ie the EditorPage.java stays as it was, and first it "tries" with spaces, gets exception, and then tries without.

michaelneale:
:bee: nice one

PR #1467, opened 2017-10-09
[FIX JENKINS-47357] release/1.3: bitbucket unable to save

kzantow:
# Description

Unable to save to bitbucket due to a classloader issue with the `apache-httpcomponents-4-api` plugin. Need ATH tests for this!!

See [JENKINS-47357](https://issues.jenkins-ci.org/browse/JENKINS-47357).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-bitbucket-pipeline/pom.xml
+++ b/blueocean-bitbucket-pipeline/pom.xml
@@ -43,9 +43,16 @@
       </exclusions>
     </dependency>
 
-    <dependency>
-      <groupId>org.jenkins-ci.plugins</groupId>
-      <artifactId>apache-httpcomponents-client-4-api</artifactId>
+    <dependency> <!-- need to use this instead of the plugin due to a classloader issue -->
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>fluent-hc</artifactId>
+      <version>4.5.3</version>
+    </dependency>
+
+    <dependency> <!-- need to use this instead of the plugin due to a classloader issue -->
+      <groupId>org.apache.httpcomponents</groupId>
+      <artifactId>httpmime</artifactId>
+      <version>4.5.3</version>
     </dependency>
 
     <dependency>


michaelneale:
On creating a pipeline - it never went to activity screen (bitbucket cloud). 

And save did not work for me (opening an existing one)

michaelneale:
Actually it did save the content, but it never goes to activity screen: 

but times out: 

```
Hnhandled Error:  TypeError: Cannot read property 'message' of undefined
    at SaveDialog.showError (jenkins-js-extension.js:5660)
    at jenkins-js-extension.js:5643
    at jenkins-js-extension.js:6368
    at SaveApi._cleanup (jenkins-js-extension.js:6750)
    at jenkins-js-extension.js:6734
```

I am not sure if the BB support has ever worked very well

michaelneale:
hrm ok - on a fresh setup it seemed to work for me in both cases... so I will chalk that up to something else. 


ok once this is green - I think :bee: (unless @vivek opposes)

vivek:
@michaelneale @kzantow I don't think `apache-httpcomponents-4-api` is not causing LinkageError. Its wrapper plugin over httpcomponent libs. Some other plugin is loading httpcomponents libraries directly. Quick look tells me git, blueocean-display-url depend on httpcomponents libraries directly. 

They should be fixed to depend on `apache-httpcomponents-4-api`, in the meantime we can exclude httpcomponents dependency from them. Basically we want to depend on `apache-httpcomponents-4-api` so that there is one copy of these classes loaded by one class loader. I am afraid this fix might cause LinkageError somewhere else 🐛 

I am going to look in to it.

kzantow:
Superseded by #1470

michaelneale:
@vivek so this one is subsumed by https://github.com/jenkinsci/blueocean-plugin/pull/1470 right?

PR #1468, opened 2017-10-10
JIRA and Github issues not showing up in REST

i386:
# Description

JIRA and Github issues were not loading via REST. On further inspection of the logs, it looks like there was a classloading error for Guava's `MoreObjects.firstNotNull`.  This switches it to use the old API and now it all works.

@vivek potential class loading problem - Guava :'(

See [JENKINS-46017](https://issues.jenkins-ci.org/browse/JENKINS-46017).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubIssue.java
+++ b/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubIssue.java
@@ -1,9 +1,7 @@
 package io.jenkins.blueocean.blueocean_github_pipeline;
 
 import com.google.common.base.Function;
-import com.google.common.base.MoreObjects;
 import com.google.common.base.Objects;
-import com.google.common.base.Predicate;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
@@ -17,9 +15,7 @@ import jenkins.scm.api.SCMSource;
 import org.jenkinsci.plugins.github.config.GitHubServerConfig;
 import org.jenkinsci.plugins.github_branch_source.GitHubSCMSource;
 import org.jenkinsci.plugins.github_branch_source.HttpsRepositoryUriResolver;
-import org.kohsuke.github.GitHub;
 
-import javax.annotation.Nullable;
 import java.util.Collection;
 import java.util.List;
 import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


i386:
@kohsuke here is one where everything is fine when the plugin is tested but does not work when installed.

michaelneale:
@i386 from what I can tell the change that triggered this class loading is https://github.com/jenkinsci/blueocean-plugin/commit/83e8302eea0cf9395f94e20d5f06c2e1b2852f67 right? 
ie 0 dependency change, just happened to load it from a different place 

Not good.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubIssue.java
```diff
@@ -65,7 +62,7 @@ public String getURL() {
                 return null;
             }
             GitHubSCMSource gitHubSource = (GitHubSCMSource)source;
-            String apiUri = MoreObjects.firstNonNull(gitHubSource.getApiUri(), GitHubServerConfig.GITHUB_URL);
+            String apiUri = Objects.firstNonNull(gitHubSource.getApiUri(), GitHubServerConfig.GITHUB_URL);
```
michaelneale: couldn't this just be trivial code vs a guava import?

michaelneale:
```
String apiUri = (gitHubSource.getApiUri() == null) ? gitHubSource.getApiUri() : GitHubServerConfig.GITHUB_URL;
```

vivek: This sounds fine to me 👍. However there are other instances where we are using guava, those uses can surprise us. Either we should somehow enforce so that development time we see version of guava coming from core, or at least compile time it generates error if there are usage from later version.

Or stop using guava or shade it or start using jdk8 for some of its functionalities are over there or write such utilities ourselves.

michaelneale:
hey @i386 https://github.com/jenkinsci/blueocean-plugin/pull/1469 is same but sans guava. 

@NicuPascu does https://github.com/jenkinsci/blueocean-plugin/pull/1469 also work with JIRA etc?

NicuPascu:
@michaelneale I'll check now

NicuPascu:
Just tested and confirmed that https://github.com/jenkinsci/blueocean-plugin/pull/1469 also fixes the issues obj problem I had.

michaelneale:
Closing as other non guava one also does the job (will merge that shortly)

PR #1474, opened 2017-10-11
Showcase of example output from prettier, across core-js, JDL and Dashboard

sophistifunk:
This PR is for discussion purposes, not to be merged. It does not contain the code that performs the transform, for clarity's sake.


blueocean-core-js/src/js/Fullscreen.jsx
```diff
@@ -3,11 +3,7 @@ import React, { PropTypes } from 'react';
 /**
  * Simple component to render a full-page overlay without any decoration
  */
-const Fullscreen = ({ className, children }) => (
-    <div className={`fullscreen ${className}`}>
-        {children}
-    </div>
-);
+const Fullscreen = ({ className, children }) => <div className={`fullscreen ${className}`}>{children}</div>;
```
kzantow: already don't like this...

.eslintrc
```diff
@@ -12,7 +12,7 @@
   "rules": {
     "eol-last": 1,
     "no-unused-vars": [1],
-    "max-len": [1, 160, 4],
+    "max-len": [0],
```
kzantow: is prettier using these rules? i think if we set a max length, it would help some of the formatting I already don't like

blueocean-core-js/src/js/NotFound.jsx
```diff
@@ -11,11 +11,21 @@ const translate = i18nTranslator('blueocean-web');
 export default () => (
     <Fullscreen className="errorscreen not-found">
         <div className="message-box">
-            <h3>{translate('Not.found.heading', {
-                defaultValue: 'Page not found (404)',
-            })}</h3>
-            <div className="message">{translate('Not.found.message', { defaultValue: 'Jenkins could not find the page you were looking for. Check the URL for errors or press the back button.' })}</div>
-            <div className="actions"><Link to="/" className="btn">{translate('Open.dashboard', { defaultValue: 'Open Dashboard' })}</Link></div>
+            <h3>
+                {translate('Not.found.heading', {
```
kzantow: adding space like this _occasionally_ can be a problem

blueocean-core-js/src/js/components/ReplayButton.jsx
```diff
@@ -49,7 +48,8 @@ export class ReplayButton extends Component {
             replaying: true,
         });
 
-        runApi.replayRun(this.props.latestRun)
+        runApi
+            .replayRun(this.props.latestRun)
```
kzantow: ick?

blueocean-core-js/src/js/components/ResultPageHeader.jsx
```diff
@@ -29,14 +20,7 @@ CloseButton.propTypes = {
...
 export const ResultPageHeader = props => {
-    const {
-        status = 'unknown',
-        title,
-        onCloseClick,
-        startTime,
-        estimatedDurationInMillis,
-        t,
-    } = props;
+    const { status = 'unknown', title, onCloseClick, startTime, estimatedDurationInMillis, t } = props;
```
kzantow: 👎

blueocean-core-js/src/js/components/RunButton.jsx
```diff
@@ -86,14 +84,18 @@ export class RunButton extends Component {
...
-        }) : translate('toast.stop', {
-            defaultValue: 'Stop',
-        });
+        const runLabel =
+            this.props.runText ||
+            translate('toast.run', {
+                defaultValue: 'Run',
+            });
+        let stopLabel = this.state.stopping
+            ? translate('toast.stopping', {
```
kzantow: oh how i hate `translate()`

.eslintrc
sophistifunk: No, it's to stop lint handling things that are done automatically. Prettier is deterministic based on the AST.

blueocean-core-js/src/js/NotFound.jsx
sophistifunk: React should just add a `<!-- some react magic cookie -->` But if it is a problem we can address it.

blueocean-core-js/src/js/components/ReplayButton.jsx
sophistifunk: IMO better. Also, the point isn't pleasing people, it's consistency and mechanical transformation.

blueocean-core-js/src/js/components/RunButton.jsx
sophistifunk: Yes.

blueocean-core-js/src/js/UrlBuilder.js
```diff
@@ -77,9 +77,7 @@ export const buildRunDetailsUrl = (run) => {
         throw new Error('Could not extract URI components');
     }
 
-    return `/organizations/${organizationName}` +
-        `/${encodeURIComponent(fullName)}/detail` +
-        `/${detailName}/${runId}/pipeline`;
+    return `/organizations/${organizationName}` + `/${encodeURIComponent(fullName)}/detail` + `/${detailName}/${runId}/pipeline`;
```
michaelneale: this one looks odd, but then should all be one string anyway probably

scherler:
```
return `/organizations/${organizationName/${encodeURIComponent(fullName)}/detai/${detailName}/${runId}/pipeline`;
```

Is how it should be, but it had been bad to start with, so IMO the formatting is not to be blamed on the tool

blueocean-core-js/src/js/components/ReplayButton.jsx
scherler: love it

blueocean-core-js/src/js/components/ResultPageHeader.jsx
scherler: well, I guess as long it fits in one line, it is the preferid formatting but I guess that may be configurable.

blueocean-core-js/src/js/components/ReplayButton.jsx
cliffmeyers: It looks like the rule is "multiple chained methods each go on their own line" which I am fine with.

blueocean-core-js/src/js/components/RunButton.jsx
cliffmeyers: We really ought to be able to remove most of the `{ defaultValue: 'my label' }` stuff since the `i18n` is supposed to load first. Maybe not the case when there isn't an active extension point on the screen? We do need to get that fixed and then kill `defaultValue` with fire, it really hurts readability IMO.

blueocean-core-js/src/js/fetch.js
```diff
@@ -185,11 +188,13 @@ export const FetchFunctions = {
                     future = future.then(FetchFunctions.checkRefreshHeader);
                 }
 
-                future = future.then(FetchFunctions.checkStatus)
-                    .then(FetchFunctions.parseJSON, FetchFunctions.parseErrorJson);
+                future = future.then(FetchFunctions.checkStatus).then(FetchFunctions.parseJSON, FetchFunctions.parseErrorJson);
```
cliffmeyers: Interesting, I expected each `then` to be on its own line. Is this because it's not exceeding the max line length?

.eslintrc
cliffmeyers: Related to above, does Prettier automatically add a newline to end of each file? If so the `eol-last` rule can probably be removed.

blueocean-core-js/src/js/components/ReplayButton.jsx
kzantow: yeah, i can live with this just fine

blueocean-core-js/src/js/components/ResultPageHeader.jsx
kzantow: this is an open bug / discussion on this: https://github.com/prettier/prettier/issues/2550

blueocean-core-js/src/js/UrlBuilder.js
kzantow: agree with @scherler on this, if we use 160 lines, it should be a really long string

blueocean-core-js/src/js/fetch.js
kzantow: maybe because it doesn't have a nested function like here https://github.com/jenkinsci/blueocean-plugin/pull/1474/files#diff-4ef4ba5e27a29ca99e7cf854e77f5805R19

blueocean-core-js/src/js/i18n/i18n.js
```diff
@@ -190,7 +191,7 @@ export default function i18nTranslator(pluginName, namespace, onLoad) {
             }
 
             const fixedT = I18n.getFixedT(detectedLang, namespace);
-            translator = function (i18nKey, i18nParams) {
+            translator = function(i18nKey, i18nParams) {
```
kzantow: is this an option? goes against the jenkins project preferred formatting

blueocean-core-js/src/js/jwt.js
```diff
@@ -65,11 +66,12 @@ export default {
             publicKeyStore = fetch(url, { credentials: 'same-origin' })
                 .then(FetchFunctions.checkStatus)
                 .then(FetchFunctions.parseJSON)
-                .then(cert => this.verifyToken(token, cert)
-                    .then(payload => ({
+                .then(cert =>
+                    this.verifyToken(token, cert).then(payload => ({
```
kzantow: i like this better but i'm confused about the consistency here, why not
```
this
    .verifyToken(...)
    .then...
```
?

blueocean-core-js/src/js/utils/UrlUtils.js
```diff
@@ -145,9 +143,7 @@ export const calculateLogUrl = (config) => {
  */
 export function calculateNodeBaseUrl(config) {
     const { name, runId, branch, _appURLBase, isMultiBranch, organization } = config;
-    const baseUrl =
-        `${_appURLBase}/rest/organizations/${encodeURIComponent(organization)}/` +
-        `pipelines/${name}`;
+    const baseUrl = `${_appURLBase}/rest/organizations/${encodeURIComponent(organization)}/` + `pipelines/${name}`;
```
kzantow: would be nice if it was smart enough to fix these...

AlanFoster:
I'm a big fan of Prettier :+1:

FWiW - If you upgrade to the latest version of prettier and re-run the tool, I believe some of the edge-cases that were pointed out originally have now been tweaked to align with the latest community standards 🎉

PR #1478, opened 2017-10-11
JENKINS-46474 Make the docker inputs larger in the editor

NicuPascu:
Makes the docker inputs in the editor expand on whole row and make args input auto-expand vertically

# Description

See [JENKINS-46474](https://issues.jenkins-ci.org/browse/JENKINS-46474).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


NicuPascu:
Old:
recording_2017-10-11_191900

New:
recording_2017-10-11_192153

i386:
Great work 👍

bitwiseman:
👍 👍 👍

blueocean-pipeline-editor/src/main/js/components/editor/InputTextArea.jsx
```diff
@@ -0,0 +1,42 @@
...
+    event.target.style.height = 'auto';
+
+    if (event.target.scrollHeight > event.target.offsetHeight) {
+        event.target.style.height = (event.target.scrollHeight + 4) + 'px';
+    } else {
+        event.target.style.height = 'auto';
+    }
+}
+
+export default function InputText({ defaultValue, onChange, isRequired, hasError, onBlur }) {
```
kzantow: `InputTextArea` ?

blueocean-pipeline-editor/src/main/js/components/editor/InputTextArea.jsx
```diff
@@ -0,0 +1,42 @@
+// @flow
```
kzantow: not necessary if you're not using flow annotations

blueocean-pipeline-editor/src/main/js/components/editor/AgentConfiguration.jsx
```diff
@@ -142,15 +143,25 @@ export class AgentConfiguration extends Component<DefaultProps, Props, State> {
             {selectedAgent && selectedAgentMetadata && <div className="agent-parameters">
                 {selectedAgentMetadata.parameters.filter(agentConfigParamFilter(selectedAgent)).map(param => {
                     const val = this.getRealOrEmptyArg(param.name).value.value;
+                    const inputElem = (param.name === 'args' ?
```
kzantow: Might be good to move this to a function in order to support different changes that get requested, e.g. `_getAgentInputControl(param)`

blueocean-pipeline-editor/src/main/js/components/editor/AgentConfiguration.jsx
```diff
@@ -142,15 +143,25 @@ export class AgentConfiguration extends Component<DefaultProps, Props, State> {
...
                     return (<div className="agent-param">
                         <label key={selectedAgent.type + '/' + param.name}>
                             <div>{param.capitalizedName}{param.isRequired ? '*' : ''}</div>
                             <div>
-                                <InputText hasError={param.isRequired && !pristine && !val}
-                                    isRequired={param.isRequired}
-                                    defaultValue={val}
-                                    onChange={val => { this.setAgentValue(param.name, val); param.isRequired && this.setState({ pristine: false }); }}
-                                    onBlur={e => param.isRequired && this.setState({ pristine: false })} />
+                                {inputElem}
```
kzantow: and here call `_getAgentInputControl(param)`, would make the intent of the code a little more clear, I think

blueocean-pipeline-editor/src/main/js/components/editor/AgentConfiguration.jsx
```diff
@@ -113,6 +114,22 @@ export class AgentConfiguration extends Component<DefaultProps, Props, State> {
...
+                isRequired={param.isRequired}
+                defaultValue={val}
+                onChange={val => { this.setAgentValue(param.name, val); param.isRequired && this.setState({ pristine: false }); }}
+                onBlur={e => param.isRequired && this.setState({ pristine: false })} />
+        } else {
+            return <InputText hasError={param.isRequired && !pristine && !val}
+                isRequired={param.isRequired}
+                defaultValue={val}
+                onChange={val => { this.setAgentValue(param.name, val); param.isRequired && this.setState({ pristine: false }); }}
+                onBlur={e => param.isRequired && this.setState({ pristine: false })} />
```
cliffmeyers: Is there a more elegant way to avoid all of the duplicated props here? Could we create an intermediate object and then explode those props onto the desired component using the spread operator?

```
const inputProps = {
   hasError: param.isRequired && !pristine && !val,
   // continue other props
}
```

and then

```
if (param.name === 'args') {
    return <InputTextArea {...inputProps} />
} else {
   return <InputText {...inputProps} />
}
```

It might not be quite _that_ simple, need to watch out for the more complex function bindings there... but to me this kind of code dup is a 🐛  unless there is literally no other way to make it work.

blueocean-pipeline-editor/src/main/js/components/editor/InputTextArea.jsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+
+function cn(...args) {
```
cliffmeyers: Is there a better method name?

blueocean-pipeline-editor/src/main/js/components/editor/InputTextArea.jsx
```diff
@@ -0,0 +1,41 @@
...
+    let out = null;
+    for (let i = 0; i < args.length; i++) {
```
cliffmeyers: Alternatively I think you could simplify this method down to `args.filter(arg => !!arg).join(' ')` ?

blueocean-pipeline-editor/src/main/js/components/editor/InputTextArea.jsx
```diff
@@ -0,0 +1,41 @@
...
+            } else {
+                out += ' ' + args[i];
+            }
+        }
+    }
+    return out;
+}
+
+function autoResizeHeight() {
+    event.target.style.height = 'auto';
```
cliffmeyers: Is `event` automatically in scope here by virtue of this being an event handler? or should it be a method param?

NicuPascu: yes, it is automatically in scope, no need to set it as a method param

PR #1484, opened 2017-10-12
[ATH IMPROVEMENT] Allow local Jenkins instance, simplified API

kzantow:
Adding ability to run the ATH locally against a locally running dev instance.

esp. @imeredith


acceptance-tests/src/main/java/io/blueocean/ath/sse/SSEClientRule.java
```diff
@@ -93,7 +94,7 @@ public void setLogEvents(boolean logEvents) {
     public void connect() throws UnirestException, InterruptedException {
         SecureRandom rnd = new SecureRandom();
         String clientId = "ath-" + rnd.nextLong();
-        HttpResponse<JsonNode> httpResponse = Unirest.get(baseUrl + "/sse-gateway/connect?clientId=" + clientId).asJson();
+        HttpResponse<JsonNode> httpResponse = Unirest.get(baseUrl + "/sse-gateway/connect?clientId=" + clientId).basicAuth(LoginPage.getUsername(), LoginPage.getPassword()).asJson();
```
kzantow: might need to add some logic when to do this

acceptance-tests/src/main/java/io/blueocean/ath/pages/classic/LoginPage.java
```diff
@@ -35,36 +35,24 @@ public LoginPage(WebDriver driver) {
...
-
-        wait.until(By.xpath("//*/button[contains(text(), 'log')]")).click();
-        wait.until(driver -> {
-            if(driver.getCurrentUrl().contains("loginError")) {
-                throw new RuntimeException("Error logging in");
-            }
-            return ExpectedConditions.urlToBe(base + "/").apply(driver);
-        });
-        logger.info("Logged in as alice");
+        find("input[name=j_username]").sendKeys(getUsername());
```
kzantow: This was a 30 minute example of how my prior testing frameworks looked... well, similar.

michaelneale:
nice

acceptance-tests/src/main/java/io/blueocean/ath/WebElementUtils.java
```diff
@@ -27,7 +27,7 @@ public static void clearText(WebElement element) {
      */
     public static void setText(WebElement element, String text) {
         checkTextElement(element);
-        element.sendKeys(Keys.chord(Keys.CONTROL, "a"));
+        element.clear();//sendKeys(Keys.chord(Keys.CONTROL, "a"));
```
kzantow: is there a reason it was doing `CTRL-a` instead of `clear` before? this definitely doesn't work on Mac ;)

michaelneale: ctrl-a goes back to the start of a text cell (emacs keys - may they never die) which I gather is not the intent here?

acceptance-tests/src/main/java/io/blueocean/ath/sse/SSEClientRule.java
michaelneale: if running unauthed I guess it may ignore those basic headers anyway (maybe...)

acceptance-tests/src/main/java/io/blueocean/ath/WebElementUtils.java
cliffmeyers: Sorry, somehow I completely missed the presence of `clear()` in the docs. That is much easier. :)

acceptance-tests/src/main/java/io/blueocean/ath/pages/classic/LoginPage.java
```diff
@@ -35,36 +35,24 @@ public LoginPage(WebDriver driver) {
...
+        find("input[name=j_password]").sendKeys(getPassword());
+        find("//button[contains(text(), 'log')]").click();
```
cliffmeyers: I like this shorthand overall. Per discussion in Gitter, I think we want a static `click` method that implements the same retry logic as here:  https://github.com/jenkinsci/blueocean-plugin/pull/1412/files#diff-9d7946f8ccd8a4191146940351b51999

cliffmeyers:
I like where this PR is headed although I would like to land #1412 first. We'd also need to check to make sure that the changes to make ATH run better in Chrome do not fail in Firefox - perhaps we can do that in this PR?

Down the road we probably want to take a look at Selenium Grid for running multiple browsers, maybe as part of the weekly CI builds we do now? https://github.com/SeleniumHQ/selenium/wiki/Grid2

acceptance-tests/src/main/java/io/blueocean/ath/pages/classic/LoginPage.java
kzantow: Yeah, absolutely. This is why I said it was a 30 minute implementation. The real logic typically ends up being more complex, and yeah, definitely agree what we do is add utility methods like `click` with retry and `setText` which should be doing what `sendKeys` is doing now, e.g. you just want to set the text of an input, but you may want to send individual keys for other purposes e.g. testing an autocomplete

kzantow:
@cliffmeyers yes, this is just a bit of an example (and something working that makes it easier to run tests locally - e.g. no need for docker, etc.). Certainly land #1412 first and then look to revisit this and combine the changes in a sensible way moving forward.

acceptance-tests/src/main/java/io/blueocean/ath/pages/classic/LoginPage.java
imeredith: Only thing I would put here is that I would make find use "By" class. Ie find(By.id(..)), good otherwise. Could default to use css for strings though

imeredith:
I like what is going on here.

acceptance-tests/src/main/java/io/blueocean/ath/sse/SSEClientRule.java
imeredith: We can set the logged in user in the state fairly easily (when using @Login)

acceptance-tests/src/main/java/io/blueocean/ath/pages/classic/LoginPage.java
kzantow: I suppose it doesn't hurt to have an override, but I would definitely not use `By` as the only way to specify things since you can specify a find by id using css like `#my-fancy-id`, which would work with this already. The only thing missing is find by link text, which you can use xpath for, but is a little more cumbersome. We really shouldn't be finding anything by text in these tests, anyway, tho

acceptance-tests/src/main/java/io/blueocean/ath/BasePage.java
```diff
@@ -0,0 +1,93 @@
...
+import org.openqa.selenium.support.ui.FluentWait;
+
+import javax.inject.Inject;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Provides utility methods to downstream test classes
+ */
+public class BasePage {
```
kzantow: This should be named something else...

acceptance-tests/src/main/java/io/blueocean/ath/AthModule.java
```diff
@@ -4,20 +4,15 @@
...
 import com.google.inject.name.Names;
 import com.offbytwo.jenkins.JenkinsServer;
-import io.blueocean.ath.factory.ActivityPageFactory;
-import io.blueocean.ath.factory.BranchPageFactory;
-import io.blueocean.ath.factory.FreestyleJobFactory;
-import io.blueocean.ath.factory.MultiBranchPipelineFactory;
-import io.blueocean.ath.factory.ClassicPipelineFactory;
-import io.blueocean.ath.factory.RunDetailsArtifactsPageFactory;
-import io.blueocean.ath.factory.RunDetailsPipelinePageFactory;
+import io.blueocean.ath.factory.*;
```
cliffmeyers: I seem to recall that using wildcard imports violates some style rules in Jenkins projects, I think @vivek pointed that out to me once. If you're using IDEA and search for "class count to use import with" in settings, you'll find an option. I set it to 100.

acceptance-tests/src/main/java/io/blueocean/ath/BasePage.java
cliffmeyers: Perhaps `BasePageObject` or `AbstractPageObject` and made abstract if not meant to be instantiated directly.

acceptance-tests/src/main/java/io/blueocean/ath/BasePage.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    public void goTo(String url) {
+        getDriver().get(url);
+    }
+
+    /**
+     * Navigates to a relative url to the base url
+     * @param url where to go
+     */
+    public void go(String url) {
```
cliffmeyers: I like the idea of two methods to navigate to relative and absolute paths, although it's not clear which one is which from the names `go` and `goTo`

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -0,0 +1,284 @@
...
+        return getElement().isDisplayed();
+    }
+
+    @Override
+    public void click() {
+        forEach(e -> e.click());
+    }
+
+    @Override
+    public void submit() {
```
cliffmeyers: We don't really have forms that truly submit in the app, should we just remove this?

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -0,0 +1,284 @@
...
+        }
+    }
+
+    public boolean isVisible() {
+        return getElement().isDisplayed();
+    }
+
+    @Override
+    public void click() {
+        forEach(e -> e.click());
```
cliffmeyers: Do want to support clicking multiple elements? It seems that it could have unintended consequences if the selector is too loose. Perhaps warn if multiple elements are matched?

We should bake in the retry logic from `WaitUtil.click` here I believe.

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    @Override
+    public void submit() {
+        forEach(e -> e.submit());
+    }
+
+    @Override
+    public void sendKeys(CharSequence... charSequences) {
+        forEach(e -> e.sendKeys(charSequences));
```
cliffmeyers: Same question as in `click`

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -0,0 +1,284 @@
...
+    public boolean isSelected() {
+        WebElement el = getElement();
+        if (el == null) {
+            return false;
+        }
+        return el.isSelected();
+    }
+
+    @Override
+    public boolean isEnabled() {
```
cliffmeyers: Would this style of method implicitly wait for the element to be enabled, or do an immediate check for whether it's enabled? If the latter, it seems like it could be subject to timing failures. Is there logic we can borrow from `ExpectedConditions` here? I honestly don't know the answer but the stuff in `WaitUtil` generally seems to wait the right way, although maybe the API can be improved here?

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -0,0 +1,284 @@
...
+        WebElement el = getElement();
+        if (el == null) {
+            return false;
+        }
+        return el.isEnabled();
+    }
+
+    @Override
+    public String getText() {
```
cliffmeyers: Similar question as above, say for example the `textToBePresentInElement` method on `ExpectedConditions` seems useful here. Again I'm a little foggy on these API's as to whether that's appropriate in the `WebElement` impl or if we need a util method on `BasePage` instead? Sorry if I'm muddying the waters.

acceptance-tests/src/main/java/io/blueocean/ath/BasePage.java
kzantow: agreed, maybe could just inspect the url if it starts with `http(s)` then don't use the base

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
kzantow: it's necessary to implement the interface

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
kzantow: yeah, probably not, might be good to dump a warning for those; i'll make those changes - or throw an error?

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
kzantow: just checks for it being enabled. sometimes you might want to check for things enabled/disabled based on user input. could add more methods as needed e.g. `waitUntilEnabled`

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
kzantow: i really hope we can't find an element without text while in the background somehow it gets filled in. that kind of behavior would be a bit ridiculous, even for react. these types of things we can at least identify and bake in to this API as needed instead of cluttering up the tests with all kinds of wait conditions

michaelneale:
I think this is ok if no one else has objections, not sure what the details are though - perhaps README could be updated?

acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/GithubCreationPage.java
```diff
@@ -105,7 +103,7 @@ public void selectPipelineToCreate(String pipeline){
     }
 
     public void clickCreatePipelineButton() {
-        wait.until(ExpectedConditions.elementToBeClickable(createPipelineButton)).click();
+        click(".button-create");
```
michaelneale: looks nice

michaelneale:
FWIW: https://github.com/jenkinsci/blueocean-plugin/pull/1503 seemed to pass (fixing the click/co-ordinates problem) but will leave that one in favour of this if this one is better and works.

michaelneale:
I was trying this locally - `testEditor` fails for me (times out) - is @cliffmeyers or @kzantow able to run it locally? (master + https://github.com/jenkinsci/blueocean-plugin/pull/1503/files does work for me though). Not sure if intermittent or something else, but I am wary of changes if I can't run them locally

michaelneale:
yep :bee: from me.

sophistifunk:
Is this the fix for the missing button?

kzantow:
@sophistifunk this is a number of improvements to ATH and that fix, yes

sophistifunk:
Sweet, when it's merged I'll pull it into my branch in the quest for fields verdant.

michaelneale:
@sophistifunk yes this will fix it

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -59,8 +81,15 @@ protected WebDriver getDriver() {
      * Gets the first matching element
      * @return see description
      */
-    public WebElement getElement() {
-        return getElements().iterator().next();
+    public WebElement getElement() throws NoSuchElementException {
+        List<WebElement> elements = getElements();
+        if (elements == null || elements.isEmpty()) {
+            throw new NoSuchElementException("Nothing matched to click on for: " + expr);
```
cliffmeyers: The reference to "click" here might be copy/paste error?

kzantow: ah yes, will fix; oops

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -73,23 +102,66 @@ public void forEach(java.util.function.Consumer<WebElement> fn) {
...
+                    logger.info(String.format("retry click successful for %s", by.toString()));
+                }
+                return;
+            } catch (WebDriverException ex) {
+                if (ex.getMessage().contains("is not clickable at point")) {
+                    logger.warn(String.format("%s not clickable: will retry click", by.toString()));
+                    logger.debug("exception: " + ex.getMessage());
+                    try {
+                        // typically this is during an animation, which should not take long
+                        Thread.sleep(500);
```
cliffmeyers: I had assumed I'd need to add a sleep to this method, although it worked very reliably without it (at least in my local env). I suppose there's no harm in being more conservative by having it pause longer before retry.

kzantow: Well, what I'd like to do (and maybe what I did before, I don't quite remember) is get rid of the WebDriver `wait` and replace it with some smart polling / retries of our own. It should speed things up in some cases. We'll want to add some other utilities to try to have some explicit waits for a css selector, for example waiting for github to scan repos and such, we know that will take a while.

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeTest.java
```diff
@@ -2031,6 +2032,21 @@ public void testBlockedStep() throws Exception {
                 assertEquals(1, stepsResp.size());
                 assertEquals("QUEUED", stepsResp.get(0).get("state"));
             }
+        } else {
+            // Avoid spurious code coverage failures
```
cliffmeyers: Nice, had these fail from time to time myself... 👍

kzantow:
... merging once green :)

kzantow:
Last 3 runs all green, I think this fixes the latest flakiness on master.

PR #1492, opened 2017-10-16
JENKINS-45393: Bitbucket server ATH

imeredith:
# Description

@michaelneale Can you please try running the ath locally? bitbucket server docker container has stoppped starging up for me and I would like someone else to try.

See [JENKINS-45393](https://issues.jenkins-ci.org/browse/JENKINS-45393).


acceptance-tests/run.sh
```diff
@@ -57,4 +59,6 @@ if [ "${RUN_SELENIUM}" == "true" ]; then
     ./runner/scripts/stop-selenium.sh
 fi
 
+./runner/scripts/stop-bitbucket-server.sh
```
kzantow: Is there anything that would make this script fail before it gets here?

michaelneale: or at least need to clean it up before starting (which we do for selenium)

michaelneale:
if this container doesn't work can use https://developer.atlassian.com/platform/marketplace/timebomb-licenses-for-testing-server-apps/ with an official one?

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,66 @@
...
+        String repo = "testJenkinsfileCreate";
+        client.api().repositoryApi().delete(project, repo);
+        client.api().repositoryApi().create(project, CreateRepository.create(repo, true));
+
+        dashboardPage.open();
+        wait.until(ExpectedConditions.visibilityOf(dashboardPage.newPipelineButton))
+            .click();;
+        LOGGER.info("Clicked on new pipeline button");
+
+        wait.until(By.xpath("//span[text()='Bitbucket Server']")).click();
```
kzantow: Any way of not using text for this? I know it _should_ be fine, using the english translation, but still

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,66 @@
...
+        wait.until(By.cssSelector(".button-add-server")).click();
+        LOGGER.info("Clicked addserver");
+
+        wait.until(By.xpath("//input[@placeholder='My Bitbucket Server']")).sendKeys("bitbucketserver");
+        wait.until(By.xpath("//input[@placeholder='https://mybitbucket.example.com']")).sendKeys("http://127.0.0.1:7990/bitbucket");
+        wait.until(By.cssSelector(".button-create-server")).click();
+        wait.until(By.cssSelector(".button-next-step")).click();
+        LOGGER.info("Entered server details");
+
+        wait.until(By.xpath("(//input[ @class='TextInput-control'])[1]")).sendKeys("admin");
```
kzantow: Similarly here, any better xpath? Or just css? `.credentials-new input[type=text]` & `.credentials-new input[type=password]`?

kzantow: Nice to have `find(".credentials-new input[type=text]").setText("admin")`

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,66 @@
...
+    @Inject
+    WaitUtil wait;
+
+    @Inject
+    DashboardPage dashboardPage;
+
+    @Test
+    public void testJenkinsfileCreate() {
+        BitbucketClient client = BitbucketClient.builder()
+            .endPoint("http://127.0.0.1:7990/bitbucket")
```
kzantow: Configurable via system properties or some other method? We should probably find a way to put a bunch of properties that are easily injected into the app, some config file then have a `SystemProperties.lookup` sort of thing that checks a few places and allows overrides down to the actual `System.getProperty` level

kzantow: maybe i'll just add that to my PR. it would be useful for me to just set it once and forget it

michaelneale: @imeredith this wont' work on CI IIRC - as the code running the ATH runs in one docker container, the bitbucket one in a peer container. On your desktop, the BB container is running locally. So I think the Jenkinsfile needs to change to link in the network/port from BB. 

Locally this test worked for me, but not on CI.

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,66 @@
...
+        LOGGER.info("Selected bitbucket server");
+
+        wait.until(By.xpath("//span[text()='Bitbucket Server']")).click();
+        LOGGER.info("Selected bitbucket server");
+
+        wait.until(By.cssSelector(".button-add-server")).click();
+        LOGGER.info("Clicked addserver");
+
+        wait.until(By.xpath("//input[@placeholder='My Bitbucket Server']")).sendKeys("bitbucketserver");
+        wait.until(By.xpath("//input[@placeholder='https://mybitbucket.example.com']")).sendKeys("http://127.0.0.1:7990/bitbucket");
```
michaelneale: @imeredith this will need to change as well (or the network stuff resolved)

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
michaelneale: not actually sure how to make this work on CI myself - the selenium one works as it uses the --net param to docker, but I dont' think we can do that for this. The IP would be different, but not sure what it would be.

cliffmeyers:
Just a note that I merged #1412 to master earlier today, so this should get a remerge of master soon since there were a handful of subtle changes made to existing ATH for it to pass in Chrome, and there may be similar issues for these new tests.

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
cliffmeyers: I agree we should stop selecting on text values. Please :) They are brittle, as we were originally using "Github" in our UI and then changed it once someone pointed out that "GitHub" is the proper branding. I don't see any reason why we shouldn't be making using liberal use of semantic CSS class names in our UI code (e.g. "bitbucket-server-scm-button") and selecting on that. If they are missing me, we should feel free to add them. Also, clicking on a `span` can result in "element not clickable" errors, we should really click on `button` or anchor elements IMO.

michaelneale:
I think the Jenkinsfile will need to start the BB container at the start - and then link the network somehow (like it does with selenium)

imeredith:
@michaelneale can you please give this a whirl?

michaelneale:
@imeredith I wasn't able to run this - it was very slow and then timed out. (running just the bitbucket test) locally

Jenkinsfile
```diff
@@ -18,6 +18,7 @@ node() {
       sh 'mv $FILE acceptance-tests/bo-ath.key'
     }
     sh "./acceptance-tests/runner/scripts/start-selenium.sh"
+    sh "./acceptance-tests/runner/scripts/start-bitbucket-server.sh"
```
michaelneale: :bug: needs to have a stop bitbucket like with selenium...

michaelneale: :bug: needs to have a stop bitbucket like with selenium...

michaelneale:
@imeredith I can't seem to get this running locally. 

I can see it start the BB container, but it isn't responding on port 0.0.0.0:7990 even looking via docker ps. 

The test fails as 127.0.0.1:7990 isn't responding... I am on High Sierra - not sure if that is a thing that matters? 

I can run other tests (eg github one) fine, and the selenium container routes to the outside world (I can see a http page from it).

imeredith:
@michaelneale ok i guess i need to upgrade to high sierra and see if that changes tihngs. Didnt get around to it after buying the usb sticks the other day

michaelneale:
@imeredith I tried on my sierra machine - and it passed there... so there is something about high sierra and docker that is preventing that bitbucket container from routing traffic... (or perhaps it was the docker for mac version.. most odd). I am using "latest" docker for mac, might be better if I switch to stable? 

anyway, looks close... very close (also there is a :bug: that the `Jenkinsfile` needs to shut down the BB container at the end).

michaelneale:
@imeredith update: not related to high sierra, but latest edge docker seems to break it... (ie if you just udpate your docker for macos you may see this). Reproduced on sierra, so no need to upgrade to see it.

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,69 @@
...
+            .credentials("admin:admin").build();
+
+        String project = "BLUE";
+        String repo = "testJenkinsfileCreate";
+        client.api().repositoryApi().delete(project, repo);
+        client.api().repositoryApi().create(project, CreateRepository.create(repo, true));
+
+        dashboardPage.clickNewPipelineBtn();
+
+        wait.until(By.xpath("//span[text()='Bitbucket Server']")).click();
```
cliffmeyers: As discussed in a prior PR, can we select on class names instead? They should have `className` specified in the `BbServerScmProvider` and `BbCloudServerProvider` components... right now "undefined" is rendered as a class name.

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,69 @@
...
+        LOGGER.info("Selected bitbucket server");
+
+        wait.until(By.cssSelector(".button-add-server")).click();
+        LOGGER.info("Clicked addserver");
+
+        wait.until(By.xpath("//input[@placeholder='My Bitbucket Server']")).sendKeys("bitbucketserver");
```
cliffmeyers: `By.css(".text-name input")` should work

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,69 @@
...
+        String repo = "testJenkinsfileCreate";
+        client.api().repositoryApi().delete(project, repo);
+        client.api().repositoryApi().create(project, CreateRepository.create(repo, true));
+
+        dashboardPage.clickNewPipelineBtn();
+
+        wait.until(By.xpath("//span[text()='Bitbucket Server']")).click();
+        LOGGER.info("Selected bitbucket server");
+
+        wait.until(By.cssSelector(".button-add-server")).click();
```
cliffmeyers: Probably just want to use `click` here for consistency.

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,69 @@
...
+        LOGGER.info("Clicked addserver");
+
+        wait.until(By.xpath("//input[@placeholder='My Bitbucket Server']")).sendKeys("bitbucketserver");
+        wait.until(By.xpath("//input[@placeholder='https://mybitbucket.example.com']")).sendKeys("http://127.0.0.1:7990");
```
cliffmeyers: `By.css(".text-url input")` should work

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,69 @@
...
+        click(".button-create-server");
+        Thread.sleep(      3000);
```
cliffmeyers: I bet if you moved this above the `click(".button-create-server")` statement on prior line that you'd be able to see the error message about duplicate server.

We need avoid explicit sleeps though. CI seems to slow down a lot in 8-10 PM UTC timeframe and I bet this sleep still wouldn't be long enough to work reliably.

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,69 @@
...
+        if (!getDriver().findElements(By.cssSelector(".FormElement.u-error-state")).isEmpty()) {
```
cliffmeyers: I really think we should get away from conditional logic in the "click streams" of our ATH. To me it's a code smell that we haven't ensured the test has consistent state before it's run. In this case we're trying to defend against the scenario where the server's already been created but I think it's the wrong way to do it.

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,69 @@
...
+            click("//button[text()='Cancel']");
```
cliffmeyers: select on className

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,69 @@
...
+            click(".Dropdown-button.Dropdown-placeholder");
+            click("//a[text()='bitbucketserver']");
```
cliffmeyers: Selecting on text here is ok in my opinion since we are looking for a specific named element. However you probably want to scope it with a parent class since the rule is a bit loose.

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,69 @@
...
+            LOGGER.info("Server already exists");
+        }
+
+        click(".button-next-step");
+
+        wait.until(By.xpath("(//input[ @class='TextInput-control'])[1]")).sendKeys("admin");
```
cliffmeyers: `By.css(".text-username input")` ?

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,69 @@
...
+        wait.until(By.xpath("(//input[ @class='TextInput-control'])[2]")).sendKeys("admin");
```
cliffmeyers: `By.css(".text-password" input)` ?

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,69 @@
...
+        wait.until(By.cssSelector(".button-create-credental")).click();
```
cliffmeyers: Probably want to use `click` here, also see my comment here: https://github.com/jenkinsci/blueocean-plugin/pull/1529#pullrequestreview-74205628

michaelneale:
@cliffmeyers are you able to run the test locally? I seem to have some fundamental docker problem

cliffmeyers:
@michaelneale yes, it does run. I'm on Docker 17.09.0-ce-mac35 (19611) on El Capitan... have avoided upgrading my OS because I don't trust Apple OS upgrades lol

michaelneale:
@vivek @i386 could you build and run the new BitbucketServerTest in the ATH on your machine?

vivek:
@michaelneale I will try it out

michaelneale:
@imeredith  ok - this works now. Sorry I had not updated the branch to get the latest docker image of ivan's (I blame jetlag). 

So - this works locally for me, however I would like to see @cliffmeyers's feedback addressed, and also it needs to cleanup the container at the end of the `Jenkinsfile` before we can merge this

vivek:
@michaelneale @imeredith ./run.sh --dev fails for me.

```
Starting Selenium Docker container...

docker: Error response from daemon: driver failed programming external connectivity on endpoint blueo-selenium (83c44a7200e13d7fb3c6ed846f3fc81926c495a592cead60f165f06b1a7d12c0): Error starting userland proxy: Bind for 0.0.0.0:7999 failed: port is already allocated.

**************************************************************
**** Docker container with Selenium, browser and VNC running.
****   Selenium server listening on :4444
****   
****   To connect and view with VNC, run:
****   $ open vnc://:secret@localhost:15900
**************************************************************

 Stopping old Bitbucket Docker container, if any. This may take a few seconds...
      ... stopped


 Starting Bitbucket Docker container...

076f8ca67f1be32a05e6a2afe475b8e370fcce644b71291740fa3ab5eef26e87
docker: Error response from daemon: cannot join network of a non running container: c6bd21b5cff9d3d5a770d169126de45ebe506a67aaca771c97b10f0ebb2859f8.
Waiting for bitbucket server to start
Waiting for bitbucket server to start
Waiting for bitbucket server to start
Waiting for bitbucket server to start
Waiting for bitbucket server to start
```

michaelneale:
@vivek very differenet failure to mine - update your docker version? (check there aren't things already running?)

michaelneale:
ping @imeredith been a few days - could fix this up and get it in ?

imeredith:
@michaelneale working on it now

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
imeredith: created JENKINS-47959 to deal with this

imeredith:
@cliffmeyers I fixed up what you have mentioned. One issue is that for whatever reason i cant do what github enterprise does for muliple servers as the url has to be unique, whereas github enterprise it doesnt have to be unique

michaelneale:
getting there, but please do make it shut down the BB container at end of Jenkinsfile, lest it leak

Jenkinsfile
```diff
@@ -69,6 +69,7 @@ node() {
       } finally {
         stage('Cleanup') {
           sh "${env.WORKSPACE}/acceptance-tests/runner/scripts/stop-selenium.sh"
+          sh "${env.WORKSPACE}/acceptance-tests/runner/scripts/stop-bitbucket-serve.sh"
```
michaelneale: typo :bug:

michaelneale:
update this to master and check it is ok @imeredith ? @vivek when you get a chance, see if it runs locally for you (maybe hit up ivan if it is still failing for you)

cliffmeyers:
@imeredith since we can't use the "random query string arg looks like unique server" trick, should we add a REST API for deleting an existing Bitbucket Server and improve the ATH test case so that it ensures the server is deleted before the test runs? I did something similar for the favorites API in #1540

michaelneale:
@imeredith can you update this and try again?

michaelneale:
ok, want to see if this runs for @vivek too

michaelneale:
well, if @vivek can run this locally - I think we are ok to merge this in - would like to know

cliffmeyers:
@michaelneale IMO, if the new test can't be run multiple times in succession and only works when run in a totally clean environment then we should not merge this PR. It will create an unreasonable burden on anyone who needs to maintain the test at a later date. If the PR absolutely needs to go in, then we need to cut a JIRA ticket to improve the test so it's re-runnable.

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,64 @@
...
+
+        wait.until(By.cssSelector(".text-name input")).sendKeys("bitbucketserver");
+        wait.until(By.cssSelector(".text-url input")).sendKeys("http://127.0.0.1:7990");
+        click(".button-create-server");
+
+        click(".button-next-step");
+
+        wait.until(By.cssSelector(".text-username input")).sendKeys("admin");
+        wait.until(By.cssSelector(".text-password input")).sendKeys("admin");
+        click(".button-create-credental");
```
cliffmeyers: 🐛  this test is incomplete, stopping after the "create credential" step. It needs to select the project, then repository, then click create and do some asserts on activity or editor. Just calling it out so it's not missed.

acceptance-tests/src/test/java/io/blueocean/ath/offline/BitbucketServerTest.java
```diff
@@ -0,0 +1,64 @@
...
+        LOGGER.info("Selected bitbucket server");
+
+        click(".button-add-server");
+        LOGGER.info("Clicked addserver");
+
+        wait.until(By.cssSelector(".text-name input")).sendKeys("bitbucketserver");
+        wait.until(By.cssSelector(".text-url input")).sendKeys("http://127.0.0.1:7990");
+        click(".button-create-server");
+
+        click(".button-next-step");
```
cliffmeyers: 🐛  running the test a second time hangs on this step because the button is still disabled, since the `click(".button-create-server")` can't complete due to duplicate server name. It would probably be good to add a wait between these two clicks that ensures the "add server dialog" is no longer present.

vivek:
I am going to try it sometime today. @cliffmeyers has some comments as well.

cliffmeyers:
FYI I took this PR off @imeredith 's hands and plan to implement the items I raised above.

PR #1500, opened 2017-10-20
JENKINS-47446 add missing default value for rerun title

NicuPascu:
# Description

See [JENKINS-47446](https://issues.jenkins-ci.org/browse/JENKINS-47446).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-core-js/src/js/components/ReplayButton.jsx
+++ b/blueocean-core-js/src/js/components/ReplayButton.jsx
@@ -73,12 +73,11 @@ export class ReplayButton extends Component {
         const isReplayable = this.props.latestRun.replayable;
         const hasPermission = permit(this.props.runnable).start();
 
-        const replayLabel = translate('run.rerun');
+        const replayLabel = translate('run.rerun', { defaultValue: 'Rerun' });
 
         if (!isFinished || !isReplayable || !hasPermission) {
             return null;
         }
-
 
         return (
             <div className={`replay-button-component ${outerClass}`} onClick={(event => stopProp(event))}>


blueocean-core-js/src/js/components/ReplayButton.jsx
```diff
@@ -73,13 +73,12 @@ export class ReplayButton extends Component {
         const isReplayable = this.props.latestRun.replayable;
         const hasPermission = permit(this.props.runnable).start();
 
-        const replayLabel = translate('run.rerun');
+        const replayLabel = translate('run.rerun', { defaultValue: 'Rerun' });
```
sophistifunk: Seems like this should have a hyphen in it? Paging @i386

sophistifunk:
Looks good, just a punctuation question 🐝

i386:
Why is setting a default value necessary at all? Aren't the defaults just the english translation?

sophistifunk:
Always set defaults IMO. Nothing is worse than missing labels at runtime, and it makes testing easier.

NicuPascu:
I got the default value from the messages.properties file, should I change it in both places for consistency? https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-web/src/main/resources/jenkins/plugins/blueocean/web/Messages.properties#L13

i386:
@sophistifunk not quite what I was asking about. So here we have `run.rerun` getting interpolated at runtime depending on the language:

- For english: `run.rerun` -> `Rerun`
- For french: `run.rerun` -> `Rediffusion`

We run the ATH in english, so the interpolated value would be `Rerun`.

If I had a new language, say Basque, and I didn't have a translation for `run.rerun`, would I see a fallback to english `Rerun` or just see the value `run.rerun`?

sophistifunk:
Not sure, the runtime behaviour in prod is hidden somewhere. I would hope it fell back to english-from-resources before falling back to 'hardcoded default' before falling back to 'dumping out the resource name'

NicuPascu:
From my testing I see it does not fall back to english if the default is not set, this was the case here

NicuPascu:
@i386 should I change the text from Rerun to Re-run?

i386:
@NicuPascu perhaps change both to "Rerun" which is more correct?

NicuPascu:
Rerun is set currently so no change necessary, ok to merge @i386?

i386:
I still don't have a good explanation on why the user was seeing `run.rerun`?

NicuPascu:
@i386 The bug started happening for me yesterday afternoon (maybe I wasn't up to date with master). I tried to see if translation is loaded and indeed the function was loaded correctly. Then I noticed there was no default value and I know in all other instances where I saw the translation being used that a default value was also added, I added it and the bug went away. I am not sure why it started happening out of the blue.

i386:
@NicuPascu I suspect it not having a default value is a bit of a red herring. Theres a translation for your browsers locale that has `run.rerun` defined, but its not loading - which is why `run.rerun` is being shown. Could that suggest the translation isn't loading in some cases? e.g. a bigger problem

NicuPascu:
@i386 I will investigate more and see if I can understand the problem better

i386:
Thanks @NicuPascu. FWIW, @scherler spent a fair bit of time on translations so he might be worth having a chat to :)

blueocean-core-js/src/js/components/ReplayButton.jsx
cliffmeyers: Has anyone tested this change with a different locale? It looks if the user has it set to Spanish or German that they're going to get "Rerun" as the text. If so I am not sure the bug is really fixed via this code change. This is probably a i18n loading / timing issue but it would help to understand why... I had thought that stuff was fixed a long ways back. Maybe it will be made more resilient with the bundling changes coming.

cliffmeyers: Sorry, I realize my comment is duplicative to some of the conversation below. I share the same concerns: `defaultValue` really shouldn't be used because it is probably just covering up a timing issue with load of i18n bundles.

NicuPascu:
I am closing this as the issue has been fixed by Keith in another PR

PR #1501, opened 2017-10-20
Add (inactive) support for auto-formatting on commit

sophistifunk:
# Description

Add some scripting that when activated will automatically format **staged** JS files on commit.

This should have very little effect on anybody unless you specifically symlink the `/bin/pre-commit.js` file from `.git/hooks/pre-commit` -- and once this goes into master, I'll be looking for one or two ~suckers~ volunteers to do so, to help me stress-test it for a week before we recommend it to everybody.

I have also removed the submodule-specific `.eslintrc` files, as they only confuse things.

Current plan:

1. Get this looked at by you mob :)
2. Merge this to master, and I (and hopefully a volunteer) will do some unrelated development during the next week or so to give it a shake-down, and make sure there's no unintended effects.
3. Get the rest of the team to symlink the hook in after the next time they pull / rebase from master.
4. From this point on, files we're editing will get consistently formatted.
5. After a few weeks, which will allow everybody's changes to go through the formatting to cut down on merge conflicts, I'll reformat the entire source with a batch-change pull-request

I know that most of us are unhappy about the way prettier currently formats object-destructure assignments when they're not quite wide enough to exceed the line limit. We can address this through the prettier.js dev team (and I'll look into how hard it is to cook up a PR for that functionality). Meanwhile in particularly egregious instances, you can insert a `//prettier-ignore` comment above the statement and that AST node should remain unmolested.


.prettierrc.yaml
```diff
@@ -0,0 +1,13 @@
+# .prettierrc
+printWidth: 160
```
kzantow: I am still opposed to this. Wider than github, makes reviewing changes in PRs more irritating.

bin/package.json
```diff
@@ -10,9 +10,9 @@
   "license": "ISC",
   "devDependencies": {
     "async": "^2.0.1",
+    "graceful-fs": "4.1.11",
+    "prettier": "1.7.4",
     "prompt": "^1.0.0"
   },
-  "dependencies": {
```
kzantow: wow, can't believe this was missed so long...

bin/pre-commit.js
```diff
@@ -0,0 +1,48 @@
+#! /usr/bin/env node
```
kzantow: Will this work if you don't have node installed globally?

bin/pretty.js
```diff
@@ -0,0 +1,226 @@
+#! /usr/bin/env node
```
kzantow: Ditto

bin/pre-commit.js
sophistifunk: Probably not, but neither will anything else in our build process.

michaelneale: if it is for ci environment and not part of mvn command cycle, is ok as we use that elsewhere... 

but I guess this does require devs to have npm/node installed (not unreasonable for development, as long as it isn't required for mvn commands right @kzantow?)

michaelneale:
this has upset the eslint rule checks, so I guess they have to be removed in favour of prettier?

sophistifunk:
@michaelneale the editor had a few semantic rules turned off that are on in the other areas, I'll revert that one for now.

bin/pre-commit.js
sophistifunk: For now it's client-side only, so it only needs to run on dev machines.

.prettierrc.yaml
michaelneale: Let's see how it goes in practice I guess?

bin/pretty.js
michaelneale: see above - is dev workstation only script and opt-in for now

michaelneale:
nice @sophistifunk 

Lets see if master can get stable with either https://github.com/jenkinsci/blueocean-plugin/pull/1484 or https://github.com/jenkinsci/blueocean-plugin/pull/1503 - then perhaps merge it in if @cliffmeyers is ok?

sophistifunk:
Green! Anybody with objections to merging, speak now or forever hold your peace ;-)

kzantow:
160 :)

sophistifunk:
Stop doing so much with each statement :)

bin/pretty.js
dianapavaluc26: This conversion is pre-existing: the base branch already calls `CountryCode.valueOfIgnoreCase` in `ApiUriConfigFactory.get(String)`. The order-history path supplies `request.tenant.name` from the `UserTenant` enum, so it cannot introduce an arbitrary unsupported tenant here. The new provider preserves that existing contract; handling invalid external tenant input belongs at the request parsing boundary rather than this routing change.

PR #1502, opened 2017-10-21
Usage analytics

i386:
# Description

* Start tracking usage statistics for viewing specific pages in Blue Ocean
* Stop sending data to `/rest/analytics/track` when Usage Statistics is disabled
* Backend will not send event data if usage stats is off. However this change is just so users don't get scared when its been disabled and they see events are POSTed back to the server.
* Fixes a few issues with the `AnalyticsService` not sending back the `TrackRequest` body properly
* Fixes an issue where `AnalyticsRoute` was not deserialising the `TrackRequest`
* Add `pageview` event to Classic pages

## Events

The following events have been added

| Event Name  | Description |
| ------------- | ------------- |
| dashboard_visited | Send when user visits the Dashboard  |
| pageview | Send every time the page is visited or the view is changed  |
| pipeline_activity_visited  | Send every time the Pipeline activity tab is visited  |
| pipeline_creation_started  | Sent when the user clicks the `New Pipeline` button  |
| pipeline_branches_visited  | Send every time the Pipeline branches tab is visited  |
| pipeline_pull_requests_visited  | Send every time the Pipeline pull requests tab is visited  |
| pipeline_run_visited  |  Send every time the Pipeline run result is visited |
| pipeline_run_changes_visited  |  Send every time the Pipeline run changes is visited |
| pipeline_run_tests_visited  |  Send every time the Pipeline run tests is visited |

See [JENKINS-47579](https://issues.jenkins-ci.org/browse/JENKINS-47579).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
exciting stuff

i386:
Ready for review @michaelneale @vivek @kzantow

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/BlueOceanUrlAction.java
```diff
@@ -35,4 +37,8 @@ public String getUrlName() {
     BlueOceanUrlObject getBlueOceanUrlObject(){
         return blueOceanUrlObject;
     }
+
+    public boolean isAnalyticsEnabled() {
```
kzantow: Maybe refactor these to a single bit of logic to determine enabled/not? e.g. could add flags, options, etc. in the future. E.g. `Analytics.isEnabled()`

blueocean-rest-impl/src/main/webapp/scripts/analytics.js
```diff
@@ -0,0 +1,10 @@
+
+window.addEventListener('load', function() {
+    var eventData = { name: 'pageview', properties: { mode: 'classic' } };
```
kzantow: mode `classic`?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/BlueOceanUrlAction.java
i386: Done

blueocean-rest-impl/src/main/webapp/scripts/analytics.js
i386: This event will send pageviews events only when classic pages are loaded.

blueocean-dashboard/src/main/js/analytics.js
```diff
@@ -0,0 +1,38 @@
+import { analyticsService } from '@jenkins-cd/blueocean-core-js';
+
+/** codifies all events send to analytics service in this module */
+class Analytics {
+    trackPageView() {
+        analyticsService.track('pageview', { mode: 'blueocean' });
```
i386: @kzantow this method is used for page views in Blue Ocean. The mode property allows us to segment blue vs classic on the same event.

i386:
@kzantow yes, thats right. Ive let downstream know how to send their data to the right place.

blueocean-dashboard/src/main/js/components/MultiBranch.jsx
```diff
@@ -28,6 +29,7 @@ export class MultiBranch extends Component {
             this.pager = this.context.pipelineService.branchPager(organization, pipeline);
         }
         this._countExtensions();
+        analytics.trackPipelineBranches();
```
i386: @kzantow is `componentWillMount` the right place to send events, if the intent is to send the event every time the branch tab is clicked?

blueocean-rest/src/main/java/io/jenkins/blueocean/analytics/Analytics.java
```diff
@@ -33,17 +41,18 @@ public TrackRequest(String name, Map<String, Object> properties) {
...
                 return input != null && input.isEnabled();
             }
         }, null);
-        if (analytics == null) {
-            throw new IllegalStateException("No analytics instance available");
-        }
-        return analytics;
+    }
+
+    public static boolean isAnalyticsEnabled() {
```
vivek: Missing docs

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/analytics/KeenConfiguration.java
```diff
@@ -35,12 +35,12 @@ KeenProject project() {
     public static class DefaultKeenConfiguration extends KeenConfiguration {
         @Override
         public String getProjectId() {
-            return "595421390ee24f5b59032f79";
+            return "5950934f3d5e150f5ab9d7be";
```
vivek: project id and key can change in future as well? key might rotate in future? In that case all the previous instances will stop emitting events, no?

blueocean-dashboard/src/main/js/components/MultiBranch.jsx
kzantow: I was looking at this, I would probably push all this stuff into the router instead of the components. `componentWillMount` will work for what you want, though, yes.

blueocean-rest/src/main/java/io/jenkins/blueocean/analytics/Analytics.java
i386: Fixed

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/analytics/KeenConfiguration.java
i386: Thats right. Not sure of a good way of solving this yet. Thoughts?

blueocean-dashboard/src/main/js/components/MultiBranch.jsx
i386: Moved everything to the router level. Much cleaner.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/analytics/KeenConfiguration.java
vivek: Maybe if there is error returned (403/401) then check for new values from jenkins.io or from public s3 bucket?

Other issue, the keen.io credentials are known so the analytics data is prone to getting spammed by malicious users. Not sure what we can do in this case. Its easier to do such tings for saas services.

michaelneale:
Hrm I wonder if de dupe for non get is hurting other things accidentally (probably not)

kzantow:
@michaelneale it's definitely possible this is/was causing some other issues, but probably not too pervasive

i386:
There's still something screwy with the Fetch here. Calls are made to the fetch but the requests are never issued.

i386:
Looks like DevTools is tricking us. Ill merge and do some more testing.

michaelneale:
as a rule probably better to hold off merging until master was healthy - other PRs have been held up for it.

i386:
@michaelneale my apologies - did not realise.

vivek:
@michaelneale @i386 Looks like fixes related to fetch fixed ATH. Master is green with this merge: https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/master/1418/pipeline is green.

i386:
How odd

PR #1503, opened 2017-10-23
testing is very hard

michaelneale:
# Description

@kzantow this may help - but may be redundant with your upcoming changes to fix the testEditor and other things

### DIFF
--- a/acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/GithubCreationPage.java
+++ b/acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/GithubCreationPage.java
@@ -108,7 +108,8 @@ public class GithubCreationPage {
     }
 
     public void clickCreatePipelineButton() {
-        wait.until(ExpectedConditions.elementToBeClickable(createPipelineButton)).click();
+        wait.until(ExpectedConditions.elementToBeClickable(createPipelineButton));
+        wait.until(ExpectedConditions.visibilityOf(createPipelineButton)).click();
     }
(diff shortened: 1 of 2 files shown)


michaelneale:
shouldn't need this one...

acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/GithubCreationPage.java
```diff
@@ -108,7 +108,8 @@ public void selectPipelineToCreate(String pipeline){
     }
 
     public void clickCreatePipelineButton() {
-        wait.until(ExpectedConditions.elementToBeClickable(createPipelineButton)).click();
+        wait.until(ExpectedConditions.elementToBeClickable(createPipelineButton));
+        wait.until(ExpectedConditions.visibilityOf(createPipelineButton)).click();
```
cliffmeyers: Which error are you trying to defend against here?

kzantow: @cliffmeyers this was an alternate fix to what I included in #1484 the issue is the first click fails because the button is moving

cliffmeyers: Yeah I don't think this code change will help, it could have used `WaitUtil.click` but to your point, we can just merge in #1484 instead.

kzantow:
Superseded by #1484

PR #1508, opened 2017-10-24
[JENKINS-47583] BlueOcean cannot be build on windows

scherler:
# Description

See [JENKINS-47583](https://issues.jenkins-ci.org/browse/JENKINS-47583).

screenshot_2017-10-24_13-38-55

BUT 

screenshot_2017-10-24_17-06-08

Not sure why ATM, works on debian so I guess more path fun :(

Depends on https://github.com/jenkinsci/js-builder/pull/22/

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
ping @kzantow

kzantow:
@scherler is this actually working or are you requesting help to get it working?

kzantow:
@scherler update to master and it should fix the flaky test

scherler:
@kzantow it does build now on windows however when doing `mvn hpi:run` you will see https://user-images.githubusercontent.com/596701/31951685-1957aa4a-b8df-11e7-994d-ae723329c32f.png and the rebundle is broken as well on windows. Help is requested to get `mvn hpi:run` working.

kzantow:
@scherler yeah, last year this was the same spot I was in. I tried to get this working for a while... the issue is `browserify`, something about how it constructs the windows paths with backslashes, I'm pretty sure. I couldn't figure out a fix for it at the time, it will likely require some modification to `js-builder` to handle this. How much of a priority is it? If it can wait a bit, it might not be worth spending time on; I have some other changes coming in the next few weeks that will probably solve this.

scherler:
@alvarolobato is setting the priorities on this one. I am doing this for @jtnord  :wink:

kzantow:
Right, point is... I spent at least a few days debugging browserify and js-builder and got nowhere. Maybe you'll have better success, but if you can wait a few weeks, it will probably make this irrelevant.

scherler:
@michaelneale @kzantow we ( @alvarolobato and I) think we will declare this PR as good as it gets under the given circumstances. However the PR fixes at least that you can build BO, so one step forward.

scherler:
That said please review as well https://github.com/jenkinsci/js-builder/pull/22

kzantow:
I disagree that this is an improvement. What's built isn't functional, so it's not actually "building".

scherler:
@kzantow the **not building functional script** is another issue and with this PR at least it builds and not fails. People may be interested to build on windows a specific jar and not the front end code. Why not allow them to do so? I do not understand which changes you request with your veto!

kzantow:
If you want to build Java code on Windows, the right way to do it is disable the `frontend-maven-plugin` in that context, not pretend that the build succeeded. The issue I have is the more likely thing is someone will try to build and run and it will fail to run because the frontend fails, so this isn't a solution to "building".

cliffmeyers:
I think it might be better stick with current behavior so that the build "fails fast" on Windows rather than covering up the fact that a broken bundle was built. If there is an urgent need for a workaround, perhaps it's better to add a property that disables the `frontend-maven-plugin`. If someone just wants to do that in their local Windows environment, they can achieve that just by deleting the `.mvn_exec_node` file I believe?

kzantow:
Right, to be more specific about what I'm talking about... you can disable a maven profile for an OS: https://stackoverflow.com/a/9285352

PR #1512, opened 2017-10-25
JENKINS-47490 Make the step indicator animate

NicuPascu:
Make the step indicator animate again (used to rely on css animation that didn't work on IE and was removed)

# Description

See [JENKINS-47490](https://issues.jenkins-ci.org/browse/JENKINS-47490).

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code


sophistifunk:
Why not change it to use `<StatusIndicator/>` instead of `getGlyphFor()`? Duplicating the animation logic (and detection of when it should spin) in here doesn't seem great.

NicuPascu:
@sophistifunk I just wanted to fix the missing animation, I didn't want to refactor the component because of this. TBH I don't know why the component doesn't use <StatusIndicator /> as you suggested, I'll try to do that and see if anything breaks.

michaelneale:
@NicuPascu any thoughts on how to test this (automated)? probably an ATH thing?

sophistifunk:
@michaelneale you kiss you mother with that mouth?

i386:
@NicuPascu I think an enzyme test to ensure the right component (indeterminate progress) gets put in the ResultItem when it is "running" would be important to the fix here. As a general rule, we should be adding tests like these to ensure we don't get a future regression.

If we were writing a new component, I'd expect these sorts of tests would be written before the component would be considered done.

michaelneale:
yeh probably not an ATH thing - but this has regressed, so where there is a regression,there is probably a missing test.

sophistifunk:
That should definitely be a unit test, not an ATH test

michaelneale:
@sophistifunk agree

sophistifunk:
Looks good, given the curry @NicuPascu got trying to re-use the existing component I suggested. Next time I need to make changes to `ResultItem` I might have a timeboxed crack at it, but it's not a huge (or super complex) duplication so it's no biggie.

bb-tf

jenkins-design-language/src/js/components/ResultItem.jsx
```diff
@@ -109,12 +134,21 @@ export class ResultItem extends Component {
             },
         };
 
+        if (resultClean === "running" && !this.infiniteRotationRunning) {
+            requestAnimationFrame(this.infiniteLoadingTimer);
```
scherler: Should that not better go into `componentDidMount` and validate it in `componentWillReceiveProps`?

NicuPascu:
@scherler after a discussion with @sophistifunk we decided that the code is more readable in this form, is there a reason I might be missing of why it would be worth it to do it like that?

scherler:
http://busypeoples.github.io/post/react-component-lifecycle/ the render method comes before didMount and generates the DOM that is then fully rendered in didMount https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame says:

> You should call this method whenever you're ready to update your animation onscreen.

That is only th case when your DOM has finished rendering hence adding it to render is wrong.

NicuPascu:
Right, but that statement assumes we want to manipulate the DOM before the next browser re-paint, in our case we do not change the DOM in any way, we just increment the rotation degrees variable that is being used in the render function. Even if the first render would show the icon rotated by a few degrees (which is not the case), it would not be noticeable to the human eye, we rotate the icon 60 times per second.

sophistifunk:
I vote for leaving it with the simplified version for now, and if it becomes a performance issue we deal with it more cleanly at that point. React should deal with the single attr change without any problem, and since we're using `requestAnimationFrame` the browser will self-throttle the animation if we're causing any sort of a ruckus.

michaelneale:
Ruckus is a word not used enough: https://www.youtube.com/watch?v=hR9F8t39Az4

PR #1515, opened 2017-10-26
[FIX JENKINS-47592] - only auto-bundle things that look like blueocean plugins

kzantow:
# Description

Only auto-bundle plugins that are downstream from blueocean.

To test, `hpi:hpl` a plugin with a `package.json` in the root, e.g. `pipeline-stage-view-plugin` to a typical blueocean clone and `hpi:run` 

See [JENKINS-47592](https://issues.jenkins-ci.org/browse/JENKINS-47592).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-web/src/main/java/io/jenkins/blueocean/dev/RunBundleWatches.java
+++ b/blueocean-web/src/main/java/io/jenkins/blueocean/dev/RunBundleWatches.java
@@ -1,5 +1,6 @@
 package io.jenkins.blueocean.dev;
 
+import hudson.PluginManager;
 import hudson.PluginWrapper;
 import hudson.init.InitMilestone;
 import hudson.init.Initializer;
@@ -86,8 +87,12 @@ public class RunBundleWatches {
         System.out.println("Running in development mode, watching bundles...");
 
         int buildNumber = 0;
-        List<PluginWrapper> plugins = Jenkins.getInstance().pluginManager.getPlugins();;
+        PluginManager pluginManager = Jenkins.getInstance().pluginManager;
+        List<PluginWrapper> plugins = pluginManager.getPlugins();
         for (final PluginWrapper p : plugins) {
+            if (!isBlueoceanPlugin(pluginManager, p)) {
+                continue;
+            }
             try {
                 final File projectDir = findPluginWorkDir(new File(p.baseResourceURL.getPath()));
                 if (projectDir != null) {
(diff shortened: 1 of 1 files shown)


michaelneale:
LGTM - @cliffmeyers ?

i386:
@kzantow your run hung for a few hours https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/PR-1515/1/pipeline

blueocean-web/src/main/java/io/jenkins/blueocean/dev/RunBundleWatches.java
```diff
@@ -370,6 +375,21 @@ public void run() {
         }
     }
 
+    private static boolean isBlueoceanPlugin(PluginManager pluginManager, PluginWrapper p) {
+        if (p == null) {
+            return false;
+        }
+        if (p.getShortName().contains("blueocean")) {
```
scherler: that means that third party plugins would not be bundled if they are not contain `blueocean` 

That would have the consequence that our company plugin would not bundle, not sure if that is what we want.

kzantow: @scherler it inspects all dependencies. If your plugin does not have a dependency on any `blueocean` plugin, then it won't get automatically bundled. If it does - e.g. `blueocean-core-js`, then it will be considered a "blueocean" plugin

kzantow: that said, i think this may be excluding `jdl`, let me see if there's a better way

cliffmeyers: Might also want to consider whether we need to check all transitive deps for a blueocean-plugin dependency. Not sure about specifics of @scherler 's setup, but I seem to recall that there are two projects that depend on a "common" project, and that depends on blueocean? If so if we aren't checking transitive deps, then auto-build would work for "common" but not for the other two?

kzantow: It's already checking transitive dependencies. But it's a bit of a hackish way, I need to find a better way.

vivek:
@kzantow safe to close this one?

halkeye:
as per @vivek's comment, since its been 5+ months without comments/commits, and doesn't seem ready to go, i'm going to close it.

PR #1517, opened 2017-10-26
Display all pipeline nodes when there are many

benlangfeld:
# Description

The BlueOcean nodes API is paginated, returning results in pages of (by default) 100 items. For large pipelines with more nodes than this, the UI would display only the first 100, making no attempt to fetch further pages.

I'm not aware of any tests for the existing functionality which I could have extended to demonstrate the issue, and so I also did not add any new tests for this.

In order to manually test this, I used [a pipeline with > 300 items](https://github.com/benlangfeld/jenkins-test/blob/master/Jenkinsfile). On the blueocean-plugin master branch, only the first 100 will be displayed. On this branch, all nodes will be displayed.

While a build is in progress, I do encounter several errors in the browser console:

```
blueocean-core-js.js:3184 [ERROR - io.jenkins.blueocean.dashboard.karaoke.Pager.Pipeline] Error fetching page TypeError: Already read
```

I believe this is to do with overlapping fetches because we do not wait for all pages to be fetched before making another attempt, but I am unsure of the best approach to fix this. It has no negative impact on the user experience, however.

See [JENKINS-41205](https://issues.jenkins-ci.org/browse/JENKINS-41205).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/package.json
+++ b/blueocean-dashboard/package.json
@@ -41,6 +41,7 @@
     "gulp-mocha": "3.0.1",
     "mocha": "3.1.0",
     "nock": "8.0.0",
+    "parse-link-header": "1.0.1",
     "react-addons-test-utils": "15.4.2",
     "react-router": "2.8.1",
     "reactify": "1.1.1",
--- a/blueocean-dashboard/src/main/js/components/karaoke/rest/KaraokeApi.js
+++ b/blueocean-dashboard/src/main/js/components/karaoke/rest/KaraokeApi.js
@@ -1,4 +1,4 @@
-import { capabilityAugmenter, Fetch, FetchFunctions, logging } from '@jenkins-cd/blueocean-core-js';
+import { capabilityAugmenter, Fetch, logging } from '@jenkins-cd/blueocean-core-js';
 import debounce from 'lodash.debounce';
 import { generateDetailUrl } from '../urls/detailUrl';
 import { getNodesInformation } from '../../../util/logDisplayHelper';
@@ -63,7 +63,6 @@ export class KaraokeApi {
         const href = generateDetailUrl(pipeline, branch, runId);
         logger.debug('Fetching href', href);
         return Fetch.fetchJSON(href, { fetchOptions })
-            .then(FetchFunctions.checkStatus)
             .then(data => capabilityAugmenter.augmentCapabilities(data));
     }
 
@@ -84,7 +83,6 @@ export class KaraokeApi {
                     resolve();
                 }
                 resolve(Fetch.fetch(finalHref, { fetchOptions })
-                    .then(FetchFunctions.checkStatus)
                     .then(parseMoreDataHeader)
                     .then(parseNewStart));
             }, 200)();
@@ -103,7 +101,6 @@ export class KaraokeApi {
                     resolve();
                 }
                 resolve(Fetch.fetchJSON(href, { fetchOptions })
-                    .then(FetchFunctions.checkStatus)
                     .then(data => capabilityAugmenter.augmentCapabilities(data))
                     .then(getNodesInformation));
             }, 200)();
--- a/blueocean-dashboard/src/main/js/components/karaoke/rest/KaraokeApi.js
+++ b/blueocean-dashboard/src/main/js/components/karaoke/rest/KaraokeApi.js
@@ -1,5 +1,6 @@
-import { capabilityAugmenter, Fetch, logging } from '@jenkins-cd/blueocean-core-js';
+import { capabilityAugmenter, Fetch, FetchFunctions, logging } from '@jenkins-cd/blueocean-core-js';
 import debounce from 'lodash.debounce';
+import parse_link_header from 'parse-link-header';
 import { generateDetailUrl } from '../urls/detailUrl';
 import { getNodesInformation } from '../../../util/logDisplayHelper';
 
@@ -47,6 +48,30 @@ function parseNewStart(response) {
     const hasMore = contentLength > maxLength;
     response.hasMore = hasMore; // eslint-disable-line
     return response;
+}
+
+function fetchAndParseNodes(href, fetchOptions) {
+  return Fetch.fetch(href, { fetchOptions })
+      .then(result => {
+        const links = parse_link_header(result.headers.get('Link'));
+        return FetchFunctions.parseJSON(result).then(data => {
+          return { rawNodes: data, links: links };
+        });
+      });
+}
+
+function recursivelyFetchNodes(href, fetchOptions) {
+  return fetchAndParseNodes(href, fetchOptions).then(result => {
+    if (!result.links.next || result.rawNodes.length < 1)
+      return result;
+
+    logger.debug("Fetching next node batch from", result.links.next.url);
+    return new Promise(resolve => {
+      recursivelyFetchNodes(result.links.next.url, fetchOptions).then(nextBatch => {
+        resolve({ rawNodes: result.rawNodes.concat(nextBatch.rawNodes) });
+      });
+    });
+  });
 }
 
 export class KaraokeApi {
@@ -103,6 +128,25 @@ export class KaraokeApi {
                     .then(getNodesInformation));
             }, 200)();
         });
+    }
+
+    getAllNodes(href) {
+      // creating a new promise to be able to debounce the fetching
+      return new Promise((resolve) => {
+          debounce(() => {
+              const fetchOptions = prepareOptions();
+              logger.debug('Fetching with json enabled parsing the following href', href);
+              if (!href || href === undefined) {
+                  // leave a logger warning and abort
+                  logger.warn('could not fetch with empty href');
+                  resolve();
+              }
+              resolve(recursivelyFetchNodes(href, { fetchOptions })
+                  .then(data => capabilityAugmenter.augmentCapabilities(data.rawNodes))
+                  .then(getNodesInformation)
+              );
+          }, 200)();
+      });
     }
 
     getSteps(href) {
--- a/blueocean-dashboard/src/main/js/components/karaoke/rest/KaraokeApi.js
+++ b/blueocean-dashboard/src/main/js/components/karaoke/rest/KaraokeApi.js
@@ -50,28 +50,25 @@ function parseNewStart(response) {
     return response;
 }
 
-function fetchAndParseNodes(href, fetchOptions) {
-  return Fetch.fetch(href, { fetchOptions })
-      .then(result => {
-        const links = parse_link_header(result.headers.get('Link'));
-        return FetchFunctions.parseJSON(result).then(data => {
-          return { rawNodes: data, links: links };
-        });
-      });
-}
-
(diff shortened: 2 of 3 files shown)


benlangfeld:
I believe the test failure here is unrelated to my changes.

blueocean-dashboard/src/main/js/components/karaoke/rest/KaraokeApi.js
```diff
@@ -62,9 +84,7 @@ export class KaraokeApi {
         const fetchOptions = prepareOptions();
         const href = generateDetailUrl(pipeline, branch, runId);
         logger.debug('Fetching href', href);
-        return Fetch.fetchJSON(href, { fetchOptions })
-            .then(FetchFunctions.checkStatus)
-            .then(data => capabilityAugmenter.augmentCapabilities(data));
```
scherler: I understand the first drop `.then(FetchFunctions.checkStatus)` but not `.then(data => capabilityAugmenter.augmentCapabilities(data))` AFAIR that is needed but did not looked at the code for a while

benlangfeld: @scherler `Fetch.fetchJSON` does it internally, it's not necessary.

benlangfeld:
@scherler 

```
Nice work, some observation:

you would fetch 1k of nodes if there would be, but I doubt that it a) performs and b) would result in an usable graph
IMO you would need to implement a "fetchMore" button to trigger more
What happens if you have serial 300 nodes, AFAIR that leads to a real weird graph
```

To be honest, I think these concerns are all secondary to actually showing all the data. If I implement a pipeline with 1k nodes, I don't think it's Jenkins' place to tell me I can't see them all, nor to judge that the resultant graph is weird. As the Pipeline author, that's my responsibility to deal with.

A "show more" button is a nice to have; at this point my main concern is that my team can't easily find their failures and it's causing us serious productivity issues.

sophistifunk:
Looks OK to me 👍

blueocean-dashboard/package.json
```diff
@@ -41,6 +41,7 @@
     "gulp-mocha": "3.0.1",
     "mocha": "3.1.0",
     "nock": "8.0.0",
+    "parse-link-header": "1.0.1",
```
michaelneale: looks like a small dep - ok for you @sophistifunk ?

sophistifunk: Yeah I checked it out it's focused and only has one other small dependency.

michaelneale:
nice one @benlangfeld !

blueocean-dashboard/src/main/js/components/karaoke/rest/KaraokeApi.js
scherler: `.then(data => capabilityAugmenter.augmentCapabilities(data))` no that is not done in `Fetch.fetchJSON`

benlangfeld: What is this then, if it’s not the same thing? https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-core-js/src/js/fetch.js#L290

scherler: You are right I confused it with https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-dashboard/src/main/js/components/karaoke/services/Augmenter.js sorry

blueocean-dashboard/src/main/js/components/karaoke/rest/KaraokeApi.js
```diff
@@ -49,6 +50,27 @@ function parseNewStart(response) {
     return response;
 }
 
+function recursivelyFetchNodes(href, fetchOptions) {
```
cliffmeyers: Would it be possible to add 1-2 unit tests for this method? You can see an example of how we've done some similar unit tests with mocks or the `nock` library in the `blueocean-core-js` module. Check out the specs for `Capabilities` and `Fetch`.

blueocean-dashboard/src/main/js/components/karaoke/rest/KaraokeApi.js
```diff
@@ -103,13 +122,30 @@ export class KaraokeApi {
...
+    getAllNodes(href) {
+      // creating a new promise to be able to debounce the fetching
+      return new Promise((resolve) => {
+          debounce(() => {
+              const fetchOptions = prepareOptions();
+              logger.debug('Fetching with json enabled parsing the following href', href);
+              if (!href || href === undefined) {
+                  // leave a logger warning and abort
+                  logger.warn('could not fetch with empty href');
+                  resolve();
```
cliffmeyers: I wonder if it'd be more appropriate to reject the promise in the event that the `href` is empty? If I were writing synchronous code then `throw new Error()` feels very natural when receiving an invalid arg... so `reject` feels more correct to me.

cliffmeyers:
cc @vivek we had discussed this concept of fetching all of the nodes for the pipeline graph when the number of nodes is large (n > 100) a while back. This PR implements that in the UI by fetching each page successively using backend's default page size (100). Do you have any insight into how increasing the requested page size would impact backend perf?  Would we be better served by using a larger page size, say 250?

michaelneale:
@cliffmeyers I think I am ok with this functionality wise. If people are going nuts with 100000000s of nodes/stages/branches... well they aren't goign to have a good time anyway (and we may have to address it by giving up and just showing a flat log... I dunno).

blueocean-dashboard/src/main/js/components/karaoke/rest/KaraokeApi.js
benlangfeld: I'm just mirroring the behaviour of the existing `getNodes` here to cause minimum impact. Would you prefer I change both, diverge or maintain parallel behaviour?

cliffmeyers: No problem, let's stick with current behavior to avoid any unintended issues. Could you respond to the other two questions I posted? Thanks

vivek:
>cc @vivek we had discussed this concept of fetching all of the nodes for the pipeline graph when the number of nodes is large (n > 100) a while back. This PR implements that in the UI by fetching each page successively using backend's default page size (100). Do you have any insight into how increasing the requested page size would impact backend perf? Would we be better served by using a larger page size, say 250?

At present blueocean backend doesn't do any caching. I had some discussion about it with @svanoort in past, maybe some level of caching is done in bismuth. @svanoort is there any caching done in bismuth? If not then reading flow node graph lets say 250 or larger at times (page size) is going to very expensive (there are persistent actions that are read from disk) and every time a page is asked, entire flow node graph is loaded in memory and specific page is served.

michaelneale:
@vivek right - so what we have now (before this PR) is perhaps too restrictive, it is fine to show far more nodes than we fetch now. however, loading them ALL regardless of how large is clearly going to be a problem (as this PR is now). Is it possible to know how large it is BEFORE we start loading? as if it is > SOME_NUMBER we can bail, and just show the log, otherwise we can let it fetch things as it is now. 

Is such a thing possible?

vivek:
@michaelneale Depends on graph and dsl involved/persistent actions. Since 100 have been working ok, atleast no complaints so far. We should bump it to lets say 250 and see how it works?

svanoort:
Are you paging logs data or just the graph information, or graph data containing logs for each node?

There's caching in the FlowNode storage so you're not taking as many disk hits for graph info.  The bigger issue is that IIRC BlueOcean does a full scan through the flowgraph every time you load the data, so your paged requests multiply the work of analyzing that graph.   Better to serve the entire structure in one request and cache it clientside and only lazy-render what you care about.  It'll be a bit fluffy and would benefit from LZF compression probably (GZIP is too CPU-expensive) but will avoid multiple fetches and multiple scan-through-entire-graph hits to the Jenkins master.

(Again, IIRC and I may not) you're digesting the entire graph, so pagination is offering minimal benefits for system load, you're only reducing the content served a bit. 

If you're hitting logs... that's going to hurt, because that can be quite a lot of data for the nodes.  But IIRC that's separate.  

By the way, I have a confirmed user with 8000+ FlowNodes as of the other day... but the norm is more like several hundred.

vivek:
>Are you paging logs data or just the graph information, or graph data containing logs for each node?

Logs are not touched in serving node graph. 

> There's caching in the FlowNode storage so you're not taking as many disk hits for graph info. The bigger issue is that IIRC BlueOcean does a full scan through the flowgraph every time you load the data, so your paged requests multiply the work of analyzing that graph. 

in blueocean the graph is built bottoms up using bismuth, events are fired last node to first (except parallel branches IIRC). Even if it was not, since events are fired in streaming starting from last to first, till we build the graph hard to build paginated nodes without building entire graph. 

> Better to serve the entire structure in one request and cache it clientside and only lazy-render what you care about. It'll be a bit fluffy and would benefit from LZF compression probably (GZIP is too CPU-expensive) but will avoid multiple fetches and multiple scan-through-entire-graph hits to the Jenkins master. (Again, IIRC and I may not) you're digesting the entire graph, so pagination is offering minimal benefits for system load, you're only reducing the content served a bit.

+1. You are right, not much benefit with pagination. Best if we serve entire graph in compressed form. stapler automatically gzip compressed. I have not looked in to generic support for LZF or its support by browsers in general.

> If you're hitting logs... that's going to hurt, because that can be quite a lot of data for the nodes. But IIRC that's separate.

Thats right. logs are fetched using log API, not touched thru node graph APIs.

> By the way, I have a confirmed user with 8000+ FlowNodes as of the other day... but the norm is more like several hundred.

👍

benlangfeld:
So you want to just fetch without pagination rather than doing this? How does one override the default page size to be infinite?

vivek:
@benlangfeld If we agree to do this, this behavior needs to change in backend API, that is disable paginated behavior for flow node graph.

benlangfeld:
I’m happy to submit a PR to that effect if someone can make the decision :)

michaelneale:
I would suggest change the page size to something very large but finite - but not sure what they value should be.

vivek:
That sounds most reasonable approach, based on conversation with Sam, we can start with 10,000?

michaelneale:
Sure - surely that is a reasonable number. ANy more than that...  really can’t show anything sensible.

benlangfeld:
Will prep a PR for that in the morning :)

benlangfeld:
@michaelneale @vivek Overriding the default page size for a particular collection type is non-trivial because the collection is never aware it's in the default scenario: https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-rest/src/main/java/io/jenkins/blueocean/rest/pageable/PagedResponse.java#L48 Could you indicate your preferred approach such that I might submit something mergeable?

benlangfeld:
Even if this doesn't make it in, the first two commits very likely should: https://github.com/jenkinsci/blueocean-plugin/pull/1532

PR #1520, opened 2017-10-27
JENKINS-46490# Fix graph visualization for single parallel branch

vivek:
# Description
Existing karaoke/JDL(PipelineGraph) treats parallel branches if number of edges > 1. Visualization breaks down if there is just one parallel branch.

To fix it, I am introducing new element type to `BluePipelineStep` and `BluePipelineNode.Edge`. Front end uses type to detect whether a node is stage or parallel.

For given pipeline script:
```
node {
    
    stage('test' ) {
        println('pre parallel step')
    }
    
    
    stage('parallel stage') { 
        
        def array = ["a_1"]
    
        def builds = [:]
    
        array.each {
        
            test = it
            builds[it] = {
                
                node {
                    stage (it) {
                        println(it)   
                        sh 'date'
                    }
                }
            }
            
            
        }
        parallel builds
    
    }    
    
    stage('test2') {
        println('post parallel step')
    }
}
```

##Existing rendering

image

## Rendering after fix

image

See [JENKINS-46490](https://issues.jenkins-ci.org/browse/JENKINS-46490).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/util/logDisplayHelper.js
+++ b/blueocean-dashboard/src/main/js/util/logDisplayHelper.js
@@ -49,9 +49,10 @@ export const getNodesInformation = (nodes) => {
         /*
          * are we in a node that indicates that we have parallel nodes?
          */
-        if (item.edges && item.edges.length >= 2) {
-            parallelNodes = item.edges.map((itemParallel) => itemParallel.id);
+        if (item.edges) {
+            parallelNodes = item.edges.filter((edge) => edge.type === 'PARALLEL');
         }
+
         // in case we had been in a parallel node before, we will indicate it and remove the id of the parallel array
         const indexParallel = parallelNodes.indexOf(item.id);
         const isParallel = indexParallel !== -1;
@@ -72,6 +73,7 @@ export const getNodesInformation = (nodes) => {
             key: key || undefined,
             id: item.id,
             edges: item.edges,
+            type: item.type,
             displayName: item.displayName,
             displayDescription: item.displayDescription,
             title: title || `runId: ${item.id}`,
--- a/blueocean-dashboard/src/main/js/util/logDisplayHelper.js
+++ b/blueocean-dashboard/src/main/js/util/logDisplayHelper.js
@@ -50,7 +50,7 @@ export const getNodesInformation = (nodes) => {
          * are we in a node that indicates that we have parallel nodes?
          */
         if (item.edges) {
-            parallelNodes = item.edges.filter((edge) => edge.type === 'PARALLEL');
+            parallelNodes = item.edges.filter((edge) => edge.type === 'PARALLEL').map((edge) => edge.id);
         }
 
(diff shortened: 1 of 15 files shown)


sophistifunk:
Code looks good! Could you please add a small test to `PipelineGraph-spec.js` that exercises the changes to `PipelineGraphLayout` just to make sure we don't regress it in future though?

sophistifunk:
What's the semantic meaning of a "parallel" branch with only one node though?

vivek:
@sophistifunk Sure, will add a spec for this.

>What's the semantic meaning of a "parallel" branch with only one node though?

This should not have been treated as single vs multiple nodes inside parallel. That was the bug in current code where it rendered nodes inside parallel as parallel visualization only if it was >1. 

Fix is such that it treats them all the same by detecting whether the node is parallel or not.

michaelneale:
ok with this is @sophistifunk is

blueocean-dashboard/src/main/js/util/logDisplayHelper.js
```diff
@@ -49,9 +49,10 @@ export const getNodesInformation = (nodes) => {
         /*
          * are we in a node that indicates that we have parallel nodes?
          */
-        if (item.edges && item.edges.length >= 2) {
-            parallelNodes = item.edges.map((itemParallel) => itemParallel.id);
+        if (item.edges) {
+            parallelNodes = item.edges.filter((edge) => edge.type === 'PARALLEL');
```
cliffmeyers: Changing this from a `map` to a `filter` is going to change the structure of the array  (from an array of ID's to an array of nodes, I believe). Later below on line 57 there is an `parallelNodes.indexOf(item.id)` check that seems like it will fail 100% of the time if we aren't still calling `map` to project the ID's into the array. I am not really familiar with the code so maybe it doesn't matter.

michaelneale:
ping @sophistifunk @vivek  this may not be right based on comments from @cliffmeyers

blueocean-dashboard/src/main/js/util/logDisplayHelper.js
vivek: Good catch. Fixing it.

vivek:
@sophistifunk Added spec to test single parallel node. @cliffmeyers logDisplayHelper bug has been fixed as well. PTAL.

sophistifunk:
Looks like we've triggered some failures in `blueocean-dashboard/src/test/js/pipeline-graph-converter-spec.js`, which deals with converting the jenkins-generated data into the format expected by the graph rendering widget. Most likely it'll just need the `type` property added to the input JSON files.

cliffmeyers:
Code changes look good just need to get test failures nailed down.

vivek:
@cliffmeyers @sophistifunk ah, fixing dashboard tests. They need to be type aware.

sophistifunk:
bbm

PR #1521, opened 2017-10-30
Performance improvements for LogConsole component

AlanFoster:
# Description

Note: This PR is best viewed with [?w=1](https://github.com/jenkinsci/blueocean-plugin/pull/1521/files?w=1)

When the following component receives new log rows: 

image

Currently when new logs are received for a given step, it forces an entire re-render of the LogConsole component's children. Unfortunately on a rather large build, this becomes quite laggy.

This PR updates the LogLine to be a PureComponent to avoid un-needed re-renders of previously rendered console lines.

# Submitter checklist
- [x] Link to JIRA ticket in description, if appropriate:
    - No ticket was created for this, let me know if this is an issue
- [x] Appropriate unit or acceptance tests or explanation to why this change has no tests:
    - This is a performance tweak, all functionality should remain the same
- [x] Reviewer's manual test instructions provided in PR description:
    - Ensure that the pipeline detail view renders as before. Note, I tested this commit on top of  the 1.3 release branch, and it worked as expected. I have rebased this PR onto of master now, so it should still work if master is stable :+1:

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
+++ b/blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
@@ -1,4 +1,4 @@
-import React, { Component, PropTypes } from 'react';
+import React, { Component, PureComponent, PropTypes } from 'react';
 import { Progress, Linkify } from '@jenkins-cd/design-language';
 import { logging } from '@jenkins-cd/blueocean-core-js';
 
@@ -13,34 +13,39 @@ const RERENDER_DELAY = 17;
 
 const logger = logging.logger('io.jenkins.blueocean.dashboard.karaoke.LogConsole');
 
-const LogLine = ({ prefix, line, index, router, location }) => {
-    const tokenized = tokenizeANSIString(line);
-    const lineChunks = makeReactChildren(tokenized);
+class LogLine extends PureComponent {
+    onClick = () => {
+        const { prefix, index, router, location } = this.props;
 
-    const onClick = () => {
         const loc2 = location; // For eslint 🙄
         loc2.hash = `#${prefix || ''}log-${index + 1}`;
         router.push(loc2);
     };
 
-    return (
-        <p key={index + 1} id={`${prefix}log-${index + 1}`}>
-            <div className="log-boxes">
-                <a
-                    className="linenumber"
-                    key={index + 1}
-                    href={`#${prefix || ''}log-${index + 1}`}
-                    name={`${prefix}log-${index + 1}`}
-                    onClick={onClick}
-                >
-                </a>
-                {
-                    React.createElement(Linkify, { className: 'line ansi-color' }, ...lineChunks)
-                }
-            </div>
-        </p>
-    );
-};
+    render() {
+        const { prefix, line, index } = this.props;
+        const tokenized = tokenizeANSIString(line);
+        const lineChunks = makeReactChildren(tokenized);
+
+        return (
+            <p id={`${prefix}log-${index + 1}`}>
+                <div className="log-boxes">
+                    <a
+                        className="linenumber"
+                        key={index + 1}
+                        href={`#${prefix || ''}log-${index + 1}`}
+                        name={`${prefix}log-${index + 1}`}
+                        onClick={this.onClick}
+                    >
+                    </a>
+                    {
+                        React.createElement(Linkify, { className: 'line ansi-color' }, ...lineChunks)
+                    }
+                </div>
+            </p>
+        );
+    }
+}
 
 LogLine.propTypes = {
     prefix: PropTypes.string,
@@ -170,7 +175,7 @@ export class LogConsole extends Component {
             </div>}
 
             {!isLoading && <div className="log-body"><pre>
-                {hasMore && <div key={0} id={`${prefix}log-${0}`} className="fullLog">
+                {hasMore && <div id={`${prefix}log-${0}`} className="fullLog">
                     <a className="btn-link inverse"
                        key={0}
                        target="_blank"
@@ -180,7 +185,8 @@ export class LogConsole extends Component {
                     </a>
                 </div>}
                 {!isLoading && lines.map((line, index) => (
-                    <LogLine prefix={prefix}
+                    <LogLine key={index}
+                             prefix={prefix}
                              index={index}
                              line={line}
                              router={router}


blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
```diff
@@ -13,33 +13,38 @@ const RERENDER_DELAY = 17;
 
 const logger = logging.logger('io.jenkins.blueocean.dashboard.karaoke.LogConsole');
 
-const LogLine = ({ prefix, line, index, router, location }) => {
-    const tokenized = tokenizeANSIString(line);
-    const lineChunks = makeReactChildren(tokenized);
+class LogLine extends PureComponent {
```
kzantow: Why `PureComponent`?

blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
```diff
@@ -13,33 +13,38 @@ const RERENDER_DELAY = 17;
...
+
+        return (
+            <p id={`${prefix}log-${index + 1}`}>
+                <div className="log-boxes">
+                    <a
+                        className="linenumber"
+                        key={index + 1}
+                        href={`#${prefix || ''}log-${index + 1}`}
+                        name={`${prefix}log-${index + 1}`}
+                        onClick={this.onClick}
```
kzantow: this needs to be something like `() => this.onClick()`

blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
AlanFoster: The default implementation of shouldComponentUpdate is useful in this scenario:

> If your React component’s render() function renders the same result given the same props and state, you can use React.PureComponent for a performance boost in some cases.
> https://reactjs.org/docs/react-api.html#reactpurecomponent

blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
```diff
@@ -180,7 +185,8 @@ export class LogConsole extends Component {
                     </a>
                 </div>}
                 {!isLoading && lines.map((line, index) => (
-                    <LogLine prefix={prefix}
+                    <LogLine key={index}
```
kzantow: I'm confused about the `key` usage vs. `PureComponent`. Also using indices for keys isn't particularly useful - the line text would be more appropriate.

AlanFoster: > I'm confused about the `key` usage vs. `PureComponent`.

PureComponent is the main fix of this PR, however I noticed that there was a `key` attribute on the `p` element of `LogLine`, which will have no effect on react's rendering.

> using indices for keys isn't particularly useful - the line text would be more appropriate.

Definitely! Generally indexes for keys are not valid and are considered an anti-pattern. However since this array is never ordered, or filtered, and is monotonically increasing, the index is the best key here. I decided to define the key explicitly here, rather than rely on a default key provided by React.

With regards to using the line text instead of the index for the key, there may be a problem with that - for instance a user may have the line "hello world" twice in their logs, which could cause issues

Let me know if i've misunderstood anything :+1:

blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
AlanFoster: Because we're using `PureComponent` now, rather than a stateless function, we can directly pass in `this.onClick` to the `a` react element without needing to wrap it in a arrow function. The functionality still works as expected, without the need for creating a new function on each render.

Let me know if I've mis-understood, or if it's the preferred code style of blueocean to wrap callbacks in arrow functions :+1:

AlanFoster:
@kzantow Thanks for the review! I've added some comments to your responses, let me know your thoughts! :+1:

blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
kzantow: Pretty sure that `this` gets lost in the execution context when using ES6 unless you use `.bind(this)`, does it not?

blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
kzantow: Good point about the line as a key. But why have a key at all? Is it required for the `PureComponent`?

blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
AlanFoster: Check out line 17:

```js
onClick = () => {
```

It's doing binding for us :)

kzantow: ah heh, missed that 😬

blueocean-dashboard/src/main/js/components/karaoke/components/LogConsole.jsx
AlanFoster: @kzantow Keys are only needed when react elements are contained within arrays, and are unrelated to Pure components 

The react docs are great in explaining this in more detail than I possibly could! Unintentionally the react docs describe this exact scenario too:

https://reactjs.org/docs/lists-and-keys.html#extracting-components-with-keys

kzantow: Right, I understand why to use keys. Question is... are they actually necessary with the `PureRender` component? They are to find the original components, I'm guessing. So good. :)

i386:
Thanks for the contribution @AlanFoster 👍

michaelneale:
very cool @AlanFoster

sophistifunk:
This is a Good-Thing™️

kzantow:
Thanks for the contribution, @AlanFoster

AlanFoster:
@kzantow Hey; I was just wondering if there's any timescales involved for this PR to land? 🏆 

If there's still performance issues with this change on my Jenkins CI setup I was thinking of sending further PRs, perhaps making use of the great [react-virtualize](https://github.com/bvaughn/react-virtualized) if need be :)

michaelneale:
@alanfoster this is merged to master - and hoping to commence beta releases soon. If you are able to build from master and see if it helps that would be well worthwhile to know if there is more work to do. 

Having said that, and more improvements welcome. Also I wonder if this woikd cherry pick nearly to the 1.3 branch so could be released immediately?

michaelneale:
(trying to get this in a new release... ping me next week if it still isn't avialable)

cliffmeyers:
@AlanFoster we are always happy to get contributions :)  We had discussed virtualizing the log console in the past. One possible downside to that is that it would probably break the ability to do a simple cmd/ctrl-F search against the logs, since only part of the DOM would be present. A workaround for that could be to override the browser's "find" functionality, similar to how Google Docs works.

I certainly don't want to discourage contributions, just trying to call out a potential pitfall with that approach. I'm happy to discuss it further with you if you like.

AlanFoster:
> One possible downside to that is that it would probably break the ability to do a simple cmd/ctrl-F search against the logs

@cliffmeyers I was thinking for a better UX it would be possible to add an additional row for components related to the virtualized list, such as providing a search input field, sorting functionality, and a quick means of jumping to specific rows etc. That would solve this particular issue in an intuitive way, and it opens the possibility for the server-side to implement the searching for rows, without needing to send everything over to the client to begin with.

One concern I had was actually the multiple collapsible Step components, i.e. how to best handle the height of multiple collapsible infinite list components, as well as handling event bubbling when scrolling to the end of a virtualized log, as well as the best UX for scrolling to the bottom when new messages come in, but not when the user is intentionally wanting to scroll up the page. For instance, currently on some of my larger builds, a CI log with fast output constantly forces a scroll to the bottom of the page - which I would like to tweak.

If you folk have already thought about all of this, I'd love to hear about it :+1:

sophistifunk:
We have a bunch of tickets for the log / run viewing, it seems to have become my white whale.

PR #1522, opened 2017-10-30
JENKINS-47580 Center align show more button

NicuPascu:
increase css specificity of the show more button so the it center alignes 

See [JENKINS-47580](https://issues.jenkins-ci.org/browse/JENKINS-47580).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/less/core.less
+++ b/blueocean-dashboard/src/main/less/core.less
@@ -17,7 +17,7 @@
 }
 
 button.btn-show-more {
-    display: table;
+    display: flex;
     margin: 10px auto;
 }
 
--- a/blueocean-dashboard/src/main/less/core.less
+++ b/blueocean-dashboard/src/main/less/core.less
@@ -16,7 +16,7 @@
     bottom: auto;
 }
 
-.btn-show-more {
+button.btn-show-more {
     display: table;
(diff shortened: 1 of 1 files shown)


blueocean-dashboard/src/main/less/core.less
```diff
@@ -16,7 +16,7 @@
     bottom: auto;
 }
 
-.btn-show-more {
+button.btn-show-more {
     display: table;
```
sophistifunk: I know you didn't do this @NicuPascu but could you try this with flexbox props on the button? Maybe `align-self:center`? The main container should be `display:flex` already. If it's not, don't worry about it, no need to rejig the whole layout or anything.

sophistifunk:
Just a small question above, but if it's more than a couple of lines to ditch the `display:table` then it's no biggie, I'm just easily offended and should probably HTFU ;-)

🐝

NicuPascu:
I have changed the display: table to flex but can't use `align-self: center` because the container is block, changing that to flex crashed to whole layout so... not sure it is worth it at this moment

sophistifunk:
All good, it's not a big deal anyway. I'll see myself out :)

PR #1528, opened 2017-11-03
JENKINS-47241 move the material icons from the JDL to a different npm module

NicuPascu:
Moved the material icons from the JDL to jenkins-cd/blueocean-material-icons

See [JENKINS-47241](https://issues.jenkins-ci.org/browse/JENKINS-47241).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


NicuPascu:
a test already exists here https://github.com/jenkinsci/blueocean-plugin/blob/master/jenkins-design-language/test/js/components/Icon-spec.js#L8

blueocean-material-icons/gulpfile.js
```diff
@@ -0,0 +1,95 @@
...
+        .pipe(lint())
+        .pipe(lint.format())
+        .pipe(lint.failAfterError())
+));
+
+
+//
+// Create the main bundle.
+//
+builder.bundle('src/js/components/index.js', 'blueocean-material-icons.js')
```
cliffmeyers: This is maybe worthwhile to get input from @kzantow on whether this should be a HPI project or a just a place old npm package. It's unlikely to change very often at all, so the latter might be the way to go? In which case we could just delete the bundle stuff?

blueocean-material-icons/gulpfile.js
```diff
@@ -0,0 +1,95 @@
+"use strict";
+
+process.env.SKIP_BLUE_IMPORTS = 'YES';
+process.env.NODE_ENV = 'production';
+
+/*
+ Build file for Jenkins Design Language theme.
```
cliffmeyers: typo

blueocean-material-icons/package.json
```diff
@@ -0,0 +1,68 @@
...
+  },
+  "author": "Nicolae Pascu <npascu@cloudbees.com> (https://www.cloudbees.com/)",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/jenkinsci/blueocean-plugin.git"
+  },
+  "dependencies": {
+    "linkifyjs": "2.1.4",
+    "lodash.debounce": "4.0.8",
```
cliffmeyers: Can we trim down the deps and devDeps here to the minimum that's actually needed?

jenkins-design-language/src/js/components/Icon.jsx
```diff
@@ -1,5 +1,5 @@
 import React, { PropTypes } from 'react';
-import * as IconId from './material-ui/svg-icons';
+import * as IconId from '@jenkins-cd/blueocean-material-icons';
```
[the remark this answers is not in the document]
NicuPascu: The way I made the blueocean-material-icons package is that it exports all the icons individually because they do not change, this also means anyone can use the icons separately if needed (and not have to use the Icon component). Also the Icon component is how we use these icons in blueocean, making it part of the JDL in my opinion, this component is what we will continue to tweak in the future and I wanted to keep in the JDL because of that. Also this way it is very easy to iterate through all the icons for the Icon stories so we have a way to see them all in one place.

jenkins-design-language/src/js/stories/IconStories.jsx
```diff
@@ -25,7 +29,7 @@ function AllIcons() {
                 <div style={{ display: 'flex', flexFlow: 'row wrap', marginTop: '20px' }}>
                     { Object.keys(IconId).sort().map(shape => (
                         <div key={shape} style={{marginTop: '20px', textAlign: 'center', width: '19%'}}>
-                            <Icon icon={shape} color="#4A90E2" />
+                            <Icon icon={shape} style={iconStyle} />
```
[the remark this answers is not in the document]
NicuPascu: This has been out of date and all the icons were invisible in the IconStories (white icons over white background). I made this change to make them blue agains the white background. There is no API change to the Icon component so no risk to other teams.

NicuPascu:
Thanks @cliffmeyers for the review, I tried implementing as much of the feedback as I could, can you take another look pls? Also I have benchmarked the build time in master vs this branch and is substantial, in master it takes my machine about 40 sec to re-build the JDL for any change I do, in this branch it takes about 18 sec.

NicuPascu:
Also per my discussion with @michaelneale we decided to make the icons a separate npm package instead of a HPI project because they will never change, we might only need to add new icons that might appear in the future if they are needed.

blueocean-material-icons/package.json
```diff
@@ -38,28 +36,16 @@
...
-    "gulp-less": "3.1.0",
-    "gulp-rename": "1.2.2",
     "gulp-sourcemaps": "2.1.1",
-    "gulp-svgmin": "1.2.3",
-    "harp": "0.24.0",
-    "less": "2.7.2",
-    "normalize.css": "5.0.0",
-    "octicons": "4.4.0",
     "react-addons-test-utils": "15.4.2",
     "react-router": "2.8.1",
```
cliffmeyers: You might be able to remove this one too.

blueocean-material-icons/gulpfile.js
cliffmeyers: @NicuPascu per your comment above, I think it's fine that this isn't an HPI module. The `builder.bundle` call can probably be removed then. We'll still want `builder` to run lint, unless you'd prefer to build that up manually in the `gulpfile`

jenkins-design-language/src/js/components/Icon.jsx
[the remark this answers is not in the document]
cliffmeyers: Cool, that works for me 👍

michaelneale:
yes, this makes sense as a regular package - for others to use. 

This is a tricky one - what is best review strategy? 

Mostly I think acceptance of this should be @scherler - need a review if this is better than things were?

blueocean-material-icons/gulpfile.js
```diff
@@ -0,0 +1,80 @@
...
+// Options, src/dest folders, etc
+
+const config = {
+    clean: ["dist", "target"],
+    react: {
+        sources: "src/**/*.{js,jsx}",
+        dest: "dist"
+    },
+    test: {
+        sources: 'test/js',
```
scherler: So no tests are run, because you are storing them in the component. :bug:

scherler: BTW why do you need a custom gulp?

scherler:
I agree BTW that this should never be a hpi, not only because it never changes but further it is a *npm* package that IMO actually should go to a rep on its own.

michaelneale:
Yes the idea is this could be its own repository too yes.

blueocean-material-icons/gulpfile.js
cliffmeyers: `js-builder` builds bundles which is not what we want if this is going to be consumed in non-Blue Ocean environments. Your point about cleaning up some of the references to test sources and possibly other items is valid and should be addressed. cc @NicuPascu

michaelneale:
@kzantow may be needed for help by @NicuPascu to get this into its own repo/module (if we all agree that is best way, if we are going to do all this work anyway...)

cliffmeyers:
If all that remains is splitting this into its own repository and publishing as a npm module, I've done that before and can probably help move it along. Just requires some support from ops team.

kzantow:
@cliffmeyers @michaelneale not sure it's necessary to split this to another repository. that can always be done later. js-extensions is already in a similar situation, where it's in this same repo but only available after publishing to npm (or doing other local dev stuff, which might be useful for testing new icons anyway)

blueocean-material-icons/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@jenkins-cd/blueocean-material-icons",
+  "version": "0.0.4",
+  "description": "Material icons for the Jenkins Design Language",
+  "main": "dist/js/components/index.js",
+  "scripts": {
+    "lint": "gulp lint",
+    "gulp": "gulp build",
+    "mvnclean": "gulp clean",
```
kzantow: this isn't actually being built by maven, i'd clean this `package.json` up to have only the minimum stuff

blueocean-material-icons/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "@jenkins-cd/blueocean-material-icons",
+  "version": "0.0.4",
+  "description": "Material icons for the Jenkins Design Language",
+  "main": "dist/js/components/index.js",
+  "scripts": {
+    "lint": "gulp lint",
```
kzantow: you pretty much just want a `prepublish` that is something like `babel -src src -dst dist` or whatever, don't think it's necessary to have anything complicated here

michaelneale:
@kzantow ok your call on splitting out, maybe it is easier in phases... but the main thing is to keep things compatible as possible for anyone using the icons.

blueocean-material-icons/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License
+
+Copyright (c) 2016 CloudBees Inc and a number of other of contributors
```
kzantow: it's 2017

blueocean-material-icons/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@jenkins-cd/blueocean-material-icons",
+  "version": "0.0.5",
+  "description": "Material icons for the Jenkins Design Language",
+  "main": "dist/js/components/index.js",
+  "scripts": {
+    "babel": "node_modules/.bin/babel src -d dist"
```
kzantow: mentioned before, but should be `prepublish` : `npm test && babel src -d dist` or something similar

blueocean-material-icons/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/jenkinsci/blueocean-plugin.git"
+  },
+  "dependencies": {
+  },
+  "devDependencies": {
+    "babel-cli": "6.26.0",
+    "babel-polyfill": "6.16.0",
```
kzantow: Don't think this is needed, pretty sure it's a runtime dep already provided

PR #1529, opened 2017-11-03
Task/jenkins 36641 smart web element improvements

cliffmeyers:
# Description
While trying to write more comprehensive tests for favorites, I found some limitations in `SmartWebElement` and `WebDriverMixin` API's (see immediately below):
- When retrieving a `SmartWebElement`, chained calls to `findElement/s` method would use raw Selenium API's that don't use the polling/wait behavior making them very subject to timing issues. We definitely want to be able to use these API's so we can write a simple selector for a top level element, and then do searches within that element (keeping individual selector expressions simple).
- The `WebDriverMixin.find` API always returned a single `SmartWebElement` which might actually correspond to multiple elements. There are times you want to interact with multiple elements, and the rest of `WebDriver` API's use a `List<WebElement>` in many, so it doesn't make sense IMO to conflate "one element" and "many elements" into `SmartWebElement` abstraction.

My attempt to improve these are described by these changes:
- Ensure that SmartWebElement.findElement/s (1) uses waits internally when finding descendent elements, and (2) returns SmartWebElements, otherwise chaining calls won't work well since we'll lose implicit wait behavior.
- Remove getElement/s since it overlapped with findElements/s and was only visible on SmartWebElement class... ideally we can just code against `WebElement` interface.
- Change `WebDriverMixin.find` API to differentiate between find/findMany so that client code can work with single element or list of elements
- Add `WebDrivenMixin.findNow` so we have ability to do queries that don't use an implicit wait, so we can check that elements aren't present
- Add `WebDrivenMixin.untilCondition` we can leverage full power of `ExpectedConditons.*` utility methods. These are powerful and we should offer full-class support.

**PLEASE NOTE** Still some internal cleanup in this PR that needs to happen but I think it's getting close. Please note the TODOs.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


acceptance-tests/src/main/java/io/blueocean/ath/LocalDriver.java
```diff
@@ -1,20 +1,23 @@
...
+import java.util.stream.Collectors;
 
 /**
- * Wrapper around an underlying WebDriver that
- * consistently handles waits automatically.
- *
- * Accepts expressions for css and xpath, if the provided lookup starts with a /, XPath is used
+ * Provides access to the current WebDriver instance.
+ * Contains some utility methods for working with waits and SmartWebElement.
+ * TODO: rename this class as it's *not* a WebDriver but rather more of a DriverHolder.
```
cliffmeyers: All of the client code was calling `LocalDriver.getDriver()` which returned the thread local'd `RemoteWebDriver`. No need for `LocalDriver` to impl `WebDriver` as far as I can see.

acceptance-tests/src/main/java/io/blueocean/ath/LogEntryLogger.java
```diff
@@ -18,11 +20,31 @@
...
+
+        if (usefulEntries.iterator().hasNext()) {
+            logger.info("browser console output below:");
+            for (LogEntry entry : usefulEntries) {
+                LogEntryLogger.recordLogEntry(entry);
+            }
+        } else {
+            logger.info(String.format("nothing useful written to browser console; %s entries were hidden", entries.size()));
+        }
+    }
```
cliffmeyers: Overall these changes make the browser logging stuff a bit more explicit about when 1. nothing is logged, 2. nothing is logged but useless messages, and 3. there are actually useful messages.

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -66,78 +46,25 @@ protected WebDriver getDriver() {
...
+    private FluentWait<WebDriver> buildWait() {
+        return LocalDriver.buildWait();
     }
 
     @Override
     public void click() {
         for (int i = 0; i < RETRY_COUNT; i++) {
             try {
-                WebElement e = getElement();
+                WebElement e = buildWait().until(ExpectedConditions.elementToBeClickable(element));
```
cliffmeyers: This got lost in #1484 and needs to be there to make sure we don't click on elements that are transitioning from disabled to enabled. Frequent cause of flakiness in creation tests when robot picks a repo and clicks "Create Pipeline" while it's still disabled, resulting in a hung test.

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -171,9 +98,9 @@ public void sendKeys(CharSequence... charSequences) {
      */
     public <T> T eval(String script) {
         String js = "return (function(el,elements){" + script + "})(arguments[0],arguments[1])";
-        List<WebElement> elements = getElements();
-        WebElement el = elements.iterator().next();
-        return (T)((JavascriptExecutor)getDriver()).executeScript(script, el, elements);
+        // TODO: validate this change is correct
+        WebElement el = getElement();
+        return (T)((JavascriptExecutor)getDriver()).executeScript(script, el);
```
cliffmeyers: @kzantow perhaps you can weigh in whether this change is appropriate.

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -271,20 +198,24 @@ public String getText() {
...
 
     @Override
-    public WebElement findElement(By by) {
-        WebElement e = getElement();
-        return e.findElement(by);
+    public WebElement findElement(By locator) {
+        return buildWait().until((WebDriver driver) -> {
+            WebElement matchedElement = getElement().findElement(locator);
+            return LocalDriver.wrapElement(getDriver(), matchedElement, locator);
+        });
```
cliffmeyers: Please have a look at this code in particular to make sure it makes sense.

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -271,20 +198,24 @@ public String getText() {
...
     }
 
     @Override
     public boolean isDisplayed() {
-        WebElement e = getElement();
+        WebElement e = buildWait().until(ExpectedConditions.visibilityOf(element));
         return e.isDisplayed();
```
cliffmeyers: I think in its current state this method can only ever return true or throw an exception. I might be wrong. Not sure if this should have some error handling baked in, or maybe left alone.

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
michaelneale: interesting

michaelneale:
Looks interesting to me... cc @imeredith maybe take a look on monday?

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
kzantow: why change this?

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -16,45 +15,26 @@
...
- * Accepts expressions for css and xpath, if the provided lookup starts with a /, XPath is used
+ * Wrapper around an underlying WebDriver that automatically handles waits and common gotchas.
  */
 public class SmartWebElement implements WebElement {
     private static Logger logger = Logger.getLogger(SmartWebElement.class);
-    public static final int DEFAULT_TIMEOUT = Integer.getInteger("webDriverDefaultTimeout", 3000);
     public static final int RETRY_COUNT = 3;
 
     private WebDriver driver;
+    protected WebElement element;
```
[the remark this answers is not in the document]
cliffmeyers: Let me see if I can reintroduce that lazy logic so we don't hold a ref to any WebElement.

acceptance-tests/src/main/java/io/blueocean/ath/LocalDriver.java
```diff
@@ -46,106 +49,31 @@ public static void setUrlBase(String base) {
...
+        return new SmartWebElement(driver, element, locator.toString());
     }
 
     /**
-     * Push a specific driver into context and execute the proc
-     * @param driver new driver in context
-     * @param proc procedure to execute
+     * Wraps the supplied elements in SmartWebElement if not already a SWE
      */
-    public static void use(WebDriver driver, Procedure proc) {
```
kzantow: I would not remove this, it can be useful to write tests that rely on multiple windows

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -66,78 +46,25 @@ protected WebDriver getDriver() {
...
-            throw new NoSuchElementException("Too many elements returned for: " + expr);
-        }
-        return elements.get(0);
-    }
-
-    /**
-     * Iterates over all found elements with the given function
-     * @param fn to perform on the elements
-     */
-    public void forEach(java.util.function.Consumer<WebElement> fn) {
```
kzantow: So this provided that way to operate on a collection of elements. You mentioned there are cases where we need to pass the list of the base `WebElement` to webdriver functions, are there examples where we're doing that?

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
```diff
@@ -66,78 +46,25 @@ protected WebDriver getDriver() {
...
-        for (WebElement e : getElements()) {
-            fn.accept(e);
-        }
-    }
-
-    /**
-     * Determines if the element is visible
-     * @return true if visible, false if not
-     */
-    public boolean isVisible() {
```
kzantow: i guess this is just duplicate noise, but i really don't like `isDisplayed` this mirrors jQuery's terminology

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
kzantow: right, definitely need this change after talking with Cliff, this was a stupid mistake on my part

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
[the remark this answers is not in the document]
kzantow: Now I finally understand the motivation, so you can wrap the other returned `WebElement`s with this same logic. Ok, let me stew on this for just a little while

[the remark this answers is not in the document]
cliffmeyers: Right. Otherwise can't really chain `findElement` calls together because we'll be returning a mix of `SmartWebElement` and `RemoteWebElement` which I think would lead to a lot of flake since some `WebElement`'s have wait behavior baked in and others do not.

The alternative with original code is that we'd need to do all our finds through `WebDriverMixin.find` to ensure we always get a SWE, which could lead to really lengthy selectors:

`find(".favorites-card-stack .pipeline-card[data-fullname="cliffs-pipeline"] .Checkbox.Favorite > label").click()`

 rather than

```
find(".favorites-card-stack .pipeline-card[data-fullname="cliffs-pipeline"]
   .findElement(".Checkbox.Favorite > label")
   .click()
```

I find the latter really useful because we can write individual methods on our page objects to grab specific elements, and we can compose those methods just by chaining the `findElement` calls.

I wonder if we could add some exception handling into SWE for `StaleElementException` that retries the locator. It wouldn't necessarily work for elements fetched via `findMany` since the expression stored to fetch them would match multiple elements. I can't say I've run into any `StaleElementException` yet but that's part of why I am working on a branch derived from this one to add really thorough tests to personalization as a testing ground.

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
cliffmeyers: `elements` no longer exists

cliffmeyers: Can you let me know if this change seems ok in light of the fact that there is no List from `getElements()` to pass down into `executeScript` ?

cliffmeyers: @kzantow can you let me know if you think this change is okay given that `getElements()` no longer exists, therefore passing the elements array as the last arg isn't possible anymore

cliffmeyers:
@kzantow I've made some updates to this PR. Based on our past conversations, can you let me know if you think these changes are suitable or should I continue to iterate? Perhaps we can touch base in a meeting if that's easier to discuss.

acceptance-tests/src/main/java/io/blueocean/ath/LocalDriver.java
kzantow: ... still, I would not remove this

acceptance-tests/src/main/java/io/blueocean/ath/SmartWebElement.java
kzantow: I also don't think removing getElements() is necessary or the right change. Again, making `getElements()` support a parent `SmartWebElement` will ensure that everything is always operating on the current state of the page when it's called.

cliffmeyers:
@kzantow indicated he wants to add some similar features as in this PR but with less code, so I am closing this PR. I will carry over the improvements to browser console logging in another PR.

PR #1534, opened 2017-11-06
[JENKINS-45947] Show Open BlueOcean to everyone

imeredith:
# Description

This PR removes the permissions check for the Open BlueOcean link on the side bar. The rational is that if you can see the sidebar at all, then you should be able to see the link. 

See [JENKINS-45947](https://issues.jenkins-ci.org/browse/JENKINS-45947).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
I think this makes sense, but I have a feelilng that was introduced for a reason... any memory of why @i386 or @vivek ? 

If they are ok with it, cool.

i386:
Sounds fine from my PoV but ill let @vivek speak last.

vivek:
@imeredith Hmm, I thought it works accordingly. I tried matrix based authorization with overall read permission to a user and I can see Open Blue Ocean link when logged in as that user. For anonymous user it works as well (see ci.blueocean.com). So what case it fails that this PR is trying to fix?

PR #1538, opened 2017-11-08
JENKINS-47513 allow test logs to wrap properly

NicuPascu:
This problem happened only for logs (the pipeline logs wrap fine).

Given the fact that this change is only css plus and extra container, I do not think any test is necessary, I will add a print-screen showing the fix in action

See [JENKINS-47513](https://issues.jenkins-ci.org/browse/JENKINS-47513).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/testing/TestDetails.jsx
+++ b/blueocean-dashboard/src/main/js/components/testing/TestDetails.jsx
@@ -6,7 +6,11 @@ import { observer } from 'mobx-react';
 const ConsoleLog = ({ text, className, key = 'console' }) =>
     <div className={`${className} console-log insert-line-numbers`}>
         {text.trim().split('\n').map((line, idx) =>
-            <div className="line" id={`#${key}-L${idx}`} key={`#${key}-L${idx}`}>{line}</div>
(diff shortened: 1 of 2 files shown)


NicuPascu:
fail_log

scherler:
> Given the fact that this change is only plus and extra container, I do not think any test is necessary, I will add a print-screen showing the fix in action

Do we have generally test for this? If not would be a good moment to add some.

sophistifunk:
Seems like the kind of thing for which a test would be a boatload of work, and also fragile AF.

NicuPascu:
In the description I wanted to say that the change is CSS only, somehow I forgot the word "CSS". I honestly don't know how I could write a test for this but will spend some more time on it just to see if I can come up with something.

scherler:
@NicuPascu I understand and agree with @sophistifunk in this case it may be just too much. I recommend to cover the whole log component with test, but it does not need to be in this PR.

michaelneale:
@scherler we have tests to check log tailing etc... fetching and so on - but not visual things like wrapping at all (nor position of things). Would be interesting to think of a strategy for it though. 

Nice one @NicuPascu  :bee:

PR #1540, opened 2017-11-08
Tasks/jenkins 36641 delete favorites rest api

cliffmeyers:
# Description
- To ensure that we can run ATH covering favorites in a clean state, we needed an API to delete all of the user's favorites.
- See [JENKINS-36641](https://issues.jenkins-ci.org/browse/JENKINS-36641).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/FavoriteContainerImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/FavoriteContainerImpl.java
@@ -9,7 +9,11 @@ import io.jenkins.blueocean.rest.model.BlueFavorite;
 import io.jenkins.blueocean.rest.model.BlueFavoriteContainer;
 import io.jenkins.blueocean.service.embedded.util.FavoriteUtil;
 import jenkins.model.Jenkins;
+import org.kohsuke.stapler.StaplerResponse;
+import org.kohsuke.stapler.WebMethod;
+import org.kohsuke.stapler.verb.DELETE;
 
+import javax.servlet.http.HttpServletResponse;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/FavoriteContainerImpl.java
```diff
@@ -58,4 +62,15 @@ public BlueFavorite get(String name) {
     public Link getLink() {
         return self;
     }
+
+    /**
+     * Delete all of the user's favorites.
+     */
+    @WebMethod(name = "") @DELETE
+    public void doDelete(StaplerResponse resp) throws Favorites.FavoriteException {
+        for (final Item favorite: Favorites.getFavorites(user.user)) {
```
i386: I think you can replace this loop with a Java 8 oneliner http://www.deadcoderising.com/java-8-no-more-loops/

cliffmeyers: `Favorites.getFavorites` returns an `Iterator`, so I think I'd need to do some `Spliterator` junk to get a stream and then operator on the items. I am not sure that's actually more readable? But if you know a better way I am all ears.

cliffmeyers: Looks like I could use this:

```
StreamSupport.stream(Favorites.getFavorites(user.user).spliterator(), false)
            .forEach(favorite -> Favorites.removeFavorite(user.user, favorite));
```

Although unfortunately `Favorites.removeFavorite` throws a checked exception, so I'd have to add a try/catch within the lambda... or change to it to a runtime exception.

i386: bloody iterators and streams 👎  not sure why they made this harder and worse to read than guava. 

DW about it

cliffmeyers: We could probably write our own little util class to make it a touch more readable

`StreamHelper.stream(Favorites.getFavorites(user.user))`

where

`public stream(Iterable iterable) { return StreamSupport.stream(iterable.spliterator(), false); }` 

what do you think about the favorites exception being checked? I am not really sure those errors are something we should try to force the user to recover from. I can look at the API more in subsequent PR to see if can be cleaned up?

i386: I think its ok to get rid of the checked exception if you want to do that.

cliffmeyers: I have some more work to do in a follow-up PR, so I'll implement it then.

PR #1541, opened 2017-11-08
Allow extensible run details tabs

kzantow:
# Description

See https://issues.jenkins-ci.org/browse/JENKINS-47929

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-dashboard/src/main/js/PipelineRoutes.jsx
```diff
@@ -128,6 +126,27 @@ function onLeaveCheckBackground() {
 
 const trends = AppConfig.isFeatureEnabled('trends');
 
+class NotFoundAction extends React.Component {
```
scherler: :ant:  For such component (simple presentation) the preferred form is: 

```javascript
const NotFoundAction = () => <NotFound/>;
```

scherler: Actually why do you wrap `<NotFound/>` anyway?

blueocean-dashboard/src/main/js/components/RunDetails.jsx
```diff
@@ -138,9 +141,14 @@ class RunDetails extends Component {
...
         const { router, params } = this.context;
         if (this.opener) {
-            // step back in the history to the item that's prior to initial load of RunDetails
-            const offsetToInitialRoute = this.initialHistoryLength - history.length;
+            // move back in the history to the item that's prior to initial load of RunDetails
+            const offsetToInitialRoute = this.initialHistoryLength - locationService.navCount;
             router.go(offsetToInitialRoute - 1);
+            setTimeout(() => {
+                if (locationService.current !== this.opener) {
+                    router.replace(this.opener);
```
scherler: Why `replace`? Especially because we first go to a specific history `router.go(offsetToInitialRoute - 1)` and then change that history entry.

blueocean-web/src/main/js/main.jsx
```diff
@@ -217,12 +217,16 @@ function startApp(routes, stores) {
...
+    window.addEventListener('popstate', e => locationService.navCount-=2); // gets a route change AND popstate
+    const onRouteChange = () => locationService.navCount++;
+
     sseService._initListeners();
 
     // Start React
     render(
         <Provider store={store}>
-            <Router history={history}>{ makeRoutes(routes) }</Router>
+            <Router history={history}><Route onChange={onRouteChange}>{ makeRoutes(routes) }</Route></Router>
```
scherler: can`t we move `onChange` into the parent?

scherler:
BTW See JENKINS-XXXXX. is not helping

blueocean-dashboard/src/main/js/PipelineRoutes.jsx
kzantow: It fails providing the `NotFound` function to the router for some reason, it needs to be a class.

blueocean-dashboard/src/main/js/components/RunDetails.jsx
kzantow: It's a fail-safe. There was a bug that was causing this code to fail because the browser seems to be capping its actual history to 50, so when the page was first entered it was initially 50, then if the user clicked any of the tabs to navigate, it was still 50 and it would only go back 1 to an incorrect url. This is why I added tracking at the topmost router level. It's possible that could miss navigations, so this fail safe restores the correct url in the case it's not right.

blueocean-web/src/main/js/main.jsx
kzantow: Doesn't seem to support the `onChange` on the `Router` element, only a `Route`

blueocean-dashboard/src/main/js/components/RunDetails.jsx
kzantow: FYI the result of the incorrect URL caused excessive full page reloads when doing anything with any tabs after the browser history got too large.

cliffmeyers: I know some of the code here was to handle a couple of bugs where closing run details didn't update the route correctly, resulting in the UI not actually re-rendering and displaying the cloned UI underneath (which caused that full page refresh after click). We just want to keep an eye on ATH, and maybe see if there's anything we could add there that would give us more confidence in this change.

blueocean-dashboard/src/main/js/components/RunDetailsArtifacts.jsx
```diff
@@ -6,9 +6,10 @@ import mobxUtils from 'mobx-utils';
 import { logging, UrlConfig } from '@jenkins-cd/blueocean-core-js';
 
 const logger = logging.logger('io.jenkins.blueocean.dashboard.artifacts');
+const t = require('@jenkins-cd/blueocean-core-js').i18nTranslator('blueocean-dashboard');
```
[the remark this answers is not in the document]
kzantow: copy/paste :)

blueocean-dashboard/src/main/js/components/RunDetails.jsx
```diff
@@ -178,27 +186,20 @@ class RunDetails extends Component {
...
-                { t('rundetail.header.tab.tests', { defaultValue: 'Tests' }) }
-                { testsBadge }
-            </TabLink>,
-            <TabLink to="/artifacts" { ...base }>{ t('rundetail.header.tab.artifacts', {
-                defaultValue: 'Artifacts',
-            }) }</TabLink>,
         ];
 
+        actions.map(action => {
+            const badgeText = action.getBadgeText && action.getBadgeText(currentRun);
```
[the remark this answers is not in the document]
kzantow: sure, would be fine to do

michaelneale:
Good progress... 

```
dashboard/src/main/js/components/RunDetailsTests.jsx
[INFO]   49:57  error  't' is missing in props validation  react/prop-types
```

michaelneale:
@scherler I have no idea what you are talking about ;) https://issues.jenkins-ci.org/browse/JENKINS-47929 (you should also police the 200 other repos I see, where no one bothers to make tickets for things that really could do with a ticket... and kind of core jenkins things too).

michaelneale:
failed to load the executor plugin  ATH failure - which means something is amiss here...

PR #1545, opened 2017-11-10
Trying newer versions of declarative and favorite

michaelneale:
# Description

Just bumping some versions to be newer to see if tests still pass

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -330,7 +330,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.20</version>
+            <version>2.22</version>
         </dependency>
 
         <!--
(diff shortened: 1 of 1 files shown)


michaelneale:
cc @vivek for the PCT

michaelneale:
@vivek interesting failure - so it won't allow those versions without some others being juggled? (so juggling some more...)

vivek:
@michaelneale we don't need to upgrade the declarative and workflow apis. The specific failing test PipelineMetadataServiceTest.declarativeAgents() was fixed in 1.3.1, see https://github.com/jenkinsci/blueocean-plugin/pull/1496.

michaelneale:
@Vivek ok what about the autofav failure?

vivek:
Looking at other test failure `io.jenkins.blueocean.autofavorite.FavoritingScmListenerTest.testAutoFavoriteForRegisteredUser` needs to be fixed in blueocean-autofavorite-plugin not blueocean. So I don't think we need this PR.

michaelneale:
@vivek yes this PR was just to try things. Why would it suddenly start failing? (It has not changed for months). I bumped the fav plugin as I wondered if it was it.

vivek:
@michaelneale declarativeAgent test failure was reported by @alvarolobato and fixed in blueocean 1.3.1. Other test failure got nothing to do with blueocean, its in different plugin. Perhaps the test is failing with upgrade of autofavorite plugin. I am checking whats going on in that plugin.

PR #1549, opened 2017-11-13
some version changes for  1.3

michaelneale:
Test out some minor upgrades to the latest plugins

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -315,7 +315,7 @@
         <dependency>
             <groupId>org.jenkinsci.plugins</groupId>
             <artifactId>pipeline-model-definition</artifactId>
-            <version>1.2</version>
+            <version>1.2.2</version>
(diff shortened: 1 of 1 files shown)


michaelneale:
@vivek am bumping dependencies to cope with the constraints, and what is in URR - but this means bumping JVM version, so this may be a no-no for 1.3.3 I think? A bit much churn for a point release. 

cc @i386

michaelneale:
@vivek not sure what I aimed to do here, but lots of test *errors* now... in any case, I am not sure it is that useful to do this here... just wanted to check what versions may work together.

vivek:
This needs upgrading 1.3 to JDK8, that is cherry picking #1479 and then upgrade the version. Not sure we want to do for 1.3 though as it upgrades many dependencies in point release version.

PR #1551, opened 2017-11-13
Task/jenkins 36641 request builder replacement

cliffmeyers:
# Description
- Per discussion with @vivek  rather than refactoring the `BaseTest.RequestBuilder` utility into a shared module, it made sense to write a similar utility with some improvements he had in mind.
- **NOTE: I will add Javadoc once this API is finalized**

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
+++ b/blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
@@ -26,7 +26,34 @@ import java.util.Map;
 import static io.jenkins.blueocean.util.HttpRequest.Method.GET;
 
 /**
- * TODO: needs javadoc once API is final
+ * Utility for making requests (usually in the context of JUnit tests).
+ * There are two key methods:
+ * - One of the "HTTP method" methods (Get, Post, etc) which support uri template syntax,
+ * - One of the "as" methods (as(Void.class), as(Map.class), asText(), etc) to execute.
+ *
+ * Sample usages:
+ *
+ * Map org = new HttpRequest("http://localhost:8080/jenkins/blue/rest")
+ *     .Get("/organizations/{org}")
+ *     .urlPart("org", "jenkins")
+ *     .as(Map.class);
+ *
+ * new HttpRequest("...")
+ *     .Post("/organizations/jenkins/pipelines")
+ *     .auth("username", "password")
+ *     .header("X-Foo", "Bar")
+ *     .bodyJson(ImmutableMap.of("name", "my-pipeline"))
+ *     .as(Void.class);
+ *
+ * Map error = new HttpRequest("...")
+ *      .Post("organizations/jenkins/pipelines")
+ *      .status(401)
+ *      .as(Map.class)
+ *
+ * To use it, add "blueocean-rest-impl" as a dep with <type>test-jar</type>
+ * then add fluent-hc, handy-uri-templates and stapler as deps.
+ *
+ * @see <a href="https://github.com/damnhandy/Handy-URI-Templates">https://github.com/damnhandy/Handy-URI-Templates</a>
  * @author cliffmeyers
  */
 public class HttpRequest {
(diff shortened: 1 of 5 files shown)


blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
```diff
@@ -0,0 +1,225 @@
...
+
+import static io.jenkins.blueocean.util.HttpRequest.Method.GET;
+
+/**
+ * TODO: needs javadoc once API is final
+ * @author cliffmeyers
+ */
+public class HttpRequest {
+
+    private Logger logger = Logger.getLogger(getClass());
```
vivek: `private static final ...`

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
```diff
@@ -0,0 +1,225 @@
...
+    private static String JSON = "application/json";
```
vivek: `private static final ...`

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
```diff
@@ -0,0 +1,225 @@
...
+
+    public byte[] asBytes() throws IOException {
+        return executeInternal().asBytes();
+    }
+
+    public void execute() throws IOException {
+        executeInternal();
+    }
+
+    public Content executeInternal() throws IOException {
```
vivek: private?

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
```diff
@@ -0,0 +1,225 @@
...
+        builder.baseUrl(baseUrl);
+        return builder;
+    }
+
+    public HttpRequest baseUrl(@Nonnull String baseUrl) {
+        this.baseUrl = baseUrl;
+        return this;
+    }
+
+    public HttpRequest GET(String url) {
```
vivek: Might be easier API to expose methods (HTTP methods) as public static and then remove `build` method and move the code that was there in to `execute`? 

Something along the line of:
```
Get("http://localhost:8080/jenkins/blue/rest/organizations/).auth("","").expectedStatus(200).as(Map.class).execute();
```

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
```diff
@@ -0,0 +1,225 @@
...
+        this.method = GET;
+        this.requestUrl = url;
+        return this;
+    }
+
+    public HttpRequest POST(String url) {
```
vivek: Capitalize or all caps?

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
```diff
@@ -0,0 +1,225 @@
...
+        }
+        return this;
+    }
+
+    public HttpRequest status(int statusCode) {
+        this.expectedStatus = statusCode;
+        return this;
+    }
+
+    public <T> T asObject(Class<T> clazz) throws IOException {
```
vivek: Its using template class, so maybe call it just `as`?

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
cliffmeyers: Fixed in 3f13a8a

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
cliffmeyers: Fixed in 3f13a8a

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
cliffmeyers: Fixed ffc6c0d

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
cliffmeyers: Good catch, silly mistake, fixed in 360cfd8

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
cliffmeyers: I think it's good to leave this name longer so it's more easily differentiated between `asText`, `asBytes` and `asInputStream` methods.

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
cliffmeyers: That's similar style as to Apache Fluent, however having the entry point me the static "method" methods require us to specify the absolute URL each time. I'd argue that in almost every test class, there is a baseUrl that is the same and specifying it in each test method (even with uri templates) clutter things up. e.g.

Get("__http://localhost:8080/jenkins/blue/rest__ /organizations/jenkins")
vs
Get("__{base}__ /organizations/jenkins") __.urlPart("base", baseUrl)__
vs just
Get("/organizations/jenkins")

In 5fcba06 I made a small change so entrypoint is either static `build()` or `build(String baseUrl)` methods, so if you prefer the first style you just need `build()` before the `Get(...)`. If you feel strongly we should go the other way let me know and we can discuss.

cliffmeyers:
@vivek right now we have `HttpRequest` in the `blueocean-rest-impl` test jar. Downside to this is for project that needs to use `HttpRequest` (such as ATH) that is doesn't inherit transitive dependencies for the test jar. So I need to add `fluent-hc`, `stapler` and `handy-uri-templates` as extra dependencies. To me that's not ideal, do you have any other ideas?

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/util/HttpRequest.java
vivek: Except its not Object. Its generic typed return where as others are specific type return like String or InputStream. Fact that it doesn't return Object so calling it asObject() is confusing.

cliffmeyers: It returns an instance that inherits from Object, no? ;)  I see your point. Would `asType` be better? I just think a method name of `as` isn't great, but maybe it's fine since it always includes a type as an arg?

vivek:
@cliffmeyers LGTM 🐝 ponding successful CI run.

PR #1552, opened 2017-11-15
attempt to repair a known flaky test

michaelneale:
@vivek instead of waiting for a string in the log, can just wait for it to be running right?


michaelneale:
@vivek this is one of the failures we saw in the PCT

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/AbstractRunImplTest.java
```diff
@@ -159,16 +159,19 @@ public void testArtifactZipFileLink() throws Exception {
...
 
         Run r = p.scheduleBuild2(0).waitForStart();
 
         String url = "/organizations/jenkins/pipelines/project/runs/" + r.getId() + "/";
         Map m = request().get(url).build(Map.class);
 
-        // Wait 'til we're out of queue and actually on the node.
-        j.waitForMessage("Running on master", r);
+        // Wait until we are up and running before continuing with the test. 
+        while (!m.get("state").equals("RUNNING")) {
```
vivek: Perhaps add a timeout? Who knows it can wait for ever due to some server error?  Also breaking while loop if state is FINISHED is good too. its possible in 1 sec sleep state transitioned to FINISHED, in that case it can keep running for ever. `while (!m.get("state").equals("RUNNING") && !m.get("state").equals("FINISHED"))`

michaelneale: @vivek there is a timeout on the unit test via annotation - is that not enough to fail it?

vivek:
Ah, looks like I missed the timeout via annotation. It’s all good for timeout. However I think it should check for FINISHED state as well in he first while loop for better test resiliency.

michaelneale:
@vivek if it finds it as "finished" state in first time run - should it just skip out of the checking running bit? the original ticket was for running state, so if it skips to finished quickly, we can't really run the test can we? ie it would pass but he kind of a not useful pass? (mabe I miss the point of the test, but that is how I read the linked `@Issue` thing).

vivek:
@michaelneale you are right. I think we should only keep second while loop and check for assertions if state is running. Something like this:

```java
        String url = "/organizations/jenkins/pipelines/project/runs/" + r.getId() + "/";
        Map m = request().get(url).build(Map.class);
        while (!"FINISHED".equals(m.get("state").toString())) {
            if("RUNNING".equals(m.get("state").toString()){
                Assert.assertEquals("RUNNING", m.get("state"));
                Assert.assertEquals("UNKNOWN", m.get("result"));
            }
            Thread.sleep(1000);
            m = request().get(url).build(Map.class);
        }
```

michaelneale:
@vivek review again?

vivek:
@michaelneale LGTM 🐝

PR #1553, opened 2017-11-15
Favoriting a multi-configuration job breaks blueocean favourites

sophistifunk:
# Description

See [JENKINS-47835](https://issues.jenkins-ci.org/browse/JENKINS-47835).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/util/Capabilities.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/util/Capabilities.java
@@ -4,6 +4,7 @@ import hudson.model.Item;
 import io.jenkins.blueocean.rest.annotation.Capability;
 
 import javax.annotation.Nonnull;
+import java.util.*;
 
 /**
  * Helpers to query @Capability annotations server-side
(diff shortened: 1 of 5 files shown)


michaelneale:
very neat code @sophistifunk looks good

code review @imeredith @vivek ?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/util/Capabilities.java
```diff
@@ -0,0 +1,47 @@
...
+     *
+     * @param item a Jenkins Item / BO object
+     * @param capability the string name of the capability you wish to search for
+     * @return true if annotation is found
+     */
+    public static boolean hasCapability(Object item, String capability) {
+        return item != null && capability != null && Capabilities.hasCapability(item.getClass(), capability);
+    }
+
+    private static boolean hasCapability(@Nonnull Class<?> clazz, @Nonnull String capability) {
```
imeredith: Imo this should be public. Its oossible that the caller only has the class instance and not the object (although unlikely)

michaelneale:
the failure doesn't seem related to this change. It is to do with editing and parallel: 



(screen shot from the artifacts tab)

running it again...

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/util/Capabilities.java
vivek: +1

michaelneale:
:bee:

PR #1561, opened 2017-11-19
JENKINS-46740 - Add link to latest run

tinexw:
# Description

See [JENKINS-46740](https://issues.jenkins-ci.org/browse/JENKINS-46740).

This is a first draft, please check if the general approach is ok. Missing:
- Tests
- Maybe a link on the pipeline page?

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


michaelneale:
thanks for this - how is existing test coverage for this? 

We may want to add an ATH test to cover this or two... 
the main concern is the encoding of URIs (they seem to bite all the time... hardest problem in com sci ;)

@cliffmeyers @kzantow you want to take a look at this code review wise?

blueocean-core-js/src/js/services/ActivityService.js
```diff
@@ -72,7 +72,7 @@ export class ActivityService extends BunkerService {
                 // Should really have dedupe on methods like these, but for now
                 // just clone data so that we dont modify other instances.
                 const run = utils.clone(data);
-                return this.setItem(run);
+                return this.setItemWithKey(href, run);
```
imeredith: why is this needed? shouldnt the runh ave the self href in it?

michaelneale: yes I did want to know why this is here. 

Maybe @cliffmeyers 

@tinexw does this fix some latent bug you saw?

i386:
@tinexw can we get a screenshot of the change please?

michaelneale:
@i386 there is no visual change yet - it just allows /latestRun to work - if one was to link to it ... so you don't need a run number...

i386:
@michaelneale ahh I though there was a visual change... never mind :)

michaelneale:
hrm - I can confirm this works with a url like: 

`http://localhost:8080/jenkins/blue/organizations/jenkins/app-store-demo/detail/master/latestRun/pipeline/`

however, if I run a new build of the master branch, it doesn't refresh to show the actual latest one - is that by design? 

The use case is for keeping it open on a monitor, but it would need to refresh itself to meet that ? also when it does pick it up and follow along, the header (for some reason) never switches to the finished state (may be related)

So looks like a good start, but the auto refreshing I think needs to work. Maybe @cliffmeyers or @kzantow has some ideas on how to make it happen? 
This is a very cool feature, would be nice to make it work, as it would help a bunch of people out

blueocean-core-js/src/js/services/ActivityService.js
tinexw: It does have the self href but it points to the actual run url, not the latest url.
E.g. in the example from the JIRA ticket ```https://ci.blueocean.io/blue/rest/organizations/jenkins/pipelines/ATH-Jenkinsfile/branches/master/latestRun/```contains 

```
_links: {
    self: {
        _class: "io.jenkins.blueocean.rest.hal.Link",
        href: "/blue/rest/organizations/jenkins/pipelines/ATH-Jenkinsfile/branches/master/runs/2433/"
},
```

The latest URL is used to as key to fetch and get the latest run from the activity service (in ``RunDetails.jsx``) but it is stored internally in the bunker service with the run id URL. Thus, it is not found from the run details.

Another option would be to change the backend response - I briefly looked into it but it seemed more complicated and kind of wrong anyway.

tinexw:
Thanks for the quick feedback. I will look into the tests next. @michaelneale Should I cover the encoding stuff with ATH or unit tests? Could you give me some pointers where I find similar tests regarding the encoding?

tinexw:
Regarding the refresh: I agree that it would be vey nice to have. But maybe it can be done in a separate PR? At least in my team this change would already help even without auto refresh because we can install a browser extension that just reloads the page (not perfect but better than before)
In any case: I would be willing to implement it but I would need some help as to where and how to implement it.

michaelneale:
@tinexw ok - yes could be another PR. 

When you use this does it show the completed header? 

As for tests - there is a multi branch test case (in java) or some js tests in the “log-karaoke” that this probably could be tacked on too. I’ll take a look tomorrow if you don’t get a chance, should be pretty easy. 

Give the scope a test that just verifies that the lastestRun loads successfully woikd be fine

blueocean-core-js/src/js/paths/rest.js
```diff
@@ -26,6 +26,14 @@ export default {
         return `${this.apiRoot()}/organizations/${encodeURIComponent(organization)}/pipelines/${pipeline}/runs/${branchStr}`;
     },
 
+    latestRun({ organization, pipeline, branch}) {
```
kzantow: NIT: space after `branch` surprised this didn't fail lint

blueocean-core-js/src/js/services/ActivityService.js
kzantow: We shouldn't do this, why not just redirect to the actual run url and avoid this bug altogether?

kzantow: I'd much rather not introduce known bugs.

tinexw: What known bug do you mean? 
I thought about just doing a redirect too but then wouldn't you show the redirected URL in the browser? How would the refresh work?

blueocean-core-js/src/js/paths/rest.js
tinexw: Maybe it did? I didn't look into tests, linting etc yet. Just wanted to get first feedback if the approach was ok or not.

tinexw:
> Either creating a REST call or redirecting to the actual latest run url seems like a way to avoid the known bug that the latest run is never refreshed.

@kzantow  Could you point me to the known bug? I don't get the suggestion to create a REST call without knowing what the bug is about.

cliffmeyers:
I think it's worth discussing what the desired end state of the code should be to support "auto refresh" before we land this PR, just so we're sure that we don't need extensive changes to make it work.

We need some kind of clear flag to the UI that the screen is in "auto refresh" or "latest" mode, otherwise we risk introducing bugs where Run Details begins auto refreshing itself when the user doesn't want it. We can probably do this in two ways:
- A: URL like:  `/organizations/jenkins/mypipeline/detail/mybranch/latest` where the `runId` route param uses a special value like "latest" which unambiguously puts the screen into auto refresh mode, or
- B: URL like:  `/organizations/jenkins/mypipeline/detail/mybranch/10?latestRun=true` which does the same.

Once the auto refresh mode is set the UI needs it needs to be able to establish a SSE listener to be informed when a new run starts, which then either (A) refreshes the existing UI with new data, or (B) pushes the user to the new URL.

It would be good if we could introduce this behavior by wrapping the existing `RunDetails` component tree in a new component which handles the listening and data refresh or redirect.

I think B would be the less obtrusive change since the RunDetails component tree won't have to account for a special value of `runId`, and the redirect would be pretty similar to how the "Replay" button works today.

kzantow:
REST call to get the latest run and not caching it in the bunker. easier just to create a redirect. If you don't do a `window.location`, but instead use the `router.replace(url)` or similar method, it will not refresh the browser and it will find the run cached by the actual href instead of `latestRun` if you just put it in the bunker normally.

cliffmeyers:
Yeah, I'm in agreement with @kzantow 's last point as doing a UI redirect should make this work without any changes in the data layer at all.

It's possible you could get the notification of the new run by some `@computed` magic in MobX, but I think using a SSE listener is a little less "magical" and easier to understand... just make sure to clean up listeners.

blueocean-core-js/src/js/services/ActivityService.js
kzantow: known bug that the bunker has a `latestRun` cached and won't re-fetch it?

tinexw:
Ok, thanks for the feedback. I'm mostly doing Java backend development and have never used react and hardly any JavaScript so I cannot promise that I will be able to do the changes soon... But I will give it a shot.

blueocean-core-js/src/js/services/ActivityService.js
tinexw: Ok, didn't realize this. I only tested manually that the newest run is displayed when I refreshed the latest run page and started a new run in between.

tinexw: @kzantow Maybe just to clarify: What I described above is a bug that would be there **without** the changes to the ActivityService and BunkerService. With these changes the bug does not exist. I just tried to explain why I made those changes i.e. to prevent this bug from happening.

tinexw:
Closing this for now because it needs a different approach. I will create a new PR if I manage to implement it ;-)

cliffmeyers:
@tinexw feel free to reach out to me on Twitter if you'd like any pointers on getting started on the UI side. Appreciate your efforts in contributing!

michaelneale:
@kzantow @cliffmeyers wouldn't the redirect prevent any future effort to make it auto refresh for new runs?

tinexw:
@michaelneale Just to answer your question

> When you use this does it show the completed header?

No, only after a refresh. Good catch :-) It would require additional changes in `DefaultSSEHandler` The run id URL is used as key in `updateJob`, `queueCancel` etc. (Just for testing I changed it to the lastest Run URL in `updateJob` and used this the href as key in the activity service in `_updateRun` and the header was updated correctly right after the job finished successfully.

cliffmeyers:
@michaelneale it shouldn't. My intent was that the redirect to a new `RunDetails` URL would include  the `latestRun=true` query param so that it would refresh perpetually.

tinexw:
@michaelneale @cliffmeyers I personally don't like version B with the URL like `/organizations/jenkins/mypipeline/detail/mybranch/10?latestRun=true`

If I set up a new monitor and have to enter the URL for the first time I would like to enter a fixed URL (like in version A) and not open a run and add the query parameter.

I would prefer to have version A with a URL like: `/organizations/jenkins/mypipeline/detail/mybranch/latest` that internally redirects to the run id but keeps showing the same URL in the browser (if that is possible)

kzantow:
@tinexw @cliffmeyers @michaelneale yeah, I agree with @tinexw that if we are going to show a latest run page that continually updates, it should just be `/organizations/jenkins/mypipeline/detail/mybranch/latest-run` or something similar, having a run id is wrong

cliffmeyers:
@tinexw @kzantow that's a fair point in terms usability. Certainly either of the two URL schemes is possible. 

Just be aware there will be some more involved refactoring to get the Run Details component tree to play nicely with a non-numeric `runId` value. A lot of the downstream components probably assume the `runId` won't change after they've mounted so they may not behave properly without non-trivial changes. For example if the user selects the "Tests" tab, and later the "latest run" changes, the component may not refresh properly by default and would instead show stale data from the prior run. You have to shift logic from `componentWill/DidMount` methods to `componentWillReceiveProps` instead.

You might be able simplify this by adding a`mybranch/latest-run` route which does an internal redirect to `mybranch/10?refresh=true` when then advances the `runId` - kind of a hybrid approach of the two we've discussed.

michaelneale:
hey @tinexw sounds like this was close - feel free to re-open and add to this vs throwing away. 

I think if this is focused only on /latestRun, then perhaps we don't need to have it too broad, if it is going to be a URL only used by info radiators...

tinexw:
@michaelneale @cliffmeyers @kzantow Hi, I won't be able to work on this in the near future. I was hoping to find some time at the end of last year but I didn't :-( Sorry!

PR #1570, opened 2017-11-29
Jenkins-47649&48268 fix activity and branch rows ellipsis issues and clicking areas

NicuPascu:
# Description

See [JENKINS-47649](https://issues.jenkins-ci.org/browse/JENKINS-47649), [JENKINS-48268](https://issues.jenkins-ci.org/browse/JENKINS-48268).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/LinkifiedText.jsx
+++ b/blueocean-dashboard/src/main/js/components/LinkifiedText.jsx
@@ -18,12 +18,14 @@ const LinkifiedText = (props) => {
             const partialLinksRegExpString = new RegExp(`(${partialLinksIdString.slice(0, -1)})`, 'gi');
 
             for (let commitMsgPart of text.split(partialLinksRegExpString)) {
+                commitMsgPart = commitMsgPart.replace(/ /g, '\u00A0');
+
                 if (partialLinksObj[commitMsgPart]) {
-                    textWithPartialLinks.push(<a href={partialLinksObj[commitMsgPart]} target="_blank">{commitMsgPart}</a>);
+                    textWithPartialLinks.push(<a href={partialLinksObj[commitMsgPart]} target="_blank"><span>{commitMsgPart}</span></a>);
                 } else {
                     if (commitMsgPart) {
                         if (textLink) {
-                            textWithPartialLinks.push((<Link to={textLink} className="unstyled-link" >{commitMsgPart}</Link>));
+                            textWithPartialLinks.push((<Link to={textLink} className="unstyled-link" ><span>{commitMsgPart}</span></Link>));
                         } else {
                             textWithPartialLinks.push(commitMsgPart);
                         }
@@ -33,13 +35,13 @@ const LinkifiedText = (props) => {
         }
 
         if (!textWithPartialLinks.length) {
-            textWithPartialLinks.push((<Link to={textLink} className="unstyled-link" >{text}</Link>));
+            textWithPartialLinks.push((<Link to={textLink} className="unstyled-link" ><span>{text}</span></Link>));
         }
 
         return (<span>{textWithPartialLinks}</span>);
     }
 
-    return (textLink ? (<Link to={textLink} className="unstyled-link" >{text}</Link>) : (<span>{text}</span>));
+    return (textLink ? (<Link to={textLink} className="unstyled-link" ><span>{text}</span></Link>) : (<span>{text}</span>));
 };
 
 LinkifiedText.propTypes = {
--- a/blueocean-dashboard/src/main/js/components/LinkifiedText.jsx
+++ b/blueocean-dashboard/src/main/js/components/LinkifiedText.jsx
@@ -18,14 +18,14 @@ const LinkifiedText = (props) => {
             const partialLinksRegExpString = new RegExp(`(${partialLinksIdString.slice(0, -1)})`, 'gi');
 
             for (let commitMsgPart of text.split(partialLinksRegExpString)) {
-                commitMsgPart = commitMsgPart.replace(/ /g, '\u00A0');
+                commitMsgPart = commitMsgPart.replace(/ /g, '\u00A0'); //replace empty space chars with &nbsp;
 
                 if (partialLinksObj[commitMsgPart]) {
-                    textWithPartialLinks.push(<a href={partialLinksObj[commitMsgPart]} target="_blank"><span>{commitMsgPart}</span></a>);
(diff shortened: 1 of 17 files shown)


michaelneale:
ping @sophistifunk review please? 

@kzantow @scherler any changes that steel may care about: https://github.com/jenkinsci/blueocean-plugin/pull/1570/files#diff-b3504b456d78305945152d66f089d4a8L108 ?

sophistifunk:
I don't really have anything to add other than we need to be careful with doing things like this, it seems like it's likely to break ellipses somewhere else in tables. Every 2 months we see that some different thing has had its text-collapse broken and we make another change like this, adding or removing spans and divs and making the styles more complex and it fixes the issue but turns out it broke something else.

I have no idea what we can do to fix the basic issue, I'm just bitching I guess. But I worry that we're turning our styles into an abominable mess over time, and I'm buggered if I can think of a good way to test it.

We could code up some renderings using components from Dashboard that don't need a server, but that's a huge amount of work given how many magic globals that are used all over the place for convenience which make assumptions about the environment they're in :(

So... for this PR, I say 🐝

blueocean-dashboard/src/main/js/components/LinkifiedText.jsx
```diff
@@ -18,12 +18,14 @@ const LinkifiedText = (props) => {
             const partialLinksRegExpString = new RegExp(`(${partialLinksIdString.slice(0, -1)})`, 'gi');
 
             for (let commitMsgPart of text.split(partialLinksRegExpString)) {
+                commitMsgPart = commitMsgPart.replace(/ /g, '\u00A0');
```
cliffmeyers: Could you please add a comment as to what this unicode character is? Or just assign it to a variable so it self-documents.

blueocean-dashboard/src/main/less/run-message-cell.less
```diff
@@ -1,12 +1,45 @@
...
+                flex-grow: 1;
+                white-space: nowrap;
+                text-overflow: ellipsis;
+                overflow: hidden;
+
+                &:not(:first-child) > span{
+                    width: 10px;
+                }
+
+                > span {
```
cliffmeyers: Seems like rather than applying this rule to the `span` we should use a semantic className here so the rule reads better, and so it's clearer in the JSX why we are wrapping the text values in the first place. Something like `.ellipsis-text` would probably be fine?

blueocean-dashboard/src/main/less/run-message-cell.less
```diff
@@ -1,12 +1,45 @@
...
+                    display: table-cell;
+                    flex-grow: 1;
+                    vertical-align: middle;
+                    height: auto;
+                    width: 300px;
```
cliffmeyers: Can you explain why this has width=300 on it? Is that the default width we apply to the Message column?

jenkins-design-language/less/components/jtable.less
```diff
@@ -105,10 +105,20 @@ a.JTable-row--href {
...
+    display: flex;
+    align-items: center;
+    width: 100%;
+    height: 100%;
     white-space: nowrap;
     overflow: hidden;
     text-overflow: ellipsis;
+
+    > code,
+    .text-with-ellipsis-container {
```
cliffmeyers: We are repeating this sequence of CSS rules in a bunch of places:

```
white-space: nowrap;
overflow: hidden;
text-overflow: ellipsis;
```

Perhaps we should make this a discrete and more generic className like `.ellipsis` - the name is brief, but I think the semantics are clear. Then we can apply it where needed, and also use it as a mixin in the other less rules so they are a bit more concise.

blueocean-dashboard/src/main/less/run-message-cell.less
NicuPascu: in this case, the actual value of the width property does not matter as long as it is smaller than the width of the container, it can be set to "width: 0px" and it would still work the same, flex-grow: 1 makes the width as large as possible. The width needs to be set to a value for text-overflow: ellipsis to work. I have set it to 300px because of a quirk in IE11 in which in some cases the text rendered correctly but the ellipsis is rendered where the width would have finished.

cliffmeyers: Cool. Adding a comment to that effect would be good. Whenever I need style rules for a browser-specific bug I usually add it, to discourage anyone from accidentally removing it later.

jenkins-design-language/less/components/jtable.less
```diff
@@ -105,10 +105,16 @@ a.JTable-row--href {
...
-    white-space: nowrap;
-    overflow: hidden;
-    text-overflow: ellipsis;
+    display: flex;
+    align-items: center;
+    width: 100%;
+    height: 100%;
+    .text-overflow();
+
+    > code,
```
sophistifunk: Which component means we need `> code` here? Won't this complicate things if we want to put wrapping blocks of logs in a table some time in the future?

NicuPascu: That is for CommitID (used in the commit column), I will add an extra class there so the css rule is not using the tag name

NicuPascu: I didn't actually need to make another class, I just used the one I use in the message column so everything should be ok now

blueocean-dashboard/src/main/less/mixins/text-overflow.less
```diff
@@ -1,5 +1,6 @@
 // Text overflow
 // Requires inline-block or block for proper styling
+// Works with flexbox as well as long as the min-width property is set to something else than the default 100% that is set for flexbox
```
cliffmeyers: In retrospect, I should have named this differently, perhaps `overflow-ellipsis`. I can't insist you rename something that I myself named poorly, but if you want to obsess over names like I do, go ahead and change it :)

jenkins-design-language/src/js/components/CommitId.jsx
```diff
@@ -9,8 +9,14 @@ const GIT_HASH_REGEX = /\b[0-9a-f]{5,40}\b/;
  */
 export class CommitId extends Component {
     render() {
-        const {commitId, url, title} = this.props;
+        const {className, commitId, url, title} = this.props;
         let displayValue;
+        const classNames = ['hash'];
+
+        if (className) {
```
cliffmeyers: Nice, this guard ensures omitting `className` prop won't fail. I think we have unit tests that verify that. 👍

michaelneale:
@NicuPascu this one ready to go in?

NicuPascu:
@michaelneale I'll just change the name of that mixin as @cliffmeyers suggested, wait for it to go green again and I'll merge it

PR #1571, opened 2017-11-29
[JENKINS-48285] - Make UserImplPermissionTest work with >= 2.77 cores

alvarolobato:
# Description

See [JENKINS-48285](https://issues.jenkins-ci.org/browse/JENKINS-48285).

So this change ([PR ](https://github.com/jenkinsci/jenkins/pull/2999)- [JENKINS-46610](https://issues.jenkins-ci.org/browse/JENKINS-46610)) moved all the hasPermission methods from several classes to a default method in the interface. In the test the method wasn't mocked because it was final and we didn't need to. Not the method is not final so it needs to be mocked.
I had to relay in a hackish check because pre 2.77 will fail the mock given that the method is final. 

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

@reviewbybees

### DIFF
--- a/blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/rest/UserImplPermissionTest.java
+++ b/blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/rest/UserImplPermissionTest.java
@@ -49,6 +49,9 @@ import org.kohsuke.stapler.StaplerRequest;
 import org.kohsuke.stapler.StaplerResponse;
 import org.kohsuke.stapler.WebMethod;
 import org.kohsuke.stapler.verb.DELETE;
+import org.mockito.Mockito;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
 import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.PowerMockRunner;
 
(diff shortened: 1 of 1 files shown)


blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/rest/UserImplPermissionTest.java
```diff
@@ -111,6 +115,21 @@ public boolean hasPermission(Authentication a, Permission permission) {
         when(Jenkins.getAuthentication()).thenReturn(authentication);
         when(Jenkins.getInstance()).thenReturn(jenkins);
 
+        try {
+            // After Jenkins 2.77 hasPermission is no longer in Node.class and is not final so we need to mock it
+            // prior to it is called as being final and mocking it will fail for the same reason.
+            // TODO remove after core base line is >= 2.77
+            Node.class.getDeclaredMethod("hasPermission", Permission.class);
+        } catch (NoSuchMethodException e) {
```
alvarolobato: Not proud of this but if we want it to work with the current core and the PCT we need to check it.

michaelneale: We've all done worse things than this ;) it is test code anyway - and lovingly documented, wonderful catch @alvarolobato !

michaelneale:
:bee: 

@alvarolobato thank you so much! this is amazing.

PR #1573, opened 2017-12-01
1.3 small fixes for compatibility and performance

michaelneale:
# Description

Cherry pick of fixes for unit tests for newer core, a flaky test that PCT picked up, and log performance improvements. 

https://github.com/jenkinsci/blueocean-plugin/pull/1521
https://github.com/jenkinsci/blueocean-plugin/pull/1552
https://github.com/jenkinsci/blueocean-plugin/pull/1571

### DIFF
--- a/blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/AbstractRunImplTest.java
+++ b/blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/AbstractRunImplTest.java
@@ -148,7 +148,7 @@ public class AbstractRunImplTest extends PipelineBaseTest {
 
     @Test(timeout = 20000)
     @Issue("JENKINS-44736")
-    public void earlyUnstableStatusShouldReportPunStateAsRunningAndResultAsUnknown() throws Exception {
+    public void earlyUnstableStatusShouldReportRunStateAsRunningAndResultAsUnknown() throws Exception {
         WorkflowJob p = j.createProject(WorkflowJob.class, "project");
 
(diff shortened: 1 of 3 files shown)


michaelneale:
a few diff fixes rolled up - in case we want to do a 1.3.4 release. 

can I have a review from say: 

@sophistifunk @kzantow @i386 as appropriate (most of the code is from master, but some had to be slightly rewritten if you diff it with master for the apis in use here)

michaelneale:
cc @AlanFoster too - if you could review your change made it safely in here for log improvements (so we can cut a release)

AlanFoster:
@michaelneale Thanks for looking into this; Running this locally everything seems good to me :+1:

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/AbstractRunImplTest.java
```diff
@@ -148,7 +148,7 @@ public void testArtifactZipFileLink() throws Exception {
 
     @Test(timeout = 20000)
     @Issue("JENKINS-44736")
-    public void earlyUnstableStatusShouldReportPunStateAsRunningAndResultAsUnknown() throws Exception {
```
sophistifunk: TODO: Dream up a flimsy reason we should have something called a `PunState`

sophistifunk:
Seems pretty straightforward to me 🐝

michaelneale:
@vivek  if ok with you - can we merge this and cut a 1.3.4 ?

vivek:
LGTM 🐝

PR #1575, opened 2017-12-04
Downstream job links (backend)

sophistifunk:
# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/DownstreamJobListener.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/DownstreamJobListener.java
@@ -1,21 +1,11 @@
 package io.jenkins.blueocean.listeners;
 
-import com.google.common.collect.ImmutableList;
 import hudson.Extension;
-import hudson.console.ModelHyperlinkNote;
-import hudson.model.Action;
-import hudson.model.Run;
-import hudson.model.TaskListener;
+import hudson.model.*;
 import hudson.model.listeners.RunListener;
-import org.jenkinsci.plugins.displayurlapi.DisplayURLProvider;
-import org.jenkinsci.plugins.workflow.steps.StepContext;
+import io.jenkins.blueocean.service.embedded.DownstreamJobAction;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.logging.Level;
 import java.util.logging.Logger;
-
-import org.jenkinsci.plugins.workflow.support.steps.build.BuildTriggerAction;
 
 @Extension
 public class DownstreamJobListener extends RunListener<Run<?, ?>> {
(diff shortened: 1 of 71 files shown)


sophistifunk:
This is just the backend part, I wanted to get it into PR first so I can seek some feedback, make sure I'm doing things in the right way. Any pointers on how to go about a meaningful test for `DownstreamJobListener` would be sweet, also 😁

i386:
@sophistifunk looks reasonable to me. 

Does the action persist across restarts? If not, you might have to call `run.save()` after you call the action.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/DownstreamJobAction.java
```diff
@@ -0,0 +1,71 @@
...
+ * Added to a job to provide details of downstream jobs triggered by it, as an inverse of Cause.UpstreamCause which
+ * is added to the downstream job during its creation.
+ */
+public class DownstreamJobAction implements Action {
+
+    private final String downstreamProject;
+    private final int downstreamBuild;
+
+    public DownstreamJobAction(Run<?, ?> run) {
+        this.downstreamProject = run.getParent().getName();
```
imeredith: I'm concerned this wont work with folders

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
```diff
@@ -257,6 +265,45 @@
         public abstract Object getCause();
     }
 
+    @ExportedBean
+    public static final class BlueDownstreamRun {
```
imeredith: I would like to see this have a `_links` field. It should link to the downstream project. With this we can use existing code to generate a blueocean url.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
```diff
@@ -257,6 +265,45 @@
...
+        private final String downstreamProject;
```
imeredith: Again this has nothing about folders I think?

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
sophistifunk: This was my initial intention, but `_links` is a map, so we could only have one "downstream" link. I talked it over with Mic and he agreed adding a specific property was the best way to go. I'm open to suggestions tho.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/DownstreamJobAction.java
sophistifunk: Any ideas on how I could verify this? How can I cook up a minimal folder project? I haven't messed with them.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
sophistifunk: I based this on the way the upstream causactions work, do they fail with folders?

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
sophistifunk: We could extend `_links` to allow list of urls? So long as we only send an array when there's multiple values we shouldn't break existing stuff. But then, we'd have to be checking for single values as well in cases where we're expecting to see multiples. What do you think, @i386 ?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/DownstreamJobAction.java
imeredith: Yes a backend test is what you need. Look at existing ones. Does getName return the path? i thought it was getFullName

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
imeredith: Yes, you have to use upstreamUrl

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
imeredith: no, in this case you are gonna have one _links field per job. So you dont n eed to have multiple links

sophistifunk: Yeah but there will be multiple downstream links for a single job, and the links field is a map

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/DownstreamJobAction.java
jglick: You can use `Item.getFullName`, or more simply a single field with `Run.getExternalizableId`.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/DefaultRunImpl.java
```diff
@@ -6,9 +6,13 @@
 import io.jenkins.blueocean.rest.factory.BlueRunFactory;
 import io.jenkins.blueocean.rest.model.BlueOrganization;
 import io.jenkins.blueocean.rest.model.BlueRun;
+import io.jenkins.blueocean.service.embedded.DownstreamJobAction;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 
+import java.util.ArrayList;
+import java.util.Collection;
```
jglick: Why?

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
```diff
@@ -257,6 +265,45 @@
...
+        private final int downstreamBuild;
+
+        public BlueDownstreamRun(String downstreamProject, int downstreamBuild) {
+            this.downstreamProject = downstreamProject;
+            this.downstreamBuild = downstreamBuild;
+        }
+
+        @Exported(name="downstreamProject")
```
jglick: Gratuitous `name`—this would be the default anyway.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
```diff
@@ -257,6 +265,45 @@
...
+        public String getDownstreamProject() {
+            return downstreamProject;
+        }
+
+        @Exported(name="getDownstreamBuild")
```
jglick: This `name` is probably wrong.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/DownstreamJobListener.java
```diff
@@ -0,0 +1,34 @@
...
+
+    @Override
+    public void onStarted(Run<?, ?> run, TaskListener listener) {
+        for (CauseAction action : run.getActions(CauseAction.class)) {
+            for (Cause cause : action.getCauses()) {
+                if (cause instanceof Cause.UpstreamCause) {
+                    Run triggerRun = ((Cause.UpstreamCause) cause).getUpstreamRun();
+                    if (triggerRun != null) {
+                        // Add a link from the upstream job to this newly spawned one, to be retrieved later
+                        triggerRun.addAction(new DownstreamJobAction(run));
```
jglick: So this is going to modify the upstream `build.xml`.

Not a problem exactly, but not what we would usually do to record Pipeline build metadata. The better approach would be for `pipeline-build-step` to record a `[FlowNode]Action` on the node corresponding to the `build` step, so that you have a record of exactly when this downstream build was started. (Currently it adds only a `LabelAction`.) Then you could retrieve a flat list of all downstream builds of a given upstream build, as in this PR, _or_ more detailed information about (say) downstream builds in a given branch.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
imeredith: I dont see that in this model?

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
sophistifunk: Yes, thanks

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/DownstreamJobListener.java
sophistifunk: I will look into this, thanks.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
sophistifunk: It's a collection: https://github.com/jenkinsci/blueocean-plugin/pull/1575/files#diff-9c0aa8614646845672acf40f503966b2R202

sophistifunk: I patterned it after the existing `causes` property

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRun.java
sophistifunk: It seems that way to me also, but it matches the rest of the code in that file so I'm going for consistency.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/DefaultRunImpl.java
sophistifunk: Just an oversight, cheers.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/DownstreamJobListener.java
abayer: I can get on that tomorrow, if you’d like.

abayer: Worth mentioning, though, that there is always a bit of a challenge in associating the upstream `build` with the actual downstream run, which may not start for a while (freestyle waiting on an agent, a project not allowing concurrent builds, etc...), which is part of why we’ve historically tracked the upstream/downstream relationship from the downstream run/job rather than from the upstream.

sophistifunk: @abayer that's why I took a similar approach to the downstream-build listener in `pipeline-build-step`. I think we want to continue doing it this way, but look at attaching the data to the flownode rather than the run itself, as Jesse suggested. Not really sure how I'll go about that yet, but that's my current thinking.

sophistifunk:
Git has shat the bed on this, I'm assuming it's something I did, so I'm killing it

michaelneale:
@sophistifunk has the work moved somewhere else?

sophistifunk:
I'm creating a new branch from master and moving my changes over a file at a time. It shouldn't take me long. My commit history on this branch is mostly cockups and uninteresting so it's no biggy.

michaelneale:
👏

sophistifunk:
Once I move the changes I'll redo my manual testing, so that'll take the longest, what with the compiles and such :)

sophistifunk:
New branch and PR: https://github.com/jenkinsci/blueocean-plugin/pull/1587

PR #1576, opened 2017-12-04
Bug/jenkins 48307 access token spaces

sophistifunk:
# Description

See [JENKINS-48307](https://issues.jenkins-ci.org/browse/JENKINS-48307).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/acceptance-tests/src/test/java/io/blueocean/ath/live/GithubEditorTest.java
+++ b/acceptance-tests/src/test/java/io/blueocean/ath/live/GithubEditorTest.java
@@ -136,6 +136,28 @@ public class GithubEditorTest {
     }
 
     /**
+     * This test covers e2e usage of the editor, with a token that has whitespace added.
+     *
+     * Creates a blank github repo, and then uses editor to create a simple pipeline.
+     */
+    @Test
+    public void testEditorWithSpace() throws IOException {
+        creationPage.createPipeline(" " + token + " ", organization, repo, true);
+        MultiBranchPipeline pipeline = mbpFactory.pipeline(repo);
+        editorPage.simplePipeline();
+        ActivityPage activityPage = pipeline.getActivityPage().checkUrl();
+        driver.navigate().refresh();
+        sseClient.untilEvents(pipeline.buildsFinished);
+        sseClient.clear();
+        BranchPage branchPage = activityPage.clickBranchTab();
+        branchPage.openEditor("master");
+        editorPage.saveBranch("new - branch");
+        activityPage.checkUrl();
+        activityPage.getRunRowForBranch("new-branch");
+        sseClient.untilEvents(pipeline.buildsFinished);
+    }
+
+    /**
      * This test covers e2e usage of the editor. Does so with a parallel pipeline.
      *
      * Creates a blank github repo, and then uses editor to create a parallel pipeline.
--- a/acceptance-tests/src/test/java/io/blueocean/ath/live/GithubEditorTest.java
+++ b/acceptance-tests/src/test/java/io/blueocean/ath/live/GithubEditorTest.java
@@ -1,6 +1,7 @@
 package io.blueocean.ath.live;
 
 import io.blueocean.ath.ATHJUnitRunner;
+import io.blueocean.ath.CustomJenkinsServer;
 import io.blueocean.ath.Login;
 import io.blueocean.ath.factory.MultiBranchPipelineFactory;
 import io.blueocean.ath.model.MultiBranchPipeline;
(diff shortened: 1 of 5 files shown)


sophistifunk:
This is currently broken, I'm trying to locate the ATH tests so I can add a case to prevent regressions.

sophistifunk:
Found the test, added a case, and borked my test-finding canary :D

sophistifunk:
@cliffmeyers if this looks like I've fixed it in the wrong place, or if you think I can shorten the new test in order to only test the actual fix without leaving anything in a wonky state, please let me know

acceptance-tests/src/test/java/io/blueocean/ath/live/GithubEditorTest.java
```diff
@@ -135,6 +135,28 @@ public void testEditor() throws IOException {
...
+        driver.navigate().refresh();
+        sseClient.untilEvents(pipeline.buildsFinished);
+        sseClient.clear();
+        BranchPage branchPage = activityPage.clickBranchTab();
+        branchPage.openEditor("master");
+        editorPage.saveBranch("new - branch");
+        activityPage.checkUrl();
+        activityPage.getRunRowForBranch("new-branch");
+        sseClient.untilEvents(pipeline.buildsFinished);
+    }
```
cliffmeyers: I think doing a full e2e creation + editor test just to validate the token trimming is overkill. Since the ATH can be flaky, I think we need to avoid creating overly long tests when we're really just testing one piece of functionality. I think we should add this as a test method on `GithubCreationTest`, the `testTokenValidation_failed` method would be good to emulate... we just need to feed in a token value padded with spaces and then assert that we moved on to the next step in the flow.

sophistifunk:
I don't think we should put the change on the backend, it seems like that's kind of unexpected behaviour in that arena. But as it's definitely a UI issue we would still want to *also* do it in the UI as well. I had a quick look at the other creation flows and it looks like they shouldn't suffer from it, as they're either actual passwords you'd type (or get from 1p), and the key entry textareas should be looking for magic markers that delineate the keys.

acceptance-tests/src/test/java/io/blueocean/ath/live/GithubEditorTest.java
cliffmeyers: Right on. `GithubCreationTest` has an `@After` method to delete the repo, and that will fire to do cleanup even if the test fails or errors out. Your test will create and save a valid token but I am pretty sure that `GithubCreationTest` and `GithubEditorTest` both have conditional logic in the underlying page object so that it can handle either a pristine creation flow w/ no token saved, or the flow when one is already saved. Arguably that's not the right way to do it but you shouldn't break anything. Famous last words :)

sophistifunk:
...Might be a good idea to also trim it at the API boundary of the backend tho, I'll have a look at the code.

sophistifunk:
Crap. I'll need to do more tweaking to this, I'll see about getting this ATH test to run locally.

cliffmeyers:
@sophistifunk you'll probably want to make sure the token is deleted before the test, otherwise it won't work consistently on successive runs locally (or if the tests are run in a different order in CI). `GithubCreationTest.testTokenValidation_failed` has a good example of this that should work.

sophistifunk:
@cliffmeyers that explains the bad pass. Thanks mate.

sophistifunk:
@cliffmeyers could you take another look at this mate? Cheers.

sophistifunk:
Yeah I can't see a good reason to commit npm-shrinkwrap but not package-lock?

cliffmeyers:
@sophistifunk I did a little research and discovered an interesting bit:

```
The most curious will also ask what happens when both of the files are present. In that case, NPM will completely ignore package-lock.json and just use npm-shrinkwrap.json. The situation should not happen however when manipulating the files just using NPM.
```

https://jpospisil.com/2017/06/02/understanding-lock-files-in-npm-5.html

That said, we haven't committed shrinkwrap in `acceptance-tests` but probably should just to avoid inconsistent behavior. Having only package-lock isn't great, as if you're using npm5 you will get a static dep tree but users of npm3/4 won't since it can't use package-lock.

In general I think it's a bad idea for our team to be mixing/matching usage of npm3/4 with npm5... just given track record of npm having unpredictable behavior resulting in subtle bugs that require days to fix. I realize keeping npm3/4 as the global lib on your system is a pain, and I guess you don't have to if you don't mind using something like `./node/npm/bin/npm` when you need to change deps.

At any rate, not making a big stink over this, but I think we should either (A) avoid mix/match of npm versions or (B) just bite the bullet and try to upgrade to npm5 or yarn if A is too much of hassle for folks.

sophistifunk:
nvm has never worked for me on this machine, so at some point I'll have to take a half-day to tear out all my node stuff and try again. I'll git remove the package-lock next time I make a change in ATH if nobody beats me to it tho, for consistency.

PR #1577, opened 2017-12-04
[JENKINS-41502] Link to upstream trigger pipelines

imeredith:
# Description
Links to upstream build if triggered, from the run header and activity message cell.

To test create 2 pipelines, one the triggers another.

See [JENKINS-41502](https://issues.jenkins-ci.org/browse/JENKINS-41502).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-dashboard/src/main/js/components/RunDetailsHeader.jsx
```diff
@@ -155,8 +155,21 @@ class RunDetailsHeader extends Component {
             </div>
         );
 
-        const causeMessage = (run && run.causes.length > 0 && run.causes[run.causes.length - 1].shortDescription) || null;
```
i386: @imeredith I think you should split this out into a `CauseCell` component that gets reused in both the header and within the `RunMessageCell`.

@kzantow causes have `_class` - should we be making the rendering for different causes an extension point?

cliffmeyers: Agree, we should absolutely not have teh same code in two places.

imeredith: Am doing that now

imeredith:
Completion of this is incoming.

michaelneale:
cc @alvarolobato - this is one of the things you brought up some time ago (there are more coming).

alvarolobato:
@michaelneale thanks!!!!

PR #1578, opened 2017-12-06
Send pipeline_step_used event

i386:
# Description

Has the following properties:
* `type` e.g. org.jenkinsci.plugins.workflow.steps.EchoStep
* `timesUsed` count of times executed in the pipeline
* `isDeclarative` used with a declarative Pipeline or not
* `runResult` the result of the job it was used with

See [JENKINS-48392](https://issues.jenkins-ci.org/browse/JENKINS-48392).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
@@ -92,10 +92,10 @@ public class PipelineStepImpl extends BluePipelineStep {
     }
 
     @Override
-    public String getAnalyticsType() {
+    public String getStepType() {
         if (node.getNode() instanceof StepAtomNode) {
             StepDescriptor descriptor = ((StepAtomNode) node.getNode()).getDescriptor();
-            return descriptor.getId();
+            if (descriptor != null) return descriptor.getId();
         }
         return "unknown";
     }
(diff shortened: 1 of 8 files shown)


blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
```diff
@@ -89,6 +91,15 @@ public String getType() {
         return node.getType().name();
     }
 
+    @Override
+    public String getAnalyticsType() {
+        if (node.getNode() instanceof StepAtomNode) {
```
abayer: I’d include `StepStartNode` in a perfect world, but we aren’t tracking those yet here in the first place.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BluePipelineStep.java
```diff
@@ -57,6 +57,11 @@
     @Exported(name = TYPE)
     public abstract String getType();
 
+    /**
+     * @return the type of step used for analytics purposes
+     */
+    public abstract String getAnalyticsType();
```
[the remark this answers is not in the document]
i386: Fixed

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
i386: @abayer I've updated this method to take into account `StepStartNode` as you have suggested.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BluePipelineStep.java
```diff
@@ -57,6 +57,11 @@
     @Exported(name = TYPE)
     public abstract String getType();
 
+    /**
+     * @return the type of step used for analytics purposes
```
vivek: Remove ‘used for analytics purposes’, doesn’t need to be here.

i386:
Replaced by https://github.com/jenkinsci/blueocean-plugin/pull/1579/

PR #1580, opened 2017-12-07
JENKINS-48392 record the name of the plugin used

i386:
# Description

* Record the name of the plugin used
* Use our own chunk visitor that has less overhead than the one Blue Ocean uses

See [JENKINS-48392](https://issues.jenkins-ci.org/browse/JENKINS-48392).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/analytics/PipelinePluginAnalytics.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/analytics/PipelinePluginAnalytics.java
@@ -2,6 +2,8 @@ package io.jenkins.blueocean.rest.impl.pipeline.analytics;
 
 import com.google.common.collect.ImmutableMap;
 import hudson.Extension;
+import hudson.PluginManager;
+import hudson.PluginWrapper;
 import hudson.model.Result;
 import hudson.model.TaskListener;
 import hudson.model.listeners.RunListener;
@@ -10,12 +12,26 @@ import io.jenkins.blueocean.analytics.Analytics.TrackRequest;
 import io.jenkins.blueocean.rest.hal.Link;
 import io.jenkins.blueocean.rest.impl.pipeline.NodeGraphBuilder;
 import io.jenkins.blueocean.service.embedded.analytics.Tally;
+import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.pipeline.modeldefinition.actions.DeclarativeJobAction;
+import org.jenkinsci.plugins.workflow.cps.nodes.StepEndNode;
+import org.jenkinsci.plugins.workflow.flow.FlowExecution;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.jenkinsci.plugins.workflow.graph.StepNode;
+import org.jenkinsci.plugins.workflow.graphanalysis.ForkScanner;
+import org.jenkinsci.plugins.workflow.graphanalysis.StandardChunkVisitor;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.jenkinsci.plugins.workflow.pipelinegraphanalysis.StageChunkFinder;
+import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 
 @Extension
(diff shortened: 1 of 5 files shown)


blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/analytics/PipelinePluginAnalyticsTest.java
```diff
@@ -33,6 +33,7 @@ public void testGenerateAnalyticsEventWithDeclarative() throws Exception {
         Assert.assertEquals("timesUsed", 2, properties.get("timesUsed"));
         Assert.assertEquals("isDeclarative", true, properties.get("isDeclarative"));
         Assert.assertEquals("runResult", "SUCCESS", properties.get("runResult"));
+//        Assert.assertEquals("plugin", "", properties.get("plugin"));
```
i386: so it turns out `pluginManager.whichPlugin` doesn't work inside of JenkinsRule tests but does work when you run the app normally. Go figure :(

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/analytics/PipelinePluginAnalytics.java
```diff
@@ -27,23 +43,51 @@ public void onCompleted(WorkflowRun workflowRun, @Nonnull TaskListener listener)
...
             props.put("timesUsed", value);
             props.put("isDeclarative", isDeclarative);
             props.put("runResult", resultAsString);
+            String pluginName = visitor.stepToPlugin.get(key);
+            props.put("plugin", pluginName != null ? pluginName : "unknown");
             analytics.track(new TrackRequest("pipeline_step_used", props.build()));
         });
     }
+
+    private class PipelineStepVisitor extends StandardChunkVisitor {
```
i386: @abayer decided it was more efficient to do things with my own chunk visitor

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/analytics/PipelinePluginAnalyticsTest.java
i386: Any ideas @vivek or @jglick ?

i386:
Thanks @abayer

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/analytics/PipelinePluginAnalytics.java
```diff
@@ -27,23 +43,51 @@ public void onCompleted(WorkflowRun workflowRun, @Nonnull TaskListener listener)
...
             props.put("timesUsed", value);
             props.put("isDeclarative", isDeclarative);
             props.put("runResult", resultAsString);
+            String pluginName = visitor.stepToPlugin.get(key);
+            props.put("plugin", pluginName != null ? pluginName : "unknown");
             analytics.track(new TrackRequest("pipeline_step_used", props.build()));
         });
     }
+
+    private static final class PipelineStepVisitor extends StandardChunkVisitor {
```
svanoort: If you're just collecting a Tally, use https://github.com/jenkinsci/workflow-api-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowNodeVisitor.java and 
https://github.com/jenkinsci/workflow-api-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/AbstractFlowScanner.java#L373

Bismuth is meant for this kind of work, and meant to make it simple -- you need not just collect on AtomNodes, we can get info on which block-scoped steps were run as well -- caveat: you might get some duplication due to nesting of blocks.

I'd actually imagined doing counts of steps by type but nobody asked for it, so I didn't code it up.

svanoort: On second thought, better approach here is probably to use a GraphListener so you're not doing a complete extra scan through the FlowGraph.

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/analytics/PipelinePluginAnalyticsTest.java
svanoort: @Jesse claimed to add something to the ArgumentsActionImpl that will give you a more direct way to get at this and may not need the plugin manager -- I haven't tried it out though.

svanoort:
Left a couple comments -- in general I think this approach *should* give a result but I think the way it's structured could use a bit of polish.

vivek:
@i386 Analytics has been disabled in blueocean: https://github.com/jenkinsci/blueocean-plugin/pull/1618. Is this PR relevant any more?

i386:
@vivek still relevant. I haven't had the chance to circle back yet.

vivek:
@i386 ok, thanks.

PR #1584, opened 2017-12-11
[JENKINS-47158] Ensure that synthetic FlowNode is not saved.

abayer:
# Description

See [JENKINS-47158](https://issues.jenkins-ci.org/browse/JENKINS-47158).

This is just a bandaid - as discussed on JENKINS-47158, the usage of
synthetic `FlowNode`s for visualization-only purposes is not
ideal. But at a minimum, this prevents the serialization of the
synthetic `FlowNode`s, making JENKINS-45892 warnings go away, and not,
well, serializing a bunch of stuff we almost certainly shouldn't be.

No manual testing really needed here, but if you want, run on core 2.75 or later, use a parallel Pipeline like [the one specified in this JIRA comment](https://issues.jenkins-ci.org/browse/JENKINS-47158?focusedCommentId=316531&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-316531), and make sure you don't see anything about JENKINS-45892 in the Jenkins log.

Assuming this is accepted, it should go into the next 1.3.x release as well as 1.4 - this is a critical issue.

# Submitter checklist
- [x] Link to JIRA ticket in description, if appropriate.
- [x] Change is code complete and matches issue description
- [x] Appropriate unit or acceptance tests or explanation to why this change has no tests
- [x] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below.

# Reviewer checklist
- [ ] Run the changes and verified the change matches the issue description
- [ ] Reviewed the code
- [ ] Verified that the appropriate tests have been written or valid explanation given

cc @jglick, who isn't eligible to be a reviewer.

PR #1586, opened 2017-12-12
release 1.3: [JENKINS-47158] Ensure that synthetic FlowNode is not saved. (#1584)

michaelneale:
A cherry pick from master of https://github.com/jenkinsci/blueocean-plugin/pull/1584

* [JENKINS-47158] Ensure that synthetic FlowNode is not saved.

This is just a bandaid - as discussed on JENKINS-47158, the usage of
synthetic `FlowNode`s for visualization-only purposes is not
ideal. But at a minimum, this prevents the serialization of the
synthetic `FlowNode`s, making JENKINS-45892 warnings go away, and not,
well, serializing a bunch of stuff we almost certainly shouldn't be.

* Add note as to why we're not saving.


vivek:
Ah, I opened one as well, https://github.com/jenkinsci/blueocean-plugin/pull/1586. I will close it in favor of this.

michaelneale:
ok merging...

PR #1587, opened 2017-12-13
feature/JENKINS-38339-downstream-job-links-2 * Annotate flow nodes wi…

sophistifunk:
# Description

Annotate flow nodes with pointers to any downstream jobs created, and add the data to the rest model.

Requires an update to `pipeline-build-step-plugin` here: https://github.com/jenkinsci/pipeline-build-step-plugin/pull/18/

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


sophistifunk:
This won't build on CI, as it needs changes to `pipeline-build-plugin`.

The code here: https://github.com/jenkinsci/blueocean-plugin/pull/1587/files#diff-c52f200b9ffd06a46ad9695f5aff5065R72 is to find the node that is actually returned to the UI. I don't like it, and I think it might be a better idea to collect the actions while walking the graph and creating the NodeWrapper, but I wanted to get some feedback.

I also don't know a good way to unit test this?

michaelneale:
@sophistifunk is this to review functionality that is visible or just code for now?

sophistifunk:
There's no UI changes for this yet, but if you pull and install the pipeline branch you can see the result in the JSON. I'll paste a sample in here in a second.

sophistifunk:
Example single node output JSON:

```
{
    "_class": "io.jenkins.blueocean.rest.impl.pipeline.PipelineNodeImpl",
    "_links": {
        "self": {
            "href": "/blue/rest/organizations/jenkins/pipelines/Upstream wait/runs/29/nodes/22/"
        },
        "actions": {
            "href": "/blue/rest/organizations/jenkins/pipelines/Upstream wait/runs/29/nodes/22/actions/"
        },
        "steps": {
            "href": "/blue/rest/organizations/jenkins/pipelines/Upstream wait/runs/29/nodes/22/steps/"
        }
    },
    "actions": [],
    "displayDescription": null,
    "displayName": "Double-downstream",
    "durationInMillis": 19873,
    "id": "22",
    "input": null,
    "result": "SUCCESS",
    "startTime": "2017-12-13T18:18:18.714+1000",
    "state": "FINISHED",
    "type": "STAGE",
    "causeOfBlockage": null,
    "downstreamBuilds": [
        {
            "description": "downstream1 #43",
            "link": {
                "href": "/blue/rest/organizations/jenkins/pipelines/downstream1/runs/43/"
            }
        },
        {
            "description": "downstream2 #55",
            "link": {
                "href": "/blue/rest/organizations/jenkins/pipelines/downstream2/runs/55/"
            }
        }
    ],
    "edges": []
}
```

michaelneale:
@cliffmeyers if you get a chance, can you code review/try this (wlil need to build the linked plugin to get the changes needed)

cliffmeyers:
@sophistifunk I have this build and running but I do not see the `downstreamBuilds` array populated with anything in my simple test pipeline:

https://github.com/histos/pipeline-samples/blob/downstream/Jenkinsfile

(NOTE: in my environment, downstream1 and downstream2 are freestyle jobs that just echo their name)

I built this and pipeline-build-step-plugin locally. PBSP was actually 2.7-SNAPSHOT in the repo, so I switched our parent POM to specify 2.7-SNAPSHOT in the dependency management area and omitted the version tag from deps in blueocean/pom.xml and blueocean-pipeline-api-impl. Not sure if that is causing anything to be weird - perhaps @michaelneale could confirm whether that change is appropriate - but in my Jenkins config I see a 2.7-SNAPSHOT installed.

I do eventually see the `DownstreamJobListener.onStarted` method fire, although only once for the 'downstream2' job. `PipelineNodeImpl.getDownstreamBuilds` always return an empty array.

I'll keep tinkering but perhaps @sophistifunk might have an idea if I'd set something up wrong.

sophistifunk:
@cliffmeyers I'll try with your pipeline instead of mine, see what I get. I have serious doubts about my code that walks back up the graph trying to guess which `FlowNode` will be used to construct the nodewrapper.

sophistifunk:
Update: it's definitely not working correctly with Cliff's pipeline, I'm looking into fixing this with more flexibility in the way we find the actions.

vivek:
@sophistifunk 

>This won't build on CI, as it needs changes to pipeline-build-plugin.

You can deploy snapshot version and use the timestamped version in the pom file. I can help you with that when you are online.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/DownstreamJobListener.java
```diff
@@ -0,0 +1,93 @@
...
+
+                try {
+                    node = execution.getNode(action.getUpstreamNodeId());
+                } catch (IOException e) {
+                    LOGGER.warning("Could not retrieve upstream node: " + e);
+                    continue;
+                }
+
+                // Look up the nearest StepStartNode which is where we want to add our action so BO will find it
+                node = findContainingStepStartNode(node);
```
vivek: `StepStartNode` is ignored and not captured in DAG by blueocean. It only includes stages and parallel branches. Instead what it should do is to walk up ancestors till it finds either parallel branch or a stage. So it could use `PipelineNodeUtil.isStage(node) && PipelineNodeUtils.isParallelBranch(node)` on each ancestor.

sophistifunk: That didn't fix it either, I'm going to have to collect the actions while building the graph. Which is intimidating, but I've wanted to get a handle on how that works for about a year now, so I'm looking forward to the other side of it :)

vivek: Oops, should be `PipelineNodeUtil.isStage(node) || PipelineNodeUtils.isParallelBranch(node)`. Anyways, look forward to your proposed changes.

vivek: >That didn't fix it either
Where ever solution lives, place we want to attach to either stage or parallel node, which ever is immediate parent of triggering step.

sophistifunk:
@vivek I've made some changes to `PipelineNodeGraphVisitor` to collect the actions as it works, and put them into `FlowNodeWrapper` when they're created. Please have a look and let me know your thoughts.

sophistifunk:
@cliffmeyers this version works for your example pipeline as well as mine (at least for me), if you get a chance to take a look please let me know how it goes.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/DownstreamJobListener.java
```diff
@@ -0,0 +1,62 @@
...
+                    LOGGER.warning("Could not retrieve upstream node (null)");
+                    continue;
+                }
+
+                // Add an action on the triggerRun node pointing to the currently executing run
+                String description = run.getDescription();
+                if (description == null) {
+                    description = run.getFullDisplayName();
+                }
+                node.addAction(new NodeDownstreamBuildAction(run.getExternalizableId(), description));
```
vivek: Why not bake the Link right here `LinkResolver.resolveLink(run)`, makes things simpler later on.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/NodeDownstreamBuildAction.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Annotates a FlowNode to point to a downstream build triggered by said node. Applied by
+ * io.jenkins.blueocean.listeners.DownstreamJobListener in blueocean-pipeline-api-impl
+ */
+public class NodeDownstreamBuildAction extends InvisibleAction implements FlowNodeAction, BluePipelineAction {
+
+    private final String runExternalizableId;
+    private final String description;
+
+    public NodeDownstreamBuildAction(String runExternalizableId, String description) {
```
vivek: It should take link reference to run. This way, it will be self contained and won't need patching link later on.

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BluePipelineNode.java
```diff
@@ -92,9 +97,58 @@
...
+     * Downstream builds that were kicked off by this node, in the Blue Ocean sense. The triggering FlowNode is a
+     * child of this.node.getNode()
+     *
+     * @return downstream builds (with links)
+     */
+    @Exported(inline = true)
+    public abstract Collection<BlueDownstreamBuild> getDownstreamBuilds();
+
+    @ExportedBean
+    public static class BlueDownstreamBuild {
```
vivek: Don't think we need it. If `NodeDownstreamBuildAction` can be used directly.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
```diff
@@ -96,6 +99,22 @@ public String getStartTimeString(){
         return edges;
     }
 
+    @Override
+    public Collection<BlueDownstreamBuild> getDownstreamBuilds() {
```
vivek: Don't see why we need additional method, we already have `getActions()`, we returned filtered actions. You do not need `BlueDownstreamBuild` as well if you can construct `NodeDownstreamBuildAction` with link.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/BluePipelineAction.java
```diff
@@ -0,0 +1,13 @@
...
+import hudson.model.Action;
+
+/**
+ * Marker interface for actions we want to coalesce into BluePipleNodes from the various pipeline FlowNodes which they
+ * represent.
+ *
+ * Allows us to get the actions we're interested in promoting via Item.getActions(Class), without having to get all
+ * actions which could be an arbitrarily large amount of wasted work.
+ */
+public interface BluePipelineAction extends Action {
```
vivek: Not sure if we need it. Take a look at `ActionProxiesImpl` and `BlueActionProxy`.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/NodeDownstreamBuildAction.java
```diff
@@ -0,0 +1,54 @@
...
+import org.jenkinsci.plugins.workflow.actions.FlowNodeAction;
+import org.jenkinsci.plugins.workflow.graph.FlowNode;
+
+import java.util.Objects;
+
+/**
+ * Annotates a FlowNode to point to a downstream build triggered by said node. Applied by
+ * io.jenkins.blueocean.listeners.DownstreamJobListener in blueocean-pipeline-api-impl
+ */
+public class NodeDownstreamBuildAction extends InvisibleAction implements FlowNodeAction, BluePipelineAction {
```
vivek: It could also be `BlueActionProxy`. Consistent with how we are handling actions and presenting them as resources.

sophistifunk:
Just a quick update, I'm currently working on implementing the more straightforward of Vivek's suggestions, I'll see where I get to and talk this over in tomorrow's dev meeting.

sophistifunk:
Using `PipelineNodeUtil.isParallelBranch || PipelineNodeUtil.isStage` to guess which node will be used for the cut-down graph sent via REST doesn't work because `isStage()` is true for some preceding parents in the flow graph that aren't in it when you build Cliff's example pipeline.

If you look at the graph I generated, you'll see that the Parallel nodes 14 and 15 are what are used for the cut-down graph, and that the trigger nodes are 20 and 21, but that 17 and 19 also return true for `isStage()`

webgraphvis-cliff

sophistifunk:
@vivek I've made those changes and it seems to be working for all my test pipelines, could you please take another look when you get a chance?

vivek:
@sophistifunk wow, Impressive graph to demonstrate this problem! How did you generate it?

>Using PipelineNodeUtil.isParallelBranch || PipelineNodeUtil.isStage to guess which node will be used for the cut-down graph sent via REST doesn't work because isStage() is true for some preceding parents in the flow graph that aren't in it when you build Cliff's example pipeline.
>If you look at the graph I generated, you'll see that the Parallel nodes 14 and 15 are what are used for the cut-down graph, and that the trigger nodes are 20 and 21, but that 17 and 19 also return true for isStage()

Ah, right, that condition won't detect nested stages or parallels. Could you get around this problem?

sophistifunk:
Yes, the lookup code is here: https://github.com/jenkinsci/blueocean-plugin/pull/1587/files#diff-c52f200b9ffd06a46ad9695f5aff5065R99

It works for the test cases I'm using now, but I'm not sure how flexible it will be long-term, the flexibility of pipeline allows the same thing to be done with many different scripts (and hence graphs). It will almost certainly need to be updated for multi-stage parallels.

The graph was generated by webgraphviz, I wrote a class that walks the execution graph and spits out .dot code, but I didn't include it as it's mostly duplicating some existing functionality I didn't know about when I started.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/DownstreamJobListener.java
```diff
@@ -0,0 +1,130 @@
...
+                    } catch (IOException e) {
+                        LOGGER.severe("Could not persist node: " + e);
+                    }
+                }
+            }
+        }
+    }
+
+    /**
+     * Walks the execution graph from the originating node for a downstream build (actionNode), looking for the
```
vivek: Looks like you are finding closest parallel or stage. What about the case where closest parallel or stage is nested? pipeline graph in UI shoes only top level parallel or stage. Given this, I think this should be looking for farthest parallel or stage?

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/rest/NodeDownstreamBuildActionTest.java
```diff
@@ -0,0 +1,17 @@
+package io.jenkins.blueocean.service.embedded.rest;
+
+import io.jenkins.blueocean.rest.hal.Link;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class NodeDownstreamBuildActionTest {
```
vivek: It will be good to have test where a free style job is triggered from a pipeline. You can refer to PipelineNodeTest for existing functional test.

blueocean-pipeline-api-impl/pom.xml
```diff
@@ -106,6 +106,13 @@
             <artifactId>plain-credentials</artifactId>
         </dependency>
 
+        <!-- Needed to read the "upstreamNode" action -->
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>pipeline-build-step</artifactId>
+            <version>2.7-SNAPSHOT</version>
```
vivek: Version already defined under pluginManagement in parent pom. Not needed here.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/DownstreamJobListener.java
sophistifunk: To be honest I can't think of a way to do this without re-implementing most of `PipelineNodeGraphVisitor` which is why I wrote the version that collected the actions while building the graph itself.

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/rest/NodeDownstreamBuildActionTest.java
sophistifunk: Will do

blueocean-pipeline-api-impl/pom.xml
sophistifunk: I thought so too, but it won't compile if I take this out.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/DownstreamJobListener.java
vivek: @sophistifunk I have taken crack at a utility that gives you enclosing top level stage or parallel. See https://github.com/jenkinsci/blueocean-plugin/blob/1313ec172b180cdae71f3494ad6b1dc4dde300c2/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeUtil.java#L217.

There are tests to demonstrate nested, orphan cases working. PTAL. If it looks good you can merge PR #1626 in to this and we can close that.

pom.xml
```diff
@@ -425,7 +425,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>pipeline-build-step</artifactId>
-            <version>2.5.1</version>
+            <version>2.7-SNAPSHOT</version>
```
vivek: @sophistifunk Can we deploy SNAPSHOT version of pipeline-build-step and use the version number deployed there so that CI build passes?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/NodeDownstreamBuildAction.java
```diff
@@ -0,0 +1,60 @@
...
+import org.kohsuke.stapler.export.ExportedBean;
+
+import java.util.Objects;
+
+/**
+ * Annotates a FlowNode to point to a downstream build triggered by said node. Applied by
+ * io.jenkins.blueocean.listeners.DownstreamJobListener in blueocean-pipeline-api-impl
+ */
+@ExportedBean
+public class NodeDownstreamBuildAction extends InvisibleAction implements FlowNodeAction, Reachable, BluePipelineAction {
```
vivek: Any reason we need it in blueocean-rest-impl? It should be moved to blueocean-pipeline-api-impl as thats only place its needed. All Flow node specific code exists only in blueocean-pipeline-api-impl.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/BluePipelineAction.java
vivek: @sophistifunk we can delete this action now?

blueocean-pipeline-api-impl/pom.xml
vivek: Hmm, this is strange. I will try it out locally.

blueocean/pom.xml
```diff
@@ -115,6 +115,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>pipeline-build-step</artifactId>
+            <version>2.7-SNAPSHOT</version>
```
vivek: This shouldn't be needed. Maybe its due to SNAPSHOT version?

sophistifunk: No idea. It makes sense to me that we shouldn't need it, but without it, kaboom. I'd say that there's some sort of intermediary in the lookup hierarchy between this pom and the root pom which is resetting pipeline-build-step back to the prod release. But debugging multi-pom projects is definitely something I suck at :(

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/NodeDownstreamBuildAction.java
sophistifunk: There *was* a reason, but I can't remember it and it may no longer be valid so I'll move it and see if there's any sort of earth-shattering kaboom :)

sophistifunk:
@vivek regarding `BluePipelineAction` it's just to differentiate the actions we care about, so that we don't end up collecting a whole bunch of things that may be added by plugins, chewing up processor and memory on the backend and also bloating the response. We could replace it with a hardcoded `instanceof` test for classes we care about, like we do when constructing the `ActionProxiesImpl`, but it seems to me more fragile and complicated to do it that way. I'm certainly happy to be convinced otherwise if there's a reason for it that I just don't know :)

sophistifunk:
@vivek I would definitely like some help getting the snapshot of `pipeline-build-step` into dogfood so we can try and get a green tick :)

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/BluePipelineAction.java
```diff
@@ -1,4 +1,4 @@
-package io.jenkins.blueocean.service.embedded.rest;
+package io.jenkins.blueocean.listeners;
```
vivek: We should really move these over inside `io.jenkins.blueocean.rest.impl.pipeline`. These are strictly implementation specific and for the purpose of node API implementation so should be put inside `io.jenkins.blueocean.rest.impl.pipeline.listener`?

vivek:
@sophistifunk 

>regarding BluePipelineAction it's just to differentiate the actions we care about, so that we don't end up collecting a whole bunch of things that may be added by plugins, chewing up processor and memory on the backend and also bloating the response. We could replace it with a hardcoded instanceof test for classes we care about, like we do when constructing the ActionProxiesImpl, but it seems to me more fragile and complicated to do it that way. 

If we go with BluePipelineAction then the only way we can consume and expose other plugins action is by asking them to implement this interface. This goes against idea of being able to consume other plugins actions as they are as long as these are `ExportedInterface`.

vivek:
@sophistifunk 

>I would definitely like some help getting the snapshot of pipeline-build-step into dogfood so we can try and get a green tick :)

Ok, let me try it.

vivek:
@sophistifunk build is green. I think left over items are:

1.  couple of comments related to BluePipeineAction, package location
2. once above comments are resolved and you are happy with pipeline-build-step, we get it released
3. update blueocean pom with released version of pipeline-build-step and then merged

PR #1589, opened 2017-12-14
Hotfix/ath rest impl dep and release jdl

cliffmeyers:
# Description
- Prevent "target" from getting deleted unintentionally when mvn release:prepare and release:perform are executed

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-core-js/gulpfile.js
+++ b/blueocean-core-js/gulpfile.js
@@ -19,7 +19,7 @@ const fs = require('fs');
 // Options, src/dest folders, etc
 
 const config = {
-    clean: ["dist", "target"],
+    clean: ["dist"],
     react: {
         sources: ["src/**/*.{js,jsx}", "!**/__mocks__/**"],
         dest: "dist"
--- a/pom.xml
+++ b/pom.xml
@@ -1074,18 +1074,6 @@
 
                             <execution>
                                 <phase>initialize</phase>
-                                <id>gulp clean</id>
-                                <goals>
-                                    <goal>npm</goal>
-                                </goals>
-                                <configuration>
-                                    <!-- The package.json must define an "mvnclean" script -->
-                                    <arguments>run mvnclean</arguments>
-                                </configuration>
-                            </execution>
-
-                            <execution>
-                                <phase>initialize</phase>
                                 <id>npm install</id>
                                 <goals>
                                     <goal>npm</goal>
--- a/acceptance-tests/pom.xml
+++ b/acceptance-tests/pom.xml
@@ -57,7 +57,7 @@
         <dependency>
             <groupId>io.jenkins.blueocean</groupId>
             <artifactId>blueocean-rest-impl</artifactId>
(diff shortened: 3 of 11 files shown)


michaelneale:
won't know if this works until we try it - but need to have a review from @vivek and probably @kzantow

blueocean-core-js/gulpfile.js
```diff
@@ -19,7 +19,7 @@ const fs = require('fs');
 // Options, src/dest folders, etc
 
 const config = {
-    clean: ["dist", "target"],
+    clean: ["dist"],
```
vivek: Is there side effect of doing this? generated jar file has js code in it, can dist and target can ever get out of sync?

pom.xml
```diff
@@ -1072,18 +1072,6 @@
...
 
-                            <execution>
-                                <phase>initialize</phase>
-                                <id>gulp clean</id>
-                                <goals>
-                                    <goal>npm</goal>
-                                </goals>
-                                <configuration>
-                                    <!-- The package.json must define an "mvnclean" script -->
-                                    <arguments>run mvnclean</arguments>
```
vivek: Hmm, intent of this was to remove jar file during gulp clean. Which sounds ok considering generated jar file includes result of gulp build or maybe I misunderstand this process?

blueocean-core-js/gulpfile.js
michaelneale: I believe people do a clean anyway...

pom.xml
kzantow: ehh. not sure why this was here, I didn't add it, did I? :)

michaelneale:
so current progress: 

@vivek is going to try to do a dryRun locally with master, check can repro the release failure
Will then try it with just the change in the parent pom phase to "clean" for the gulp clean block
if that works, will open a PR just with that one line change... 

failing that, we may fall back to this.

cliffmeyers:
@michaelneale @vivek I think you will have problems firing the gulp clean in the clean phase, because it fires before initialize and therefore node and npm won't have been installed yet.

michaelneale:
@cliffmeyers I would think init is first of all based on that list.. but who knows.

cliffmeyers:
https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#Lifecycle_Reference

michaelneale:
@cliffmeyers still not clear to me - do we even need that clean in the parent pom - perhaps if that was removed would be enough?

pom.xml
cliffmeyers: The intent wasn't to remove the .jar file, I was trying to clear out the stuff in `target` that was copied by maven-resources-plugin. Deleting all of `target` was clearly a bad idea, I should have individually cleaned out the folders copied it, just more of a maintenance headache.

vivek:
Tried experimenting with dryRun to test alternate proposal from @kzantow. dryRun doesn't do anything meaningful for release:perform. We are going to merge this PR and try a release after successful master run.

PR #1590, opened 2017-12-14
remove dependency on blue test scoped utility

michaelneale:
Removing a dependency - @imeredith @cliffmeyers this is "plan B"


kzantow:
This is all a bunch of duplicated code, yes?

imeredith:
I'm split. We already use Unirest a lot in ATH. So the arguement for chaning the libs doenst really make much sense. However I like the httprequest class for tests.

I dislike the code duplication we have in tests though, I would like to see this all in a util jar.

michaelneale:
merging it in to get master healthy

PR #1591, opened 2017-12-15
Use Unirest instead of httprequst in ath

imeredith:
# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


acceptance-tests/pom.xml
```diff
@@ -53,29 +53,6 @@
             <artifactId>selenium-firefox-driver</artifactId>
             <version>2.53.0</version>
         </dependency>
-        <!-- begin HttpRequest helper -->
-        <dependency>
-            <groupId>io.jenkins.blueocean</groupId>
-            <artifactId>blueocean-rest-impl</artifactId>
-            <version>1.4.0-beta-1-SNAPSHOT</version>
```
vivek: LGTM. 

Yeah dependency from ath to snapshot version is always going to be a problem. Specially it the thing that it intends to test. Either we make it standalone released http test library or use the one already there.

imeredith:
@vivek yeah I agree in principle. In practise we already use Unirest in ATH and it was easy to use it for what is in use today. But in the future I would like to see us maybe have a test-utils module for shared test code instead of having multiple impls

vivek:
@imeredith there is competing PR #1590  from @michaelneale. We should keep only one of them. My vote goes for keeping 1590.

PR #1592, opened 2017-12-15
Revert "Hotfix/ath rest impl dep and release jdl (#1589)"

michaelneale:
This reverts commit 42188454624628a3fadb88c65b6d6e2984a6df09.

Gets master back to old state regarding gulp


michaelneale:
hey @vivek Keith @kzantow said we should probably revert this change, get master back the way it was (with the fix for ATH) - as this didn't improve things. It just moved the failure to the prepare (which may be interesting...)

PR #1594, opened 2017-12-15
Harden message enricher

michaelneale:
# Description

https://issues.jenkins-ci.org/browse/JENKINS-48587

hey @vivek @svanoort this is to handle random failures at enriching (which happens during listening).

### DIFF
--- a/blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
+++ b/blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
@@ -47,6 +47,8 @@ import org.jenkinsci.plugins.workflow.multibranch.WorkflowMultiBranchProject;
 import javax.annotation.Nonnull;
 import java.util.logging.Logger;
 
+import static java.util.logging.Level.WARNING;
+
 /**
  * @author <a href="mailto:tom.fennelly@gmail.com">tom.fennelly@gmail.com</a>
  */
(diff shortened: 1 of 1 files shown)


michaelneale:
@vivek if this is ok, can cherry pick to 1.3 and cut a release again there

blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
```diff
@@ -64,7 +66,15 @@
 
     @Override
     public void enrich(@Nonnull Message message) {
+        try {
+            maybeEnrichMessage(message);
+        } catch (Exception e) {
```
vivek: This looks fine. There is one downside though, masking exceptions could have front end side effects as many things are tied to sse events. If they suddenly stops appearing then only way to find will be to look in log for these warning messages.

Maybe we can do something like this:

```
// Add BlueEventProps.blueocean_unexpectederror enum up in this file
jobChannelMessage.set(BlueEventProps.blueocean_unexpectederror, e.getMessage());
```

Even if frontend doesn't process such error message but atleast will be easier to debug or find cause of UI not behaving as expected?

blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
```diff
@@ -64,7 +66,15 @@
...
+            LOGGER.log(WARNING, "Unable to enrich message", e);
```
vivek: Better: `LOGGER.log(WARNING, "Unable to enrich message: " +e.getMessage(), e);`

michaelneale: ok - wont' flood the log with long messages? I guess it is ok though, as if this warning happens something is wrong and we do want to know why

blueocean-events/src/main/java/io/jenkins/blueocean/events/BlueMessageEnricher.java
michaelneale: possibly - risk of flooding is a bit high though.... but like you said, it is probably broken a bit.

michaelneale: I think if things aren't working though, we tend to ask for a support bundle (easier to get than browser logs) so would tend to see a flood of this anyway so 🤷‍♀️

michaelneale:
Ok I might leave it this way for now. I have some gut reaction against putting richer code in a catch that in turn could fail ...

PR #1597, opened 2017-12-18
JENKINS-4292 add generic error reporting for failed input step

NicuPascu:
# Description

See [JENKINS-42920](https://issues.jenkins-ci.org/browse/JENKINS-42920).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


NicuPascu:
screenshot from 2017-12-18 20-29-40

NicuPascu:
Regarding tests for this, before writing any tests I wanted to get some feedback on the implementation, if everyone is happy I will proceed to write some tests before merging

michaelneale:
perhaps @vivek or @abayer could do a code review of this approach?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
```diff
@@ -234,7 +232,22 @@ public HttpResponse submitInputStep(StaplerRequest request) {
             }
 
             //XXX: execution.doProceed(request) expects submitted form, otherwise we could have simply used it
-            execution.preSubmissionCheck();
+
+            try {
+                execution.preSubmissionCheck();
+            } catch (Failure f) {
+                JSONObject obj = new JSONObject();
```
vivek: Backend error messages are in specific format. Lets keep it consistent. Following code should do it:

```java
...
rsp.getWriter().print(JsonConverter.toJson(new ErrorMessage(400, f.getMessage())));
....
```

vivek:
jacoco failure as one of the test timed out quietly resulting in to less than required coverage.  Need to fix it. Kicked off another run.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
```diff
@@ -234,7 +234,19 @@ public HttpResponse submitInputStep(StaplerRequest request) {
             }
 
             //XXX: execution.doProceed(request) expects submitted form, otherwise we could have simply used it
-            execution.preSubmissionCheck();
+
+            try {
+                execution.preSubmissionCheck();
+            } catch (Failure f) {
+                return new HttpResponse() {
```
vivek: @NicuPascu Ah, Jacoco failure is legitimate. There is no test for this code block. Please look at https://github.com/jenkinsci/blueocean-plugin/blob/2c91571484e3524a1449719f9fdce059dbf1808d/blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeTest.java#L2115 and other input related tests and add one for this case.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
```diff
@@ -234,7 +234,19 @@ public HttpResponse submitInputStep(StaplerRequest request) {
...
+                    @Override
+                    public void generateResponse(StaplerRequest req, StaplerResponse rsp, Object node) throws IOException, ServletException {
+                        rsp.setStatus(400);
+                        rsp.setContentType("application/json;charset=UTF-8");
+                        rsp.getWriter().print(JsonConverter.toJson(new ErrorMessage(400, f.getMessage())));
```
cliffmeyers: Would throwing a `ServiceException` here instead of returning `new HttpResponse()` be a simpler way to achieve the same effect? @vivek

vivek: Yes, maybe worth trying ‘BadRequestException’.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
```diff
@@ -4,10 +4,7 @@
 import hudson.ExtensionList;
 import hudson.FilePath;
 import hudson.console.AnnotatedLargeText;
-import hudson.model.Action;
-import hudson.model.FileParameterValue;
-import hudson.model.ParameterDefinition;
-import hudson.model.ParameterValue;
+import hudson.model.*;
```
michaelneale: this should be specific imports, not wildcard

michaelneale: (wildcard is ok for test code but this is prod code)

michaelneale:
code review @vivek or @imeredith ? my only feedback is on the wildcard import as noted

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepImpl.java
sophistifunk: There's a setting in Idea that will do this for you when you optimise imports.

NicuPascu: @sophistifunk I had a problem with the latest intelliJ and optimise imports wasn't doing anything, I downgraded to the previous major version and it did the trick, thanks

vivek:
@NicuPascu LGTM 🐝  and nice work!

PR #1598, opened 2017-12-19
Many dependencies need to be bumped

i386:
# Description

Bump it all

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -360,6 +360,7 @@
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
             <version>2.15</version>
+            <!-- Use 2.16 or above when baseline is >= 2.76 -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
 
   <properties>
     <java.level>8</java.level>
-    <jackson.version>2.8.9</jackson.version>
+    <jackson.version>2.8.10</jackson.version>
     <jenkins.version>2.73.3</jenkins.version> <!-- Should be kept in sync with Dockerfile FROM statement-->
     <javadoc.exec.goal>javadoc-no-fork</javadoc.exec.goal> <!-- stop initialize phase plugins executing twice -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -368,12 +369,12 @@
...
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-branch-source</artifactId>
-            <version>2.2.3</version>
+            <version>2.3.2</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-multibranch</artifactId>
-            <version>2.16</version>
+            <version>${workflow.support.version}</version>
```
kzantow: Is this always going to be the same as `workflow-support`?

vivek: Yeah this doesn't seem right. They are different plugins and can have different versions.

i386: That was my IDE being shit. Sorry about that.

michaelneale:
wow the previous failure - NOTHING worked in the ATH (probably means it didn't load).

michaelneale:
won't some of these plugins need newer version of LTS than blue ocean or ATH currently supports?

pom.xml
```diff
@@ -25,7 +25,7 @@
 
   <properties>
     <java.level>8</java.level>
-    <jackson.version>2.8.9</jackson.version>
+    <jackson.version>2.8.10</jackson.version>
     <jenkins.version>2.73.3</jenkins.version> <!-- Should be kept in sync with Dockerfile FROM statement-->
```
michaelneale: this and the ATH version in the Jenkinsfile *may* need to be bumped to latest LTS if some of these new deps require it

michaelneale:
:bug: need to try the ATH - it likely isn't loading blue ocean at all.

i386:
@michaelneale latest `workflow-job` needs the latest LTS. Ive downgraded.

i386:
Legit test failure resulted in [JENKINS-48644](https://issues.jenkins-ci.org/browse/JENKINS-48644) which is being tracked down by @svanoort

michaelneale:
nice - the system works! ish.

vivek:
Closing, not relevant anymore.

PR #1600, opened 2017-12-19
tidying assertions to actually count

michaelneale:
# Description

As discussed with @vivek  - a bit messy but can't do assertions in callbacks like this if you want them to mean anything. So this pulls them out


michaelneale:
@vivek jacoco failing here - but not sure why - it doesnt' change any test code that would affect coverage does it ?

michaelneale:
have a better approach. Ignore.

PR #1601, opened 2017-12-19
try this type of refactoring... use finals

michaelneale:
# Description

ok @vivek moving assertions to top level .. this replaces the other PR


michaelneale:
I think @vivek will be replacing this ...

vivek:
Yeah should have a PR replacing this one.

PR #1603, opened 2017-12-19
Add jdl & core-js to aggregator. Whoops!

kzantow:
# Description

Probably need to include the new modules as direct dependencies of the aggregator.

### DIFF
--- a/blueocean/pom.xml
+++ b/blueocean/pom.xml
@@ -22,6 +22,14 @@
     <dependencies>
         <dependency>
             <groupId>${project.groupId}</groupId>
+            <artifactId>jenkins-design-language</artifactId>
+        </dependency>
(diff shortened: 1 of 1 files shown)


michaelneale:
hey @vivek @i386  this look ok to you?

I did a fresh install of 1.4 beta from experiemental UC and installed fine, as core/jdl must be transitive deps as they are pulled in. However @kshultzCB picked up a problem updating an existing install, so perhaps we should, like other deps, include it in the aggregator as they are required? (even if they are also transitively pulled in?)

michaelneale:
:bee:

@kzantow @vivek  when merged - may as well roll another beta

kzantow:
@michaelneale done and done

PR #1607, opened 2018-01-05
JENKINS-48219# When creating git pipeline check Item.READ permission instead of overall READ permission

vivek:
# Description

Git creation flow checks for overall read permission in getContent() API call. It should really be checking that Item's READ permission instead.

cc: @daniel-beck @kzantow 

See [JENKINS-48219](https://issues.jenkins-ci.org/browse/JENKINS-48219).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-git-pipeline/src/main/java/io/jenkins/blueocean/blueocean_git_pipeline/GitReadSaveService.java
+++ b/blueocean-git-pipeline/src/main/java/io/jenkins/blueocean/blueocean_git_pipeline/GitReadSaveService.java
@@ -23,26 +23,23 @@
  */
 package io.jenkins.blueocean.blueocean_git_pipeline;
 
-import java.io.IOException;
-
-import hudson.model.User;
-import org.apache.commons.io.IOUtils;
-import org.apache.commons.lang.StringUtils;
-import org.apache.commons.lang3.ObjectUtils;
-import org.kohsuke.stapler.StaplerRequest;
-
 import hudson.Extension;
 import hudson.model.Item;
+import hudson.model.User;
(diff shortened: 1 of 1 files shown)


vivek:
@oleg-nenashev Good catch. This was the only place where system permission was checked. I agree Item.CONFIGURE is more appropriate.

Regarding EXTENDED_READ, I think READ is sufficient for this operation.

michaelneale:
LGTM - @oleg-nenashev can you bless the latest changes so we can get this in?

oleg-nenashev:
> Regarding EXTENDED_READ, I think READ is sufficient for this operation.

IMHO it needs more discussion. CC @daniel-beck who is a deputy(?) security officer.

michaelneale:
@oleg-nenashev fair enough - it is a fiddly area and we don't want this to be too liberal with permissions.

vivek:
@oleg-nenashev Sure, I already have Daniel as reviewer. I do not want to widen read scope to EXTENDED_READ when READ is sufficient. EXTENDED_READ lets you read secrets attached to the job Item which we do not need. 

@daniel-beck This PR narrows permission scope from overall read/write to Item.READ  and Item.WRITE on a Job. PTAL.

vivek:
@michaelneale @kzantow I have tested by creating/editing/saving GitHub repo using ssh for non-admin user without overall read or write permission, just the permission to read/write job. Works fine. Do you want to test anything else?

michaelneale:
Seems reasonable @vivek - @kzantow any tricky git scenaraios that have bitten in the past to retest that aren't already covered by automated tests?

PR #1611, opened 2018-01-09
Revert "Bug/jenkins 47219 skipped parallel stage wrong status"

michaelneale:
Reverts jenkinsci/blueocean-plugin#1610

see if this goes green...


michaelneale:
hrm @vivek @cliffmeyers this still breaks - yet I have tried reverting it back... so why would that test suddenly wig out?

cliffmeyers:
Looks like perhaps #1612 was the solution to this problem? Should we close this PR?

michaelneale:
@cliffmeyers ah yes, forgot to close this one. 

All good.

PR #1614, opened 2018-01-10
JENKINS-48872# Fix deserialization error for missing ModelObject field

vivek:
# Description

ModelObject was serialized in prior version (1.4.0-beta4 and older), it was removed as part of fix that went in PR #1609. This caused error reported in OldData. Fix is to simply put back unused ModelObject as transient. 

@michaelneale @jglick Do not like this fix, is there better ways other plugins handle it? On the other hand, we could mark this issue as will not fix as 'Discard Unreadable Data' is just fine. Just that it will surprise anyone upgrading to 1.4.0 and we might want to port #1609 to 1.3.0.

See [JENKINS-48872](https://issues.jenkins-ci.org/browse/JENKINS-48872).

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code


michaelneale:
yeah maybe @jglick know of a `@PleaseIgnore` type annotation to make this go away, but this while unpleasant might do the trick (perhaps with a comment that it can be removed in a future version)

michaelneale:
:bee:

PR #1615, opened 2018-01-14
[JENKINS-45037] Remove download link if there are no artifacts

imeredith:
# Description

This PR removes the download link if there are no artifacts to download

See [JENKINS-45037](https://issues.jenkins-ci.org/browse/JENKINS-45037).


michaelneale:
:bee:

michaelneale:
would be nice to have a test for this - to check it doesn't show

imeredith:
Adding test

PR #1618, opened 2018-01-15
Disable analytics

i386:
We need to work out how much data we are actually sending back before this gets turned on


michaelneale:
If this goes green - go for it

michaelneale:
@i386 the test fails for this - I assume it checks that it is enabled somehow? 

`io.jenkins.blueocean.service.embedded.analytics.AnalyticsTest.enableUsageStats(AnalyticsTest.java:39)` - assertion failed

i386:
@michaelneale thanks for that

PR #1624, opened 2018-01-17
JENKINS-47395 for github server pipeline creation if url is invalid try appending /api/v3

NicuPascu:
# Description
While adding a new GHE server, if the url is invalid, check if the URL ends with "/api/v3" and if not, append it and re-try adding the server. This way if the url is being re-writen by a reverse proxy it will still work.

See [JENKINS-47395](https://issues.jenkins-ci.org/browse/JENKINS-47395).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


imeredith:
I'm with Josh on this. But if there is some wonkyness going on then I guess is fine

NicuPascu:
@sophistifunk GH hasn't changed their API, this is needed because some people didn't realise they had to put the full path of the GHE api endpoint, I didn't want to always assume that a full path is always needed because I am sure some people re-write their URIs and that would make GHE unusable for them. I think in the long term this implementation will take care of the most use-cases and ultimately it fixes the problem that JENKINS-47395 refers to.

PR #1625, opened 2018-01-17
JENKINS-48198 - Add logging when a property is skipped

sophistifunk:
* Add logging when a property is skipped during JSON marshalling due to an exception

# Description

See [JENKINS-48198](https://issues.jenkins-ci.org/browse/JENKINS-48198).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/ExportInterceptor.java
+++ b/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/ExportInterceptor.java
@@ -1,6 +1,8 @@
 package io.jenkins.blueocean.commons.stapler.export;
 
 import org.kohsuke.stapler.export.Exported;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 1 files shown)


sophistifunk:
This does *not* fix the linked issue, but at least it will end the "silent" part of "silently swallowing exceptions and sending the UI partial responses" - and I want to get it into master ASAP as who knows where else we're not noticing exceptions.

I'm looking into a sensible fix.

michaelneale:
interesting - so at least if that error happens when the front end misses a field it wants - server side log would say something right? 

@vivek this could be quite useful (we don't necessarily want to skip it silently from the front end code, perhaps some nicer error message, but mostly we just want to know the failure so it can be dealt with. Usually server side logs are best for this anyway)

sophistifunk:
Will do.

sophistifunk:
Stapler PR: https://github.com/stapler/stapler/pull/136

Stapler looks to be using `java.util.logging` rather than `slf4j` like we are.

vivek:
@sophistifunk upstream stapler pr is upgraded. This pr should have exact change copied. Easy to just copy the changes file over to corresponding on in blueocean.

sophistifunk:
Updated to match upstream exactly (save package name). 

Related: It confuses the hell out of IntelliJ when you copy in a file with the right class name but the wrong package :D

sophistifunk:
@vivek Happy to merge this? I re-ran the build and got green, so it was just one of the flaky tests.

vivek:
@sophistifunk LGTM 🐝

PR #1626, opened 2018-01-17
JENKINS-38339# Utility to get enclosing top level stage or parallel

vivek:
# Description

@sophistifunk PTAL. I think this should give you stage or parallel for a given node. Handles nesting parallel/stage or combination of it. Should also work with orphan steps. 

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeUtil.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeUtil.java
@@ -5,16 +5,21 @@ import hudson.model.Action;
 import hudson.model.Queue;
 import hudson.model.queue.CauseOfBlockage;
 import io.jenkins.blueocean.rest.model.BlueRun;
+import java.util.List;
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import org.jenkinsci.plugins.pipeline.StageStatus;
 import org.jenkinsci.plugins.pipeline.SyntheticStage;
 import org.jenkinsci.plugins.workflow.actions.ErrorAction;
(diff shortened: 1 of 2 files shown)


sophistifunk:
Thanks heaps for this Vivek, I'll merge it into my branch and update my PR tomorrow morning!

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeUtil.java
```diff
@@ -206,4 +207,60 @@ public static boolean isAgentStart(@Nullable FlowNode node) {
...
+        }
+        return null;
+    }
+
+    /**
+     * Gives immediate enclosing Stage or parallel node.
+     * @param node it could be any node inside a stage or parallel
+     * @return Immediate stage or parallel node
+     */
+    private static @CheckForNull FlowNode getStageOrParallelParent(@Nonnull FlowNode node) {
```
svanoort: 🐛 No, this will generate a very uncontrolled StackOverflowException in real-world use cases and extremely bad big-O scaling -- please use: https://github.com/jenkinsci/workflow-api-plugin/pull/50 `iterateEnclosingBlocks(FlowNode)` -- that's what it's there for. :)   Also, use a Predicate to match Stage/Parallel -- probably https://github.com/jenkinsci/workflow-api-plugin/blob/master/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/NodeStepTypePredicate.java

vivek:
@svanoort Thanks. Regarding stack overflow, the code can be converted to iterative, but why do all of that when there is already an optimized way to do it:) 
@sophistifunk   take a look st Sam’s suggestion instead, of it gives what you want then all good. I might be able to look at it later in the evening.

vivek:
Closing this PR as its being implemented in #1587

PR #1627, opened 2018-01-17
Improvement/jenkins 43211 redux markdown cleanup 2

cliffmeyers:
# Description
See [JENKINS-43211](https://issues.jenkins-ci.org/browse/JENKINS-43211).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


cliffmeyers:
Input Step original
markdown-removal-1-input-step-original

Input Step updated (no Markdown)
markdown-removal-2-input-step-modified

Run Button "unsupported" dialog original
markdown-removal-3-unsupported-dialog-original

Run Button "unsupported" dialog updated (no Markdown)
markdown-removal-4-unsupported-dialog-modified

michaelneale:
nice - a lot of deleted code! Any measurable impact? 

IN the i18n docs/readme, it mentions Markdown - I guess with this change then that needs to be edited? 

Is @sophistifunk going to pick this up @cliffmeyers ? 

On removing these deps, would like to check with @scherler @kzantow that it is ok and no unexpected side esfects.

sophistifunk:
If it's green, I say go for it, once we've got confirmation from the downstream projects. Doesn't matter if they're depending on packages we've stopped requiring so much, that's an easy fix. Just if they were depending on some of the code we've removed from this project is the concern.

cliffmeyers:
@michaelneale right, I agree w/ @sophistifunk that this change is probably good to go as it is. I think porting favorites to use mobx instead of redux is better done in a separate PR, since the risk of regression there is much higher... this PR is really just removing a lot of dead code and the Markdown removal is pretty superficial.

kzantow:
@michaelneale these dependencies were not exposed in any way to downstream plugins, so shouldn't have any issues

blueocean-dashboard/package.json
```diff
@@ -66,13 +70,8 @@
...
     "react-dom": "15.4.2",
     "react-i18next": "1.10.1",
-    "react-redux": "4.4.5",
-    "react-remarkable": "1.1.1",
     "react-router": "2.8.1",
-    "recharts": "1.0.0-alpha.4",
-    "redux": "3.6.0",
-    "redux-thunk": "2.1.0",
-    "reselect": "2.5.4"
+    "recharts": "1.0.0-alpha.4"
```
scherler: why?

michaelneale: @scherler not actually an addition https://github.com/jenkinsci/blueocean-plugin/pull/1627/files/e0da8c181b8590b5716b3d7815c4f6b52a1daade#diff-0ca33611a0c8a93fe297afbfc1b1a8aaL72 has it a few lines above. Just a case of git diff getting confused

michaelneale:
@imeredith you able to take ownership of this PR?

cliffmeyers:
@kzantow the additions I believe are due to "dev" and "optional" fields being added to shinkwrap. Apparently when shrink was recreated it was using an older version of npm that didn't include those flags. Until migration to npm5 or yarn, everyone on the team should check they are using latest npm 3.x, I think it was 3.10.10 the last time I upgraded.

PR #1628, opened 2018-01-18
Fix JENKINS-49028 null protect call to getShortDescription()

fdearle:
# Description

See [JENKINS-49028](https://issues.jenkins-ci.org/browse/JENKINS-49028).

As per my comments in the Jira. I've been unable to generate an appropriate unit test for this issue. PipeLineNodeUtils does not seem to have its own independent unit tests. The fix is for an obvious NPE due to a null dereference.


michaelneale:
Running tests again - as I think those failures are unrelated.

imeredith:
Looks mergeable. I'm working on a unit test for this and will merge it when it passes

michaelneale:
nice one @imeredith @fdearle !

imeredith:
@michaelneale @fdearle I'm not entirely sure how useful this test is. But it tests all logic branches in the method at least.

fdearle:
@michaelneale @imeredith 

This is perfect. I had figured that a specific unit test for PipelineNodeUtil was what was needed but did not have the background to write one.

PR #1617, opened 2018-01-15
Add Chinese translation

LinuxSuRen:

PR #1629, opened 2018-01-20
Add Chinese localization for pipeline editor

LinuxSuRen:


LinuxSuRen:
@jenkinsci/code-reviewers Any one can help me to review this pr, tks.

imeredith:
@LinuxSuRen nice work. One change I would like to see is that the default option is removed from all JS code and the default is only in Messages.properties (thanks for already doing this part). That way we dont have 2 defaults to update in the future. I know we still have code that does this, at some point it will be removed though.

michaelneale:
@vivek this one is also ok to merge (along with other one)

vivek:
@LinuxSuRen There are conflicts as result of merging #1617. Please fix these conflicts and this should be in good shape.

LinuxSuRen:
@vivek Thanks for the tip, conflicts alread been fixed.

michaelneale:
once this is green, can get it in

PR #1636, opened 2018-01-27
[JENKINS-XXXXX] Added a dependency with blueocean-common

recena:
# Description

if `jenkins-design-language` does not have any Maven dependency with core plugins of Blue Ocean, this plugin could be installed independently of the rest of Blue Ocean plugins.

This would be an not desired behavior.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/jenkins-design-language/pom.xml
+++ b/jenkins-design-language/pom.xml
@@ -15,6 +15,14 @@
 
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Jenkins+Design+Language</url>
 
+    <dependencies>
+        <dependency>
+           <groupId>io.jenkins.blueocean</groupId>
+           <artifactId>blueocean-commons</artifactId>
+           <version>${project.version}</version>
+        </dependency>
+    </dependencies>
+
     <build>
         <resources>
             <resource>
@@ -41,18 +49,5 @@
             </resource>
         </resources>
     </build>
-
-    <repositories>
-        <repository>
-            <id>repo.jenkins-ci.org</id>
-            <url>http://repo.jenkins-ci.org/public/</url>
-        </repository>
-    </repositories>
-    <pluginRepositories>
-        <pluginRepository>
-            <id>repo.jenkins-ci.org</id>
-            <url>http://repo.jenkins-ci.org/public/</url>
-        </pluginRepository>
-    </pluginRepositories>
 
 </project>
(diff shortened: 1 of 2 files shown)


jenkins-design-language/pom.xml
```diff
@@ -42,17 +50,4 @@
         </resources>
     </build>
 
-    <repositories>
```
recena: It is not needed since it is defined in the parent.

pom.xml
```diff
@@ -143,13 +143,13 @@
   <repositories>
     <repository>
       <id>repo.jenkins-ci.org</id>
-      <url>http://repo.jenkins-ci.org/public/</url>
+      <url>https://repo.jenkins-ci.org/public/</url>
```
recena: It is a good practice.

jenkins-design-language/pom.xml
```diff
@@ -15,6 +15,14 @@
 
     <url>https://wiki.jenkins-ci.org/display/JENKINS/Jenkins+Design+Language</url>
 
+    <dependencies>
+        <dependency>
+           <groupId>io.jenkins.blueocean</groupId>
+           <artifactId>blueocean-commons</artifactId>
```
kzantow: Why is this added? JDL should not depend on this.

recena: @kzantow 

> if jenkins-design-language does not have any Maven dependency with core plugins of Blue Ocean, this plugin could be installed independently of the rest of Blue Ocean plugins. This would be an not desired behavior.

If this plugin is completely independent and has its own lifecycle, please move it out side of this multi module project and use a version naming not synced with Blue Ocean.

vivek: @recena Do not get this change. Yes this plugin has no dependency on other modules but its tied to blueocean release lifecycle and its intentional. 

>please move it out side of this multi module project and use a version naming not synced with Blue Ocean.

What motivates this change request?

michaelneale: @recena not sure if depending on commons will solve anything - it will just get people to install more unrelated plugins than just jdl

michaelneale:
the other changes other than the commons dep look great. 

@recena there are many other examples of "leaf" node plugins that may not necessarily make sense on their own (adding some behavior that doesn't make sense on its own - eg some workflow/pipeline ones, library plugins, and more)

vivek:
Closing this, not sure if its relevant.

PR #1638, opened 2018-01-31
[FIX JENKINS-44995] favorites can be a cause of slowness and they were broken and using redux

kzantow:
# Description

Refactor favorites to remove redux, fix many of the preloader problems, etc., etc.. To test just make sure favorites work in all cases. There might be an edge case with the default branch, not sure offhand how to fix that, but it's probably less of an issue than the performance is.

See [JENKINS-44995](https://issues.jenkins-ci.org/browse/JENKINS-44995).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


kzantow:
@vivek @sophistifunk PTAL

michaelneale:
is checkdeps still relevant? I guess so...

michaelneale:
but this looks awesome!

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/preload/FavoritesStatePreloader.java
```diff
@@ -57,30 +60,41 @@
...
+
+                // don't need this list when at pipeline pages
+                if (pipelineFullName != null) {
+                    return null;
+                }
 
-        if (jenkinsUser != null && organization != null) {
-            UserImpl blueUser = new UserImpl(organization, jenkinsUser);
-            BlueFavoriteContainer favoritesContainer = blueUser.getFavorites();
+                UserImpl blueUser = new UserImpl(organization, jenkinsUser, organization.getUsers());
```
kzantow: @vivek this was the crux of the problem causing the favorites to be re-fetched despite being on the page: URLs didn't match

vivek: I guess the reason we had it loaded on each page was because favorites were needed to be around on all pages and now we don't need it as we have the new `FavoriteListStatePreloader`?

blueocean-personalization/src/main/js/model/FavoriteStore.js
```diff
@@ -0,0 +1,128 @@
...
+
+        this.fetch(url, fetchOptions)
+        .then(() => {
+            const { _favoritesNames: f } = this;
+            const itemName = getItemName(item);
+            if (favorite) {
+                this._setFavoritesNames([...f, itemName]);
+            } else {
+                const idx = f.indexOf(itemName);
+                this._setFavoritesNames([...f.slice(0, idx), ...f.slice(idx + 1)]);
```
cliffmeyers: This might read slightly better as an array.filter

blueocean-personalization/src/main/js/model/FavoriteStore.js
```diff
@@ -0,0 +1,128 @@
...
+                this.fetch(url)
+                    .then(favorites => this._setFavorites(sortHelper.applyStandardSort(favorites)));
+            }
+        }
+        return this._favorites;
+    }
+
+    fetch(url, fetchOptions) {
+        return Fetch.fetchJSON(url, { fetchOptions })
+            .then(data => capabilityAugmenter.augmentCapabilities(data))
```
cliffmeyers: I believe the `augmentCapabilities` is baked into `Fetch.fetchJSON` by default

blueocean-personalization/package.json
```diff
@@ -50,33 +51,27 @@
...
+    "mobx": "2.6.0",
+    "mobx-react": "3.5.7",
     "react": "15.4.2",
     "react-addons-css-transition-group": "15.4.2",
     "react-dom": "15.4.2",
-    "react-router": "2.8.1",
-    "redux": "3.6.0",
-    "redux-thunk": "2.1.0",
-    "reselect": "2.5.4"
+    "react-i18next": "1.10.1",
```
cliffmeyers: is `react-i18next` actually required? I figured that would come in transitively from core?

michaelneale: might be ok to explicitly include it?

michaelneale:
cc @sophistifunk PTAL - I will also take a look tests

michaelneale:
hey @kzantow have analyised the test failures: 

* `testEditor`: fake news. Just flaky. Ignore it it will pass
* `io.jenkins.blueocean.config.JenkinsJSExtensionsTest.test()`: This is a real failure. @kzantow I can see some code changes to do with the assertions here, not sure what the intent was https://github.com/jenkinsci/blueocean-plugin/pull/1638/files#diff-0ac7543c805acf75a99db9a1185958d4L61 BUT the result is: 
`expected:<components/[DashboardCards]> but was:<components/[FavoritePipeline]>` - perhaps a typo in the test? 

* `testMultibranch – io.blueocean.ath.offline.personalization.FavoritesCardsTest` that is a real failure. Master doesn't seem to finish in the test: dashboardPage.checkFavoriteCardStatus(fullNameMaster, SUCCESS); (ie it is always blue - but should go green) - line 100

* On the other failure `testMultibranch – io.blueocean.ath.offline.personalization.FavoritesNavigationTest`
```
        // check the branches tab
        BranchPage branches = navigateBranches(pipeline)
            .checkFavoriteStatus(branchMaster, true)
```
fails. You can see this if you follow along - it doesn't show it up as favorited on he branches screen, so something isn't right with the data or store somehow.. 

@sophistifunk I suggest looking at the last to favorite failures. The other one @kzantow can probably sort out. Unfortunately to really see it you may need to run the ATH locally (I did) but if you just poke around favs by hand, you may see them not update correctly. I wasn't able to myself, but watching VNC the failures seem clear enough, but I am not quite sure what is going on. 

Investigation needed, but things aren't right.

michaelneale:
I can't work out why those tests fail - when if I manually run from the branch tab, it does update... all I can think is somehow the changes here affect the timing of things. I tried for far too long... but this changed something..

cliffmeyers:
@michaelneale did you test specifically with the master and non-master branches? I remember the logic with master being tricky at times.

michaelneale:
@cliffmeyers yes, tested by hand with both. In the ATH - it was the master branch that wouldn't update. Non master did (the one with the funky name) so it is kind of the opposite to the failure mode I was expecting. As best as I can follow the test, if I do the same steps by hand, it works.

NicuPascu:
I opened a PR to this branch for the testMultibranch ATH test https://github.com/kzantow/blueocean-plugin/pull/1

kzantow:
@NicuPascu thanks and merged; crossing my fingers

blueocean-personalization/src/main/js/jenkins-js-extension.yaml
```diff
@@ -2,8 +2,6 @@
 # NB: "component" currently maps to modules, not "symbols" so make sure to "export default"
 # WARNING: If you change this you'll have to change io.jenkins.blueocean.jsextensions.JenkinsJSExtensionsTest as well :(
 extensions:
-  - component: redux/FavoritesStore
```
kzantow: This is the problem causing `io.jenkins.blueocean.config.JenkinsJSExtensionsTest` to fail. We should just delete that test, it's useless.

michaelneale: yeah I agree. It made sense early on but not relevant now.

NicuPascu:
made another PR that fixes FavoritesNavigationTest.testMultibranch, this was also related to the test env history, could not replicate behaviour but when navigating to the dashboard instead of going back, the test passes https://github.com/kzantow/blueocean-plugin/pull/2

michaelneale:
@NicuPascu I think just make a new branch and new PR with your changes atop all this - see how that goes. Keith suggested getting rid of the extensions test case to get it to pass (it isn't testing anything useful now)

michaelneale:
nice one - eyes from @sophistifunk @imeredith possible?

blueocean-personalization/package.json
```diff
@@ -50,33 +51,27 @@
...
   },
   "jenkinscd": {
     "import": [
       "@jenkins-cd/blueocean-core-js@any",
       "@jenkins-cd/design-language@any",
       "@jenkins-cd/js-extensions@any",
-      "immutable@any",
-      "keymirror@any",
+      "mobx@any",
+      "mobx-react@any",
```
sophistifunk: /me *wipes a single tear of joy*

blueocean-personalization/src/main/js/model/FavoriteStore.js
sophistifunk: Faster, too. But not a big deal.

sophistifunk:
I don't have much of substance to add, but LGTM 👍

PR #1639, opened 2018-02-01
[JENKINS-49272] Use node.type==PARALLEL to determine if a node is par…

imeredith:
…allel

# Description

See [JENKINS-49272](https://issues.jenkins-ci.org/browse/JENKINS-49272).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/acceptance-tests/src/test/java/io/blueocean/ath/offline/multibranch/ParallelNavigationTest.java
+++ b/acceptance-tests/src/test/java/io/blueocean/ath/offline/multibranch/ParallelNavigationTest.java
@@ -70,9 +70,11 @@ public class ParallelNavigationTest {
         // At first we see branch one
         wait.until(By.xpath("//*[text()=\"firstBranch\"]"));
         logger.info("Found first branch");
+        wait.until(By.xpath("//*[text()=\"first branch visible\"]"));
         // and clicking on the unselected node will yield us the second branch
         wait.click(By.xpath("//*[contains(@class, 'pipeline-node')][3]"));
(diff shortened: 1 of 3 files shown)


michaelneale:
very nice @imeredith great catch. :bee:

(I am looking at the ATH to catch this in future)

michaelneale:
@imeredith ok - well I looked at the ATH - it does exercise a test very much like this problem, specifically for this case. However, both on master as well as this branch it runs fine. However clicking as a human is not fine. 

the line: 

```
wait.click(By.xpath("//*[contains(@class, 'pipeline-node')][3]"));
``` 

seems to click it successfully - but somehow this is not the case in reality - so I am at a bit of a loss regarding the ATH here. 

I think this is worthwhile merging and cherry picking for a new release as this is obviously a bad problem and needs to be addressed ASAP. 

I will keep thinking about the ATH problem

michaelneale:
ok - I found it. The test was silly - was just looking for "secondBranch" text() - which the node would satisfy. I fixed it to search for a unit step title. 

I checked this fails on the master branch, and passes here, so if this goes green - we can be good to go.

michaelneale:
@vivek @imeredith if this is all green - can someone merge this, and @vivek do a 1.4.1 release? as this is a rather annoying one.

vivek:
@michaelneale sounds good. Thanks.

michaelneale:
@vivek small one to cherry-pick to `release/1.4` I think...

PR #1644, opened 2018-02-05
Use the right ExtensionPoint class

batmat:
Sezpoz actually doesn't care, but well let's reduce the WTF/line ratio :P.

I didn't file a JIRA since this is a small and pretty obvious fix. But if you feel this ought to have one, please let me know.

@reviewbybees


michaelneale:
woah - how on earth did this happen ;)

looks ok @vivek ?

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/BitbucketApiFactory.java
```diff
@@ -14,7 +14,7 @@
  *
  * @author Vivek Pandey
  */
-public abstract class BitbucketApiFactory extends ExtensionPoint{
+public abstract class BitbucketApiFactory implements ExtensionPoint {
```
vivek: Wow! Thanks @batmat, I think luckily it work because lookup is not done using ExtensionPoint, It’s done using BitbucketApiFactory.

batmat:
You're very welcome!

PR #1645, opened 2018-02-05
1.4 cherry pick: JENKINS-48198 - Add logging when a property is skipped (#1625)

michaelneale:
# Description
A cherry pick from https://github.com/jenkinsci/blueocean-plugin/pull/1625

cc @vivek @sophistifunk I think this makes sense for a 1.4.1 release.


sophistifunk:
LGTM

PR #1647, opened 2018-02-07
JENKINS-47022 - Fix for IE dropdown losing focus on selection

NicuPascu:
# Description
This uses the proprietary IE setActive method to restore focus to the dropdown button element after applying the selection

See [JENKINS-47022](https://issues.jenkins-ci.org/browse/JENKINS-47022).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/jenkins-design-language/src/js/components/forms/Dropdown.jsx
+++ b/jenkins-design-language/src/js/components/forms/Dropdown.jsx
@@ -265,6 +265,11 @@ export class Dropdown extends React.Component {
         if (this.props.onChange) {
             this.props.onChange(option, index);
         }
+
+        //restore the focus on the button element in IE
+        if (this.buttonRef.setActive) {
+            this.buttonRef.setActive();
+        }
     }
 
     _optionToLabel(option) {


jenkins-design-language/src/js/components/forms/Dropdown.jsx
```diff
@@ -265,6 +265,11 @@ export class Dropdown extends React.Component {
         if (this.props.onChange) {
             this.props.onChange(option, index);
         }
+
+        //restore the focus on the button element in IE
+        if (this.buttonRef.setActive) {
```
vivek: Should we have check if the browser in context is IE or I guess there is no side effect for other browsers?

NicuPascu: setActive is an IE only method, that is why I checked for its existence before calling it, this ensures it runs only in IE

sophistifunk:
Code looks good to me

bee-movie-bee

jenkins-design-language/src/js/components/forms/Dropdown.jsx
scherler: @NicuPascu I just tested with firefox and chrome and the loose the focus in showing there as well. Trying to say I think you need to change the above to 

```
if (this.buttonRef && this.buttonRef.focus) {
  this.buttonRef.focus();
}
```

This will further prevent possible NPE like `undefined.focus`

NicuPascu:
I've opened a new PR (https://github.com/jenkinsci/blueocean-plugin/pull/1652) which fixes the focus problem in all browsers

PR #1653, opened 2018-02-12
JENKINS-49519 Limit the amount of analytics data being collected

i386:
# Description

This config kept on jenkins.io. It expresses that there are 1000 cohorts, of which 10 are active (1% of the Jenkins server install base).

```
{
  "cohorts": 1000,
  "activeCohorts": [
    "0-9"
  ]
}
```

The server will decide based on its server identity if it is enrolled into the range 0-9 or not.

See https://github.com/jenkins-infra/jenkins.io/pull/1387 for the configuration file
See [JENKINS-49519](https://issues.jenkins-ci.org/browse/JENKINS-49519).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/analytics/AbstractAnalytics.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/analytics/AbstractAnalytics.java
@@ -51,6 +51,7 @@ public abstract class AbstractAnalytics extends Analytics {
         AnalyticsConfig config = AnalyticsConfigService.get();
         if (!belongsToActiveCohort(config, server)) {
             LOGGER.log(Level.FINE, "Server is not in active cohort. No analytics event has been sent");
+            return;
         }
         Map<String, Object> allProps = req.properties == null ? Maps.newHashMap() : Maps.newHashMap(req.properties);
         // Enhance with additional properties
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/analytics/AbstractAnalytics.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/analytics/AbstractAnalytics.java
@@ -50,7 +50,7 @@ public abstract class AbstractAnalytics extends Analytics {
         String server = server();
         AnalyticsConfig config = AnalyticsConfigService.get();
         if (!belongsToActiveCohort(config, server)) {
-            LOGGER.log(Level.INFO, "Server is not in active cohort. No analytics event has been sent");
+            LOGGER.log(Level.FINE, "Server is not in active cohort. No analytics event has been sent");
         }
         Map<String, Object> allProps = req.properties == null ? Maps.newHashMap() : Maps.newHashMap(req.properties);
         // Enhance with additional properties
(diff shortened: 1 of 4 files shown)


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/analytics/AbstractAnalytics.java
```diff
@@ -42,6 +47,11 @@ public void track(TrackRequest req) {
         if (StringUtils.isEmpty(req.name)) {
             throw new ServiceException.BadRequestException("missing name");
         }
+        String server = server();
+        AnalyticsConfig config = AnalyticsConfigService.get();
+        if (!belongsToActiveCohort(config, server)) {
+            LOGGER.log(Level.INFO, "Server is not in active cohort. No analytics event has been sent");
```
daniel-beck: This looks like massive log spam for the majority of users, as INFO is visible by default…?

i386: Yep, that should be `Level.FINE`

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/analytics/AbstractAnalytics.java
```diff
@@ -42,6 +47,11 @@ public void track(TrackRequest req) {
         if (StringUtils.isEmpty(req.name)) {
             throw new ServiceException.BadRequestException("missing name");
         }
+        String server = server();
+        AnalyticsConfig config = AnalyticsConfigService.get();
+        if (!belongsToActiveCohort(config, server)) {
+            LOGGER.log(Level.FINE, "Server is not in active cohort. No analytics event has been sent");
```
vivek: @i386 Looks like when `! belongsToActiveCohort()` is true then it should skip sending events. Missing return here?

vivek:
@rtyler PTAL. This has impact on how we an collect stats over to keen.io by setting up configuration at jenkins.io?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/analytics/AnalyticsConfigService.java
```diff
@@ -0,0 +1,62 @@
...
+            config = loadAnalyticsConfig();
+        } catch (Throwable e) {
+            config = AnalyticsConfig.EMPTY;
+            LOGGER.log(Level.SEVERE, "Could not load analytics configuration", e);
+        }
+        return config;
+    }, 1, TimeUnit.DAYS);
+
+    private static AnalyticsConfig loadAnalyticsConfig() throws IOException {
+        URL url = new URL("https://jenkins.io/projects/blueocean/analytics.json");
```
rtyler: I believe this should be using ANALYTICS_CONFIG_URL

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/analytics/AnalyticsConfigService.java
```diff
@@ -0,0 +1,62 @@
...
+    private static final Logger LOGGER = Logger.getLogger(AnalyticsConfigService.class.getName());
+
+    private static final String ANALYTICS_CONFIG_URL = "https://jenkins.io/projects/blueocean/analytics.json";
+
+    public static AnalyticsConfig get() {
+        return configSupplier.get();
+    }
+
+    private static Supplier<AnalyticsConfig> configSupplier = Suppliers.memoizeWithExpiration(() -> {
+        AnalyticsConfig config;
```
rtyler: Would I be correct in assuming that this method is going to be executed only once per instance, on boot?

i386: Its memoized so that the stored value is refreshed every day

PR #1658, opened 2018-02-14
JENKINS-46706: Let Default UI be selectable

imeredith:
# Description

@michaelneale Can you please give this a whirl and see if its what you would expect.

See [JENKINS-46706](https://issues.jenkins-ci.org/browse/JENKINS-46706).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


imeredith:
@vivek Honestly I know know where it should go. I don't really want to make a new plugin which was my first impulse.

michaelneale:
sorry having trouble building locally 

```
[ERROR] npm ERR! Darwin 17.4.0
[ERROR] npm ERR! argv "/Users/michaelneale/projects/paas/blueocean-plugin/blueocean-dashboard/node/node" "/Users/michaelneale/projects/paas/blueocean-plugin/blueocean-dashboard/node/node_modules/npm/bin/npm-cli.js" "install"
[ERROR] npm ERR! node v6.4.0
[ERROR] npm ERR! npm  v3.10.3
[ERROR] npm ERR! code Z_BUF_ERROR
[ERROR] npm ERR! errno -5
[ERROR] 
[ERROR] npm ERR! unexpected end of file
``` - so need to clean out again..

**Comment by michaelneale** [comment]
ok functionally seems ok, some Q's: 

* If it is per user (is it?) is the settings screen the right place? (might be) - but when I tried it was all users  (not per user - is that right?)
* ATH tests appropriate for this? 
* Is there a URI switch to disable it temporarily? 
* Should the exit button on the blue ocean screen use that URI flag so it always goes to classic? 
* Should there be a global system property that makes this apply for all users? (maybe a future PR...)

Otherwise - if you can clarify the above, may be ok as it is very much opt in to start with. Nice work so far!

**Comment by imeredith** [comment]
URI switch is a good idea.

User settings exist. They override the global setting.

I actually think the correct test is a backend test now that I think about it. And it saves me having to figure out how to configure the properties without using jenkins classic

**Comment by michaelneale** [comment]
ok the on :bug: I can find is that if I use the setting in global to set it to blue - I can't override it to be Classic per user - is that by design? If so - probably should make that clear somehow ... (not sure how) or it is a :bug:

**Comment by imeredith** [comment]
@vivek @michaelneale ok i think its working properly now w/ tests

File: blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/InterfaceOption.java [hunk 1]
```diff
@@ -0,0 +1,22 @@
+package io.jenkins.blueocean.service.embedded.redirect;
+
+public class InterfaceOption {
+    public static final InterfaceOption classic = new InterfaceOption("classic", "Jenkins Classic");
```
jtnord: NIT: psf (or immutables depending on your view) should be UPPER_CASE

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/RedirectFilter.java
```diff
@@ -0,0 +1,66 @@
...
+        //noop
+    }
+
+    @Override
+    public void destroy() {
+        //noop
+    }
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
```
jtnord: I have the feeling this will make crappy performance even more crappy

evey we request will go thought this - even if the user is explicitly requesting a blue or a classic UI

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/RedirectFilter.java
```diff
@@ -0,0 +1,66 @@
...
+                defaultUI = property.getInterfaceId();
+            }
+        }
+
+        if(InterfaceOption.blueocean.getInterfaceId().equals(defaultUI) && request instanceof HttpServletRequest) {
+            HttpServletRequest httpRequest = (HttpServletRequest) request;
+
+            String noRedirect = httpRequest.getParameter("noDefaultRedirect");
+            if(noRedirect == null && "/".equals(httpRequest.getPathInfo())) {
+                ((HttpServletResponse) response).sendRedirect(httpRequest.getContextPath() + "/blue");
```
jtnord: you are doing all the lookup above, when you only want it for `/`  surely that is the first check you need to do?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/RedirectFilter.java
```diff
@@ -0,0 +1,66 @@
...
+            DefaultUserInterfaceUserProperty property = user.getProperty(DefaultUserInterfaceUserProperty.class);
+            if(property != null && !property.getInterfaceId().equals(DefaultUserInterfaceUserProperty.system.getInterfaceId())) {
+                defaultUI = property.getInterfaceId();
+            }
+        }
+
+        if(InterfaceOption.blueocean.getInterfaceId().equals(defaultUI) && request instanceof HttpServletRequest) {
+            HttpServletRequest httpRequest = (HttpServletRequest) request;
+
+            String noRedirect = httpRequest.getParameter("noDefaultRedirect");
```
jtnord: this never gets set on any page AFAICT.   if the user hates blue how can they get to classic to actually set this.  BO does not add user properties or management - so a user would logically go to the root url - at which point they will be redirected back.  Seems like an imensly sucky UX.  :bug; by my current understanding

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/RedirectFilter.java
vivek: Good point. `doFilter` is reading user property and configuration, probably off file system. Is there listener support for configuration objects that we can use to update in memory config objects - that can probably answer perf concern?

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/RedirectFilter.java
imeredith: The back to classic button in blueocean uses this flag.

imeredith:
@jtnord I've rearranged the code so that it checks to see if its at '/' before looking up the user.

Criticisms about config and default landing pages still apply. Although its an option that has to be turned on, its not on by default. And in the ticket people were already doing this with nginx so this has to better than at least that. We could also have config or a system prop to turn the system on to eliminate any performance issues for people that dont want it. Or just make it an optional plugin.

michaelneale:
@jtnord did @imeredith's latest changes help?

michaelneale:
functionality wise - this seems ok IF this is the way we want it to work. Would like to know if @jtnord thinks that it won't have bad impact things by looking up the user. 

Another option is a systemwide flag if people aren't happy with that, but kudos @imeredith for the tests!

i386:
@imeredith would you mind demoing this to me before merging to master? Any time in my calendar is fine.

michaelneale:
@i386 I could show you tomorrow

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/RedirectFilter.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Override
+    public void destroy() {
+        //noop
+    }
+
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException, ServletException {
+        HttpServletRequest request1 = (HttpServletRequest) request;
+        if (request instanceof HttpServletRequest && "/".equals(request1.getPathInfo()) && request.getParameter("noDefaultRedirect") == null) {
```
amuniz: Too late, cast already done the line above. The `instanceof` should be done before.

michaelneale: YIKES... good catch. That would be ugly blowing up.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/InterfaceOption.java
```diff
@@ -0,0 +1,22 @@
...
+    public static final InterfaceOption blueocean = new InterfaceOption("blueocean", "Blue Ocean");
+
+    private final String interfaceId;
+    private String name;
```
agentgonzo: NIT: name could also be final

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/RedirectFilter.java
```diff
@@ -0,0 +1,62 @@
...
+            User user = User.current();
+            if (user != null) {
+                DefaultUserInterfaceUserProperty property = user.getProperty(DefaultUserInterfaceUserProperty.class);
+                if (property != null && !property.getInterfaceId().equals(DefaultUserInterfaceUserProperty.system.getInterfaceId())) {
+                    defaultUI = property.getInterfaceId();
+                }
+            }
+
+            if (InterfaceOption.blueocean.getInterfaceId().equals(defaultUI)) {
+                ((HttpServletResponse) response).sendRedirect(request1.getContextPath() + "/blue");
```
agentgonzo: NIT: This "/blue" would be nicer as a constant at the top of the file.

blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/redirect/RedirectFilterTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+        Assert.assertEquals(200, response.getStatus());
+
+        user.addProperty(new DefaultUserInterfaceUserProperty("blueocean"));
+        user.save();
+        response = Unirest.get(j.jenkins.getRootUrl())
+            .basicAuth("alice", "alice")
+            .asString();
+
+        Assert.assertEquals(302, response.getStatus());
```
agentgonzo: NIT: Probably worth asserting the Location header in the response here too

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/RedirectFilter.java
agentgonzo: Is it possible to use a filter chain pattern to do this URL matching (https://docs.spring.io/spring-security/site/docs/3.0.x/reference/security-filter-chain.html - `    <url-pattern>/*</url-pattern>`? That way, you don't have to have the ugly path matching here and it would alleviate @jtnord 's concerns that the filter gets executed on every page.

agentgonzo: --> Not knowing how Jenkins/plugins interact with HttpServlets etc, this SO question _may_ help if we don't have access to the web.xml file: https://stackoverflow.com/questions/35434978/how-to-specify-url-pattern-for-servlet-filters-in-webapplicationinitializer

jtnord: The way that [PluginServletFilter](http://javadoc.jenkins.io/index.html?hudson/util/PluginServletFilter.html) is implemented and stapler I do not believe it is possible to add `Filters` to specific paths.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/RedirectFilter.java
```diff
@@ -0,0 +1,62 @@
...
+
+            User user = User.current();
+            if (user != null) {
+                DefaultUserInterfaceUserProperty property = user.getProperty(DefaultUserInterfaceUserProperty.class);
+                if (property != null && !property.getInterfaceId().equals(DefaultUserInterfaceUserProperty.system.getInterfaceId())) {
+                    defaultUI = property.getInterfaceId();
+                }
+            }
+
+            if (InterfaceOption.blueocean.getInterfaceId().equals(defaultUI)) {
```
jtnord: So I think there are a couple of :bug:s with this approach.

1)  agents / CLI in order to lookup the JNLP port perform a HEAD to `/` to get the CLI/JNLP host/port settings.[Functions.java](https://github.com/jenkinsci/jenkins/blob/006c51256e58d9b691956962d9fd624fb0c1e08b/core/src/main/java/hudson/Functions.java#L2037-L2053)  and this occurs *after* filter processing - so old agents (will not be able to connect).  OR even just looking for the `X-Jenkins` header to confirm it is actually a Jenkins instance.
[JnlpAgentEndpointResolver.java](https://github.com/jenkinsci/remoting/blob/563bff81f5639db0aad501f784a97e8af6742541/src/main/java/org/jenkinsci/remoting/engine/JnlpAgentEndpointResolver.java#L179-L202)

2) How does this interact with the `UpgradeWizard`  I believe users will no longer ever see this if the default is Blue.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/RedirectFilter.java
agentgonzo: Boooo - that's a shame

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/RedirectFilter.java
imeredith: Does the upgrade wizard show when you upgrade jenkins? I thought it was just on fresh install?

imeredith:
I'm not really sure where to go with this.

michaelneale:
@jtnord shouldn't the usual apply and it is just  redirect? (ie same headers).

michaelneale:
after some discussion, we can shelve this now - as in the ticket, people are using nginx and other things to good effect right now. 

This code won't rot, and if there is stronger/clearer demand, can bring it back (or it could be an optional plugin, as suggested).

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/redirect/RedirectFilter.java
jtnord: the "setup wizzard" is triggered in a few scenarios.  One is you are a new install,  the other is the plugins json has marked a new plugin as a "new feature"

this is done by adding an "added" with the version that added this (which gives you just the plugin install wizard)  This has not yet been used in open source AFAICT

galek:
Status of PR?

bitwiseman:
@galek  Stalled.

PR #1665, opened 2018-02-20
Fix JavaDoc warnings

sophistifunk:
* Fix up the warnings causing Javadoc to kill the build

# Description

https://issues.jenkins-ci.org/browse/JENKINS-49477

### DIFF
--- a/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/ExportInterceptor.java
+++ b/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/ExportInterceptor.java
@@ -32,10 +32,11 @@ public abstract class ExportInterceptor {
      *
      * @param property to get the value from model object
      * @param model object with this property
+     * @param config current configuration to control marshalling / exception-handling behaviour
      * @return the value of the property, if {@link #SKIP} is returned, this property will be skipped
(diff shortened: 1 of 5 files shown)


michaelneale:
thoughts @vivek ?

blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/export/ExportInterceptor.java
```diff
@@ -32,10 +32,11 @@
      *
      * @param property to get the value from model object
      * @param model object with this property
+     * @param config current configuration to control marshalling / exception-handling behaviour
      * @return the value of the property, if {@link #SKIP} is returned, this property will be skipped
      * @throws IOException if there was a problem with serialization that should prevent
      *         the serialization from proceeding
-     * @see Exported#skipNull()
+     * @see org.kohsuke.stapler.export.Exported#skipNull()
```
vivek: We do not copy `Exported` as we continue using `org.kohsuke.stapler.export.Exported` in our external interface/model classes. There is not much we can do other than keep this fix locally here till we figure out how excluded packages for javadoc run works with java8.

sophistifunk:
No need for this RN it seems.

PR #1666, opened 2018-02-20
Hide BlueOceanExportInterceptor so Javadoc can ignore .export package

sophistifunk:
# Description

Hide BlueOceanExportInterceptor so Javadoc can ignore .export package

### DIFF
--- a/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/Export.java
+++ b/blueocean-commons/src/main/java/io/jenkins/blueocean/commons/stapler/Export.java
@@ -150,7 +150,7 @@ public class Export {
 
     private Export() {};
 
-    public static class BlueOceanExportInterceptor extends ExportInterceptor{
+    private static class BlueOceanExportInterceptor extends ExportInterceptor{
         @Override
         public Object getValue(Property property, Object model, ExportConfig config) throws IOException {
             if(model instanceof Action){


sophistifunk:
Pain to bytes ratio is *way out* on this one

michaelneale:
@sophistifunk WOW - so this was it? 
And this one even attempted to take out your hard drive? the javadoc thing REALLY hates everyone doesn't it. 


cc @vivek @raul-arabaolaza

sophistifunk:
I blame Oracle for whatever the hell Idea happened to do to my kernel that required a reboot and triggered the drive hiccups. I've never seen `ps aux` give a stat of `?E` before.

michaelneale:
@sophistifunk I have seen all sorts of weird things in the early days of AWS EBS, when things would flame out in the most novel ways. 

/me stares into the jungle

sophistifunk:
...something, something Tannhauser Gate....

sophistifunk:
Re-running build

sophistifunk:
Different ATH test failing, so re-running again....

michaelneale:
OK with me if ok with @vivek - no chance anyone external is depending on this I would hope?

vivek:
@sophistifunk Great sleuthing! This is best solution. This class is only for internal use so all good. Tried locally, works well. Also tested case where there is  java doc error in non-stapler blueocean code in blueocean-commins, the doc error is reported. 

LGTM 🐝

PR #1667, opened 2018-02-20
JENKINS-48820# Fix for stack overflow in flownode sse event generation

vivek:
# Description

Fix for  stack overflow error during pipeline execution SSE event generation, has other minor improvements.

See [JENKINS-48820](https://issues.jenkins-ci.org/browse/JENKINS-48820).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
+++ b/blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
@@ -7,7 +7,7 @@ import io.jenkins.blueocean.rest.impl.pipeline.PipelineInputStepListener;
 import io.jenkins.blueocean.rest.impl.pipeline.PipelineNodeUtil;
 import java.io.IOException;
 import java.util.ArrayDeque;
-import java.util.Iterator;
+import java.util.Collection;
 import java.util.Map;
 import java.util.WeakHashMap;
 import java.util.logging.Level;
@@ -60,20 +60,20 @@ public class PipelineEventListener implements GraphListener {
             ArrayDeque<String> branch = getBranch(flowNode);
             currentStageName.put(flowNode.getExecution(), flowNode.getDisplayName());
             currentStageId.put(flowNode.getExecution(), flowNode.getId());
-            publishEvent(newMessage(PipelineEventChannel.Event.pipeline_stage, flowNode, branch.iterator()));
+            publishEvent(newMessage(PipelineEventChannel.Event.pipeline_stage, flowNode, branch));
         } else if (flowNode instanceof StepStartNode) {
             if (flowNode.getAction(BodyInvocationAction.class) != null) {
                 ArrayDeque<String> branch = getBranch(flowNode);
                 branch.add(flowNode.getId());
-                publishEvent(newMessage(PipelineEventChannel.Event.pipeline_block_start, flowNode, branch.iterator()));
+                publishEvent(newMessage(PipelineEventChannel.Event.pipeline_block_start, flowNode, branch));
             } else if (flowNode.getPersistentAction(QueueItemAction.class) != null) {
                 // Make sure we fire an event for the start of node blocks.
                 ArrayDeque<String> branch = getBranch(flowNode);
-                publishEvent(newMessage(PipelineEventChannel.Event.pipeline_step, flowNode, branch.iterator()));
+                publishEvent(newMessage(PipelineEventChannel.Event.pipeline_step, flowNode, branch));
             }
         } else if (flowNode instanceof StepAtomNode) {
             ArrayDeque<String> branch = getBranch(flowNode);
-            publishEvent(newMessage(PipelineEventChannel.Event.pipeline_step, flowNode, branch.iterator()));
+            publishEvent(newMessage(PipelineEventChannel.Event.pipeline_step, flowNode, branch));
         } else if (flowNode instanceof StepEndNode) {
             if (flowNode.getAction(BodyInvocationAction.class) != null) {
                 FlowNode startNode = ((StepEndNode) flowNode).getStartNode();
@@ -82,7 +82,7 @@ public class PipelineEventListener implements GraphListener {
                 ArrayDeque<String> branch = getBranch(startNode);
 
                 branch.add(startNodeId);
-                publishEvent(newMessage(PipelineEventChannel.Event.pipeline_block_end, flowNode, branch.iterator()));
+                publishEvent(newMessage(PipelineEventChannel.Event.pipeline_block_end, flowNode, branch));
             }
         } else if (flowNode instanceof FlowEndNode) {
             publishEvent(newMessage(PipelineEventChannel.Event.pipeline_end, flowNode.getExecution()));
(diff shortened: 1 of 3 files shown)


vivek:
@svanoort I think I am picking up right API to get immediate parent node. PTAL.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/FlowNodeWrapper.java
```diff
@@ -203,6 +204,9 @@ public void setBlockErrorAction(ErrorAction blockErrorAction) {
      * Filters by class to mimic Item.getActions(class).
      */
     public <T extends Action> Collection<T> getPipelineActions(Class<T> clazz) {
+        if(pipelineActions == null){
```
vivek: Fixes NPE happening in cases where there are no actions added.

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
```diff
@@ -58,77 +57,63 @@
...
-                return grandparent;
-            }
-        }
-
-        return null;
+        return branch;
     }
 
-    private String toPath(List<String> branch) {
+    private String toPath(Iterator<String> branch) {
```
svanoort: 🐜 Seems like we could just pass an `Iterable` rather than a one-shot iterator?

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
```diff
@@ -164,7 +149,7 @@ private static Message newMessage(PipelineEventChannel.Event event, FlowExecutio
         return message;
     }
 
-    private Message newMessage(PipelineEventChannel.Event event, FlowNode flowNode, List<String> branch) {
+    private Message newMessage(PipelineEventChannel.Event event, FlowNode flowNode, Iterator<String> branch) {
```
svanoort: 🐜  Again, perhaps we can do an Iterable or collection rather than an Iterator?

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
```diff
@@ -58,77 +57,63 @@
...
-        List<FlowNode> parents = flowNode.getParents();
+    private ArrayDeque<String> getBranch(FlowNode flowNode) {
+        ArrayDeque<String> branch = new ArrayDeque<>();
+        String parentBlockId = flowNode.getEnclosingId();
 
-        for (FlowNode parent : parents) {
-            if (parent instanceof StepStartNode) {
-                if (parent.getAction(BodyInvocationAction.class) != null) {
-                    return parent;
+        while (parentBlockId != null) {
```
svanoort: I think what you want is `myFlowNode.getEnclosingBlocks()` which returns a list of blocks -- perhaps combined with Guava or Java 8 stream/filter APIs, perhaps via  `new NodeStepTypePredicate('parallel'))` or `FlowScanningUtils.hasActionPredicate(SomeAction.class))` -- it'll do all these goodies for you.

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
svanoort: Or a collection.

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
vivek: @svanoort I looked at `myFlowNode.getEnclosingBlocks()`, it doesn't return blocks in order. This code was not written to me and I want to keep changes to minimum as it might break UI. Previous code was making sure preserving certain order, so I would prefer leaving the code with  fixing SOF problem with minor enhancements.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/FlowNodeWrapper.java
sophistifunk: Cheers!

sophistifunk:
Was this SO due to an infinite loop, or just by a complicated pipeline?

vivek:
@sophistifunk caused by large number of steps in pipelines about 4600+ as reported in https://issues.jenkins-ci.org/browse/JENKINS-48820.

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
svanoort: @vivek Could we just do a `Lists.reverse`, to maintain the order? It's iterating in inside-out order here so I think you want to simply invert that.

vivek: @svanoort Right. I was not planning to use was due to this javadoc of `GraphLookupView.findAllEnclosingBlockStarts() ` , https://github.com/jenkinsci/workflow-api-plugin/blob/a20f21fddc6831f51c3ceeefd3ed1b744b899e98/src/main/java/org/jenkinsci/plugins/workflow/graph/GraphLookupView.java#L55 that says order is not guaranteed. After clarification offline with you, I am fine using it. Please open  a PR to fix the Javadoc to make it clear that inside-out order will be maintained.

vivek:
@svanoort PTAL

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
```diff
@@ -90,45 +89,14 @@ public void onNewHead(FlowNode flowNode) {
...
-            FlowNode grandparent = getParentBlock(parent);
-            if (grandparent != null) {
-                return grandparent;
-            }
-        }
-
-        return null;
+    /* package: so that we can unit test it */ List<String> getBranch(FlowNode flowNode) {
+        return Lists.reverse(flowNode.getEnclosingBlocks().stream()
+                .map(FlowNode::getId).collect(Collectors.toList()));
```
svanoort: Oh hey, if you just wanted the IDs there's `node.getAllEnclosingIds()`

Please don't kill me :)  I've got APIs to do pretty much all the things you'd want to do with FlowNodes at this point...

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
```diff
@@ -90,45 +89,14 @@ public void onNewHead(FlowNode flowNode) {
...
     }
 
-    private String toPath(List<String> branch) {
+    private String toPath(Collection<String> branch) {
         StringBuilder builder = new StringBuilder();
-        for (String leaf : branch) {
+        for(String leaf: branch){
```
svanoort: 🐜  Can this path creation with just `branch.stream().collect(Collectors.joining("/"))` if you're feeling all Streams-y :)

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
vivek: Ha ha. This is great. Let me make this change.

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
vivek: Right, there are many places in the code we need to change to adopt java 8, keeping the scope limited to just this fix, so be making such changes some other time.

PR #1676, opened 2018-02-27
JENKINS-45334 go to classic link missing in invalid input param

NicuPascu:
# Description
This fixes the link to classic in the case of an invalid input param

See [JENKINS-45334](https://issues.jenkins-ci.org/browse/JENKINS-45334).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/karaoke/components/InputStep.jsx
+++ b/blueocean-dashboard/src/main/js/components/karaoke/components/InputStep.jsx
@@ -99,7 +99,7 @@ export default class InputStep extends Component {
 
     render() {
         const { parameters } = this.parameterService;
-        const { run, branch, pipeline } = this.props;
+        const { classicInputUrl } = this.props;
 
         // Early out
         if (!parameters) {
@@ -107,16 +107,9 @@ export default class InputStep extends Component {
         }
 
         const sanity = parameters.filter(parameter => supportedInputTypesMapping[parameter.type] !== undefined);
-        logger.debug('sanity check', sanity.length, parameters.length, this.props.classicInputUrl);
+        logger.debug('sanity check', sanity.length, parameters.length, classicInputUrl);
         if (sanity.length !== parameters.length) {
             logger.debug('sanity check failed. Returning Alert instead of the form.');
-
-            let classicInputUrl;
-            if (pipeline.branchNames) {
-                classicInputUrl = `${rootPath(pipeline.fullName)}job/${encodeURIComponent(branch)}/${encodeURIComponent(run.id)}/input`;
-            } else {
-                classicInputUrl = `${rootPath(pipeline.fullName)}${encodeURIComponent(run.id)}/input`;
-            }
 
             const alertCaption = [
                 <p>{translate('inputStep.error.message')}</p>,
@@ -162,9 +155,7 @@ const { object, shape } = PropTypes;
 InputStep.propTypes = {
     step: shape().isRequired,
     classicInputUrl: object,
-    pipeline: PropTypes.object,
-    branch: PropTypes.string,
-    run: PropTypes.object,
+    classicInputUrl: PropTypes.string,
 };
 
 InputStep.contextTypes = {
--- a/blueocean-dashboard/src/main/js/components/karaoke/components/InputStep.jsx
+++ b/blueocean-dashboard/src/main/js/components/karaoke/components/InputStep.jsx
@@ -110,7 +110,7 @@ export default class InputStep extends Component {
         logger.debug('sanity check', sanity.length, parameters.length, this.props.classicInputUrl);
         if (sanity.length !== parameters.length) {
             logger.debug('sanity check failed. Returning Alert instead of the form.');
-            
+
             let classicInputUrl;
             if (pipeline.branchNames) {
                 classicInputUrl = `${rootPath(pipeline.fullName)}job/${encodeURIComponent(branch)}/${encodeURIComponent(run.id)}/input`;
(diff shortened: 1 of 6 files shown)


blueocean-dashboard/src/main/js/components/karaoke/components/Step.jsx
```diff
@@ -130,7 +130,12 @@ export class Step extends Component {
             }}
             />);
         } else if (step.isInputStep) {
-            children = <InputStep {...{ step, key: 'step' }} />;
+            children = (<InputStep 
+                {...{ step, key: 'step' }}
+                pipeline={pipeline}
```
scherler: why not: 

```
{...{ step, key: 'step', pipeline, branch, run }}
```

blueocean-dashboard/src/main/js/components/karaoke/components/InputStep.jsx
```diff
@@ -8,6 +8,7 @@ import {
     StringUtil,
     logging,
 } from '@jenkins-cd/blueocean-core-js';
+import {rootPath} from '../../../util/UrlUtils';
```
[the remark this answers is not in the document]
NicuPascu: `buildClassicInputUrl` didn't build the correct URL and I was not sure it was not being used somewhere else and that is why I decided to not fix that function so I wouldn't break a functionality somewhere else. I also saw that in other places the URL is being built on the fly so I just replicated that behaviour (https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-dashboard/src/main/js/components/RunDetails.jsx#L46)

If you think that function is not being used anywhere I can fix and use that but my opinion is that given how few lines of code this requires, I would leave the code here for easier debugging in the future (and because it is only being used here, there is no re-usability), it is a little hard to track those functions down all the way in core-js.

blueocean-dashboard/src/main/js/components/karaoke/components/Step.jsx
NicuPascu: Sure, I'll make the change, I just really don't like that spread operator, it's just not easy to debug and looks weird, especially when used like this, I'll get used to it eventually

NicuPascu: and I just found out that the use of the spread operator in object literals is not even part of the spec yet, it is just an experimental proposal still subject to change (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Spread_in_object_literals). Is it a good idea to use it at all until it becomes part of the spec?

blueocean-dashboard/src/main/js/components/karaoke/components/InputStep.jsx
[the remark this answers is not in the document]
scherler: The problem with all this "other places the URL is being built" is that you have to scan the whole wrap in case something in the url pattern changes.  

https://github.com/jenkinsci/blueocean-plugin/search?utf8=%E2%9C%93&q=buildClassicInputUrl&type= it is ATM not used anymore, so I think the save bet is to fix the method and then use it.

blueocean-dashboard/src/main/js/components/karaoke/components/Step.jsx
scherler: I personally prefer the spread operator but you can use the other style as well but IMO you should choose either one and not mix those two.

NicuPascu: yeah you are right, will change it to one style tomorrow, thanks

blueocean-dashboard/src/main/js/components/karaoke/components/InputStep.jsx
```diff
@@ -149,6 +162,9 @@ const { object, shape } = PropTypes;
 InputStep.propTypes = {
     step: shape().isRequired,
     classicInputUrl: object,
+    pipeline: PropTypes.object,
```
scherler: I feel you should create the `classicInputUrl` in the calling component and you would not have to change this. If you prefer the current way then you should drop `classicInputUrl`  and the `this.props.classicInputUrl`  usage (only used now in comments)

blueocean-dashboard/src/main/js/components/karaoke/components/Pipeline.jsx
```diff
@@ -291,6 +291,9 @@ export default class Pipeline extends Component {
                        scrollToBottom={scrollToBottom}
                        router={router}
                        location={location}
+                       pipeline={pipeline}
```
scherler: here you should create `classicInputUrl` and pass it IMHO

blueocean-dashboard/src/main/js/components/karaoke/components/Step.jsx
sophistifunk: Spreading out a throwaway object literal in a JSX element is an abomination that we should never be doing, IMO.

sophistifunk:
LGTM if it's green 🐝

blueocean-dashboard/src/main/js/components/karaoke/components/InputStep.jsx
```diff
@@ -149,6 +154,7 @@ const { object, shape } = PropTypes;
 InputStep.propTypes = {
     step: shape().isRequired,
     classicInputUrl: object,
+    classicInputUrl: PropTypes.string,
```
scherler: that is duplicated @NicuPascu

NicuPascu: thanks for noticing, I'll make a new PR removing the duplicate code, thanks

PR #1677, opened 2018-02-27
Fix for flaky sse events tests

vivek:
# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-events/src/test/java/io/jenkins/blueocean/events/SseEventTest.java
+++ b/blueocean-events/src/test/java/io/jenkins/blueocean/events/SseEventTest.java
@@ -25,6 +25,7 @@ import jenkins.plugins.git.GitSCMSource;
 import jenkins.plugins.git.GitSampleRepoRule;
 import jenkins.scm.api.SCMSource;
 import org.jenkinsci.plugins.pubsub.ChannelSubscriber;
+import org.jenkinsci.plugins.pubsub.Events;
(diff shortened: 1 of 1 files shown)


blueocean-events/src/test/java/io/jenkins/blueocean/events/SseEventTest.java
```diff
@@ -505,24 +526,32 @@ public OrganizationImpl of(ItemGroup group) {
         }
     }
 
-    private void isEquals(String source, String target, AtomicInteger errorCount){
-        if(!source.equals(target)){
-            System.err.println(String.format("Source: %s is not equals to target: %s", source, target));
-            errorCount.incrementAndGet();
+    private static class AssertionHelper {
```
sophistifunk: What's the point of all this, if it calls `fail()`? If calling it doesn't end the test by throwing, can't we just rely on the test runner to count the errors during summation? If it does, why not just use normal asserts? I think I must be missing something here.

vivek: because, in async block unit assertions or fail doesn't fail the test. Reason I am calling fail is, so that looking at log we can find which test and where it failed.

sophistifunk:
Finally got this to make it through a full build without any weird jacoco errors. Given it was failing for me every time yesterday, I call it fixed :)

bbm-bueno

PR #1685, opened 2018-03-06
JENKINS-45634 Table action buttons hidden until hover

NicuPascu:
# Description

See [JENKINS-45634](https://issues.jenkins-ci.org/browse/JENKINS-45634).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/acceptance-tests/src/test/java/io/blueocean/ath/offline/personalization/FavoritesCardsTest.java
+++ b/acceptance-tests/src/test/java/io/blueocean/ath/offline/personalization/FavoritesCardsTest.java
@@ -8,7 +8,13 @@ import io.blueocean.ath.model.FreestyleJob;
 import io.blueocean.ath.model.MultiBranchPipeline;
 import org.apache.log4j.Logger;
 import org.eclipse.jgit.api.errors.GitAPIException;
+import org.junit.Assert;
 import org.junit.Test;
+import org.openqa.selenium.By;
+import org.openqa.selenium.JavascriptExecutor;
+import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.interactions.Actions;
 
 import javax.inject.Inject;
 import java.io.IOException;
@@ -31,6 +37,9 @@ public class FavoritesCardsTest extends AbstractFavoritesTest {
 
     @Inject
     ActivityPageFactory activityPageFactory;
+
+    @Inject
+    WebDriver driver;
 
     @Test
     public void testFreestyle() throws IOException {
@@ -70,7 +79,7 @@ public class FavoritesCardsTest extends AbstractFavoritesTest {
     }
 
     @Test
-    public void testMultibranch() throws IOException, GitAPIException {
+    public void testMultibranch() throws IOException, GitAPIException, InterruptedException {
         String branchOther = "feature/1";
 
         git.writeJenkinsFile(resources.loadJenkinsFile());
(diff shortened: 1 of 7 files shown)


sophistifunk:
LGTM. That's some hairy CSS, but isn't it always? 🐝

NicuPascu:
@kshultzCB I will try to write a unit test for this, I haven't tested hover states before, will see how it goes

acceptance-tests/src/test/java/io/blueocean/ath/offline/personalization/FavoritesCardsTest.java
```diff
@@ -100,6 +109,33 @@ public void testMultibranch() throws IOException, GitAPIException {
         dashboardPage.checkFavoriteCardStatus(fullNameMaster, SUCCESS);
         dashboardPage.checkFavoriteCardStatus(fullNameOther, SUCCESS);
 
+        //Check that favorite actions are hidden until pointer is hovering the row
+        JavascriptExecutor js = (JavascriptExecutor) driver;
+        Object backgroundColor;
+
+        backgroundColor = js.executeScript(
+            "return getComputedStyle(document.querySelectorAll('.actions-container')[0], ':after').getPropertyValue('background-color')"
```
imeredith: I feel like this could be made into something more reusable. Not needed for this PR but it would be nice

acceptance-tests/src/test/java/io/blueocean/ath/offline/personalization/FavoritesCardsTest.java
```diff
@@ -100,6 +109,33 @@ public void testMultibranch() throws IOException, GitAPIException {
...
+        Assert.assertNotEquals(backgroundColor, "rgba(0, 0, 0, 0)");
+
+        logger.info("move pointer over favorite card actions");
+
+        Actions action = new Actions(driver);
+        WebElement we = driver.findElement(By.xpath("//*[@class=\"pipeline-card success-bg-lite\"]"));
+        action.moveToElement(we).perform();
+
+        //Wait for the hover animation to finish
+        Thread.sleep(150);
```
imeredith: You should be using `wait` here. This is gonna fail at some point probably for a stupid reason. Alway better to poll for the change, rather than sleep

blueocean-dashboard/src/main/js/components/BranchDetailsRow.jsx
```diff
@@ -84,7 +77,7 @@ export class BranchDetailsRowRenderer extends Component {
                 <TableCell linkTo={runDetailsUrl}><CommitId commitId={commitId} /></TableCell>
                 <TableCell>{ runMessage }</TableCell>
                 <TableCell linkTo={runDetailsUrl}>{ completed }</TableCell>
-                { actionsCell }
+                <TableCell className="TableCell--actions">{ actions }</TableCell>
```
imeredith: why does this css class have a double dash in it? I know you didnt add it but seems weird.

NicuPascu: I am not a fan of double dashes either tbh, because it is part of the JTable components I think it was @sophistifunk decision to name it like this, I don't really mind it but I agree it looks a little strange

acceptance-tests/src/test/java/io/blueocean/ath/offline/personalization/FavoritesCardsTest.java
NicuPascu: will change the thread.sleep to a wait statement, thanks

acceptance-tests/src/test/java/io/blueocean/ath/offline/personalization/FavoritesCardsTest.java
NicuPascu: Unfortunately I haven't found any other way of asserting this hover state css... if you have any idea I don't mind changing things

imeredith: How you are doing it is fine, but a java method that takes a selector and spits out computed css could be useful

blueocean-dashboard/src/main/js/components/BranchDetailsRow.jsx
sophistifunk: Just trying to follow [a standard](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md).

Although I think I'm leaning more towards [this](http://ecss.io/).

acceptance-tests/src/test/java/io/blueocean/ath/offline/personalization/FavoritesCardsTest.java
kshultzCB: I've seen this pattern used [elsewhere in the ATH](https://github.com/jenkinsci/blueocean-plugin/blob/08b6f1b5adf5ca0f122cc9ddb6a1808915e5727c/acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/GithubEnterpriseCreationPage.java#L50):

`wait.until(ExpectedConditions.visibilityOf(githubEnterpriseCreationButton)).click();`

Maybe something along those lines?

sophistifunk: Seems that would pass as soon as the animation starts tho, rather than when it's done? 

I find quite frustrating all the uncertainty and "remoteness" of selenium testing.

imeredith: `wait.until` uses `FluentWait` which is basically just polling until something returns not null. Using this pattern anywhere you need to check something removes uncertainty. In this case, you can wait on the fact that the computed css is correct.

imeredith:
LGTM :bee:

PR #1688, opened 2018-03-06
[FIX-JENKINS-49948] vulnerabilities in bo js dependencies

scherler:
# Description

See [JENKINS-49948](https://issues.jenkins-ci.org/browse/JENKINS-49948).

Update vulnerable moment dep

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/acceptance-tests/src/main/java/io/blueocean/ath/AthModule.java
+++ b/acceptance-tests/src/main/java/io/blueocean/ath/AthModule.java
@@ -9,6 +9,7 @@ import io.blueocean.ath.factory.BranchPageFactory;
 import io.blueocean.ath.factory.ClassicPipelineFactory;
 import io.blueocean.ath.factory.FreestyleJobFactory;
 import io.blueocean.ath.factory.MultiBranchPipelineFactory;
+import io.blueocean.ath.factory.PullRequestsPageFactory;
 import io.blueocean.ath.factory.RunDetailsArtifactsPageFactory;
 import io.blueocean.ath.factory.RunDetailsPipelinePageFactory;
 import io.blueocean.ath.model.ClassicPipeline;
@@ -16,6 +17,7 @@ import io.blueocean.ath.model.FreestyleJob;
 import io.blueocean.ath.model.MultiBranchPipeline;
 import io.blueocean.ath.pages.blue.ActivityPage;
 import io.blueocean.ath.pages.blue.BranchPage;
+import io.blueocean.ath.pages.blue.PullRequestsPage;
 import io.blueocean.ath.pages.blue.RunDetailsArtifactsPage;
 import io.blueocean.ath.pages.blue.RunDetailsPipelinePage;
 import org.openqa.selenium.Dimension;
@@ -139,5 +141,9 @@ public class AthModule extends AbstractModule {
         install(new FactoryModuleBuilder()
             .implement(BranchPage.class, BranchPage.class)
             .build(BranchPageFactory.class));
+
+        install(new FactoryModuleBuilder()
+            .implement(PullRequestsPage.class, PullRequestsPage.class)
+            .build(PullRequestsPageFactory.class));
     }
 }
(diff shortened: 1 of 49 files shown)


scherler:
ShrinkCRAP ;) fixed

kzantow:
Yeah, I hate shrinkcrap, too, but it's there for now. :'(

kzantow:
🐝 from me after a passing build

michaelneale:
looks like a breaking change for moment.js

vivek:
@scherler breaking changes in `moment.duration.format()`:

```
TypeError: _moment2.default.duration(...).format is not a function at TimeManager.format (/home/jenkins/slave/workspace/blueocean_PR-1688-AHAFGL2KZN6M72CYTMXY66YILZBW3G3TCIEEPRTTH4TWNCVZ5XVQ/blueocean-dashboard/node_modules/@jenkins-cd/blueocean-core-js/dist/js/utils/TimeManager.js:46:54) at convertJenkinsNodeDetails (/home/jenkins/slave/workspace/blueocean_PR-1688-AHAFGL2KZN6M72CYTMXY66YILZBW3G3TCIEEPRTTH4TWNCVZ5XVQ/blueocean-dashboard/src/main/js/components/PipelineRunGraph.jsx:99:3380)
```

scherler:
I raised now the version of  `moment-duration-format`. 

It works fine in my browser 
screenshot_2018-03-13_18-47-32

However the test in dashboard is failing, but should not

scherler:
@vivek I think I know what is going on. I think the test needs to have a published version of the core-js since it seems that it is picking up the old version. (published: '0.0.170':  2017-10-18T20:11:52.822Z')

I tried to publish a new version, however the `dist` folder is not generated anymore hence it I cannot publish a version of core-js to use it in the tests.

vivek:
@scherler I see. @NicuPascu, I think you released JDL to npm recently, how did it worked over there?

@kzantow any idea what might be going on? My understanding was that with hpi, the tests will continue using whatever the latest version there is, no need to publish new version of `blueocean-core-js`.

NicuPascu:
I just published the JDL without re-publishing core-js, the reason I published the JDL was that if other projects would need the latest JDL they could get it from npm

kzantow:
@vivek i think @scherler is right, basically because of the sorta lax publishing of core-js and jdl, tests are getting an old version (due to the horribleness of npm and my attempted workarounds for the upstream plugins)

kzantow:
The "most correct" npm-ish fix is what I described before: spawning an npm proxy for the upstream dependencies. I have the code for that pretty much done, but it conflicts potentially with `frontend-maven-plugin`, as it will _also_ do an NPM install, potentially without the correct proxy being in place...

vivek:
@kzantow so what can we do to make this test pass?

scherler:
@vivek publish a versión for npm. ...most quick fix is bring back the dist generation and publish a new version of core-js. Then raise deps on all modules to that version. ...old school. 😉

...needless to say that I am doing the dance ATM as I adviced about that from happening, It has hit bo before us though ;)   

Anyway if no one beats me to it I wil bring back the dist generation and will go old shool to fix that, publishing a new version mañana.

kzantow:
@vivek -- the easiest thing to do is just disable the test ;)

vivek:
@kzantow hmm, not sure if we want to go that far though. Whats the downside resurrecting 'dist' generation? Such issues might come in the future, better to get it in order, no?

kzantow:
We'll just have to bump all the versions all the time again, but not as bad as it was, as we only really have to do it in isolated situations like this.

vivek:
@kzantow yeah, I think it should be ok. What changes are needed to create dist? is it matter of adding `mkdir dist` somewhere in script? Maybe suggest changes needed or open another PR to get what it needs to enable core js release?

kzantow:
@vivek it's just a... _bump the version_ then `npm run gulp` and then `npm publish` and then update all the versions everywhere to the new one, I'm sure @NicuPascu can take care of it :)

NicuPascu:
I have published and installed JDL and core-js, I also had to fix the TimeDuration-spec test because the test would substract 1ms from 178d 0h 0m 0s and expect the time to change to 177d 23h 59m 59s, in the new version that would happen only if a minimum of 501ms are substracted, for easier reading I substracted a full second, the build should succeed now

scherler:
@NicuPascu cheers. BTW `npm run gulp` is not producing the dist dir anymore, however `npm run gulp build` does.

vivek:
Thanks @NicuPascu!

PR #1691, opened 2018-03-15
ATH to add stages to existing pipeline, plus selector improvements

kshultzCB:
# Description

See [JENKINS-49927](https://issues.jenkins-ci.org/browse/JENKINS-49927). This PR includes changes to the Editor, thanks to @imeredith 's excellent help:

1. `Sheets.jsx` now surfaces which sheet in the editor is the active sheet. This allows for more reliable location of the correct `back-from-sheet` element when there are more than one of them present, e.g. when adding a new stage (sheet 1) which includes an echo step (sheet 2). Instead of having to know to click on the second one, as in `wait.click(By.xpath("(//a[@class='back-from-sheet'])[2]"))`, we can now use `wait.click(By.cssSelector("div.sheet.active a.back-from-sheet"));` and always get the sheet in the foreground.
2. `EditorPipelineGraph.jsx` now adds an `id` to the `hittarget` nodes in the editor, so that they can be targeted in a test in a more predictable, unique way
3. Addition of the `testEditorAddStage` test, along with supporting `addStageToPipeline` method in `EditorPage.java`. This tests the creation of a Jenkinsfile, and then adds a second stage to this same pipeline, with two steps in it, finally saving it to a new branch.
4. Addition of `clear` and `sendKeys` methods in WaitUtils.java, similar to existing `click` method
5. Addition of logging to the above-mentioned WaitUtils methods to show the number of retries which were used
6. All of the tests in EditorPage.java have been converted to use the WaitUtils methods.
7. Documentation improvements

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


kshultzCB:
I'm marking this as DO NOT MERGE while I try and track down the test failure. The test passes locally, but the failing test is the one which is new in this PR. So it warrants careful examination.

kshultzCB:
Yes, gosh the conflicts. Spent a week in face to face meetings (which were great) and never thought to keep things current. I'm working on untangling it all now.

FTR, I'll probably end up bumping up that `RETRY~COUNT`. That does appear to fix the intermittent problem, and feels a little more elegant than putting a `Thread.sleep` in there.

PR #1692, opened 2018-03-19
Typescript support

imeredith:
This adds typescript support to blueocean. Its still somewhat a WIP, but enough to be merged I believe.


blueocean-dashboard/package.json
```diff
@@ -89,5 +96,34 @@
       "react-router@any",
       "redux-thunk@any"
     ]
+  },
+  "jest": {
```
kzantow: might be good to have this config in the external jest config file `jest.config.js`

blueocean-dashboard/src/main/js/Dashboard.tsx
```diff
@@ -0,0 +1,50 @@
...
+}
+
+interface DashBoardContext {
+    pipelineService?: object,
+    activityService?: object,
+    params?: object,
+    location?: object
+}
+class Dashboard extends React.Component<DashboardProps, any> {
+    static childContextTypes = {
```
kzantow: erhmergerhd... i have a better solution for this, but not part of this pr

blueocean-dashboard/package.json
imeredith: That is true, i would have to update the js-builder-jest plugin though. Which i might do in another pr

bin/rebuildShrinkwrap.js
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * We will prune and install BEFORE we install the requested version to make sure that
+ * shrinkwrap will update correctly everytime. We further do an optional `mvn install`
+ * to publish the new hpi to the local .m2 repository
+ *
+ * Usage: bin/cleanInstall.js [[@scope/]package@version] [dev] [mvn]
+ *
+ * dev will install a devDependency
+ * mvn will run the optional `mvn install` afterwards
+ */
```
sophistifunk: Is this meant to replace `bin/cleanInstall.js` ?

blueocean-core-js/src/js/fetch.ts
```diff
@@ -59,26 +76,24 @@ export const FetchFunctions = {
...
      * This should only be used if not using fetch or fetchJson.
      */
-    checkStatus(response) {
+    checkStatus(response: Response): Response {
         if (response.status >= 300 || response.status < 200) {
             const message = `fetch failed: ${response.status} for ${response.url}`;
-            const error = new Error(message);
-            error.response = response;
-            throw error;
+            throw {message , response};
```
sophistifunk: We should be throwing `Error` instances, not objects. The stack trace comes from the object creation, not the throw statement.

blueocean-core-js/src/js/fetch.ts
```diff
@@ -101,7 +116,7 @@ export const FetchFunctions = {
      * Usage:
      * FetchUtils.fetch(..).then(FetchUtils.parseJSON)
      */
-    parseJSON(response) {
+    parseJSON<T>(response: Response): Promise<T> {
```
sophistifunk: Delicious typed promises 😍

blueocean-core-js/test/js/fetch-spec.js
```diff
@@ -67,6 +67,7 @@ describe('Fetch', () => {
                             assert.fail(null, null, 'should not call success handler');
                         },
                         error => {
+                            console.log('XXXXX:' ,error);
```
sophistifunk: Leftover debugging, or is this meant to be here?

blueocean-dashboard/src/main/js/Dashboard.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { pipelineService, activityService } from '@jenkins-cd/blueocean-core-js';
+import navState from './DashboardNavState';
+
+interface Props {
+    params: object
+    children?: React.ReactNode
+    location: object
+}
+
+interface DashBoardContext {
```
sophistifunk: Yay!

But that capital B should probably be a normal b :)

blueocean-pipeline-editor/tsconfig.json
```diff
@@ -0,0 +1,30 @@
+{
```
sophistifunk: Do we have anything in place yet that will help us keep these files the same?

sophistifunk:
Looks great, can't wait to leverage this when I get back to pipeline results!

bin/rebuildShrinkwrap.js
imeredith: No, i need to fix up that file

blueocean-core-js/test/js/fetch-spec.js
imeredith: No cheers

blueocean-pipeline-editor/tsconfig.json
imeredith: No.  I tried to symlink them but tooling seems to not like that.

PR #1694, opened 2018-03-20
JENKINS-49297# Capture missing input steps from post declarative block

vivek:
# Description

See [JENKINS-49297](https://issues.jenkins-ci.org/browse/JENKINS-49297).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepVisitor.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepVisitor.java
@@ -2,6 +2,18 @@ package io.jenkins.blueocean.rest.impl.pipeline;
 
 import io.jenkins.blueocean.rest.model.BlueRun;
 import io.jenkins.blueocean.rest.model.BlueRun.BlueRunResult;
+import java.io.IOException;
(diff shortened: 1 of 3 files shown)


blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineStepVisitor.java
```diff
@@ -104,8 +103,13 @@ public void chunkEnd(@Nonnull FlowNode endNode, @CheckForNull FlowNode afterChun
         super.chunkEnd(endNode, afterChunk, scanner);
         if(endNode instanceof StepEndNode && PipelineNodeUtil.isStage(((StepEndNode)endNode).getStartNode())){
             currentStage = ((StepEndNode)endNode).getStartNode();
+        } else {
```
kzantow: Yikes, is there a better way we can deal with random pipeline events? e.g. just reload the whole pipeline on the front end?

vivek: Why you think this change is bad? All we are doing is finding enclosing block, upstream library has it parsed/cached optimally. 

Reloading the pipeline will execute this very code and we won’t return the right steps.

PR #1696, opened 2018-03-21
[JENKINS-50325] Some part of the code uses object.assign which needs to be polyfill to support IE11

scherler:
Add support for polyfill and use snapshot versions for testing

# Description

See [JENKINS-50325](https://issues.jenkins-ci.org/browse/JENKINS-50325).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-material-icons/.babelrc
+++ b/blueocean-material-icons/.babelrc
@@ -1,3 +1,4 @@
 {
-    "presets": ["es2015", "react", "stage-0"]
+    "presets": ["es2015", "react", "stage-0"],
+    "plugins": ["transform-object-assign"]
 }
--- a/blueocean-material-icons/package.json
+++ b/blueocean-material-icons/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@jenkins-cd/blueocean-material-icons",
-  "version": "0.0.6",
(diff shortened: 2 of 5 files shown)


NicuPascu:
This is really strange, I have checked blueocean with these icons on IE11 and they work well.

Anyway, instead of loading a polyfill I suggest we just replace the one instance of Object.assign (https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-material-icons/src/js/components/SvgIcon/SvgIcon.js#L58) with ES5 code, if you want I can take care of this.

sophistifunk:
I think manual removal of the explicit usage is probably the right choice here, also. No need for an additional polyfill + plugin, or even manually rewriting it in ES5, either. Just replace it with a spread, and let Babel deal with it however it pleases, depending on the target level.

scherler:
@NicuPascu I agree that a rewrite may be along the lines of what @sophistifunk said. Basically I do "let babel deal with it" using "babel-plugin-transform-object-assign": "6.22.0", However we should deliver our dist to be compatible as well with IE11 out-of-the-box since dealing with it on the consumer side forces to use polyfill global hammer which is neither desired.

PR #1710, opened 2018-04-04
JENKINS-49455 & JENKINS-45486 don't step out of karaoke mode when scrolling up if run is paused

NicuPascu:
This change stops stepping out of karaoke mode if scrolling up or 'key up' is pressed when run is paused

I don't think this can or should be tested

See [JENKINS-49455](https://issues.jenkins-ci.org/browse/JENKINS-49455) [JENKINS-45486](https://issues.jenkins-ci.org/browse/JENKINS-45486).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
+++ b/blueocean-dashboard/src/main/js/components/RunDetailsPipeline.jsx
@@ -47,7 +47,7 @@ export class RunDetailsPipeline extends Component {
     }
     // we bail out on arrow_up key
     _handleKeys(event) {
-        if (event.keyCode === 38 && this.augmenter.karaoke) {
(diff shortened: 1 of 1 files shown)


sophistifunk:
It would be preferable in my mind that we explicitly return to karaoke mode when the pipeline becomes un-paused, but since this old karaoke code's days are numbered it's easier to stick with this smaller change for now 🐝

NicuPascu:
This code change doesn't take us out of karaoke mode (that was the problem) when the run is paused, when it becomes un-paused, if the user scrolls up or presses the up key the code will work as it did before

sophistifunk:
What I mean is that it *should* take you out of karaoke mode if there's input needed, but put it back in when the job resumes.

NicuPascu:
Ah, right, I didn't read your first comment properly, got it

PR #1712, opened 2018-04-05
Revert "JENKINS-48357# Fix jira-plugin binary compatibility issue"

vivek:
Its causing other integration hassles and given that in short term we are going to have help on fixing jira-plugin, lets keep the status quo and ship blueocean-jira with jira-plugin version that works.

Reverts jenkinsci/blueocean-plugin#1704


michaelneale:
SGTM

PR #1719, opened 2018-04-18
[JENKINS-50885] Make jira plugin optional

michaelneale:
# Description

JIRA plugin is still a pain, and in theory could be made optional, but still available. 
I believe this will still install it but it can then later be removed or excluded as needed. 

JIRA: https://issues.jenkins-ci.org/browse/JENKINS-50885


michaelneale:
re-running due to flake (not seen that one flake before though.. but clearly not related)

vivek:
Closing this PR as https://issues.jenkins-ci.org/browse/JENKINS-48357 is fixed and integrated in blueocean.

PR #1720, opened 2018-04-19
Git https auth round trip

sophistifunk:
# Description

Just putting this up to do some cloud ATH runs while I work locally, ignore.


sophistifunk:
supergreen

acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/GitCreationPage.java
```diff
@@ -79,4 +70,30 @@ public MultiBranchPipeline createPipeline(SSEClientRule sseCLient, String pipeli
...
+        clickGitCreationOption();
+        wait.until(By.cssSelector("div.text-repository-url input")).sendKeys(url);
+
+        wait.until(By.xpath("//*[contains(text(), 'Jenkins needs a user credential')]"));
+
+        wait.until(By.cssSelector("div.text-username input")).sendKeys(user);
+        wait.until(By.cssSelector("div.text-password input")).sendKeys(pass);
+        wait.until(By.cssSelector(".button-create-credential")).click();
+
+        wait.until(By.xpath("//*[contains(text(), 'Use existing credential')]"));
```
kshultzCB: I know this comment is in the test code, and not the UI code, but I'll ask the question here and hope it makes sense. Any chance we could have a cssSelector here? That way, if the text in messages.properties gets changed for some reason, the test will still work.

vivek:
@sophistifunk I gave it a try, have these observations:

- It correctly detects ssh/git vs http(s) protocols and displays different credential ui components
- Create Pipeline is always enabled. Is this intentional? Maybe because given http(s) repo could be a public repo? 
- Once `Create Pipeline` is clicked it screen scrolls up/down/flashes, specially if there is naming conflict. Same happens when create credential is clicked.

out

- Once a new git credential is created, it displays "Use existing credential vivek/***", but because of the expansion and closing of credential its impossible to click on Create Pipeline to proceed to next step. I wonder if the entered credential is valid, we  should move the focus over to Create Pipeline button?

blueocean-git-pipeline/src/main/java/io/jenkins/blueocean/blueocean_git_pipeline/GitScm.java
```diff
@@ -25,21 +34,97 @@
...
+            // Require a host
+            String host = uri.getHost();
+            if (host == null || host.length() == 0) {
+                return null;
+            }
+
+            // Only http(s) urls have a default credential ID keyed to the repo right now
+            String scheme = uri.getScheme();
+            if (scheme != null && scheme.startsWith("http")) {
+                return ID + ":" + normalizedUrl;
```
vivek: This is problematic as the credential id is used in classic credential management to manage the credential. e.g. Click on User->Credentials and then click on created credential id, it redirects you to `http://localhost:8080/jenkins/user/admin/credentials/store/user/domain/blueocean-git-domain/credential/git:https://github.com/vivek/pipeline-demo-test.git`. At this point stapler craps out as it tries to find route to github resulting on 404. As alternative, a hash of the computed id can work?

image

blueocean-git-pipeline/src/main/java/io/jenkins/blueocean/blueocean_git_pipeline/GitScm.java
```diff
@@ -90,60 +215,127 @@ public HttpResponse validateAndCreate(@JsonBody JSONObject request) {
...
+
+        if (credentialId == null) {
+            // Still null? Must be a bad repoURL
+            throw new ServiceException.BadRequestException("Invalid URL \"" + repositoryUrl + "\"");
+        }
+
+        // --[ Load or create credentials ]--------------------------------------------------------------------
+
+        // Create new is only for username + password
+        if (request.has("userName") || request.has("password")) {
```
vivek: Should it not be &&, what condition it should be  either or?

acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/GitCreationPage.java
sophistifunk: I really don't like adding classes just for test cases, as every time we do it we add a tiny but persistent and ever-increasing bloat to the code and maintenance load (and browser performance). But mostly because that's not what CSS classnames are for.

I'll change it if there's some sort of consensus, but it ain't right agnabbit.

blueocean-dashboard/src/main/js/creation/credentials/CreateCredentialDialog.jsx
```diff
@@ -189,7 +189,7 @@ export class CreateCredentialDialog extends React.Component {
         const disabled = this.state.creationPending;
 
         const buttons = [
-            <button className="button-create-credental" disabled={disabled} onClick={() => this._onCreateClick()}>
+            <button className="button-create-credential" disabled={disabled} onClick={() => this._onCreateClick()}>
```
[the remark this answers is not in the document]
sophistifunk: Word.

blueocean-git-pipeline/src/main/java/io/jenkins/blueocean/blueocean_git_pipeline/GitScm.java
sophistifunk: Seems like a bug in classic not urlEncoding it. Easy enough to use a hash, but if we change the ID to gibberish, how will the user know which "bob/****" credential is the right one when he wants to delete it?

blueocean-git-pipeline/src/main/java/io/jenkins/blueocean/blueocean_git_pipeline/GitScm.java
sophistifunk: I just want to make sure you get an error if you're missing one, rather than silently pretending it wasn't attempted.

sophistifunk:
> Create Pipeline is always enabled. Is this intentional? Maybe because given http(s) repo could be a public repo?

It seemed intentional to me. I don't know why the creation screen flows exactly the way it does, but I already changed a boatload of stuff in this PR, and I didn't want to include a bunch of changes to the github and BB*2 flows as well if I changed something like that.

> Once Create Pipeline is clicked it screen scrolls up/down/flashes, specially if there is naming conflict. Same happens when create credential is clicked.

I will have a look and see if I can clean up the way this works. There's supposed to be a sliding transition effect I think, but it seems to be not working.

blueocean-git-pipeline/src/main/java/io/jenkins/blueocean/blueocean_git_pipeline/GitScm.java
vivek: > Seems like a bug in classic not urlEncoding it. 

Yeah, I think its bug in credentials-plugin that does not encode the id. Anyways, user can get to it via clicking credential domain so probably not a show stopper. I am little concerned as changing credential id has backward compatibility concerns. That is in future if we change it, then we have to either migrate existing job configurations to include new changed id or somehow make system work with old id. Basically whether we use hash or current id implementation, we should make sure we think bit more about it. cc: @stephenc 

> Easy enough to use a hash, but if we change the ID to gibberish, how will the user know which "bob/****" credential is the right one when he wants to delete it?

Right. I played it little bit more. Maybe we can use description for that. The domain page shows description, so choosing a hash should probably be fine.

image

vivek:
@sophistifunk looks like other merge to master caused these conflicts, you might like to rebase with latest master.

blueocean-git-pipeline/src/main/java/io/jenkins/blueocean/blueocean_git_pipeline/GitScm.java
sophistifunk: Ack

vivek: @sophistifunk You are using repo URL, e.g. `git:https://github.com/vivek/pipeline-demo-test.git`. This means for each repo you are going to create a new credential? Should it not be just the host:port? I wonder if normalized id could be `git:https:github.com`. This can avoid hashing too. but may be there is use case where we need to create Id of repo URL?

sophistifunk:
Description with hash id:

image

sophistifunk:
Looking into the blinking issue now, it's a new problem, so shouldn't be a big change. I missed it because I was using ATH rather than manual testing for the last few changes I made Tuesday.

sophistifunk:
> You are using repo URL, e.g. git:https://github.com/vivek/pipeline-demo-test.git. This means for each repo you are going to create a new credential? Should it not be just the host:port? I wonder if normalized id could be git:https:github.com. This can avoid hashing too. but may be there is use case where we need to create Id of repo URL?

If we're auto-generating credential Ids to save the user the hassle of credential management I think we should scope it to a specific repo rather than just the host. Otherwise we could have problems in corporates where there's a boatload of repos that live on a single host, like gitlab / github enterprise in an environment that for some reason (proxies, poor network setup, etc) needs to use https instead of ssh/git

But the main reason I did it is because that's what was specified in the Jira ticket 😀

blueocean-dashboard/src/main/js/credentials/git/GitCredentialsPickerPassword.tsx
```diff
@@ -0,0 +1,278 @@
...
+import {
+    FormElement,
+    PasswordInput,
+    TextInput,
+    RadioButtonGroup,
+} from '@jenkins-cd/design-language';
+
+const t = i18nTranslator('blueocean-dashboard');
+
+interface Props {
```
imeredith: This interface needs to be exported. and I think it should be namespaced, so that people do `GitCredentialsPickerPassword.Props`. Not totally sure of conventions here, but it seems like a good idea to me

imeredith: Just looked at some code @kzantow has done and he is not exporting props, but i think its probably not a bad standard practise. What do you guys think?

sophistifunk:
Green again, w007!

sophistifunk:
@vivek @imeredith happy with current state?

vivek:
@sophistifunk I still see the Create Credential button not appearing right:

image

sophistifunk:
Yeah, this behaviour is intended, to match the BB implementation as I understand it.

vivek:
@sophistifunk The ✔️ Symbol on the grayed out button looks odd. If this is the same behavior with bitbucket then let’s leave it like that and fix in another PR?

sophistifunk:
BB Cloud:

image

vivek:
@sophistifunk merge away! Thanks for verifying.

PR #1721, opened 2018-04-20
[JENKINS-46231] Hidden Parameters Support

chen-anders:
# Description

This PR introduces compatibility with the Hidden Parameters Plugin in Jenkins for the BlueOcean UI. 

Closes: [JENKINS-46231](https://issues.jenkins-ci.org/browse/JENKINS-46231)

#### Instructions to Test
- Install hidden parameters plugin
- Create a freestyle job with hidden parameters (with some default value) + build-step that echos the hidden parameter (to see that it passes all the way through when the build is run) + execute the job to see that the parameter value is echoed out

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-core-js/src/js/parameter/commonProptypes.jsx
+++ b/blueocean-core-js/src/js/parameter/commonProptypes.jsx
@@ -19,6 +19,7 @@ export const propTypes = {
             'TextParameterDefinition',
             'StringParameterDefinition',
             'PasswordParameterDefinition',
+            'WHideParameterDefinition',
         ]),
     }),
 };
--- a/blueocean-core-js/src/js/parameter/components/Hidden.jsx
+++ b/blueocean-core-js/src/js/parameter/components/Hidden.jsx
@@ -1,0 +1,18 @@
+import React, { Component } from 'react';
+import { propTypes } from '../commonProptypes';
+import { removeMarkupTags } from '../../stringUtil';
+
+export class Hidden extends Component {
+    render() {
+        const { defaultParameterValue: { value }, name } = this.props;
+        const cleanName = removeMarkupTags(name);
+
+        return (
+            <div className="Hidden FullWidth">
+                <input type="hidden" name={cleanName} value={value}/>
+            </div>
+        );
+    }
+}
+
+Hidden.propTypes = propTypes;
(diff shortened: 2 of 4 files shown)


sophistifunk:
Why have a component at all for inputs that can't actually be set? What happens in the case where a pipeline only has hidden parameters - will you just get a creepy empty input dialog? 

Perhaps maybe we could put this parameter type on a list of ignored inputs or something? So the server won't send them to the UI I mean. We don't want people working on the assumption this is a place they can put things they don't want certain users to be able to see, and then dumping them in the JSON responses and DOM where they're easily accessible.

chen-anders:
> Perhaps maybe we could put this parameter type on a list of ignored inputs or something? 

This sounds like a better idea. The initial implementation mirrored what the `hidden-parameters` plugin did for the Jenkins Classic UI, which presented a `input` element of type `hidden` in the Build with Parameters page (which itself is not ideal for its own set of reasons - some of which you already pointed out).

Unfortunately, it seems like the live code-reloading only works for Javascript related code on my end (IntelliJ or not despite following the brief setup instructions on the repository README.md) and none of my print statements are actually printing to STDOUT if I'm assuming that we want the filtering to be done at `InputStepImpl.java:getParameters()` (just from my efforts toward tracing where the `PipelineStatePreloader` tries to load data from). I'm effectively stuck otherwise and will likely just use my own snapshot build (as the goal was to simply get around the `This pipeline uses input types that are unsupported. Use Jenkins Classic to resolve parametrized build.` error).

AnneTheAgile:
cross ref https://issues.jenkins-ci.org/browse/JENKINS-46231

vivek:
@chen-anders Is this something you are working on or should we close it?

chen-anders:
Sorry, I have not been able to work on this. Closing because I assume someone else has a working dev environment and will be able to come up with the suggested solution by @sophistifunk

fuzzmz:
@sophistifunk , was there an issue created related to having this parameter type on an ignore list? Asking because there are use cases where we want to have a "template" pipeline which uses the parameters configured in the job, but we don't want to ask for user input at every run.

PR #1722, opened 2018-04-20
JENKINS-48075 add next/prev links to run details screen

NicuPascu:
# Description

See [JENKINS-48075](https://issues.jenkins-ci.org/browse/JENKINS-48075).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


NicuPascu:
Last run:
prev_link

Next to last run:
next_prev_links

First run:
next_link

Single run:
single_run

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/AbstractRunImplTest.java
```diff
@@ -238,6 +238,37 @@ public void pipelineLatestRunIncludesRunning() throws Exception {
...
+        // Get latest run for this pipeline
+        pipeline = request().get(String.format("/organizations/jenkins/pipelines/%s/", p.getName())).build(Map.class);
+        Map secondRun = (Map) pipeline.get("latestRun");
+        String prevRunUrl = ((Map)((Map)secondRun.get("_links")).get("prevRun")).get("href").toString();
+
+        //check the run id of the second run
+        Assert.assertEquals("2", secondRun.get("id"));
+
+        //check that id in previous run url is 1
+        Assert.assertEquals("1", prevRunUrl.substring(prevRunUrl.length() - 2, prevRunUrl.length() - 1));
```
NicuPascu: This only checks for the previous run URL bcause I couldn't find a way to get any other runs except for "latestRun". Except for this problem I think everything else is ok.

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/AbstractRunImplTest.java
```diff
@@ -238,6 +238,37 @@ public void pipelineLatestRunIncludesRunning() throws Exception {
...
+        // Run until completed
+        Run r = p.scheduleBuild2(0).waitForStart();
+        j.waitForCompletion(r);
+
+        // Make the next runs queue
+        j.jenkins.setNumExecutors(0);
+
+        // Schedule another run so it goes in the queue
+        WorkflowRun r2 = p.scheduleBuild2(0).waitForStart();
+        j.waitForMessage("Still waiting to schedule task", r2);
```
vivek: Why not wait for completion `j.waitForCompletion(r2);`, like in first run? Otherwise it might introduce flakiness?

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/AbstractRunImplTest.java
```diff
@@ -238,6 +238,37 @@ public void pipelineLatestRunIncludesRunning() throws Exception {
...
+
+        // Get latest run for this pipeline
+        pipeline = request().get(String.format("/organizations/jenkins/pipelines/%s/", p.getName())).build(Map.class);
```
vivek: you might in fact fetch `/organizations/jenkins/pipelines/{di}/runs` - it should give 2 runs. Or directly `/organizations/jenkins/pipelines/{di}/runs/2` to get latest run.

jenkins-design-language/less/components/material-icon.less
```diff
@@ -1,6 +1,6 @@
 .svg-icon {
-    display: 'inline-block';
```
vivek: Good catch 👏

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
```diff
@@ -326,7 +326,21 @@ private boolean isCompletedOrAborted(){
 
     @Override
     public Links getLinks() {
-        return super.getLinks().add("parent", parent.getLink());
+        Links links = super.getLinks().add("parent", parent.getLink());
+        Run nextRun = run.getNextBuild();
+        Run prevRun = run.getPreviousBuild();
+
+        if(nextRun != null) {
+            Link nextRunLink = organization.getLink().rel(String.format("pipelines/%s/runs/%s", run.getParent().getName(), nextRun.getId()));
```
vivek: Use `LinkResolver.resolveLink(nextRun)` instead of hand crafting run link, I suspect it might not work when inside nested folders.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/AbstractRunImpl.java
```diff
@@ -326,7 +326,21 @@ private boolean isCompletedOrAborted(){
...
+            links.add("nextRun", nextRunLink);
+        }
+
+        if(prevRun != null) {
+            Link prevRunLink = organization.getLink().rel(String.format("pipelines/%s/runs/%s", run.getParent().getName(), prevRun.getId()));
```
vivek: Use `LinkResolver.resolveLink(prevRun)` instead of hand crafting run link, I suspect it might not work when inside nested folders.

jenkins-design-language/less/components/material-icon.less
dragoonis: @vivek does inline-block with quotes behave differently than without quotes?

NicuPascu: @dragoonis if the display value is written in quotes it is ignored by the browser, like this:
image

sophistifunk: Probably just caused by an oversight when somebody was moving styles from hard-coded react props into CSS.

NicuPascu: Yeah I did that, just copy-pasted and did not realise at the time that the icons were not vertically aligned (that was the only consequence in this case).

PR #1725, opened 2018-04-24
Add correct permissions for jenkins to access docker.sock

MartinMuzatko:
This is required for containers to run within jenkins

# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,4 +21,6 @@ RUN for f in /usr/share/jenkins/ref/plugins/blueocean-*.jpi; do mv "$f" "$f.over
 # let scripts customize the reference Jenkins folder. Used in bin/build-in-docker to inject the git build data
 COPY docker/ref /usr/share/jenkins/ref
 
-USER jenkins
+USER jenkins:docker
+
+ADD --chown=:docker /run/docker.sock


vivek:
@MartinMuzatko what problem this PR trying to solve? cc: @imeredith

imeredith:
@vivek This would let plugins or a master agent control docker containers outside of the jenkins docker container.

MartinMuzatko:
First of all: sorry for the sparse description. Also, login servers for JIRA were down at the time of posting this PR.

The problem I faced with jenkins running this way, was that if I use dockerfiles or docker images to build my software with, the container has no access to the underlying docker daemon socket.

This should fix this problem, so jenkins is able to make docker containers access the docker daemon for building and loading docker images

imeredith:
@vivek I'm ok with this PR, but i believe the --chown flag requires a newish docker version. What builds these images currently?

imeredith:
@MartinMuzatko shouldnt it be /var/run/docker.sock? what is the expected docker run command to start the container and mount the socket?

kzantow:
I'm also a little confused about this one... you can run even the blue ocean docker image with some parameters to allow it to further utilize docker containers; see this: https://docs.google.com/presentation/d/1eRXPuLnL5I07WaJCjtVvvJ1jag66RKD6mwfDKpJNlTY/edit?usp=sharing#slide=id.g215a7c690d_0_65

MartinMuzatko:
> @MartinMuzatko shouldnt it be /var/run/docker.sock? what is the expected docker run command to start the container and mount the socket?
`/var/run` is an alias to `/run`

image

>  you can run even the blue ocean docker image with some parameters to allow it to further utilize docker containers; see this: 

I'm running Jenkins via Portainer.

image

> https://docs.google.com/presentation/d/1eRXPuLnL5I07WaJCjtVvvJ1jag66RKD6mwfDKpJNlTY/edit?usp=sharing#slide=id.g215a7c690d_0_65
I have no access to that document :/
It would be nice to have that explained here: https://jenkins.io/doc/book/installing/#on-macos-and-linux

From what I can see, the ports are all exposed in my configuration and the volume for `var/run/docker.sock` is mapped correctly.

I have to see if portainer offers something for custom parameters with the run command.

vivek:
@MartinMuzatko Checking if you are still interested in this PR. I am not sure if this is right place you want to make docker changes in to though?

MartinMuzatko:
I'm creating my own dockerfile.

``` dockerfile
FROM jenkinsci/blueocean:latest

RUN chown :docker /var/run/docker.sock
USER jenkins:docker
```

halkeye:
@MartinMuzatko See https://issues.jenkins-ci.org/browse/JENKINS-46799
I'm going to close the PR with the same comment as #1878

I think this is a compile time fix, I don't think it'll change the permissions of the mount, but I'm not certain.

```
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock --group-add $(getent group docker | awk -F: '{print $3}') -p 8080:8080 --name jenkins_blueocean --entrypoint "" jenkinsci/blueocean:latest docker ps
CONTAINER ID        IMAGE                        COMMAND             CREATED                  STATUS                  PORTS                               NAMES
82a1ef2c8d81        jenkinsci/blueocean:latest   "docker ps"         Less than a second ago   Up Less than a second   0.0.0.0:8080->8080/tcp, 50000/tcp   jenkins_blueocean

```

PR #1733, opened 2018-05-04
Jenkins 50825: Test for stage deletion from the editor, plus parallelPipeline simplifications

kshultzCB:
# Description

See [JENKINS-50825](https://issues.jenkins-ci.org/browse/JENKINS-50825). This PR replaces the previous `testEditorAddStages` test case with `testEditorAddAndDeleteStages`. It now tests both the addition of stages, and deletion of stages.

Also included here is an `id` to support easier targeting of the stages, by their actual stage names, in the editor tests. 

Finally, there are some "while I'm here" optimizations I've made:

- EditorPage.parallelPipeline has been simplified a lot:
  - Ito longer has its own save logic. The test case which calls this - `testEditorParallel` - now calls `EditorPage.saveBranch` on its own. No sense in having the `saveBranch` logic exist in >1 place, so I'm removing extra stuff when I can.
  - It's also been changed to use a shell step which does a simple `whoami` instead of a `netstat -a`. Should help run times to do something simple here, even if it's only a little bit.
- Removal of a couple `driver.navigate().refresh()` calls, again, in the interest of cutting down execution time. Local testing showed no obvious ill effects.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


kshultzCB:
@sophistifunk - I think it was you who commented in gitter the other day about not wanting to add classes for the purpose of making things easier to automate. I needed a way to target the stages of a pipeline by name, so I added an id. 

As an aside - I'm not sure we need the very simple `testEditor()` test case anymore. The other editor tests are doing the same work of creating a pipetline, editing it, and saving it.

PR #1738, opened 2018-05-16
Task/extract interfaces

sophistifunk:
# Description

Created a few interfaces to use for TS defs, and have extracted them from my codegen branch. Want to get some feedback on the taxonomy :)

This branch is *just* the interfaces. I haven't removed redundant declarations from abstract classes, nor refactored any of the rest of the code to use the new interfaces. Aiming for baby steps.

### DIFF
--- a/blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueContainerItem.java
+++ b/blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueContainerItem.java
@@ -1,0 +1,47 @@
+package io.jenkins.blueocean.rest.model;
+
+import org.kohsuke.stapler.export.Exported;
+
+public interface BlueContainerItem {
+    /**
+     * @return Gives pipeline container
+     */
+    BluePipelineContainer getPipelines();
+
+    /**
+     * Gets nested BluePipeline inside the BluePipelineFolder
+     * <p>
+     * For example for: /pipelines/folder1/pipelines/folder2/pipelines/p1, call sequence  will be:
+     *
+     * <ul>
+     * <li>getPipelines().get("folder1")</li>
+     * <li>getPipelines().get(folder2)</li>
+     * <li>getDynamics(p1)</li>
+     * </ul>
+     *
+     * @param name name of pipeline
+     * @return a {@link BluePipeline}
+     */
+    BluePipeline getDynamic(String name);
+
+    /**
+     * @return Number of folders in this folder
+     */
+    @Exported(name = "numberOfFolders")
+    Integer getNumberOfFolders();
+
+    /**
+     * @return Number of pipelines in this folder. Pipeline is any buildable type.
+     */
+    @Exported(name = "numberOfPipelines")
+    Integer getNumberOfPipelines();
+
+    @Exported(skipNull = true)
+    BlueIcon getIcon();
+
+    /**
+     * Returns pipeline folder names present in this folder.
+     */
+    @Exported(name = "pipelineFolderNames")
+    Iterable<String> getPipelineFolderNames();
+}
--- a/blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueContainerItem.java
+++ b/blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueContainerItem.java
@@ -1,13 +1,17 @@
 package io.jenkins.blueocean.rest.model;
 
+import io.jenkins.blueocean.rest.annotation.Capability;
 import org.kohsuke.stapler.export.Exported;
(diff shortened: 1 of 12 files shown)


sophistifunk:
"Item" was the best idea Andrew and I could come up with at the time, but I'm in no way attached to it. I think we'll need to add interfaces for other things added in subclasses/impls over time. Not trying to address the whole model here, I just plan on getting us a start, and enough for the run details screen.

sophistifunk:
Green, hooray! At least I haven't broken anything in this branch :)

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueRunnableItem.java
```diff
@@ -0,0 +1,50 @@
+package io.jenkins.blueocean.rest.model;
+
+import io.jenkins.blueocean.rest.Navigable;
+import org.kohsuke.stapler.export.Exported;
+
+import java.util.List;
+
```
vivek: missing javadoc

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueMultiBranchItem.java
```diff
@@ -0,0 +1,54 @@
+package io.jenkins.blueocean.rest.model;
+
+import org.kohsuke.stapler.export.Exported;
+
+import java.util.Collection;
+
```
vivek: Missing javadoc

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueManagedSource.java
```diff
@@ -0,0 +1,11 @@
+package io.jenkins.blueocean.rest.model;
+
+import io.jenkins.blueocean.rest.Navigable;
+
```
vivek: missing javadoc

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BlueContainerItem.java
```diff
@@ -0,0 +1,47 @@
+package io.jenkins.blueocean.rest.model;
+
+import org.kohsuke.stapler.export.Exported;
+
```
vivek: Missing javadoc

vivek:
@sophistifunk @imeredith I am fine with 'item', Jenkins also has 'Item' as base interface too.

vivek:
>I'm not sure this is necessarily true any more with default implementations in Java 8, but there may be circumstances I haven't thought of.

Ah, yes of course we are on java8:) default implementation can come handy for those AbstractMethodError for newly added interface methods. Its good here:)

> When using interfaces instead of abstract base classes there's more options we can choose from to smooth over compatibility issues going forwards; creating new smaller interfaces rather than making breaking changes to existing ones, creating adapter impls or "exchange services" for retired interfaces, and the like.

Right. Java8 made it easier to avoid those backward compatibility issues that interfaces used to create.

>I was under the impression we could simply use the interface classes as capabilities when they're implemented by a concrete class?

From capability (`/classes` API) perspective, an interface of any base class having `@Capability` annotations forms as base capability for all subclasses off it. So basically if we annotation an interface with a capability, all its subclasses will inherit it and will be made available thru classes API when queried thru API.

imeredith:
I like where this is going. Excited to see typescript interfaces

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/IBlueRun.java
```diff
@@ -0,0 +1,166 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+
+import static io.jenkins.blueocean.rest.model.KnownCapabilities.BLUE_RUN;
+
+/**
+ * Common interface for job run details
+ * <p>
+ * TODO: This isn't C#. Figure out what to call io.jenkins.blueocean.rest.model.BlueRun so we can call this BlueRun
```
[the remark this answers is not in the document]
vivek: @sophistifunk Other option is to use `BlueRunItem` instead of `IBlueRun`. BlueRunItem is consistent with other `...Item` interfaces, such as `BlueRunnableItem`.

sophistifunk:
@imeredith what do you think of the name `BlueItemRun` for the base run details interface?

sophistifunk:
For the sake of this code, I'm using the term `Item` to mean "thing that lives in the Jenkins job namespace", eg pipelines, folders, multibranch containers, freestyle projects, etc.

PR #1740, opened 2018-05-24
JENKINS-51161 reduce the CPU ussage of the SVG spinners

NicuPascu:
# Description
These changes make the spinner a CSS animation instead of a JS one for non-IE browsers. This reduces the overall CPU usage especially when multiple spinners are present on the same page (we weren't using a single function call for all spinners and it didn't scale very well because of this).
Also by using "transform: translateZ(0)" some full page re-paints do not happen anymore which further reduce the CPU usage.

See [JENKINS-51161](https://issues.jenkins-ci.org/browse/JENKINS-51161).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/jenkins-design-language/less/components/result-status.less
+++ b/jenkins-design-language/less/components/result-status.less
@@ -12,6 +12,10 @@
     opacity: 0;
 }
 
+.svgResultStatus {
+    transform: translateZ(0);
+ }
+
 .svgResultStatus circle.success {
     stroke: none;
     fill: @brand-success-lite;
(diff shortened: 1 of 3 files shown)


NicuPascu:
Before:
before

After:
after

NicuPascu:
In the above screenshots I ran 35 spinners on the page so to exacerbate the problem. In the before case the rendering part was really efficient and if we could have made a single function call that would take care of all the spinners on the page that would still make that implementation the most efficient. But because we cannot, each additional spinner would make a separate "requestAnimationFrame" function call and they would add more and more stress on the browser causing high CPU ussage.

michaelneale:
if this passes - why not. Looks like a contained change, perhaps @sophistifunk wants to review the code but surely the proof of the pudding is in the eating. Nice one!

sophistifunk:
LGTM

bbm

PR #1742, opened 2018-05-29
[JENKINS-51585] JWT response should have status code 204 since it does not provide content

scherler:
Use correct status code for the empty response of the JWT. Otherwise FF will try to parse the response and throw an exception because of it.

# Description

See [JENKINS-51585](https://issues.jenkins-ci.org/browse/JENKINS-51585).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-jwt/src/main/java/io/jenkins/blueocean/auth/jwt/JwtToken.java
+++ b/blueocean-jwt/src/main/java/io/jenkins/blueocean/auth/jwt/JwtToken.java
@@ -80,7 +80,10 @@ public class JwtToken implements HttpResponse {
      */
     @Override
     public void generateResponse(StaplerRequest req, StaplerResponse rsp, Object node) throws IOException, ServletException {
-        rsp.setStatus(200);
+        rsp.setStatus(204);
+        /* https://httpstatuses.com/204 No Content
+        The server has successfully fulfilled the request and
+        that there is no additional content to send in the response payload body */
         rsp.addHeader(X_BLUEOCEAN_JWT, sign());
     }
 }


blueocean-jwt/src/main/java/io/jenkins/blueocean/auth/jwt/JwtToken.java
```diff
@@ -80,7 +80,10 @@ public String sign() {
      */
     @Override
     public void generateResponse(StaplerRequest req, StaplerResponse rsp, Object node) throws IOException, ServletException {
-        rsp.setStatus(200);
+        rsp.setStatus(204);
+        /* https://httpstatuses.com/204 No Content
```
agentgonzo: TBH, the comment is unnecessary.

vivek:
@scherler There are failing tests, looks like with 204 status `rsp.addHeader(X_BLUEOCEAN_JWT, sign());` is not working. We have failing tests that expects `X_BLUEOCEAN_JWT` header to be present but its not.

olamy:
I wonder if there is any way to add a unit test for this?

agentgonzo:
Seems that the unit test failure is due to the way that the WebClient handles a different response for 200 vs 204:
https://github.com/hudson/htmlunit/blob/master/src/main/java/com/gargoylesoftware/htmlunit/WebClient.java#L431-L433

scherler:
I worked around the webclient issue by using httpConnection directly now.

blueocean-jwt/src/test/java/io/jenkins/blueocean/auth/jwt/impl/JwtAuthenticationServiceImplTest.java
```diff
@@ -132,4 +133,18 @@ public void anonymousUserToken() throws Exception{
...
+
+    // webclient has problems with pages returning 204, so we use HttpURLConnection directly to handle the token
+    private String getToken(JenkinsRule.WebClient webClient) throws IOException {
+        URL tokenUrl = new URL(webClient.getContextPath() + "jwt-auth/token/");
+        HttpURLConnection  connection = (HttpURLConnection) tokenUrl.openConnection();
+        Set<Cookie> cookies = webClient.getCookies(tokenUrl);
+        for (Cookie cookie : cookies) {
+            connection.addRequestProperty("Cookie", cookie.getName() + "=" + cookie.getValue());
+        }
+        Assert.assertEquals(connection.getResponseCode(), 204);
```
olamy: 👍

vivek:
@scherler Nice! Looks good. Re ran tests, looks like a a flaky test. If passes merge away.

PR #1743, opened 2018-05-29
Keep the editor from blowing up when seeing sequential stages

abayer:
# Description

Note that this doesn't actually do a thing for working with sequential
stages, so editting a Pipeline with sequential stages (once that's
actually released in Declarative) will not render anything, etc. But
it keeps things from blowing up for now.

Declarative PR that's relevant: https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/227

The included test makes sure that we can convert a Declarative JSON representation that contains a `stage` with nested `stages` - you can manually verify by trying to edit a Jenkinsfile like https://github.com/abayer/pipeline-model-definition-plugin/blob/79cad4ee2026e2123dcf13c6614c1988303fc034/pipeline-model-definition/src/test/resources/topLevelStageGroup.groovy after building and installing that PR (since validation will fail otherwise).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-editor/src/main/js/services/PipelineSyntaxConverter.js
+++ b/blueocean-pipeline-editor/src/main/js/services/PipelineSyntaxConverter.js
@@ -187,6 +187,8 @@ export function convertStageFromJson(topStage: PipelineStage): StageInfo {
             const stage = convertStageFromJson(topStage.parallel[j]);
             topStageInfo.children.push(stage);
         }
+    } else if (topStage.stages) {
+        // TODO: Will need to be filled in when we actually have the ability to do something with sequential stages.
     } else {
         throw new Error('Unable to determine stage type: ' + JSON.stringify(topStage));
     }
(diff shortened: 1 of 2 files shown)


vivek:
@abayer @kshultzCB well, save is blowing up - not sure if its accepted. Is there way by which blueocean side code can determine the pipeline script has unsupported code? Guess we should add some warning text that its incompatible declarative pipeline and disable 'Save button'?

kshultzCB:
It's a fair point about Save, and that's why I brought it up. It seems a _somewhat_ likely thing for a well-intentioned user to stumble into. But I understand this PR isn't supposed to allow the user to do stuff to their sequential pipelines.

The editor can be recovered by doing a full page refresh, for what that's worth. 

What I don't know is, how reasonably we can expect users to remember that editing pipelines with sequential stages in them isn't supported _yet_, unless we tell them on the screen. I'm not taking a position on that - I'm saying that I don't know.

vivek:
Closed in favor of https://github.com/jenkinsci/blueocean-plugin/pull/1749.

PR #1749, opened 2018-06-07
Pipeline Editor - check if pipeline has sequential parallel stages and show error

NicuPascu:
# Description
This checks if the pipeline contains sequential parallel stages and in case it does it show an error explaining which stage contains them and that the editor currently does not support them.

See UX-728.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-editor/src/main/js/EditorPage.jsx
+++ b/blueocean-pipeline-editor/src/main/js/EditorPage.jsx
@@ -323,6 +323,18 @@ class PipelineLoader extends React.Component {
         });
     }
 
+    checkForNestedSequentialStages(internalModelPipeline, handler) {
+        for (let i = 0; i < internalModelPipeline.children.length; i++) {
+            for (let j = 0; j < internalModelPipeline.children[i].children.length; j++) {
+                if (typeof internalModelPipeline.children[i].children[j].stages !== 'undefined') {
+                    handler(internalModelPipeline.children[i].children[j].name); //name of stage that contains sequential parallel stages
+                }
+            }
+        }
+
+        handler(false);
+    }
+
     loadContent(onComplete) {
         const { organization, pipeline, branch } = this.props.params;
         this.contentApi
@@ -333,9 +345,24 @@ class PipelineLoader extends React.Component {
                 }
                 const pipelineScript = Base64.decode(content.base64Data);
                 this.setState({ sha: content.sha });
+
                 convertPipelineToJson(pipelineScript, (p, err) => {
                     if (!err) {
                         const internal = convertJsonToInternalModel(p);
+
+                        //check if pipeline contains sequential parallel stages and show error if it does
+                        this.checkForNestedSequentialStages(internal, errParentStageName => {
+                            if (errParentStageName) {
+                                const nestedSequentialStagesError = [
+                                    {
+                                        error: t('editor.page.common.pipeline.nestedSequentialStagesError', { 0: errParentStageName }),
+                                    },
+                                ];
+
+                                this.showLoadingError(nestedSequentialStagesError);
+                            }
+                        });
+
                         if (onComplete) {
                             onComplete(internal);
                         } else {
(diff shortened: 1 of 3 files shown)


NicuPascu:
sequential_parallel_stages
@abayer I tested this implementation with the pipeline you wrote in the test [here]( https://github.com/jenkinsci/blueocean-plugin/pull/1743/files#diff-5aa3dec388929f5fd7b8f466a135c7aeR358) and it works as intended, can you please give this PR a try and check that it covers all possibilities for parallel sequential stages?

blueocean-pipeline-editor/src/main/js/EditorPage.jsx
```diff
@@ -323,6 +323,18 @@ class PipelineLoader extends React.Component {
         });
     }
 
+    checkForNestedSequentialStages(internalModelPipeline, handler) {
+        for (let i = 0; i < internalModelPipeline.children.length; i++) {
+            for (let j = 0; j < internalModelPipeline.children[i].children.length; j++) {
+                if (typeof internalModelPipeline.children[i].children[j].stages !== 'undefined') {
```
vivek: Should we also have `internalModelPipeline.children[i].children[j].stages.length > 0` as check?

blueocean-pipeline-editor/src/main/js/EditorPage.jsx
```diff
@@ -333,9 +345,24 @@ class PipelineLoader extends React.Component {
                 }
                 const pipelineScript = Base64.decode(content.base64Data);
                 this.setState({ sha: content.sha });
+
                 convertPipelineToJson(pipelineScript, (p, err) => {
                     if (!err) {
                         const internal = convertJsonToInternalModel(p);
+
+                        //check if pipeline contains sequential parallel stages and show error if it does
+                        this.checkForNestedSequentialStages(internal, errParentStageName => {
```
vivek: We should come up with generic code at this point to handle incompatibility. do not want to hold this PR, something we should do in another PR.

blueocean-pipeline-editor/src/main/js/services/PipelineSyntaxConverter.js
```diff
@@ -187,6 +187,8 @@ export function convertStageFromJson(topStage: PipelineStage): StageInfo {
             const stage = convertStageFromJson(topStage.parallel[j]);
             topStageInfo.children.push(stage);
         }
+    } else if (topStage.stages) {
+        // TODO: Will need to be filled in when we actually have the ability to do something with sequential stages.
```
vivek: Should we log a warning in console as well?

kshultzCB:
I really like @vivek 's suggestion about another, follow up PR to better handle incompatibility. If we plan to do that, is this current PR a temporary fix?

Either way, if error handling stuff like this is going into the editor, it'd be nice to have tests in place. Doesn't necessarily have to be in this PR, if we're just going to roll this back once a more long-term solution is in place.

blueocean-pipeline-editor/src/main/js/EditorPage.jsx
NicuPascu: In the pipeline example @abayer wrote, after converting the pipeline to internalModel, internalModelPipeline.children[i].children[j].stages actually has a type of obj, not array, so it doesn't have length, it does have an array called json, but because I am not sure that will always be present, I decided against testing for it.

From my testing, internalModelPipeline.children[i].children[j].stages is always undefined for all my test cases except sequential parallel stages

internal-model-structure

blueocean-pipeline-editor/src/main/js/services/PipelineSyntaxConverter.js
NicuPascu: Done

NicuPascu:
As per my discussion with @vivek, we will create a new ticket and create an ATH test after we add support for sequential parallel stages.

Until then, I have manually verified that this change works correctly with [this Jenkinsfile](https://github.com/NicuPascu/app-store-demo/blob/sequential_parallel_stages/Jenkinsfile) using [this pipeline-model-definition](https://github.com/jenkinsci/blueocean-plugin/pull/1751/files) and everything worked as expected, CI is green so I will go ahead and merge.

seq_parallel

PR #1762, opened 2018-06-25
Fixes for typedefs

sophistifunk:
# Description

Pull in some fixes for the typedefs that have been living in a local dev branch


sophistifunk:
No ticket #, Jira seems to be on fire.

PR #1766, opened 2018-07-03
[JENKINS-52332] Node Api enhancement restartable flag and restart command

olamy:
# Description

See [JENKINS-52332](https://issues.jenkins-ci.org/browse/JENKINS-52332).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
@@ -171,15 +171,12 @@ public class PipelineNodeImpl extends BluePipelineNode {
     }
 
     @Override
-    public boolean isRestartable()
-    {
+    public boolean isRestartable() {
         RestartDeclarativePipelineAction restartDeclarativePipelineAction =
             this.run.getAction( RestartDeclarativePipelineAction.class );
-        if ( restartDeclarativePipelineAction != null )
-        {
+        if ( restartDeclarativePipelineAction != null ) {
             List<String> restartableStages = restartDeclarativePipelineAction.getRestartableStages();
-            if ( restartableStages != null )
-            {
+            if ( restartableStages != null ) {
                 return restartableStages.contains(this.getDisplayName())
                     && this.getStateObj() == BlueRun.BlueRunState.FINISHED;
             }
@@ -194,12 +191,15 @@ public class PipelineNodeImpl extends BluePipelineNode {
 
     @Override
     public HttpResponse submitInputStep(StaplerRequest request) {
+        return null;
+    }
+
+    public HttpResponse restart(StaplerRequest request) {
         try
         {
             JSONObject body = JSONObject.fromObject( IOUtils.toString( request.getReader() ) );
             boolean restart = body.getBoolean( "restart" );
-            if ( restart && isRestartable() )
-            {
+            if ( restart && isRestartable() ) {
                 LOGGER.debug( "submitInputStep, restart: {}, step: {}", restart, this.getDisplayName() );
 
                 RestartDeclarativePipelineAction restartDeclarativePipelineAction =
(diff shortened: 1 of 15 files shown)


blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
```diff
@@ -152,16 +170,83 @@ public Link getLink() {
...
+    @Override
+    public boolean isRestartable()
+    {
+        RestartDeclarativePipelineAction restartDeclarativePipelineAction =
+            this.run.getAction( RestartDeclarativePipelineAction.class );
+        if ( restartDeclarativePipelineAction != null )
+        {
+            List<String> restartableStages = restartDeclarativePipelineAction.getRestartableStages();
+            if ( restartableStages != null )
+            {
```
vivek: Nit: Curly braces on new line? Most code we have has curly braces on same line.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
```diff
@@ -152,16 +170,83 @@ public Link getLink() {
...
+                return restartableStages.contains(this.getDisplayName())
```
vivek: Should it be based on displayName or id? displayName is not unique AFAIK. cc: @abayer

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueItemImpl.java
```diff
@@ -59,7 +59,7 @@ public String getOrganization() {
 
     @Override
     public String getPipeline() {
-        return pipeline.getName();
+        return pipeline == null ? "" : pipeline.getName();
```
vivek: When will `pipeline` be null? Should it not be an error? Not sure if returning empty string is going to eventually cause confusing error otherwise?

blueocean-rest/src/main/java/io/jenkins/blueocean/rest/model/BluePipelineNode.java
```diff
@@ -69,6 +69,9 @@
     @Navigable
     public abstract BluePipelineStepContainer getSteps();
 
+    @Exported
+    public abstract boolean isRestartable();
```
vivek: Missing java doc

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
```diff
@@ -152,16 +170,83 @@ public Link getLink() {
...
+        return false;
+    }
+
     @Override
     public BlueInputStep getInputStep() {
         return null;
     }
 
     @Override
     public HttpResponse submitInputStep(StaplerRequest request) {
```
vivek: It isn't `submitInputStep`, technically it works but this doesn't look right. We should add separate method to `BluePipelineStep`?

```java
    @POST
    @WebMethod(name = "restart")
    public abstract HttpResponse restart(StaplerRequest request);
```

olamy: LGTM good idea.
@NicuPascu please follow this coming change

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/QueueItemImpl.java
olamy: Well not sure exactly when I got this. Probably because of wrong usage on my side of QueueItemImpl.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
olamy: Looking at https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/RestartDeclarativePipelineAction.java#L138 
Not sure we have the id...

olamy:
Not merging now as I'm enhancing test for a possible issue found with @NicuPascu

PR #1772, opened 2018-07-11
[JENKINS-52494] Upgrade selenium container to pick up newer chrome, selenium

kshultzCB:
# Description

See [JENKINS-52494](https://issues.jenkins-ci.org/browse/JENKINS-52494). The idea here is to make a quick change to the Selenium container we use, which will being with it upgrades to Chrome, Chromedriver, and Selenium. This is probably better for test coverage anyway, but more to the point, I'd like to see if it helps with any of the flaky tests we've got.

I've seen this work in another ci setting, so I'm filing a temporary PR to watch it work on ci.blueocean.io.

# Submitter checklist
- [x] Appropriate unit or acceptance tests or explanation to why this change has no tests: Not applicable, as this is a test plumbing change

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/acceptance-tests/runner/scripts/start-selenium.sh
+++ b/acceptance-tests/runner/scripts/start-selenium.sh
@@ -6,7 +6,7 @@ $SCRIPT_DIR/stop-selenium.sh
 echo ""
 echo " Starting Selenium Docker container..."
 echo ""
-docker run -d --name blueo-selenium -p 15900:5900 -p 7990:7990 -p 7999:7999 -p 4444:4444 -e no_proxy=localhost selenium/standalone-chrome-debug:3.6.0-bromine > /dev/null
+docker run -d --name blueo-selenium -p 15900:5900 -p 7990:7990 -p 7999:7999 -p 4444:4444 -e no_proxy=localhost selenium/standalone-chrome-debug:3.13.0-argon > /dev/null
 
 # Output the containers bridge network IP to file
 SELENIUM_IP=`docker inspect -f '{{ .NetworkSettings.IPAddress }}' blueo-selenium`


kshultzCB:
Well, the Bitbucket test passed just fine, but it already does that most of the time anyway. So that doesn't really tell us much. I've kicked off a second run.

[The failure shown in FavoritesCardTest](https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/PR-1772/1/tests) looks a lot like [a previous run](https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/PR-1771/1/tests), which didn't have my upgraded docker image in place. Seems possible that Chrome isn't getting shut down correctly - there are [others with the same problem](https://stackoverflow.com/questions/50642308/org-openqa-selenium-webdriverexception-unknown-error-devtoolsactiveport-file-d).

acceptance-tests/runner/scripts/start-selenium.sh
```diff
@@ -6,7 +6,7 @@ $SCRIPT_DIR/stop-selenium.sh
 echo ""
 echo " Starting Selenium Docker container..."
 echo ""
-docker run -d --name blueo-selenium -p 15900:5900 -p 7990:7990 -p 7999:7999 -p 4444:4444 -e no_proxy=localhost selenium/standalone-chrome-debug:3.6.0-bromine > /dev/null
+docker run -d --name blueo-selenium -p 15900:5900 -p 7990:7990 -p 7999:7999 -p 4444:4444 -e no_proxy=localhost selenium/standalone-chrome-debug:3.13.0-argon > /dev/null
```
sophistifunk: I remember back in the day, the first time I thought to name a bunch of products after the elements... I thought I was so clever and unique 😂

sophistifunk:
bless

michaelneale:
Kicked off build again - that failure looks not related.. amazing it got that far... and everything but one passing.

kshultzCB:
Yeah, that failure in the second run is one of those `StaleElementReferenceException` things. I've been seeing these occasionally. 

```
org.openqa.selenium.StaleElementReferenceException:
stale element reference: element is not attached to the page document
(Session info: chrome=67.0.3396.99)
(Driver info: chromedriver=2.40.565383 (76257d1ab79276b2d53ee976b2c3e3b9f335cde7),platform=Linux 4.4.0-1048-aws x86_64) (WARNING: The server did not provide any stacktrace information)
```

Are we doing anything like [what's described here](https://docs.seleniumhq.org/exceptions/stale_element_reference.jsp) in the UI code? Some of the stuff that article talks about includes:

- A JS library deleting an element and replacing it with another that has identical attributes
- An element changes, but still has the same locator (this one actually seems pretty plausible)
- Too-fast page refreshes or changes, the tests run too slowly to keep up(also seems plausible)

I'd actually like to start using [the `sseClientRule` stuff](https://github.com/jenkinsci/blueocean-plugin/blob/c2c513441b94219ea8495fda3d2a251417a4f4b2/acceptance-tests/src/main/java/io/blueocean/ath/sse/SSEClientRule.java) in more of the tests, [like Nicu just did](https://github.com/jenkinsci/blueocean-plugin/blob/391470898e1467377d67181e27f6c1c3d5ed39c9/acceptance-tests/src/test/java/io/blueocean/ath/offline/multibranch/RestartStageTest.java#L61) in his restartable stages PR. That seems like a pretty solid way of waiting for things to finish before letting the test proceed. But that's a different (and much more involved) thing.

sophistifunk:
> A JS library deleting an element and replacing it with another that has identical attributes

If React gets confused, or catches us messing with the DOM behind its back, quite likely.

> An element changes, but still has the same locator (this one actually seems pretty plausible)

What's a locator? But probably, changing elements as little as possible is what React is for.

> Too-fast page refreshes or changes, the tests run too slowly to keep up(also seems plausible)

Seems this has already given us grief at various times.

kshultzCB:
A locator is something that Selenium uses to find stuff on the page. For example:

- A cssSelector like is being used [here](https://github.com/jenkinsci/blueocean-plugin/blob/391470898e1467377d67181e27f6c1c3d5ed39c9/acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/EditorPage.java#L48). 
- An xpath, like a `data-` tag, such as [this one](https://github.com/jenkinsci/blueocean-plugin/blob/391470898e1467377d67181e27f6c1c3d5ed39c9/acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/EditorPage.java#L69)

There are others, but I think those are the most frequently used.

The run that @michaelneale kicked off passed, incidentally :) Still annoying that it didn't pass the first two times. But...

- The Bitbucket test hasn't flaked with the new docker image in place (yet?)
- Having a more up to date version of Chrome (and Selenium) in CI is probably a good idea anyway.

This PR might open the door to easier cross-browser testing, which is how I started down this road to begin with. I haven't spent a lot of time trying get it to work with Firefox, but [that docker image already exists](https://github.com/SeleniumHQ/docker-selenium/tree/3.13.0-argon/StandaloneFirefoxDebug). I bet it wouldn't be terrible to get it going.

michaelneale:
nice @kshultzCB  - FYI we used to use firefox until mid last year, cliff converted to chrome as it opened possibility for a few other things (I forget what specifically). 

That parallel one is known to be funky - would be nice to know why

kshultzCB:
Here's another thought. Do we need the VNC server that's built into this docker image?

It's cool if we do - personally, I don't ever use it, but I know not everyone is working on test code one at a time. But if nobody's using it, [we could use a different image instead](https://github.com/SeleniumHQ/docker-selenium/tree/3.13.0-argon/StandaloneChrome). Launching a VNC server probably isn't all that resource intensive, but if we're trying to make the tests better performing, every little bit helps.

NicuPascu:
I personally use VNC all the time while writing or modifying ath tests, it's the only way I know to see what the test is doing and "seeing".
Also, thanks for upgrading Chrome here.

michaelneale:
yeah I think VNC is used pretty often when people hack on tests, not sure how else you would do it (other than running it outside of the container, but that was always more trouble than it was worth in the past).

PR #1776, opened 2018-07-20
Move Favorites to Dashboard and remove extension points

imeredith:
Moves personalization to dashboard. Removes the extension points that it uses.

We need to noop the plugin i think?


imeredith:
@sophistifunk When you have some time can you please take a look at this branch? Basically favorites on the dashboard are not showing, but only when running in ATH, not hpi:run.

To reproduce, mvn install the project, then run `./run.sh --dev`. Create a job on the instance and go to dashboard. The favorite icon is hidden from view  but is in the dom. I can't figure out what css is doing that so i can't fix whatever the problem is.

NicuPascu:
I just realised the failing tests are genuine because the editor link is generated through that extension point on the branches page, will try to fix it tomorrow.

NicuPascu:
This branch has been merged into the 'standalone' branch

PR #1778, opened 2018-07-27
sequential stages inside parallel in Declarative syntax

olamy:
# Description

See [JENKINS-49779](https://issues.jenkins-ci.org/browse/JENKINS-49779).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


pom.xml
```diff
@@ -332,27 +332,27 @@
...
             <groupId>org.jenkinsci.plugins</groupId>
             <artifactId>pipeline-model-api</artifactId>
-            <version>1.3</version>
+            <version>1.3.1</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.46</version>
+            <version>2.53</version>
```
vivek: Change it to 2.54 (version getting in to CAP)

pom.xml
```diff
@@ -332,27 +332,27 @@
...
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.25</version>
+            <version>2.28</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.17</version>
+            <version>2.21</version>
```
vivek: Change it to 2.23.

pom.xml
```diff
@@ -362,17 +362,17 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-multibranch</artifactId>
-            <version>2.17</version>
+            <version>2.19</version>
```
vivek: Change it to 2.20

pom.xml
```diff
@@ -362,17 +362,17 @@
...
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-step-api</artifactId>
-            <version>2.14</version>
+            <version>2.15</version>
```
vivek: Change it to 2.16

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
```diff
@@ -323,20 +337,44 @@ public void parallelStart(@Nonnull FlowNode parallelStartNode, @Nonnull FlowNode
...
+                FlowNodeWrapper flowNodeWrapper = stack.pop();
+                if(stack.isEmpty()){
+                    if(nextStage!=null) {
+                        branch.addEdge(nextStage);
+                    }
+                    parallelBranches.push(branch);
+                    continue;
+                }
+                // if name is different it's not a dummy 'duplicate' stage with the same name so let's add it to the graph
+                if(!StringUtils.equals( flowNodeWrapper.getDisplayName(), branch.getDisplayName())){
```
vivek: Probably should use `flowNodeWrapper.getId()` instead of `flowNodeWrapper.getDisplayName()`?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
```diff
@@ -187,16 +189,15 @@ protected void handleChunkDone(@Nonnull MemoryFlowChunk chunk) {
...
+        boolean parallelNestedStages = false;
+
+        // it's nested stages inside parallel so let's collect them later
         if(parallelEnd != null){
-            return;
+            parallelNestedStages = true;
         }
 
         if(!nestedStages.empty()){
-            nestedStages.pop(); //we throw away nested stages
```
vivek: We should verify scripted pipeline behavior. Nested stages in scripted pipeline have been skipped. e.g. below pipeline scripted should not generated nodes in DAG for BranchStage1 and NestedBranchStage1 stages.

```
node {
   stage ('test') { 
     echo ('Testing'); 
     parallel NestedStagesBranch: {
        echo 'first Branch'
        stage('BranchStage1') {
          echo 'NestedBranchStage1'
           stage('NestedBranchStage') {
              echo 'NestedBranchStage1'
            }
        }
        echo 'first Branch end'
     }
   }    
}
```

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeTest.java
```diff
@@ -507,29 +512,34 @@ public void testNestedBlockStage() throws Exception{
                 Assert.assertEquals(1, edges.size());
                 Assert.assertEquals(rn.get("result"), "SUCCESS");
                 Assert.assertEquals(rn.get("state"), "FINISHED");
-            }else if(rn.get("displayName").equals("test")){
+            }else if(rn.get("displayName").equals("Packaging")){
```
vivek: 🐛 We do not support nested stages in scripted pipeline. Only sequential stages should appear in declarative pipeline. 

@abayer Assuming we render nested stages on in case of declarative sequential stage: is there way to find we are executing declarative pipeline vs scripted? Maybe from `WorkFlowRun` or `FlowExecution`? Or maybe from `FlowNode` corresponding to a stage being inspected while building graph in API call to exclude nested stages in scripted but show them in the declarative context for sequential stages?

abayer: `WorkflowRun#getAction(ExecutionModelAction.class)` - `ExecutionModelAction` gets attached to every Declarative run, and won't be on any Scripted run. What's more, `ExecutedModelAction` actually contains the execution plan for the Declarative Pipeline, via [`#getStages()`](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/actions/ExecutionModelAction.java#L62), which will give you the [`ModelASTStages`](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-api/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/ast/ModelASTStages.java) that will be/are being executed by the run.

michaelneale:
this is cool - does this now work with visualisation? that is probably worthy of a very loud blog post!

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
olamy: I wish but not really :(
In such case:
```
pipeline {
    agent any
    stages {
        stage("top") {
            parallel {
                stage("first") {
                   ......
                }
    .......
```
We have 2 nodes:
* StepStartNode (descriptorId: org.jenkinsci.plugins.workflow.cps.steps.ParallelStep) and ParallelLabelAction "first"
* StepStartNode (descriptorId: org.jenkinsci.plugins.workflow.support.steps.StageStep) and LabelAction "first"
id are very different. Anyway the test should happen only in this special case

vivek: I was suggesting using ‘flowNode.getId()’ which is always unique for each FlowNode, I guess you are saying that’s not the case?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
olamy: fixed

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
olamy: yup. but the need is to eliminate the extra node because we only want one node with type parallel ("first") and not the one with type stage.

PR #1791, opened 2018-08-14
get rid of hipchat send

michaelneale:
wasn't needed, and hipchat is going away...


michaelneale:
@olamy or Gitter perhaps? should be whereever people can see it publicly. Or it isn't needed at all as people can freely see the build results (can put a badge on it too probably)

PR #1803, opened 2018-08-30
[JENKINS-53132] Bitbucket redux: moves sleep into saveBranch, more consistent logging

kshultzCB:
# Description

See [JENKINS-53132](https://issues.jenkins-ci.org/browse/JENKINS-53132).

This Bitbucket test has been failing somewhat randomly for a long time now, and really, only on ci.blueocean.io. It works all the time locally and on other CI infra. The first failure is always marked by Selenium not being able to find the `textarea` where one enters their commit message. That has placeholder text of "What Changed?" An example of such a failure can be seen in the below screenshot, on L10 of its log. It reads: 
```
no such element: Unable to locate element: {"method":"css selector","selector":"textarea[placeholder="What changed?"]"}
```

image

The test gets two more `Retry` chances to get it right, but those never work either. They fail for a different reason - there's an `unexpected alert open: {Alert text : }`. That's shown here, look at L3:

image

This isn't useful logging, but we'll handle that in a different PR on a different day. Or better still, this problematic test will start behaving correctly all the time and we won't need to. 

In any case, this PR actually removes what amounts to a 3 second `Thread.sleep` from the test itself, and instead puts it in the `EditorPage#saveBranch`, which is where it probably belonged anyway.

I'm marking this one DO-NOT-MERGE because I want to see this work a lot of times before we merge it. I also don't want to get anyone's hopes up just yet that this will fix the test.

# To run locally:
1. Go to your local repo, `cd acceptance-tests`, and issue `./run.sh --dev`
2. Once everything's up and going, from a different  terminal, `cd acceptance-tests`, and run `mvn test -Dtest=BitbucketServerTest`.
3. Watch the test pass. Which it already did locally anyway...

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/EditorPage.java
+++ b/acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/EditorPage.java
@@ -121,7 +121,12 @@ public class EditorPage {
     public void saveBranch(String branch) {
         logger.info("saveBranch method called");
         wait.click(By.xpath("//*[text()='Save']"));
+        // In some CI infra, we're getting failures probably half the time. Let's
+        // sleep here as a safeguard.
+        wait.tinySleep(1000);
+        logger.info("wait.until the What Changed textarea appears");
         wait.until(ExpectedConditions.visibilityOfElementLocated(By.cssSelector("textarea[placeholder=\"What changed?\"]")));
+        logger.info("Entering commit message into What Changed textarea");
         wait.sendKeys(By.cssSelector("textarea[placeholder=\"What changed?\"]"), "ATH made changes and is saving");
         if(!Strings.isNullOrEmpty(branch)) {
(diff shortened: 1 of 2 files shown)


kshultzCB:
Three runs in a row where BitbucketServerTest didn't flake. Seems too good to be true.

kshultzCB:
Make that four successful runs in a row. (Disregard run number 2, that was another, different test that's been a little flaky lately).

image

If this keeps looking good, I'll remove the two labels and ask for a quick review. Maybe third time is the charm.

kshultzCB:
@NicuPascu , this is a tiny change I'm hoping will help the BitbucketServerTest stop flaking, or at least flake less often. Full explanation can be found up above. I've got six runs in CI without a failure in the topic test case. Seems okay to merge to master to me, but I'd love another set of eyes on it just in case. Thanks!

PR #1811, opened 2018-09-27
check good name

runzexia:
There are some problems with the name check now, for example, I can create a job named `a/b`


olamy:
LGTM but I'd like to see a unit test.

PR #1812, opened 2018-09-27
Remove unused zombie dependency

sophistifunk:
Remove unused zombie dependency that was breaking my build


sophistifunk:
Different failures every time, still trying to get a green build :(

kzantow:
@sophistifunk build is green -- Go, go, go!

PR #1818, opened 2018-09-29
[JEP-305] Incrementalify Blue Ocean

batmat:
This will allow generating and consume incremental releases for Blue Ocean in Evergreen if needed.

```
mvn incrementals:incrementalify
```

Cf. https://jenkins.io/blog/2018/05/15/incremental-deployment/ and
https://github.com/jenkinsci/jep/tree/master/jep/305 to dig into this more.

# Description

Relates to https://issues.jenkins-ci.org/browse/JENKINS-50686

# Submitter checklist
~~- [ ] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below.~~

# Reviewer checklist
- [x] Reviewed the code

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,10 @@
 #!groovy
 
 if (JENKINS_URL == 'https://ci.jenkins.io/') {
-    buildPlugin()
+    buildPlugin(
+      platforms: ['linux'],
+      tests: [skip: true]
+    )
     return
 }
 
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
 
   <groupId>io.jenkins.blueocean</groupId>
   <artifactId>blueocean-parent</artifactId>
-  <version>1.10.0-beta-1-SNAPSHOT</version>
+  <version>${revision}${changelist}</version>
   <packaging>pom</packaging>
 
   <name>Blue Ocean Parent</name>
@@ -24,6 +24,8 @@
   </licenses>
 
   <properties>
+    <revision>1.10.0-beta-1</revision>
+    <changelist>-SNAPSHOT</changelist>
     <java.level>8</java.level>
     <jenkins.version>2.121.1</jenkins.version> <!-- Should be kept in sync with the version resulting from Dockerfile FROM statement-->
     <javadoc.exec.goal>javadoc-no-fork</javadoc.exec.goal> <!-- stop initialize phase plugins executing twice -->
@@ -49,7 +51,7 @@
     <connection>scm:git:https://github.com/jenkinsci/blueocean-plugin.git</connection>
     <developerConnection>scm:git:https://github.com/jenkinsci/blueocean-plugin.git</developerConnection>
     <url>https://github.com/jenkinsci/blueocean-plugin</url>
-    <tag>HEAD</tag>
+    <tag>${scmTag}</tag>
   </scm>
 
   <url>https://wiki.jenkins-ci.org/display/JENKINS/Blue+Ocean+Plugin</url>
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -1,0 +1,7 @@
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
+  <extension>
+    <groupId>io.jenkins.tools.incrementals</groupId>
+    <artifactId>git-changelist-maven-extension</artifactId>
+    <version>1.0-beta-7</version>
+  </extension>
+</extensions>
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,2 @@
+-Pconsume-incrementals
+-Pmight-produce-incrementals
--- a/blueocean-bitbucket-pipeline/pom.xml
+++ b/blueocean-bitbucket-pipeline/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <artifactId>blueocean-parent</artifactId>
     <groupId>io.jenkins.blueocean</groupId>
-    <version>1.10.0-beta-1-SNAPSHOT</version>
+    <version>${revision}${changelist}</version>
   </parent>
 
   <name>Bitbucket Pipeline for Blue Ocean</name>
--- a/blueocean-commons/pom.xml
+++ b/blueocean-commons/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>io.jenkins.blueocean</groupId>
     <artifactId>blueocean-parent</artifactId>
-    <version>1.10.0-beta-1-SNAPSHOT</version>
+    <version>${revision}${changelist}</version>
   </parent>
 
   <artifactId>blueocean-commons</artifactId>
--- a/blueocean-config/pom.xml
+++ b/blueocean-config/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>io.jenkins.blueocean</groupId>
     <artifactId>blueocean-parent</artifactId>
-    <version>1.10.0-beta-1-SNAPSHOT</version>
+    <version>${revision}${changelist}</version>
   </parent>
 
   <artifactId>blueocean-config</artifactId>
--- a/blueocean-core-js/pom.xml
+++ b/blueocean-core-js/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>blueocean-parent</artifactId>
         <groupId>io.jenkins.blueocean</groupId>
-        <version>1.10.0-beta-1-SNAPSHOT</version>
+        <version>${revision}${changelist}</version>
     </parent>
 
     <artifactId>blueocean-core-js</artifactId>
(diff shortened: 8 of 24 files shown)


batmat:
Maven 3.3.3, 3.5+ required. Filed https://issues.jenkins-ci.org/browse/JENKINS-53842 to track this. Thanks

olamy:
@batmat you may need to rebase your pr with current master to pick up changes from #1819

batmat:
Right, just saw it's building indeed `pr-head` and not `pr-merge`. Trying.

batmat:
So this configures the JEP 305 as the Maven level, but will not publish Incremental releases as Blue Ocean is neither built on ci.jenkins.io nor uses the buildPlugin step. The latter is fixable pretty easily, I guess the former is trickier since it will not be possible to use the incrementals publishing pipeline code from https://github.com/jenkins-infra/pipeline-library/blob/e2fe66b4ee193c728b1e371a1ce2f9ff9a07d603/vars/buildPlugin.groovy#L172-L174

I guess a simple possible workaround could be to have also Blue Ocean built on ci.jenkins.io, but maybe only on master branch using `buildPlugin` as usual?

WDYT @jglick @rtyler ?

olamy:
well the build on https://ci.blueocean.io is running more test including some acceptance tests so not sure to move that and btw it's using our own dogfood.
Maybe having a duplicate build on https://ci.jenkins.io/ to push incremental releases for evergreen.
@vivek WDYT?

batmat:
@olamy I wrote **also**, so yeah, I didn't suggest to *move* the build there, just having there **too** indeed :-).

olamy:
@batmat oops my bad I missed the also :) 
do you need to modify the Jenkinsfile or can use an other one?

batmat:
That's a good question. I _think_ either we do something hackish in a single `Jenkinsfile` to detect where it's running to call either the current code, or `buildPlugin` if it detects it's running in ci.jenkins.io.

Or, we could maybe create/configure a custom job on ci.jenkins.io side, so it uses another file than `Jenkinsfile`, like `Jenkinsfile.incrementals` or something like this.

olamy:
the less hackish might be a new file called `Jenkinsfile.incrementals` ?

pom.xml
```diff
@@ -24,6 +24,8 @@
   </licenses>
 
   <properties>
+    <revision>1.9.0-beta-1</revision>
```
vivek: @batmat with this change, no need run `mvn versions:set -DnewVersion=1.9.0-SNAPSHOT`? This we do to switch from beta to final release. So the only change we need to make it edit pom to change revision property, right?

vivek:
@olamy @batmat +1 for `Jenkinsfile.incrementals`.

rtyler:
There are not custom Pipelines configured on ci.jenkins.io, so this has to be sourced or picked up from the GitHub Branch Source plugin in some way.

No snowflakes allowed anymore :smile:

batmat:
Given what Tyler says, we could just reverse the logic:
* `Jenkinsfile` would be changed to contain the usual `buildPlugin()` only
* `Jenkinsfile.blueocean` would be the renamed file with the [current Jenkinsfile content](https://github.com/jenkinsci/blueocean-plugin/blob/db945c1a4604f9532e2b6132183838f57a331dbe/Jenkinsfile)
* https://ci.blueocean.io would be changed to use not `Jenkinsfile` but `Jenkinsfile.blueocean`

I guess it makes more sense indeed to do the change this way anyway, since doing custom adaptations is more easily done on the already dedicated and custom pet-caring CI server that is https://ci.blueocean.io, than on the cattle-breeding server, handling **all** plugins, that is https://ci.jenkins.io

pom.xml
batmat: Good point. I've never tried it on a multimodule project, but given my Maven foo: yes I do not see anything that would be needed. That's another good point of the [Maven CI Friendly versions](https://maven.apache.org/maven-ci-friendly.html) I had overlooked: no need to update all children since they now use a variable that does not need to be changed when parent is bumped, given standard inheritance (adapted in 3.5+ to support variables) will apply.

WDYT @olamy ?

olamy: correct just edit this one. I don't remember how this will work with release plugin..
but just edit commit and tag and ` mvn deploy ` then change revision to next snapshot so only release process change a bit

olamy:
yup sounds good to me to reverse the logic.

vivek:
> Given what Tyler says, we could just reverse the logic:
> 
> * `Jenkinsfile` would be changed to contain the usual `buildPlugin()` only
> * `Jenkinsfile.blueocean` would be the renamed file with the [current Jenkinsfile content](https://github.com/jenkinsci/blueocean-plugin/blob/db945c1a4604f9532e2b6132183838f57a331dbe/Jenkinsfile)
> * https://ci.blueocean.io would be changed to use not `Jenkinsfile` but `Jenkinsfile.blueocean`
> 
> I guess it makes more sense indeed to do the change this way anyway, since doing custom adaptations is more easily done on the already dedicated and custom pet-caring CI server that is https://ci.blueocean.io, than on the cattle-breeding server, handling **all** plugins, that is https://ci.jenkins.io
> 
> WDYT @olamy @vivek ?

@batmat @olamy This is fine as long as we make sure it doesn't impact dogfood (ci.blueocean.io). Basically make sure `blueocean` job on the dogfood is updated to work off the renamed Jenkinsfile. @imeredith is that the only change needed or there are more places it needs to be updated?

pom.xml
jglick: See [guide](https://github.com/jenkinsci/incrementals-tools/blob/master/README.md#enabling-production-of-incrementals). I do not see any issues here.

jglick:
> I _think_ either we do something hackish in a single `Jenkinsfile` to detect where it's running to call either the current code, or `buildPlugin` if it detects it's running in ci.jenkins.io.

This is routine. Just

```groovy
if (infra.isRunningOnJenkinsInfra()) {
    // ci.jenkins.io
    buildPlugin()
} else {
    // as before…
}
```

batmat:
@jglick yes, this would work on ci.jenkins.io, but I suppose will crash the build on ci.blueocean.io.

Anyway, going to give this a try and take the opportunity to fix the conflicts above soonish.

jglick:
> I suppose will crash the build on ci.blueocean.io

Nope, see above.

Of course you will need to file an `INFRA` ticket to remove whatever is [blocking `blueocean-plugin` from being scanned](https://ci.jenkins.io/job/Plugins/computation/consoleFull).

batmat:
Proposed the `if` switch trick as a separate PR in https://github.com/jenkinsci/blueocean-plugin/pull/1838 @jglick thanks for the hint!

kzantow:
@batmat just update this to master and re-trigger CI? seems like it should be fine

batmat:
@kzantow almost, yes. But we need to remove the config on ci.jenkins.io that made Blue Ocean repo be skipped until now. Once https://issues.jenkins-ci.org/browse/INFRA-1826 is done, then yes you're right :)

batmat:
Wooooot http://repo.jenkins-ci.org/incrementals/io/jenkins/blueocean/

pom.xml
kzantow: @jglick -- just to be clear, `mvn release` works normally with these changes?

batmat: @kzantow yes, you can still use the normal release process using the release plugin. If the [parent pom is ≥3.18](https://github.com/jenkinsci/plugin-pom/blob/master/CHANGELOG.md#318) it will automatically be "reincrementalified" when using common m-release-p releases. 

So, as [Blue Ocean is already on 3.23](https://github.com/jenkinsci/blueocean-plugin/blob/a6b4aa390417e7307dfb98bfa29b6d2094a0674b/pom.xml#L8), you're good to go.

michaelneale:
looks like @vivek just needs to see if his feedback is addressed. @batmat there is a beta in the .mvn extension config, but that is probably ok right?

batmat:
>  just needs to see if his feedback is addressed. 

AFAICT it is addressed: @vivek was blocking for the `Jenkinsfile` related discussion, finally handled in https://github.com/jenkinsci/blueocean-plugin/pull/1838 dedicated PR as commented above.

> .mvn extension config, but that is probably ok right?

@michaelneale yes. beta-7 is the latest published by @jglick for https://github.com/jenkinsci/incrementals-tools and it works pretty well for already many plugins, as you can see by the releases published at https://repo.jenkins-ci.org/incrementals/, and even the jenkins WAR for publishing an incremental release for each master merges.

batmat:
> LGTM - pending @vivek comments. I trust the pom version changes will work with the current release automation that is used outside of evergreen?

Yes, see my answer to @kzantow above in https://github.com/jenkinsci/blueocean-plugin/pull/1818#discussion_r230221534 :

image

olamy:
@vivek no worries you will still be able to use the old school ` mvn release:prepare release:perform ` :)

Jenkinsfile
```diff
@@ -1,7 +1,10 @@
 #!groovy
 
 if (JENKINS_URL == 'https://ci.jenkins.io/') {
-    buildPlugin()
+    buildPlugin(
+      platforms: ['linux'],
+      tests: [skip: true]
```
vivek: @batmat We are skipping tests, this is intentional? Won't this result in to releasing blueocean incrementally with test failures?

vivek:
kicked off another ci run as there were some tests failures (most likely flaky ones)

kshultzCB:
FWIW - the test failure I see [here](https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/PR-1818/10/tests) is FavoritesCardsTest. That one's been flaky recently.

kzantow:
ditto, 'nother run

PR #1835, opened 2018-10-17
[TASK] Add pipeline and run to extension point

sven-carstens:
Extension point  _jenkins.pipeline.activity.list.action_ does not supply the _pipeline_ and the _run_ corresponding to the list entry. 

To make it more usefull we add the parameters, that are passed to the rerun button, for others to use here.

eg. the plugin example blueocean-executor-info uses this extension point and does nothing for this point  as it checks for an existing pipeline inside of the props.


imeredith:
Hi, we have work in progress to remove the current extension system and replace it with something simpler and more flexible. I will be posting on the jenkins mailing list this week.

Unforunatly this will mean that when we release the new version of blueocean current plugins will have to be updated to work. I'm currently updated the executor info plugin and will incorporate this change in the new system too.

Do you have a plugin that uses the extension system? We did some searching of github and couldnt find anything outside of executor info.

Once I'm finished with executor info plugin it will serve to be a reference impl for how they can be done.

sven-carstens-udg:
I was trying to get some functionality of promoted-builds plugin working. So there is no public repo anywhere to be found yet as I'm still in very early stages. I had some questions on glitter and it seems I need to change the action handling inside of blueocean-rest to not only have the simple ActionProxy.

imeredith:
Just to update, things have taken longer that I wanted to get everything aligned. We have finished updating blueocean on a branch with the new Extensibility layer and I'm working on some last minute refactoring and documentation, at which point I'll make a post onto the ML

timja:
@imeredith did anything ever happen with the new extensibility layer?

halkeye:
new layer is essentially a rewrite, and pretty much dead at this point

PR #1840, opened 2018-10-30
Refactored if conditions with ternary operations since it is more in …

LavaGolem:
Small refactor changes if conditions in render function were not in the spirit of React, so I switch them with conditional rendering.


halkeye:
Hey @LavaGolem, thanks for the contribution, but since we as maintainers actually prefer the if statements for readability we are going to decline this contribution.

PR #1841, opened 2018-10-31
[JENKINS-55093] fix multi branch limit offset

runzexia:
# Description

In the previous version, start limit was performed for the run of each branch.
In this case, some runs may not be able to be queried normally.
This PR concats all the runs to query,and added unit test

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultibranchPipelineRunContainer.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultibranchPipelineRunContainer.java
@@ -62,8 +62,6 @@ public class MultibranchPipelineRunContainer extends BlueRunContainer{
      */
     @Override
     public Iterator<BlueRun> iterator(int start, int limit) {
-        List<BlueRun> c = new ArrayList<>();
-
         List<BluePipeline> branches;
 
         // Check for branch filter
@@ -85,22 +83,26 @@ public class MultibranchPipelineRunContainer extends BlueRunContainer{
             sortBranchesByLatestRun(branches);
         }
 
+        List<BlueRun> allRunList = new ArrayList<BlueRun>();
         for (final BluePipeline b : branches) {
             BlueRunContainer blueRunContainer = b.getRuns();
             if(blueRunContainer==null){
                 continue;
             }
-            Iterator<BlueRun> it = blueRunContainer.iterator(0, MAX_MBP_RUNS_ROWS);
-            int count = 0;
-            Utils.skip(it, start);
-            while (it.hasNext() && count++ < limit) {
-                c.add(it.next());
+            List<BlueRun> it = Lists.newArrayList(blueRunContainer.list());
+            allRunList.addAll(it);
+            if (allRunList.size() > start + limit){
+                break;
             }
         }
 
-        Collections.sort(c, LATEST_RUN_START_TIME_COMPARATOR);
+        if (start < allRunList.size()) {
+            List<BlueRun> c = allRunList.subList(start, allRunList.size());
+            Collections.sort(c, LATEST_RUN_START_TIME_COMPARATOR);
 
-        return Iterators.limit(c.iterator(), limit);
+            return Iterators.limit(c.iterator(), limit);
+        }
+        return Iterators.emptyIterator();
     }
 
     static void sortBranchesByLatestRun(List<BluePipeline> branches) {
--- a/blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/MultiBranchTest.java
+++ b/blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/MultiBranchTest.java
@@ -2,6 +2,7 @@ package io.jenkins.blueocean.rest.impl.pipeline;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterators;
 import hudson.Util;
 import hudson.model.FreeStyleProject;
 import hudson.model.Queue;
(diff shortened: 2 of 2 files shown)


runzexia:
cc @olamy

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultibranchPipelineRunContainer.java
```diff
@@ -84,21 +84,24 @@ public BlueRun get(String name) {
...
-                c.add(it.next());
-            }
+            allRunIt = Iterators.concat(allRunIt,it);
         }
 
+
         Collections.sort(c, LATEST_RUN_START_TIME_COMPARATOR);
+        Utils.skip(allRunIt, start);
+        int count = 0;
+        while (allRunIt.hasNext() && count++ < limit) {
```
kzantow: I think it would be better to keep the limit in the `branches` loop, without that, you're potentially reading a lot more stuff from disk (and slow/etc.)

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultibranchPipelineRunContainer.java
```diff
@@ -84,21 +84,24 @@ public BlueRun get(String name) {
...
         }
+        Iterator<BlueRun> allRunIt = Collections.emptyIterator();
 
         for (final BluePipeline b : branches) {
             BlueRunContainer blueRunContainer = b.getRuns();
             if(blueRunContainer==null){
                 continue;
             }
             Iterator<BlueRun> it = blueRunContainer.iterator(0, MAX_MBP_RUNS_ROWS);
-            int count = 0;
```
kzantow: This should be moved before the loop, then break once `limit` is reached

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultibranchPipelineRunContainer.java
runzexia: Looping in a branches can keep running fast, but in some cases some runs cannot be listed correctly.

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/MultiBranchTest.java
```diff
@@ -1135,4 +1137,63 @@ public void testMultiBranchPipelineQueueContainer() throws Exception {
...
+        sampleRepo1.git("checkout", "-b", "abc");
+        sampleRepo1.write("Jenkinsfile", "echo \"branch=${env.BRANCH_NAME}\"; " + "node {" +
+            "   stage ('Build'); " +
+            "   echo ('Building'); " +
+            "   stage ('Test'); echo hello; " +
+            "   echo ('Testing'); " +
+            "   stage ('Deploy'); " +
+            "   echo ('Deploying'); " +
+            "}");
+        ScriptApproval.get().approveSignature("method java.lang.String toUpperCase");
```
halkeye: I don't see toUpperCase being used, can you remove this line from the test?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultibranchPipelineRunContainer.java
```diff
@@ -85,22 +83,26 @@ public BlueRun get(String name) {
...
-                c.add(it.next());
+            List<BlueRun> it = Lists.newArrayList(blueRunContainer.list());
+            allRunList.addAll(it);
+            if (allRunList.size() > start + limit){
+                break;
             }
         }
 
-        Collections.sort(c, LATEST_RUN_START_TIME_COMPARATOR);
+        if (start < allRunList.size()) {
```
halkeye: I'm trying to follow this, so i'm writing it out.

If start (3) is smaller than the total size (5), then grab start (3) to the end(5)
so grabs 2 element?
Then sorts that list of 2 elements
then limits it (1)
So ends up with 1 correct?

That feels right to me, though seems wasteful to grab all runs to do it, but looking at the old code it did the same thing.

halkeye:
So what was the problem with the old way? was it an off by 1 error with skip or something? I like the inclusion of a new test.

runzexia:
https://github.com/jenkinsci/blueocean-plugin/pull/1939
@halkeye There were some problems with the previous test case, I updated it, and I opened a PR separately to show that there is a problem now.

halkeye:
Can you rebase with master (git fetch origin, git rebase origin/master) so we get the code up to latest? then I'll pull and take a look at the changes.

> In this case, some runs may not be able to be queried normally.

Can you show us a screenshot of the change?  feel like now if you are looking at a single branches (master) runs, you'll get a neighbouring branch (master + feature1)'s runs.

I'm not saying you are wrong, I'm just being super careful because this is potentially used everywhere.

runzexia:
image
rebase master & screenshot blueocean project
@halkeye

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/MultibranchPipelineRunContainer.java
```diff
@@ -85,22 +83,26 @@ public BlueRun get(String name) {
...
             BlueRunContainer blueRunContainer = b.getRuns();
             if(blueRunContainer==null){
                 continue;
             }
-            Iterator<BlueRun> it = blueRunContainer.iterator(0, MAX_MBP_RUNS_ROWS);
-            int count = 0;
-            Utils.skip(it, start);
-            while (it.hasNext() && count++ < limit) {
-                c.add(it.next());
+            List<BlueRun> it = Lists.newArrayList(blueRunContainer.list());
```
sophistifunk: Why create a new ArrayList that's just thrown away? why not just `allRunList.addAll(blueRunContainer.list())`?

runzexia: blueRunContainer.list() did not actually return List, but returned Iterable

runzexia:
hi @halkeye @sophistifunk 
I created a pipeline in my own environment to show the problem, I hope to get your feedback.
http://139.198.120.87:30180/blue/organizations/jenkins/project-GrAR4EXNx820%2Fin-scm/activity/

PR #1878, opened 2018-11-13
Update Dockerfile

eprats:
# Description

See [JENKINS-46799](https://issues.jenkins-ci.org/browse/JENKINS-46799).

Permission for Jenkins user to run docker commands has been added.


jpmontez:
I'm seeing a 404 on the `continuous-integration/jenkins/pr-head` job. Can this check be rerun?

halkeye:
as per the comments in the tickets, I think --group-add is the right answer.

RUN is compile time, so *I think* it'll update the socket inside the container, not the socket mounted with -v
when you do it with the ctrl+c method, it just updates your system's docker socket to be gid 1000, which I don't think is desireable.

I'm not certain though

If you feel I'm wrong, please re-open/comment on the ticket.

PR #1880, opened 2018-11-13
JENKINS-54518 - Support SCM Systems that don't implement ObjectMetadataAction

halkeye:
Add support for pulling branch name from the SCM source, at least until all scm source implement object metadata

# Description

See [JENKINS-54518](https://issues.jenkins-ci.org/browse/JENKINS-54518).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/Caches.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/Caches.java
@@ -7,17 +7,21 @@ import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import hudson.Extension;
 import hudson.model.Item;
+import hudson.model.ItemGroup;
 import hudson.model.Job;
 import hudson.model.listeners.ItemListener;
 import io.jenkins.blueocean.rest.factory.BlueIssueFactory;
 import io.jenkins.blueocean.rest.impl.pipeline.BranchImpl.Branch;
 import io.jenkins.blueocean.rest.impl.pipeline.BranchImpl.PullRequest;
+import jenkins.branch.BranchProjectFactory;
+import jenkins.branch.MultiBranchProject;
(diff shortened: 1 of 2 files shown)


halkeye:
Just wanting to get feedback before I attempt to mock all the parts needed to write a good test.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/Caches.java
```diff
@@ -95,10 +100,24 @@ public void onDeleted(Item item) {
...
+                ItemGroup parent = job.getParent();
+                if (parent instanceof MultiBranchProject) {
+                    BranchProjectFactory projectFactory = ((MultiBranchProject) parent).getProjectFactory();
+                    if (projectFactory.isProject(job)) {
+                        SCMHead head = projectFactory.getBranch(job).getHead();
+                        url = head.getName();
+                    }
+                }
+            }
+            if (StringUtils.isEmpty(url) && pima == null) {
```
davidbees: Why do  we need to check again for `isEmpty(url)` ? Couldn't you also add an `} else if(pima == null) {` on line 115 above?

Other then that the logic looks fine. Just curious why the additional check.

halkeye: Its a long while ago that I wrote this, but I'm pretty sure its because I assumed head.getName() wasn't a) guaranteed to be returned, and b) guaranteed to get there.

If url is set to something, we can not even bother to check pima.

sophistifunk:
What are the consequences of this? Is there downstream code somewhere expecting data that won't be there? What was the missing metadata being used for?

halkeye:
The issue in the original ticket showed that when you list a multi branch job, it would sort by branch (name and and other fields), but without this code, it would only be able to pull the branch name for GitHub repos. Git repos would have random sort.

sophistifunk:
Intredasting. Code LGTM.

PR #1881, opened 2018-11-14
[JENKINS-53127][DESIGN-249] add labels to parallel sequential stage branches

NicuPascu:
# Description
This adds the name of the parallel sequential stages parent stage as a label at the beginning of each "branch"

See [JENKINS-53127](https://issues.jenkins-ci.org/browse/JENKINS-53127) [DESIGN-249].

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/jenkins-design-language/src/js/components/pipeline/PipelineGraph.tsx
+++ b/jenkins-design-language/src/js/components/pipeline/PipelineGraph.tsx
@@ -3,7 +3,17 @@ import { decodeResultValue, getGroupForResult } from '../status/StatusIndicator'
 import { strokeWidth as nodeStrokeWidth } from '../status/SvgSpinner';
 import { TruncatingLabel } from '../TruncatingLabel';
 
-import { CompositeConnection, defaultLayout, LabelInfo, LayoutInfo, MATRIOSKA_PATHS, NodeColumn, NodeInfo, StageInfo } from './PipelineGraphModel';
+import {
+    CompositeConnection,
+    defaultLayout,
+    LabelInfo,
+    LayoutInfo,
+    MATRIOSKA_PATHS,
+    NodeColumn,
+    NodeInfo,
+    StageNodeInfo,
+    StageInfo,
+} from './PipelineGraphModel';
 
 import { layoutGraph } from './PipelineGraphLayout';
 
@@ -112,7 +122,19 @@ export class PipelineGraph extends React.Component {
             marginLeft: labelOffsetH,
         };
 
-        const x = details.x;
+        const hasSequentialParallelStages = () => {
+            if (details.stage) {
+                for (const stageChild of details.stage.children) {
+                    if (stageChild.seqContainerName) {
+                        return true;
+                    }
+                }
+            }
+
+            return false;
+        };
+
+        const x = hasSequentialParallelStages() ? details.x + 35 : details.x;
         const bottom = this.state.measuredHeight - details.y + labelOffsetV;
 
         // These are about layout more than appearance, so they're inline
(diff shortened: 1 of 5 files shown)


NicuPascu:
sequential_labels
@chadbercea_twitter

chadbercea:
@NicuPascu I love this! I think it looks great and I think the hover labels are a great touch as well. I'd love to get this into the real world to find out if customers find this helpful. Double thumbs up sir.

jenkins-design-language/src/js/components/pipeline/PipelineGraph.tsx
```diff
@@ -112,7 +122,19 @@ export class PipelineGraph extends React.Component {
             marginLeft: labelOffsetH,
         };
 
-        const x = details.x;
+        const hasSequentialParallelStages = () => {
```
sophistifunk: Why is this wrapped up into a function?

jenkins-design-language/src/js/components/pipeline/PipelineGraphLayout.ts
```diff
@@ -163,6 +163,9 @@ function positionNodes(nodeColumns: Array<NodeColumn>, { nodeSpacingH, parallelS
             xp += Math.round((widestRow - row.length) * parallelSpacingH * 0.5);
 
             for (const node of row) {
+                if (!node.isPlaceholder && node.stage && node.stage.seqContainerName) {
+                    xp += 70;
```
sophistifunk: We probably want to put this as an `export const` so that it has a name, and use that value (÷2) when offsetting the top label position in PipelineGraph

jenkins-design-language/src/js/components/pipeline/PipelineGraph.tsx
NicuPascu: It was just easier at the time to do it this way instead of initialising a let and then iterating through the children array and changing the initial value of the let. I can change it if you don't like how it looks. Also I used to do this kind of "method" functions in a past project and my mind randomly reverts to thinking like this sometimes.

jenkins-design-language/src/js/components/pipeline/PipelineGraphLayout.ts
NicuPascu: Will do, thanks

NicuPascu: Done :)

jenkins-design-language/src/js/components/pipeline/PipelineGraph.tsx
sophistifunk: Why not `details.stage.children.find(child => child.seqContainerName)` if you're offended by mutating a local ;-)

sophistifunk: ... or maybe extract an exported function from layout, and use that for all the places you're checking for almost exactly the same thing?

... or leave it, I'm just tossing out ideas.

NicuPascu: No, I think this is quite the perfect place to talk about ways of writing code that is as easy to read as possible because we have concrete examples to which to debate on.

NicuPascu: For the details.stage.children.find() solution, it won't really solve the issue of the function because I would still need to check for details.stage before doing a find on its children.

And for the creation of a general function that could be re-usable, in this case we search the children of the stage to see if any of them have a "seqContainerName" and in the other parts we check the stage itself if it has a "seqContainerName".

sophistifunk:
These changes look good, I'd like to get them into master (and the widget version). Are we gonna leave this PR open until you've beaten Storybook into submission and include those changes here alongside?

PR #1882, opened 2018-11-23
[JENKINS-54889] Avoid NPE with graphlistener (null check on value before adding to the map)

olamy:
Signed-off-by: olivier lamy <olamy@apache.org>

# Description

See [JENKINS-54889](https://issues.jenkins-ci.org/browse/JENKINS-54889).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
+++ b/blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
@@ -71,8 +71,12 @@ public class PipelineEventListener implements GraphListener {
         // test whether we have a stage node
         if (PipelineNodeUtil.isStage(flowNode)) {
             List<String> branch = getBranch(flowNode);
-            currentStageName.put(flowNode.getExecution(), flowNode.getDisplayName());
-            currentStageId.put(flowNode.getExecution(), flowNode.getId());
+            if(flowNode.getExecution()!=null && flowNode.getDisplayName()!=null) {
+                currentStageName.put(flowNode.getExecution(), flowNode.getDisplayName());
+            }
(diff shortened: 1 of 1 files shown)


imeredith:
should there be a try catch around whole method? incase there is ever any other issues?

imeredith:
nvm i see cps pr, :bee:

olamy:
IMHO nope.
Reading javadoc: https://docs.oracle.com/javase/10/docs/api/java/util/concurrent/ConcurrentHashMap.html#put(K,V)
null check on values/keys should be enough. (it's not javascript code so everything is under control and documented :P )

blueocean-events/src/main/java/io/jenkins/blueocean/events/PipelineEventListener.java
```diff
@@ -71,8 +71,12 @@ public void onNewHead(FlowNode flowNode) {
         // test whether we have a stage node
         if (PipelineNodeUtil.isStage(flowNode)) {
             List<String> branch = getBranch(flowNode);
-            currentStageName.put(flowNode.getExecution(), flowNode.getDisplayName());
-            currentStageId.put(flowNode.getExecution(), flowNode.getId());
+            if(flowNode.getExecution()!=null && flowNode.getDisplayName()!=null) {
```
kzantow: My only question about this is what scenarios is `flowNode.getExecution()` ever `null`? Just curious if it's indicative of a bigger/different problem.

olamy: well sure flowNode.getExecution() shouldn't be null. I will remove but this doesn't hurt neither have a big impact....

PR #1885, opened 2018-11-30
[JENKINS-51716] _links could not be defined on some BO objects

kuisathaverat:
# Description

See [JENKINS-51716](https://issues.jenkins-ci.org/browse/JENKINS-51716).

# Submitter checklist
  * Go to the activity page
  * Open dev tool on chrome
  * Set a breakpoint on Pager.fetchNextPage method
  * Reload the page
  * Change the value of the data variable with the data.json file provided in the Jira
  * Continue the execution, now the page is loaded. 

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-core-js/src/js/services/BunkerService.js
+++ b/blueocean-core-js/src/js/services/BunkerService.js
@@ -33,7 +33,11 @@ export class BunkerService {
      * @returns {any} The key for the store.
      */
     bunkerKey(data) {
+      if(data && data._links && data._links.self && data._links.self.href){
         return data._links.self.href;
+      } else {
+        return Math.random().toString(36).substring(2) + (new Date()).getTime().toString(36);
+      }
     }
 
     /**
--- a/blueocean-core-js/test/js/data/BunkerService-data-1.json
+++ b/blueocean-core-js/test/js/data/BunkerService-data-1.json
@@ -1,0 +1,99 @@
+{
+    "_class": "io.jenkins.blueocean.rest.impl.pipeline.PipelineRunImpl",
+    "actions": [],
+    "artifactsZipFile": null,
+    "causeOfBlockage": null,
+    "causes": [{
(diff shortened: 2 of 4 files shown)


halkeye:
@kuisathaverat can you rebase this branch?

Also, can you add a test for your change? see cat test/js/services/LocationService-spec.js for an example. The json file you have in the ticket would be a great test case. Don't worry about the unrelated changes, just the one you made.

If your not comfortable, I'll see if i can help out.

kuisathaverat:
@halkeye I'm going to try to make it today.

halkeye:
Ack, Finally back to reviewing this, we've been busy trying to stabalize our tests and slowly working through older PRs

blueocean-core-js/src/js/services/BunkerService.js
```diff
@@ -33,7 +33,11 @@ export class BunkerService {
      * @returns {any} The key for the store.
      */
     bunkerKey(data) {
+      if(data && data._links && data._links.self && data._links.self.href){
         return data._links.self.href;
+      } else {
+        return Math.random().toString(36).substring(2) + (new Date()).getTime().toString(36);
```
halkeye: So asking around, all items in the bunker should have a deterministic key. Using a random string fixes the symptom, but doesn't fix the problem.

In your BunkerService-data, you have `"_class": "io.jenkins.blueocean.rest.impl.pipeline.PipelineRunImpl",` for both cases. From your testing (cause i don't really know how to reproduce it), is it only pipelinerunimpl that don't have a self link?

imeredith:
If the Rest API is not returning the self href, the backend needs to be fixed. 

The other potential source is from the SSE events. We sometimes have to generate our own hrefs https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-core-js/src/js/sse/SseBus.js

kuisathaverat:
I agree it is a symptom of some issue in the backend, this PR is a dirty workaround, the fix should be made on the backend, and it is probably made because I did not see the issue anymore since a couple of versions of so.

PR #1889, opened 2018-12-14
Task/replace graph

sophistifunk:
# Description

Updates BO to use the newest graph control and namespaced styles (from ux-widgets).


sophistifunk:
Not quite ready for merge yet, just want a PR open for running ATH.

blueocean-dashboard/package.json
```diff
@@ -60,8 +60,9 @@
   "dependencies": {
     "@jenkins-cd/blueocean-core-js": "0.0.171",
     "@jenkins-cd/design-language": "0.0.162",
-    "@jenkins-cd/js-modules": "0.0.10",
     "@jenkins-cd/js-extensions": "0.0.44",
+    "@jenkins-cd/js-modules": "0.0.10",
+    "@jenkins-cd/pipeline-graph-widget": "^0.1.2",
```
NicuPascu: caret :D

sophistifunk:
NFI why there's no option for "this is done" rather than "dismiss" but ¯\_(ツ)_/¯

NicuPascu:
LGTM :bee:

sophistifunk:
Hmm. One of these ATH failures looks like it might be legit...

sophistifunk:
`RestartStageTest` is a valid failure, still trying to lock down the exact changes to bring it back to life. Don't think the other failures are though.

PR #1897, opened 2019-01-08
Testing against Jenkins 2.138.4

sophistifunk:
Just need the PR for running ATH in the cloud for now.


sophistifunk:
Nothing to see here for now.

acceptance-tests/src/test/js/failing.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
      * @param browser
      */
     'Step 01': function (browser) {
-        browser.login();
+        // browser.login();
```
vivek: Is this intentional? I guess login step is not needed due to new login UI or perhaps due to by default anonymous user has admin permission? In any case, sone documentation to clarify why its commented out will be good.

sophistifunk: These aren't changes for merging, I'm just trying to figure out what's going on. But I did get a green, so not only is the login not required, but it's also the thing where it goes wrong. I'm gonna just ditch some/all of these tests and replace them with Java code we can debug.

PR #1910, opened 2019-02-04
Update SSE client lib used in ATH

sophistifunk:
Pull in updated SSE Client rule that supports later Jenkins versions.


sophistifunk:
Not making any changes to Jenkins version requirements here, just the ATH changes we'll need.

sophistifunk:
Thanks for all the help @halkeye

PR #1911, opened 2019-02-05
Update version of npm package of core js

scherler:


blueocean-core-js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@jenkins-cd/blueocean-core-js",
-  "version": "0.0.171",
+  "version": "1.10.3",
```
kzantow: There are lots of references to `@jenkins-cd/blueocean-core-js@0.0.171` in the various `package.json` files that probably should be updated, too.

scherler: Should their?

I really do not know since BO switched to core-js being a hpi plugin and auto-overriding those version anyway. 

I asked that changes to the core-js plugin (a js only package) should been published to npm to be consumable as npm upstream but the latest security updates to fetch had not been published, 

As I understand the only ones that need that version to be in sync are upstream consumer so no other package.json in BO is minding that version anyway, saying it, since 11 month ago I needed the version to be changed (git blamed myself) and now (no other changes).

...I actually published that version in latest version as npm package to make it usable in our company project.

halkeye:
Hey @scherler 

1) can you rebase with master
2) is this just house keeping? or does it fix a problem? what triggered this change?

NicuPascu:
I think it is safe to close this because I already released new versions of core-js and the JDL in https://github.com/jenkinsci/blueocean-plugin/pull/1916

PR #1912, opened 2019-02-08
[JENKINS-55986] Don't show the "Restart Stage" link if permissions do not allow it

NicuPascu:
Hides the "Restart Stage" link in case start permissions are not granted

See [JENKINS-55986](https://issues.jenkins-ci.org/browse/JENKINS-55986).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/StageRestartLink.tsx
+++ b/blueocean-dashboard/src/main/js/components/StageRestartLink.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
-import { RunApi as runApi } from '@jenkins-cd/blueocean-core-js';
-import { ToastUtils } from '@jenkins-cd/blueocean-core-js';
+import { Security, ToastUtils, RunApi as runApi } from '@jenkins-cd/blueocean-core-js';
 
 interface Props {
     title: string;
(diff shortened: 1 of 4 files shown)


blueocean-dashboard/src/main/js/components/StageRestartLink.tsx
```diff
@@ -74,12 +73,16 @@ class StageRestartLink extends React.Component<Props, State> {
...
-        );
+        if (Security.permit(this.props.pipeline).start()) {
+            return (
+                <a className="restart-stage" onClick={() => this._onRestartStageClick()}>
+                    <RestartStageIcon />
+                    <span>{this.props.t('rundetail.logToolbar.restartStage', { 0: this.props.title })}</span>
+                </a>
+            );
+        } else {
+            return false;
```
halkeye: won't this render "false" or something? I don't think you need the fallthrough return statement

halkeye:
@NicuPascu can you add a test of some sort to confirm this exact bug wont happen again in the future

blueocean-dashboard/src/main/js/components/StageRestartLink.tsx
sophistifunk: Nah react won't render `false` or `undefined`. You have to be careful about your truthyness tho, if you've got a comparison that can return 0 rather than false it *will* render that.

NicuPascu:
@halkeye I have added a unit test that should ensure this doesn't regress in the future

blueocean-dashboard/src/main/js/components/StageRestartLink.tsx
NicuPascu: yup, in this case there is no danger of things being rendered incorrectly, it either shows the link or it doesn't render any tag or content

halkeye:
Lgtm!

davidbees:
> @NicuPascu can you add a test of some sort to confirm this exact bug wont happen again in the future

This is probably a good idea considering we now own the addition of ensuring our stuff is properly tested.

NicuPascu:
As per our discussion, the test added is sufficient for this change so I will go ahead and merge this PR

summergan:
hi everybody, I am login Jenkins and using a scripted pipeline then can't find the ‘restart stage’ link. But if using Declarative pipeline I can find it. I hope I can use restart any stage when using Scripted pipeline, what should I do？

halkeye:
Open a new ticket, or don't use blueocean, its pretty abandoned. Don't try to revive really old tickets, those who need to see it won't be able to see it.

summergan:
Got it.Thanks

PR #1924, opened 2019-03-04
Fix ATH flaky tests

NicuPascu:
Trying to fix some tests that keep failing on CI

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -57,7 +57,7 @@ node() {
 
         stage('ATH - Jenkins 2.138.4') {
           timeout(time: 90, unit: 'MINUTES') {
-            sh "cd acceptance-tests && ./run.sh -v=2.138.4 --no-selenium --settings='-s ${env.WORKSPACE}/settings.xml'"
+            sh "cd acceptance-tests && bash -x ./run.sh -v=2.138.4 --no-selenium --settings='-s ${env.WORKSPACE}/settings.xml'"
             junit 'acceptance-tests/target/surefire-reports/*.xml'
             archive 'acceptance-tests/target/screenshots/**/*'
           }
@@ -65,19 +65,19 @@ node() {
 
         if (env.JOB_NAME =~ 'blueocean-weekly-ath') {
           stage('ATH - Jenkins 2.73.2') {
-            sh "cd acceptance-tests && ./run.sh -v=2.73.2 --no-selenium --settings='-s ${env.WORKSPACE}/settings.xml'"
+            sh "cd acceptance-tests && bash -x ./run.sh -v=2.73.2 --no-selenium --settings='-s ${env.WORKSPACE}/settings.xml'"
             junit 'acceptance-tests/target/surefire-reports/*.xml'
           }
           stage('ATH - Jenkins 2.73.3') {
-            sh "cd acceptance-tests && ./run.sh -v=2.73.3 --no-selenium --settings='-s ${env.WORKSPACE}/settings.xml'"
+            sh "cd acceptance-tests && bash -x ./run.sh -v=2.73.3 --no-selenium --settings='-s ${env.WORKSPACE}/settings.xml'"
             junit 'acceptance-tests/target/surefire-reports/*.xml'
           }
           stage('ATH - Jenkins 2.107.2') {
-            sh "cd acceptance-tests && ./run.sh -v=2.107.2 --no-selenium --settings='-s ${env.WORKSPACE}/settings.xml'"
+            sh "cd acceptance-tests && bash -x ./run.sh -v=2.107.2 --no-selenium --settings='-s ${env.WORKSPACE}/settings.xml'"
             junit 'acceptance-tests/target/surefire-reports/*.xml'
           }
           stage('ATH - Jenkins 2.121.1') {
-            sh "cd acceptance-tests && ./run.sh -v=2.121.1 --no-selenium --settings='-s ${env.WORKSPACE}/settings.xml'"
+            sh "cd acceptance-tests && bash -x ./run.sh -v=2.121.1 --no-selenium --settings='-s ${env.WORKSPACE}/settings.xml'"
             junit 'acceptance-tests/target/surefire-reports/*.xml'
           }
         }
--- a/acceptance-tests/src/main/java/io/blueocean/ath/WaitUtil.java
+++ b/acceptance-tests/src/main/java/io/blueocean/ath/WaitUtil.java
@@ -6,6 +6,7 @@ import org.openqa.selenium.By;
 import org.openqa.selenium.JavascriptExecutor;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.NotFoundException;
+import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.WebDriver;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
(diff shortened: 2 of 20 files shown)


acceptance-tests/src/main/java/io/blueocean/ath/WaitUtil.java
```diff
@@ -127,6 +128,9 @@ public void click(By by) {
                     logger.info(String.format("Retry click successful on attempt " + i + " for %s", by.toString()));
                 }
                 return;
+            } catch (StaleElementReferenceException ex) {
+                // item went away while we were trying to click (like auto complete)
+                // so try again
```
sophistifunk: Why do I feel we're swimming upstream here?

sophistifunk:
LGTM, although the need for it kinda bums me out.

acceptance-tests/src/main/java/io/blueocean/ath/WaitUtil.java
halkeye: I mean you are not wrong. But in this case, javascript destroys the element we just got before we had a chance to do something with it. I think thats a safe thing to catch most places. We'll see

halkeye:
So this evolved into a more general "fix ath" type PR

Lots of making sure sse is ready before we do anything
I finally noticed the fav test failing sometimes because of pagination, it couldnt' find the job it needed to fav.  So slowly added cleanup of itself to the other tests.

halkeye:
@NicuPascu @sophistifunk I've rebased this with master now to get all the other fixes lately.

I think its good for a review and merge cause it seems to fix the big bad ones. 

I know I added a wait function that is no longer used, do you think its worth deleting now? or just keeping it for usefulness later?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
```diff
@@ -243,6 +240,18 @@ public HttpResponse restart(StaplerRequest request) {
         return null;
     }
 
+    protected static WorkflowRun getRun(WorkflowJob job, long itemId) throws InterruptedException {
+        WorkflowRun restartRun = QueueUtil.getRun(job, itemId);
+        if (restartRun == null) {
+            long startTimeMs = System.currentTimeMillis();
+            while (restartRun == null && System.currentTimeMillis()-startTimeMs < 1000) {
```
olamy: could we make this `1000` configurable? like ` Integer.getInteger("wait.job.inqueue", 1000) `

halkeye: Done

Jenkinsfile
```diff
@@ -57,27 +57,27 @@ node() {
 
         stage('ATH - Jenkins 2.138.4') {
           timeout(time: 90, unit: 'MINUTES') {
-            sh "cd acceptance-tests && ./run.sh -v=2.138.4 --no-selenium --settings='-s ${env.WORKSPACE}/settings.xml'"
+            sh "cd acceptance-tests && bash -x ./run.sh -v=2.138.4 --no-selenium --settings='-s ${env.WORKSPACE}/settings.xml'"
```
sophistifunk: 👍

acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/GithubCreationPage.java
```diff
@@ -26,18 +26,6 @@ public GithubCreationPage(WebDriver driver) {
         PageFactory.initElements(driver, this);
     }
 
-    @FindBy(css = "button.github-creation")
```
sophistifunk: I think we probably want to keep the strings themselves as final consts, but these aren't the kind of magic I like to purge anyway, as you could find them using Java resolution rather than relying on the implicit namespace where you're stringly-typed and using `@someNameYouCantResolve` inside selector strings.

I'll see mysef out.

acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/GithubCreationPage.java
```diff
@@ -131,10 +117,12 @@ public void beginCreationFlow(String jobName) throws IOException {
     }
 
     public void completeCreationFlow(String apiKey, String org, String pipeline, boolean createJenkinsFile) {
-        if(wait.until(wait.orVisible(
-            driver -> apiKeyInput,
-            driver -> driver.findElement(getOrgSelector(org)))) == 1) {
-
+        if(wait.until(
```
sophistifunk: howcome we have to nest these two things called "wait"?

acceptance-tests/src/test/java/io/blueocean/ath/offline/multibranch/CommitMessagesTest.java
```diff
@@ -41,12 +47,19 @@
     @Inject
     MultiBranchPipelineFactory mbpFactory;
 
+    @Inject
+    ClassicJobApi jobApi;
+
+    @After
+    public void tearDown() throws IOException {
```
sophistifunk: Do we really want this? It might make debugging harder if you can't mess with jenkins via VNC. The whole thing gets splatted at the end, on account of it being a container that lives only until run.sh ends

sophistifunk:
LGTM 🐝

acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/GithubCreationPage.java
halkeye: The annotation was just a wrapper around finding the element, the problem was the element was found when the page object was created, not when it the property was requested, so sometimes the element wasn't ready or had gone stale or whatever.

I'm not sold on constants for all the selectors, but its an easy change if desireable

sophistifunk: > the problem was the element was found when the page object was created, not when it the property was requested

Ew. Carry on 😀

acceptance-tests/src/main/java/io/blueocean/ath/pages/blue/GithubCreationPage.java
halkeye: The "wait" object seems to contain a bunch of dom related utils. It probably should be refactored to waitUntilOneIsVisible, but i honestly don't know where else its being used. orVisible returns 0 or 1 depending on which path is true.

Its kinda gross, but I rather save it to another pass

acceptance-tests/src/test/java/io/blueocean/ath/offline/multibranch/CommitMessagesTest.java
halkeye: Yea, sadly, some of the tests, like favs, are actually clicking on items, and based on some of the github creation tests retrying, sometimes there's too many items and the fav items show up on the next page. So I got a little agressive on cleaning up tests.

I agree, it does make things a bit harder, but you can always comment it out or something right?

sophistifunk: Yeah makes sense. Just checking there's a reason is all :)

PR #1926, opened 2019-03-04
Add most of the developers to the codeowners list

halkeye:
Update Codeowners to match the current list of developers.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -2,8 +2,11 @@
 # Each line is a file pattern followed by one or more owners.
 # See: https://github.com/blog/2392-introducing-code-owners for details
 
+# Everything else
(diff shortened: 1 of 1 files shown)


halkeye:
@vivek do you want to stay on it?
@michaelneale do you want to be added

I'm wondering if its worth adding `@jenkinsci/blueocean-plugin-developers` (https://github.com/orgs/jenkinsci/teams/blueocean-plugin-developers/members) so it doesn't need to be updated all the time

michaelneale:
I don't mind staying

halkeye:
> I don't mind staying

you were apparently never on it

vivek:
>do you want to stay on it?
@halkeye yes, don’t mind staying on it.

PR #1932, opened 2019-03-07
[JENKINS-51716] Prevent non numeric run id from throwing exception when a run is actually found

halkeye:
# Description

See [JENKINS-51716](https://issues.jenkins-ci.org/browse/JENKINS-51716).

Super hard to test manually, since you need to have runs that have non numeric id. I believe jenkins has deprecated id, so most jobs will be null or a number.

If you can get that, then testing is just hitting any rest interface that returns the run. The front page will try and return the latest run for a job.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


sophistifunk:
bbm-bueno

PR #1933, opened 2019-03-14
Fix for JENKINS-56383 without rolling back fix for JENKINS-38339

sophistifunk:
# Description

Due to some shortcomings in our unit tests (and me not doing enough manual testing it seems) the fix for some edge cases where downstream links weren't showing broke rendering of (some) pipelines with parallel branches.

We rolled back the change as a hotfix (v 1.13.2) as the regression was a more egregious bug than the missing downstream links, but this patch builds on the original fix for 38339 to address the regression.

This patch includes a bunch of new regression tests that explicitly check every node and connection in the graphs, as well as checking for downstream actions where they should be. And a 3 line fix for the regression :) 

See [JENKINS-56383](https://issues.jenkins-ci.org/browse/JENKINS-56383).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/test/resources/io/jenkins/blueocean/rest/impl/pipeline/restartStage.jenkinsfile
+++ b/blueocean-pipeline-api-impl/src/test/resources/io/jenkins/blueocean/rest/impl/pipeline/restartStage.jenkinsfile
@@ -41,7 +41,7 @@ pipeline {
     }
     stage('Deploy') {
       parallel {
-        stage('Deploy') {
+        stage('DeployX') {
           steps {
             echo 'mvn source:jar package -Dmaven.test.skip'
           }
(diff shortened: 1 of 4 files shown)


sophistifunk:
Every time I say "regression", drink!

/me falls off chair

sophistifunk:
ATH failure in `RestartStageTest` *might* be real, I'm looking into it while I wait for a rebuild.

halkeye:
If it's saying 500 error in the JS logs, I think it's the same issue I've
been poking at since Friday. Nicu said it used to show up sometimes on Linux

sophistifunk:
Never mind!

supergreen

blueocean-pipeline-api-impl/src/test/resources/io/jenkins/blueocean/rest/impl/pipeline/restartStage.jenkinsfile
```diff
@@ -41,7 +41,7 @@ pipeline {
     }
     stage('Deploy') {
       parallel {
-        stage('Deploy') {
+        stage('DeployX') {
```
halkeye: What's this change about?

sophistifunk: The new unit tests are based on stage names for readability, so they won't work on pipelines that have multiple stages with the same name.

sophistifunk:
@halkeye just using existing jenkinsfiles from existing tests. Those tests are looking for different and more specific things, whereas the new tests are explicitly just for ensuring the graph structure. I didn't include any test pipelines that don't finish, or do weird / duplicate stuff.

halkeye:
Then yea looks good to me. Was curious

PR #1934, opened 2019-03-20
Fix executor plugin

halkeye:
One of the recent merges seemed to have deleted a bunch of files in the executor plugin.

This PR just adds them back in.


halkeye:
ath *seems* unrelated, but master is so hard to tell

PR #1938, opened 2019-03-21
JENKINS-46193 - Add support for filtering out non regressed failures.

halkeye:
Summary does it, frontend needs to do it too.

Also included ability to filter by age

# Description

See [JENKINS-46193](https://issues.jenkins-ci.org/browse/JENKINS-46193).

# Submitter checklist
- [] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below.

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/testing/TestService.js
+++ b/blueocean-dashboard/src/main/js/components/testing/TestService.js
@@ -3,18 +3,6 @@ import TestLogService from './TestLogService';
 
 const PAGE_SIZE = 100;
 
-// Source: https://stackoverflow.com/questions/286141/remove-blank-attributes-from-an-object-in-javascript/38340730#38340730
-const stripNullAndUndefined = obj =>
-    Object.keys(obj)
-        .filter(k => obj[k] !== null && obj[k] !== undefined) // Remove undef. and null.
-        .reduce(
-            (newObj, k) =>
-                typeof obj[k] === 'object'
-                    ? Object.assign(newObj, { [k]: stripNullAndUndefined(obj[k]) }) // Recurse.
-                    : Object.assign(newObj, { [k]: obj[k] }), // Copy value.
-            {}
-        );
-
 export default class TestService extends BunkerService {
     constructor(pagerService) {
         super(pagerService);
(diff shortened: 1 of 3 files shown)


halkeye:
I feel bad about including the age filter. I ended up doing the !REGRESSED support instead as its closer to summary. But I also didn't want to throw it away.

halkeye:
[JENKINS-46193.zip](https://github.com/jenkinsci/blueocean-plugin/files/2993480/JENKINS-46193.zip)

if you want to test, i just pulled the job that was originally mentioned in the ticket locally

blueocean-dashboard/src/main/js/components/testing/TestService.js
```diff
@@ -3,6 +3,18 @@ import TestLogService from './TestLogService';
 
 const PAGE_SIZE = 100;
 
+// Source: https://stackoverflow.com/questions/286141/remove-blank-attributes-from-an-object-in-javascript/38340730#38340730
+const stripNullAndUndefined = obj =>
```
sophistifunk: If you end up needing this, don't do it like this :) Just `obj => JSON.parse(JSON.stringify(obj))` is easier might be faster, and instead of hanging up will throw if a cycle is detected.

Also IMO (I'm not signing the cheques so feel free to ignore, just a style nitpick) top level functions should only be in arrow form if they're React components, and then only to make it clear they're not constructors.

blueocean-dashboard/src/main/js/components/testing/TestService.js
```diff
@@ -47,17 +59,12 @@ export default class TestService extends BunkerService {
         return this._logs;
     }
 
-    static createURL({ run, status, state }) {
-        if (status && state) {
-            throw new Error('status and state are exclusive');
-        }
+    static createURL({ run, status, state, age }) {
```
sophistifunk: This isn't exported, so I'm not sure why exactly the signature is what it is. I'd change the signature to this:

`static createURL(run, searchParams)`

Then explicitly check `searchParams` for "at least one" if that check is really needed, and then just hand that object off to `new URLSearchParams()`. Make the calling functions stop putting null props into it in the first place.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/BlueTestResultContainerImpl.java
```diff
@@ -63,62 +63,93 @@ public BlueTestResult get(final String name) {
...
-            results = resolved.results.iterator();
+            String age = request.getParameter("age");
+            return getBlueTestResultIterator(resolved.results, status, state, age);
+
+        }
+        return resolved.results.iterator();
+    }
+
+    @VisibleForTesting
+    public Iterator<BlueTestResult> getBlueTestResultIterator(Iterable<BlueTestResult> results, String status, String state, String age) {
```
sophistifunk: You've updated the backend to gracefully handle missing search params here it seems, probably can remove the exception at the front-end if they're missing, also.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/BlueTestResultContainerImpl.java
```diff
@@ -147,4 +178,29 @@ public boolean apply(@Nullable BlueTestResult input) {
             return input != null && input.getTestState().equals(state);
         }
     }
+
+    static class AgePredicate implements Predicate<BlueTestResult> {
```
sophistifunk: This doesn't appear to bind to anything in the parent class, that I can see (it's hard to be sure using GH instead of Idea, so ignore me if I'm wrong) but if it can be moved to file scope rather than nested it's probably nicer to do so. It surely makes little difference in this case, but no need to bloat classes if we don't need to.

halkeye: Agreed, but also super lazy, and since they were there already, Imma going to leave them.

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/BlueTestResultContainerImpl.java
halkeye: Agreed, done

blueocean-dashboard/src/main/js/components/testing/TestService.js
halkeye: Agreed, I like your idea way better, so done. Way cleaner

blueocean-dashboard/src/main/js/components/testing/TestService.js
halkeye: Not needed now anyways. Done

sophistifunk:
LGTM 🐝 I have NFI how to make github record an approval, I've never found the button. Not sure if I'm missing a chromosome, or UI is just that poorly thought through.

notsure

PR #1940, opened 2019-03-22
Allow HTML tags inside Alert messages...

dvdliao:
such as adding `<a>` for links

# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/jenkins-design-language/src/js/components/Alerts.jsx
+++ b/jenkins-design-language/src/js/components/Alerts.jsx
@@ -11,7 +11,7 @@ export class Alerts extends Component {
                 <div className="border">&nbsp;</div>
                 <div className="title-div">
                     <span className="title">{title}</span>
-                    <span className="message">{message}</span>
+                    <span className="message" dangerouslySetInnerHTML={{ __html: message }} />
                 </div>
             </div>
         );


dvdliao:
We're currently trying to get our artifacts stored in GCS to appear on the artifacts page (if you know a work around for this please point in the direction 🙂) but since that didnt workout we decided to link to the bucket path `https://gcs/jobname/buildnumber/` but this url isnt clickable in blueocean. Its currently already clickable in normal jenkins view.

They message is something along the lines of `Artifacts are available <a href="link">here</a>`

halkeye:
So this would open us up to XSS (user contributed bad javascript). Where is the message being generated? the element should already (according to the storybooks i'm trying to fix) handle incoming jsx with html, but not user contributed random html.

dvdliao:
We are setting the build description in our own jenkins library that does the gcs uploading
`currentBuild.description = "Artifacts are available <a href='link'>here</a>"` Definitely cant be putting JSX in here

jpigree:
Hey. This will be useful for us too.

If you are concerned about the security risks, why not do exactly like the formatting inside the vanilla Jenkins UI?

It enables the user to display user populated text in "plain text" or in "Safe HTML" (Manage Jenkins => Configure global security => Markup Formatter).

halkeye:
Yea, that would be a lot better solution than to allow random html in the UI library. It could be done in the specific location (build description). 

There's an open ticket if you are interested in contributing something.

If we could safely guarentee that the run.description is safe html, then 
https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-dashboard/src/main/js/components/karaoke/components/RunDescription.jsx#L16
could be updated to something like:

```
<Alerts title={this.props.t('rundetail.pipeline.description')} message={<span dangerouslySetInnerHTML={{ __html: this.props.run.description }}</span>} />
```

But I don't think any of us are comfortably blindly trusting any user input at all to be html. It would have to go through the sanitizer first.

jpigree:
@halkeye . I agree.

I don't see any issues on this github repository however.  Where is the open ticket you are talking about?

I don't know how difficult this fix would be (SafeHTML + dangerouslySetInnerHTML on fields like description).
Do you have an estimation?

Thanks

halkeye:
Oh my bad, i had the tab open but forgot to copy and paste - https://issues.jenkins-ci.org/browse/JENKINS-55599 (I should transcribe a bunch of these notes)

I don't know the API offhand, but a quick search of jenkins codebase I found:

https://github.com/jenkinsci/jenkins/blob/0795e89b308ec7fcbda097858d58763d8531be8c/core/src/main/java/hudson/model/ParameterDefinition.java#L141-L152

So i'd say the blueocean rest api needs to be updated to escape description everywhere, then ui needs to be updated to trust description everywhere.

I'd say its a pretty quick task, maybe a day or two, perfect for a new contributor. the CB blueocean people probably won't get to it anytime soon, we are working on stability and bug fixes mainly for the immediate future.

jpigree:
Okay. I will look into it. Thanks for all the informations. :)

halkeye:
I would also make a new @Exported field that is like filteredDescription or
something, so we don't ever accidently forget a description somewhere and
accidently cause xss

PR #1948, opened 2019-04-09
JENKINS-55497 - Add ChangeSet information to its own unique container

halkeye:
UI Element updated to use the preload/pager data
Left all previous data behind, but we should still delete all the
run.changeSet stuff

cc @stuartrowe 
# Description

See [JENKINS-55497](https://issues.jenkins-ci.org/browse/JENKINS-55497).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ChangeSetResource.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ChangeSetResource.java
@@ -107,6 +107,6 @@ public class ChangeSetResource extends BlueChangeSetEntry {
 
     @Override
     public Link getLink() {
-        return parent.getLink().rel("changeset/"+getCommitId());
+        return parent.getLink().rel(getCommitId());
     }
 }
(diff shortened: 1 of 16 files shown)


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ChangeSetResource.java
```diff
@@ -107,6 +107,6 @@ public BlueChangeSetEntry setCheckoutCount(int checkoutCount) {
 
     @Override
     public Link getLink() {
-        return parent.getLink().rel("changeset/"+getCommitId());
+        return parent.getLink().rel(getCommitId());
```
sophistifunk: Shouldn't this just be `parent.getLink().rel("changeSet")`?

halkeye: The container is "changeSet", this is the Id of the actual element inside of the container.
so parent.getlink() === "changeSet"
...
I think at least.

sophistifunk: Right, I see. Poorly named classes (not your doing I know) confused me :(

PR #1950, opened 2019-04-12
JENKINS-57209 - Use UUID for bitbucket

halkeye:
# Description

See [JENKINS-57209](https://issues.jenkins-ci.org/browse/JENKINS-57209).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/cloud/model/BbCloudTeam.java
```diff
@@ -30,17 +30,17 @@ public BbCloudTeam(@Nonnull @JsonProperty("username") String teamName,
         }
     }
 
-    public BbCloudTeam(String teamName,
+    public BbCloudTeam(String uuid,
```
vivek: Does it also need any changes in the UI, it might get used in the pipeline creation wizard. I also wonder if username field should be preserved, their migration guide is clear about preserving username in the URL but silent about whether response will have username removed.

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/model/BbServerUser.java
```diff
@@ -14,16 +14,14 @@
...
 public class BbServerUser extends BbUser {
-    private final String name;
     private final String displayName;
     private final String slug;
     private final String emailAddress;
     private final String avatar;
 
     @JsonCreator
-    public BbServerUser(@Nonnull @JsonProperty("name") String name, @Nonnull @JsonProperty("displayName") String displayName,
+    public BbServerUser(@Nonnull @JsonProperty("displayName") String displayName,
```
vivek: No changes in server APIs, we should not change this.

halkeye: You suggested we remove getName() (as its not used anywhere). The constructor isn't data bound, so I figured it was safe to remove the field.

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/cloud/model/BbCloudTeam.java
halkeye: I'll have to look more monday, We are calling getUser() which returns who they are, but there's probably at least one place that calls getUser("name")

halkeye:
Apparently not needed, I was confused as to what was changing. Closing

halkeye:
Everything seems to work after @vivek found out how to force the new behavior for testing, it didn't without this PR, so i've re-opened it.

The only negative thing about this change is it loses the username in the job that was created. This is because the return values no longer return username. UUID works just fine, and if someone changes it to username, it'll work fine too (until atlassian changes apis again)

image (1)

vivek:
There is slight UX side effect, where bitbucket cloud pipelines created by blueocean will have owner field as uuid in the classic ui. Since username is removed there isn't much we can do in this case.

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/model/BbUser.java
```diff
@@ -11,12 +11,6 @@
  */
 @Restricted(NoExternalUse.class)
 public abstract class BbUser {
-    /**
-     * @return User name
-     */
-    @JsonProperty("name")
-    public abstract String getName();
-
```
sophistifunk: Is this added back in for non-cloud?

halkeye: Hrm.. i forgot now. I think we found out its not actually used anywhere.

blueocean-bitbucket-pipeline/src/test/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/cloud/BbCloudWireMock.java
```diff
@@ -11,6 +11,8 @@
  */
 public class BbCloudWireMock extends BitbucketWireMockBase {
 
+    protected static String USER_KEY = "{1c5c9255-d59f-47e2-b5c7-52269c0332b9}"; // vivekp7;
```
sophistifunk: Maybe change these from `_KEY` to `_UUID`? Also probably nothing added by showing the old texts in comments, it might be confusing to future maintainers thinking it requires live real stuff set up by Vivek?

halkeye: yea fair enough, the comments were for me because i was starting to get confused by the uuid's everywhere. That's vivek's uuid so we didn't have to generate more fixtures.

sophistifunk: Right, but it's not actually going to the cloud right, that's the point of having the `wiremock`?

halkeye: Correct, i can delete the comment now that its done, i needed the comment to know which value to replace with which uuid at the time. The constants helped a lot. I'll rename and delete the comment.

PR #1954, opened 2019-04-18
JENKINS-39203 - Only show blocks containing a WarningAction as unstable

dwnusbaum:
# Description

See https://github.com/jenkinsci/workflow-api-plugin/pull/91, https://github.com/jenkinsci/pipeline-graph-analysis-plugin/pull/25, https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/83, and [JENKINS-39203](https://issues.jenkins-ci.org/browse/JENKINS-39203). (I am intentionally not updating the status of that ticket or assigning it to myself until I am ready to merge/release all relevant changes to avoid giving users premature hope of a fix and so that I have clear answers for what will and won't work and when any fixes will be released.)

This PR adjusts the nodes used when calling `StatusAndTiming.computeChunkStatus2` for stages. Previously, `before` was the stage step start, `firstNode` was the first atom node in the stage, `lastNode` was the end of the stage step's body, and `after` was the stage step end. Now, `firstNode` is the start of the stage step's body. This change matches the standard behavior of `StageChunkFinder`, which means that we can create tests in pipeline-graph-analysis that actually match the way the APIs are used by Blue Ocean rather than being subtly different, and it makes it easier to look through the current block inside of `computeChunkStatus2` (I am a little curious about whether `firstExecuted` can be removed entirely, but haven't tested that out yet).

This PR also adds regression tests for [JENKINS-43292](https://issues.jenkins-ci.org/browse/JENKINS-43292).

Screenshot with the changes (without the changes, all of the yellow exclamation points would be green check marks:

Screen Shot 2019-04-24 at 17 02 02

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


pom.xml
```diff
@@ -27,7 +27,7 @@
     <revision>1.15.0</revision>
     <changelist>-SNAPSHOT</changelist>
     <java.level>8</java.level>
-    <jenkins.version>2.121.1</jenkins.version> <!-- Should be kept in sync with the version resulting from Dockerfile FROM statement-->
+    <jenkins.version>2.138.4</jenkins.version> <!-- Should be kept in sync with the version resulting from Dockerfile FROM statement-->
```
dwnusbaum: Required for the workflow-api update. Current LTS is 2.164.2, not sure if you want to go all the way to that version though.

pom.xml
```diff
@@ -45,6 +45,7 @@
     <git.version>3.8.0</git.version>
     <access-modifier-checker.failOnError>true</access-modifier-checker.failOnError>
     <workflow-support.version>3.2</workflow-support.version>
+    <pipeline-model-definition.version>1.3.7</pipeline-model-definition.version>
```
dwnusbaum: Moved to a property to make it easy to keep everything in sync, and the update is required for the JENKINS-43292 regression tests to work.

pom.xml
```diff
@@ -341,32 +342,37 @@
...
         </dependency>
         <dependency>
             <groupId>org.jenkinsci.plugins</groupId>
             <artifactId>pipeline-model-api</artifactId>
-            <version>1.3.2</version>
+            <version>${pipeline-model-definition.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkinsci.plugins</groupId>
+            <artifactId>pipeline-model-extensions</artifactId>
```
dwnusbaum: Just one of the other Declarative plugins, not sure why it wasn't here already.

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/GraphBuilderTest.java
```diff
@@ -216,15 +216,16 @@ public void downstreamBuildLinksSeq() throws Exception {
...
         WorkflowRun run = createAndRunJob("SillyLongName",
                                           "earlyUnstableStatusShouldReportPunStateAsRunningAndResultAsUnknown.jenkinsfile",
                                           Result.UNSTABLE);
         NodeGraphBuilder graph = NodeGraphBuilder.NodeGraphBuilderFactory.getInstance(run);
         List<FlowNodeWrapper> nodes = graph.getPipelineNodes();
 
-        assertStageAndEdges(nodes, "stage 1 marked as unstable", BlueRun.BlueRunState.NOT_BUILT, BlueRun.BlueRunResult.NOT_BUILT, "stage 2 wait");
-        assertStageAndEdges(nodes, "stage 2 wait", BlueRun.BlueRunState.FINISHED, BlueRun.BlueRunResult.UNSTABLE);
+        assertStageAndEdges(nodes, "stage 1 marked as unstable", BlueRun.BlueRunState.FINISHED, BlueRun.BlueRunResult.UNSTABLE, "stage 2 wait");
+        assertStageAndEdges(nodes, "stage 2 wait", BlueRun.BlueRunState.FINISHED, BlueRun.BlueRunResult.SUCCESS);
```
halkeye: This test changed, does it mean existing behaviour might break?

dwnusbaum: Yeah this test in particular is the exact case described in JENKINS-39203. `stage 2 wait` succeeded, but previously it would have been shown as unstable. The fix is that only the stage that actually set the status to unstable should be marked as unstable. I'm not sure why stage 1 was previously computed to be NOT_BUILT, but I will see if I can figure it out.

halkeye: Its probably all good, i just get super nervous when tests change. Josh will be able to give better feedback next week

dwnusbaum: Yeah this test is definitely expected to change since it was asserting exactly the behavior that we don't want anymore (i.e. it was testing the bug). No rush on reviewing, this is on top of 3 other PRs and will likely change over time, so probably better to wait to review until I upgrade it from draft status to real PR status anyway.

blueocean-pipeline-api-impl/src/test/resources/io/jenkins/blueocean/rest/impl/pipeline/earlyUnstableStatusShouldReportPunStateAsRunningAndResultAsUnknown.jenkinsfile
```diff
@@ -1,6 +1,6 @@
 node {
     stage ("stage 1 marked as unstable") {
-        currentBuild.result = 'UNSTABLE';
+        unstable('foo')
```
jglick: We should look into whether `RunWrapper.setResult` can be deprecated. Tricky because this was done as a method on a POJO rather than a `Step`, so we have no `TaskListener` available. Ideally `currentBuild` would be deprecated as a whole: https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/71

pom.xml
jglick: So do you need to update some `Dockerfile` as the comment asks?

halkeye:
I can't seem to reply to the comment, but I think the comment was resolved a long time ago as https://github.com/jenkinsci/blueocean-plugin/blob/master/docker/official/Dockerfile says lts? maybe?

pom.xml
dwnusbaum: The Dockerfile just [uses `FROM jenkins/jenkins:lts`](https://github.com/jenkinsci/blueocean-plugin/blob/d0f12cb922e608105a190ebd4126ff0ef4f3b9c7/Dockerfile#L10), and that version is currently 2.164.2. I wasn't sure if we want to update to such a new version, hence the comment. Perhaps the comment should just be deleted now that the Dockerfile uses the latest LTS.

jglick: > Perhaps the comment in pom.xml should just be deleted now that the Dockerfile always tracks the latest LTS.

Or the `Dockerfile` should be made to pin a version. Depends on what it is being used for.

halkeye: I would however update the one in Jenkinsfile (but that looks like its already 2.138.4, so it doesn't matter)

dwnusbaum: Ah good point, [this line](https://github.com/jenkinsci/blueocean-plugin/blob/d0f12cb922e608105a190ebd4126ff0ef4f3b9c7/Jenkinsfile#L43) needs to be deleted/or modified since we wouldn't be able to test on a baseline older than 2.138.4 anymore.

halkeye: Go ahead.

I did learn recently when i cleaned up the jenkinsfile, the weekly ath isn't run anymore anyways so no fears of breaking things

halkeye:
Hrm, want to label this correctly, is it a bug or enhancement? I'm thinking bug fix more than new feature

dwnusbaum:
@halkeye I guess it's somewhat on the border between them, but from user's perspective it probably makes the most sense to treat it as a bug.

halkeye:
I wonder what release drafter will do if its labeled as both :)

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/GraphBuilderTest.java
```diff
@@ -276,6 +277,39 @@ public void parallelStagesNonNested() throws Exception {
...
+    }
+
+    @Test
+    @Issue("JENKINS-43292")
+    public void parallelFailFastDeclarative() throws Exception {
+        WorkflowRun run = createAndRunJob("parallelFailFastDeclarative", "parallelFailFastDeclarative.jenkinsfile", Result.FAILURE);
+        NodeGraphBuilder graph = NodeGraphBuilder.NodeGraphBuilderFactory.getInstance(run);
+        List<FlowNodeWrapper> nodes = graph.getPipelineNodes();
+
+        // top should be BlueRunResult.FAILURE, but the status is currently computed from the beginning of top to the
```
dwnusbaum: This is because `StandardChunkVisitor` doesn't handled nested stages. See further discussion and an ignored test here: https://github.com/jenkinsci/pipeline-graph-analysis-plugin/pull/25/files#r277844096.

pom.xml
dwnusbaum: Done in d5861ae.

dwnusbaum:
Closing and reopening to see if those ATH failures were flakes or not.

halkeye:
Restart stage is still slightly flakey, but the other one is a new error

dwnusbaum:
Tried reproducing the `PipelineNodeTest.nodeWithPartialParallels` failure but wasn't able to. I'll try to track down the source of the ATH tests, perhaps they are looking for a specific result or something that has intentionally changed.

dwnusbaum:
All downstream changes have been released. Strictly speaking, the only necessary non-test change in this plugin is the one in `NodeRunStatus.java` that will change the per-step status displayed to support `WarningAction`. The change in `PipelineNodeGraphVisitor.java` described in the PR description could be backed out, but the change makes the usage of `StatusAndTiming.computeChunkStatus2` here much easier to test elsewhere and more consistent with no drawbacks as far as I can tell.

halkeye:
When I get a few min I'll poke at the test and see if I can figure out what's wrong

dwnusbaum:
Ok, so from the last 3 builds, which don't have any interesting changes between them, it looks like there are some test flakes.

In build 8, the following 2 tests failed:

* AllTest.javaTest
* RestartStageTest.restartStage

In build 9, 4 tests failed:

* AllTest.javaTest
* FavoritesCardsTest.testFreestyle
* PipelineNodeTest.nodesWithPartialParallels
* RestartStageTest.restartStage

In Build 10, 2 tests failed:

* AllTest.javaTest
* FavoritesCardsTest.testFreestyle

The AllTest.javaTest failure is consistent, but I don't understand what that test is doing. It looks like it just reruns the integration test using `mvn test`, but as part of the ATH? Or maybe that test is just a wrapper for all of the other tests, so it fails if _any_ ATH test fails?

Based on the [screenshot](https://ci.blueocean.io/job/blueocean/job/PR-1954/10/artifact/target/screenshots/io.blueocean.ath.offline.personalization.FavoritesCardsTest_testFreestyle.png) for `FavoritesCardTest`, it looks like maybe the build completed before the status was detected as running, so the test timed out because the build was already done, but I'm not sure.

halkeye:
yea, All test is just a wrapper to run all the other tests. So it always
fails if the others fail.
Restart stage and fav's have been flaking for a while, probably a 90%
success rate
PipelineNodeTest.nodesWithPartialParallels is new to me

> Based on the screenshot for FavoritesCardTest, it looks like maybe the
build completed before the status was detected as running, so the test
timed out because the build was already done, but I'm not sure.

Yep thats about right, we've cleaned it up a lot, but there's still a
couple races we've missed.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
```diff
@@ -242,9 +242,9 @@ protected void handleChunkDone(@Nonnull MemoryFlowChunk chunk) {
             if (parallelNestedStages && chunk.getFirstNode().isActive()) {
                 status = new NodeRunStatus(chunk.getFirstNode());
             } else {
-                FlowNode nodeBefore = chunk.getNodeBefore(), lastNode = chunk.getLastNode(), nodeAfter = chunk.getNodeAfter();
+                FlowNode nodeBefore = chunk.getNodeBefore(), firstNode = chunk.getFirstNode(), lastNode = chunk.getLastNode(), nodeAfter = chunk.getNodeAfter();
                 status = new NodeRunStatus(
-                    StatusAndTiming.computeChunkStatus2(run, nodeBefore, firstExecuted, lastNode, nodeAfter));
+                    StatusAndTiming.computeChunkStatus2(run, nodeBefore, firstNode, lastNode, nodeAfter));
```
dwnusbaum: I thought this change wasn't strictly necessary, but it turns out that it is, and it fixes things like [JENKINS-57579](https://issues.jenkins-ci.org/browse/JENKINS-57579). I can't quite tell what `firstExecuted` is expected to be, and it seems like it will vary based on context, but `firstNode` is _always_ (even for in-progress builds), going to be the start of a stage step block or parallel branch block, and that is what we expect over in `StatusAndTiming.computeChunkStatus2` anyway.

My guess as to the problem is that in some cases `firstExecuted` ends up being identical to or later than `lastNode` in the graph, and so setting up the graph traversal [here in StatusAndTiming.findWarningsBetween](https://github.com/jenkinsci/pipeline-graph-analysis-plugin/blob/a34a23633dc8ebd4c74012498335653b69b45206/src/main/java/org/jenkinsci/plugins/workflow/pipelinegraphanalysis/StatusAndTiming.java#L326) fails and the method returns null.

dwnusbaum:
@halkeye Anything in particular you want me to look into here? It looks like we are just seeing a bunch of different flaky test failures. I looked into the `PipelineNodeTest.nodesWithPartialParallels` failure since you hadn't seen it before, but I haven't been able to reproduce it, and looking at the test I don't really see how my changes would make it flaky unless it was already flaky.

halkeye:
nope, not right now, I've been poking at trying to fix them (#1962) but they are mostly timing based e2e tests that are failing.

I've moved projects so I only have personal time to commit to this project now.

vivek:
Failing test in latest build#12 has been flakily failing on master as well across multiple build: https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/PR-1954/12/tests/. I can't get ATH to run locally to reproduce to see why this test is flaky. In any case, fixing this flaky test is out of scope of this PR.

Given wider interest in this PR, I propose we merge it and address flaky ATH tests separately. @michaelneale  @halkeye @kshultzCB

michaelneale:
looks like a good improvement - not sure of impact of dependency changes (and the new pipeline dependency - but likely it is fine). Looks like an improvement of things anyway.

michaelneale:
@dwnusbaum so mostly flakes that fail that you see?

dwnusbaum:
> and the new pipeline dependency

pipeline-model-extensions was already a dependency through pipeline-model-definition, it just wasn't declared explicitly in pom.xml.

> so mostly flakes that fail that you see?

Yes, as far as I can tell the ATH failures are not consistent. The one non-ATH failure that seems a bit sketchy is `PipelineNodeTest.nodesWithPartialParallels`, but I haven't been able to reproduce that failure locally. I recreated the test in a live instance based on this PR, and still wasn't able to reproduce it. 

I went ahead and pushed 629d57f, which changes that test from guessing how long to wait using `Thread.sleep` to using the `semaphore` step which was designed exactly for this purpose for testing. Now, we know that when we make the other calls in the test the Pipeline must have progressed to the point where there are definitely 3 steps in the left branch and 2 steps in the right branch. If we don't get there, then the test will time out, showing us that something has gone wrong earlier. (If we actually cared about the inputs in the test or the visualization, changing them wouldn't work, but as it is they are just used to stop the execution for the test)

dwnusbaum:
Rebuilding due to infra failure:

> Could not transfer artifact org.jenkins-ci.plugins:pipeline-graph-analysis:jar:1.10 from/to repo.jenkins-ci.org (http://repo.jenkins-ci.org/public/): Failed to transfer file: http://repo.jenkins-ci.org/public/org/jenkins-ci/plugins/pipeline-graph-analysis/1.10/pipeline-graph-analysis-1.10.jar. Return code is: 502 , ReasonPhrase:Bad Gateway.

dwnusbaum:
Latest test failure is `PipelineNodeTest.sequentialParallelStagesLongRun`. Not sure what is wrong with that one, but I haven't been able to reproduce it locally, and since it doesn't have anything to do with statuses I don't think it's related to my change. Best guess is some unusual race condition being hit in [this loop](https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-pipeline-api-impl/src/test/resources/io/jenkins/blueocean/rest/impl/pipeline/sequential_parallel_stages_long_run_time.jenkinsfile) that runs every 100ms. Closing and reopening for a fresh build.

michaelneale:
woo hoo green !

rodrigc:
@dwnusbaum nice job, and nice blog post: https://jenkins.io/blog/2019/07/05/jenkins-pipeline-stage-result-visualization-improvements/

PR #1959, opened 2019-05-09
JENKINS-57386 - Use Date instead of timers for duration tracking

sanniassin:
# Description

This PR keeps duration tracking accurate even if the browser tab is inactive.

See [JENKINS-57386](https://issues.jenkins-ci.org/browse/JENKINS-57386).

Since it only affects accuracy and doesn't change any behavior, I didn't add tests.


halkeye:
I can't seem to get the tests to pass, so I'm thinking something changed (possibly the test is depending on broken behavior)

I have rerun a few times. I havnt had time to dig into it which is why no review yet

sanniassin:
I'm not sure if failure is related to this PR since the `Jenkins Design Language` is fine and it's the `Common API for Blue Ocean` stage that fails. It also seems that other PRs, like #1960, are also affected by the same issue.

halkeye:
Agreed, I just wanted to sorta explain why its not merged. I've spent a lot of time cleaning up the tests, so don't want to merge something that potentially makes them worse without looking into it.

So once I get a chance, i'll dig into it, then merge

halkeye:
can you merge master to fix the ci.jenkins build issue?

PR #1960, opened 2019-05-14
[JENKINS-57473] Switch confusing second x to up arrow

jsoref:
# Description

See [JENKINS-57473](https://issues.jenkins-ci.org/browse/JENKINS-57473).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


halkeye:
Can you add a screenshot/image of the new behavior to the ticket?

Otherwise looks good

jsoref:
I did a bit more testing and worked to make the arrow more in line w/ the dimensions/placement of the x -- and, oops, apparently I had the wrong color, so I've added `white`.

halkeye:
blueocean build is failing due to javascript errors.

React/js uses camel case for html properties (so strokeWidth not stroke-width) cause the svg files are being processed by react :

```
[INFO] /home/jenkins/agent/workspace/blueocean_PR-1960/blueocean-core-js/src/js/components/ResultPageHeader.jsx

[INFO]   11:23  error  Unknown property 'stroke-width' found, use 'strokeWidth' instead  react/no-unknown-property
```

jsoref:
Grr. I hit that, fixed it, played w/ it, and then copied stuff out of the DOM into the file and forgot to refix it. Thanks.

halkeye:
can you merge master to fix the ci.jenkins build issue?

jsoref:
Sure. Probably tomorrow

halkeye:
great! I'll merge as soon as joe gets a chance to comment on the ticket

josephbrueggen:
Hello @jsoref, thank you for this contribution!

I strongly agree that utilizing the same 'x' icon to represent two different things is misleading and should be fixed. However, I feel that the proposed replacement icon may not be the right path forward. 

While I can appreciate the conceptual value of moving _up_ a stage when closing the pipeline view, I feel that we should avoid changing the standard convention of using an 'x' to indicate a 'close' functionality.

I would propose that we keep the 'close' icon as-is, and instead replace the failed build indicator with a slightly modified icon. I'd be happy to mock this up and provide the replacement icon. I'll add a comment here with the asset early next week.

jsoref:
Sure. It was a toss up for me as to which to replace. I picked the up arrow because it was an icon I could reasonably identify. I couldn't think of a reasonable replacement for the build failure.

jennbriden:
I agree with @josephbrueggen. @jsoref, thank you for helping here! It's great to see community members participating in Blue Ocean. 💃

josephbrueggen:
@jsoref I think the best solution here is to replace the build failure icon with the same one used on the _Activity_ tab. This reinforces visual consistency and manages to distinguish the 'x' icons somewhat without breaking the 'close' icon convention. Here's the solution mocked up for mobile and the associated icon asset in SVG format:
 
Mobile Visualization

[Build_Failure_Icon.zip](https://github.com/jenkinsci/blueocean-plugin/files/3283666/Build_Failure_Icon.zip)

jsoref:
Ah, yeah, well, err. Yes, that should be done -- that's indeed the thing I would have been looking for but couldn't find.

I'd claim that you haven't really solved the problem. You're still using an `x` in two places. Sure, there's a circle around it, but lower in the page you might have:

image

I think it's going to be necessary to replace the `x` on the right as well.

2/3 contexts use an `x` to mean _fail_, and one to mean _close_, that's still not great, and in one of those _fail_ cases, they use a `v` to mean _close_.

cliffmeyers:
Perhaps replacing the "Close" X icon with a ghost button (similar to Logout) that says either "Close" or "Exit" would be to the most obvious way to disambiguate the two.

josephbrueggen:
@jsoref and @cliffmeyers thanks for the feedback!

I share your concern around having multiple `x` icons throughout the interface. You have a valid point and my proposed solution is not a perfect fix. However, the solution should not be to replace the convention of using `x` for exiting this view. This convention plays an important role in usability, and replacing it with a button similar to `Logout` may lead to confusion between the two.

Concerning the most recent screenshot, the placement of this `x` in the interface disambiguates it from the others we're discussing. I understand that it's less than ideal to have it appear in another place in the UI, but I don't think it's pivotal to the challenge at hand.

Returning to the top bar, there is additional evidence to suggest that my previous proposal is the path forward given the unique constraints at play.
- A different solution to replace the failed build `x` icon in the upper left would break consistency with the UI for other build statuses, such as _succeeded_, which also presents an icon in the upper left. This icon should remain, albeit slightly modified as suggested.
- If a user that is new to Blue Ocean attempts to back out of the build view by selecting the failed build `x` icon in the upper left, nothing will happen. Avoiding this confusion from the outset would definitely be ideal, but there is a low learning curve associated with discovering the proper action. Additionally, experienced users will not make this mistake.

Given the low likelihood that these icons would lead to anything more damaging than momentary confusion, and especially given the unique constraints of the challenge at hand, I recommend that we move forward with my previous suggestion.

Please let me know if there is anything else I can do to help out here, and thank you again!

jsoref:
You could help out by making a PR...

Sorry. I don't have enough anything to invest in your proposal. And I'm at most a volunteer.

PR #1961, opened 2019-05-17
Sequential parallel branches appear incorrectly while build in-progress

sophistifunk:
# Description

See [JENKINS-53816](https://issues.jenkins-ci.org/browse/JENKINS-53816).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


sophistifunk:
Example pipeline that can be used to verify before/after changes: https://gist.github.com/sophistifunk/7b75a86658b80e82715d213820b673c7

sophistifunk:
This doesn't address the shortcomings in the front-end code that are mainly responsible for JENKINS-55605 but some occurences of that problem will likely go away due to this fix.

sophistifunk:
Build 5 was green, but I renamed some stuff so I'll try and get another green before I merge. You know, and a review.

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeTest.java
```diff
@@ -2462,6 +2402,78 @@ public void parallelStagesGroupsAndNestedStages() throws Exception {
...
+                                                  .collect(Collectors.joining(", "));
+
+            assertEquals("running node names", completeNodeNames, runningNodeNames);
+
+            loopCount++;
+
+        } while (run2.isBuilding());
+
+        // Sanity check, make sure we're *actually* checking stuff.
+        assertTrue("Checked multiple times while building", loopCount > 5);
```
olamy: I wonder if maybe we could have lower numbers on very slow/not powerful machines?

sophistifunk: I don't think it's a good idea. What use is a green test that's not really testing anything. I'd like to be able to use some other mechanism that time delays in the pipeline, something deterministic, but I don't know what that could really be. FWIW, the count running on my machine was > 90, so 5 seems pretty low.

PR #1962, opened 2019-05-23
Attempt to fix the flaky FavoritesCardsTest ATH Test

halkeye:
Add waits for favs test instead of depending on sleeps


sophistifunk:
Nice, I've gotta remember this next time I'm trying to test something along those lines.

bb-modern

halkeye:
Yea I think I've peaked and i'll never be this useful again

PR #1977, opened 2019-05-29
update guava from 11.0.1 to 11.0.2

based2:
https://mvnrepository.com/artifact/com.google.guava/guava
https://github.com/google/guava/releases


halkeye:
as with #1976, why are you wanting to upgrade this?

PR #2015, opened 2019-08-15
address error.responseBody is undefined

heisian:
When confirming on the user input step, this error is thrown: `TypeError: error.responseBody is undefinedjenkins-js-extension.js:95465:21`.

This code should handle an error in the case that there is no `responseBody` in the `error` object.

# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/karaoke/components/InputStep.jsx
+++ b/blueocean-dashboard/src/main/js/components/karaoke/components/InputStep.jsx
@@ -88,9 +88,13 @@ export default class InputStep extends Component {
         const parameters = this.parameterService.parametersToSubmitArray();
 
         parameterApi.submitInputParameter(href, id, parameters).catch(error => {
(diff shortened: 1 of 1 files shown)


halkeye:
What is the response coming back? can you see in your network tab?

heisian:
```json
{
  "message" : "Stage 13 not found in pipeline Continuous Deployment.",
  "code" : 404,
  "errors" : [ ]
}
```

heisian:
I think generally if the error output is not normalized across the entire API we shouldn't rely on assuming that a certain property will be there, in this case, `responseBody`, or even `errors`, if it is the case that somewhere upstream we iterate through the `errors` array and that gets passed to this area.

halkeye:
There has been a bad habit in the past of patching and not fixing the actual problem.

In this case I think it's the right solution

heisian:
ty!

PR #2017, opened 2019-08-21
[JENKINS-58085] Fix issue where stages are shown as waiting to start when they are running/completed

dwnusbaum:
# Description

See [JENKINS-58085](https://issues.jenkins-ci.org/browse/JENKINS-58085)/[JENKINS-53816](https://issues.jenkins-ci.org/browse/JENKINS-53816) and some tickets linked from those.

This PR tweaks the fix for JENKINS-53816 added in #1961.

In some cases, when a build's flow graph changes from one build to the next (because the Pipeline was changed, or perhaps due to the result of a `when` condition or some dynamic Pipeline behavior in a Scripted Pipeline), but stays similar enough that Blue Ocean considers the overall graph structure the same, the node ID for one of the nodes in the current Pipeline would use the node ID for what Blue Ocean thought the corresponding node was in previous builds. If the node ID for that node was not actually the same in both builds, the new node would be shown as having never started, and would not have any steps associated with it (because the API calls for that node ID would be for some other node entirely, perhaps one that wasn't actually a stage, so it had no children). Once the build completed, the node ID would be replaced with the actual node ID from the build, and things would be displayed correctly.

I'm not sure if this covers all of the issues users describe in JENKINS-58085/JENKINS-53816, but it fixes at least one problem I found. Just opening a draft PR for now while I try to come up with a regression test for the bug.

<details>
<summary>Here is the Pipeline I was using for testing</summary>

```
pipeline {
    agent none
    stages {
        stage("Before Parallel") {
            steps {
                echo "starting"
                sleep 15
                echo "done"
            }
        }
        stage("Parallel") {
            parallel {
                stage("Par 1") {
                    stages {
                        stage("Par 1 - Seq 1") {
                            steps {
                                echo "starting"
                                sleep 5
                                echo "done"
                            }
                        }
                        stage("Par 1 - Seq 2") {
                            steps {
                                echo "starting"
                                sleep 10
                                echo "done"
                            }
                        }
                    }
                }
                stage("Par 2") {
                    // Comment and uncomment this `when` from one build to the next to see the bug I am trying to fix.
                    when {
                        expression { true == true }
                    }
                    stages {
                        stage("Par 2 - Seq 1") {
                            steps {
                                echo "starting"
                                sleep 10
                                echo "done"
                            }
                        }
                        stage("Par 2 - Seq 2") {
                            steps {
                                echo "starting"
                                sleep 10
                                echo "done"
                            }
                        }
                    }
                }
                stage("Par 3") {
                    steps {
                        echo "starting"
                        sleep 5
                        echo "done"
                    }
                }
            }
        }
        stage("After Parallel") {
            steps {
                echo "starting"
                sleep 15
                echo "done"
            }
        }
    }
}
```
</details>

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
@@ -687,7 +687,8 @@ public class PipelineNodeGraphVisitor extends StandardChunkVisitor implements No
 
         boolean graphsAreCompatible = true;
 
-        Map<String, FlowNodeWrapper> newNodes = new HashMap<>(previousNodes.size()); // indexed by id
+        // A map from the ID of a node from previousNodes to the corresponding node for the current build.
+        Map<String, FlowNodeWrapper> newNodes = new HashMap<>(previousNodes.size());
 
         // Start with the currently-executing nodes
         for (FlowNodeWrapper currentNodeWrapper : nodes) {
@@ -710,14 +711,14 @@ public class PipelineNodeGraphVisitor extends StandardChunkVisitor implements No
 
                 // New wrapper object based on current execution
                 FlowNodeWrapper newNodeWrapper = new FlowNodeWrapper(
-                    oldNodeWrapper.getNode(), // Use old node because we want to show the final id not intermediary
+                    currentNodeWrapper.getNode(),
                     currentNodeWrapper.getStatus(),
                     currentNodeWrapper.getTiming(),
                     currentNodeWrapper.getInputStep(),
                     run
                 );
 
-                newNodes.put(nodeId, newNodeWrapper);
+                newNodes.put(oldNodeWrapper.getId(), newNodeWrapper);
 
             } else {
                 // Node does not exist in previous graph, user probably changed pipleine definition.
@@ -729,9 +730,9 @@ public class PipelineNodeGraphVisitor extends StandardChunkVisitor implements No
         // Walk the old graph, create new wrappers for any stages not yet started
         if (graphsAreCompatible) {
             for (FlowNodeWrapper oldNodeWrapper : previousNodes) {
-                final String nodeId = oldNodeWrapper.getId();
+                final String oldNodeId = oldNodeWrapper.getId();
 
-                if (!newNodes.containsKey(nodeId)) {
+                if (!newNodes.containsKey(oldNodeId)) {
                     FlowNodeWrapper newNodeWrapper = new FlowNodeWrapper(
                         oldNodeWrapper.getNode(),
                         new NodeRunStatus(null, null),
@@ -740,7 +741,7 @@ public class PipelineNodeGraphVisitor extends StandardChunkVisitor implements No
                         run
                     );
 
-                    newNodes.put(nodeId, newNodeWrapper);
+                    newNodes.put(oldNodeId, newNodeWrapper);
                 }
             }
         }
@@ -748,9 +749,9 @@ public class PipelineNodeGraphVisitor extends StandardChunkVisitor implements No
         // Re-create edges and parentage based on previous run
         if (graphsAreCompatible) {
             for (FlowNodeWrapper oldNodeWrapper : previousNodes) {
-                final String nodeId = oldNodeWrapper.getId();
+                final String oldNodeId = oldNodeWrapper.getId();
 
-                FlowNodeWrapper newNodeWrapper = newNodes.get(nodeId);
+                FlowNodeWrapper newNodeWrapper = newNodes.get(oldNodeId);
 
                 for (FlowNodeWrapper oldEdge : oldNodeWrapper.edges) {
                     FlowNodeWrapper newEdge = newNodes.get(oldEdge.getId());
@@ -773,8 +774,8 @@ public class PipelineNodeGraphVisitor extends StandardChunkVisitor implements No
             // code that makes too many assumptions
 
             for (FlowNodeWrapper oldNodeWrapper : previousNodes) {
-                final String nodeId = oldNodeWrapper.getId();
-                FlowNodeWrapper newNodeWrapper = newNodes.get(nodeId);
+                final String oldNodeId = oldNodeWrapper.getId();
+                FlowNodeWrapper newNodeWrapper = newNodes.get(oldNodeId);
                 newList.add(new PipelineNodeImpl(newNodeWrapper, () -> parent, run));
             }
 
--- a/blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/GraphBuilderTest.java
+++ b/blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/GraphBuilderTest.java
@@ -4,11 +4,13 @@ import com.google.common.base.Charsets;
 import com.google.common.io.Resources;
 import hudson.model.Result;
 import io.jenkins.blueocean.listeners.NodeDownstreamBuildAction;
+import io.jenkins.blueocean.rest.hal.Link;
 import io.jenkins.blueocean.rest.model.BlueRun;
 import org.apache.commons.lang3.StringUtils;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.job.WorkflowJob;
 import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.jenkinsci.plugins.workflow.test.steps.SemaphoreStep;
 import org.junit.Assert;
 import org.junit.Test;
 import org.jvnet.hudson.test.Issue;
@@ -16,6 +18,7 @@ import org.jvnet.hudson.test.Issue;
 import java.net.URL;
 import java.util.Collection;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import static org.junit.Assert.assertEquals;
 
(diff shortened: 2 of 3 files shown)


blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
```diff
@@ -710,14 +711,14 @@ public BluePipelineStep getPipelineNodeStep(String id, Link parent) {
 
                 // New wrapper object based on current execution
                 FlowNodeWrapper newNodeWrapper = new FlowNodeWrapper(
-                    oldNodeWrapper.getNode(), // Use old node because we want to show the final id not intermediary
+                    currentNodeWrapper.getNode(),
```
dwnusbaum: This is the main fix. The old comment doesn't make sense to me. My understanding is that in this case, we have a real node from the current Pipeline. Its ID will never change, even when the build completes, so we already have the "final" ID. Because the code checks for a match using `findNodeWrapperByParentageIn` if it doesn't find a direct ID match, the ID of the real node might not match the ID of the corresponding node from the old build. This leads to a bug when we use that ID in reference to the current build in other contexts where we expect it to be a stage/branch or whatever, since that ID might be something totally different in the current build.

dwnusbaum: Maybe we should just delete `newNodeWrapper` and use `currentNodeWrapper` directly now that we aren't changing the fields?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
```diff
@@ -687,7 +687,8 @@ public BluePipelineStep getPipelineNodeStep(String id, Link parent) {
 
         boolean graphsAreCompatible = true;
 
-        Map<String, FlowNodeWrapper> newNodes = new HashMap<>(previousNodes.size()); // indexed by id
+        // A map from the ID of a node from previousNodes to the corresponding node for the current build.
+        Map<String, FlowNodeWrapper> newNodes = new HashMap<>(previousNodes.size());
```
dwnusbaum: I thought it would be useful to clarify that the IDs here come from the old build, not the new build, and I updated the names of the variables that get put into the map to try to make that clearer as well.

halkeye:
AFAIK, Josh has worked the most on this space, so added him as a reviewer to get feedback.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
sophistifunk: The root problem is the node ids *do* change during a run, thanks to the way this code walks the graph backwards instead of forwards.

sophistifunk: Changing this will fix some of the bad graphs, and break others that are currently working. I used the old node because it's from a *finished* run, which has the "final" node ids.

dwnusbaum: > The root problem is the node ids do change during a run, thanks to the way this code walks the graph backwards instead of forwards.

The ID for an actual Pipeline `FlowNode` in a build will never change once it is added to the flow graph. Whether we walk the graph in reverse or not does not matter. In this particular loop, as far as I can tell, we are looping through `FlowNodeWrappers`s for `FlowNode`s that already exist in the flow graph of the current build, so their IDs will never change.

If you mean the problem is that a single `FlowNodeWrapper` that reference nodes from the current build can end up referencing a different node once the build progresses further because we are mutating `FlowNodeWrapper` or something is unstable in the visitor, that seems like a bug, and I am happy to look into fixing that here too if we can come up with a reproduction case.

> I used the old node because it's from a finished run, which has the "final" node ids.

The problem is that expecting the node IDs from a different build to match the node IDs in the current build is absolutely not something you can count on. It happens to work in some cases, but `findNodeByParentageIn` is not strict enough to guarantee that if two nodes match by that criteria their IDs also match, and nothing in a Pipeline will change the ID of a `FlowNode` once it has been added to the flow graph to make them match after the fact or something.

Further on in this code, where we use the IDs from the old build in what we return, is also a bit suspect, because one of the old IDs might refer to a different node in the current Pipeline, but since those nodes are marked as having not started it seems to not really matter. It would be safer to make the node IDs for the fake `FlowNodeWrapper`s some ID that could not possibly be a valid ID in the current Pipeline, but I don't know what assumptions the rest of the code makes about the IDs for the fake `FlowNodeWrapper`s.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
```diff
@@ -710,14 +711,14 @@ public BluePipelineStep getPipelineNodeStep(String id, Link parent) {
...
-                    oldNodeWrapper.getNode(), // Use old node because we want to show the final id not intermediary
+                    currentNodeWrapper.getNode(),
                     currentNodeWrapper.getStatus(),
                     currentNodeWrapper.getTiming(),
                     currentNodeWrapper.getInputStep(),
                     run
                 );
 
-                newNodes.put(nodeId, newNodeWrapper);
+                newNodes.put(oldNodeWrapper.getId(), newNodeWrapper);
```
dwnusbaum: Ok, so with some more testing, it turns out that this is actually the only change needed to fix the test case I am working with. The problem is that `nodeId` is an ID from the current build, but all other `get`s/`put`s on `newNodes` use an  ID from the old build, so when the IDs don't match across builds the logic in the rest this method can get mixed up. Regardless of the discussion in https://github.com/jenkinsci/blueocean-plugin/pull/2017#discussion_r316394955, I think we should use the ID from the old node here so that all of the keys in `newNodes` are IDs from the old build.

blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/GraphBuilderTest.java
```diff
@@ -409,6 +412,53 @@ public void unstableSmokes() throws Exception {
...
+        SemaphoreStep.success("third/1", null);
+        j.waitForCompletion(b1);
+        j.assertBuildStatus(Result.SUCCESS, b1);
+        // Run the second build. Its graph as computed by PipelineNodeGraphVisitor will be created by combining the current build with the previous build.
+        WorkflowRun b2 = p.scheduleBuild2(0).waitForStart();
+        SemaphoreStep.waitForStart("second/2", b2);
+        {
+            List<FlowNodeWrapper> nodes = unionPipelineNodes(b1, b2);
+            assertStageAndEdges(nodes, "first", BlueRun.BlueRunState.FINISHED, BlueRun.BlueRunResult.SUCCESS, "second");
+            assertStageAndEdges(nodes, "second", BlueRun.BlueRunState.RUNNING, BlueRun.BlueRunResult.UNKNOWN, "third");
```
dwnusbaum: If you revert the changes in `src/main`, all assertions expecting `BlueRunState.RUNNING` and `BlueRunResult.UNKNOWN` fail because they are `null` instead.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
sophistifunk: Yes I know the underlying canonical graph doesn't change, but due to the `PipelineNodeGraphVisitor` working backwards (and relies on events around structure rather than just the nodes themselves), it sometimes picks a different set of nodes to represent the same graph depending on where exactly the job is during a build. This is the underlying issue causing all of these problems, because the UI needs to be able to identify a node by id, both for navigation and selection, and (I believe) for updating the model during builds when SSE events come in. 

The drift of node ids causes the difficulty in merging the old and new graphs on the server-side, and also causes the UI state to not get correctly updated all the time, as well as user events not being tied to ids when the model held in the front-end happens to have the transiently-bad ids, the user clicks something, and then that id is no longer attached to that node by the time of the round-trip to the server looking for more info.

sophistifunk: I'm not trying to prescribe solutions, just trying to dump out my knowledge before it all bitrots away in my head.

dwnusbaum: > it sometimes picks a different set of nodes to represent the same graph depending on where exactly the job is during a build

Ok, do you have any particular scenarios in mind that might cause this, so I can investigate them and hopefully at least add a test case for the future? 

One specific case with weird behavior that I am aware of is that if we have a sequential stage, the status computation for the `FlowNodeWrapper` for that stage will use the start of the sequential stage as the start (correct), and the start of the first stage inside of the sequential stage as the end (it should probably be the end of the sequential stage like for regular stages, but `StageChunkFinder` doesn't work for sequential stages). I don't think that would affect the stability of the listener the way you are describing (and it doesn't really matter because we don't display a status bubble for sequential stages), but it falls under the "weird behavior with PipelineNodeGraphVisitor" umbrella.

dwnusbaum:
Test failures on ci.blueocean.io are `PipelineNodesTest. sequentialParallelStagesLongRun` and `GithubEditorTest.testEditor`/`GithubEditorTest.testEditorSetEnvironmentVariables`. The latter two seem unrelated. The former could be related to the issues discussed in https://github.com/jenkinsci/blueocean-plugin/pull/2017#discussion_r316394955, so I'll look into it, but that test was failing before these changes as well, see https://github.com/jenkinsci/blueocean-plugin/pull/1954#issuecomment-499536518.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
dwnusbaum: Ok, I came up with two minimal test cases where nodes that show up while the build is in progress do not appear in the final graph. Both of them involve a `parallel` step with only one branch, so they're not particularly realistic, but they should help to figure out what is going on (and to check if my change affects the behavior or not).

Here is the Declarative test case:

```
pipeline {
  agent none
  stages {
    stage('outer') {
      parallel {
        stage('inner') {
          steps {
            sleep 5
          }
        }
      }
    }
  }
}
```

Here is the scripted test case:

```
stage('outer') {
  parallel inner: {
    stage('inner') {
        sleep 5
    }
  }
}
```

In both cases, `outer` is shown in a separate bubble preceding `inner` until the build completes, at which point only `inner` is shown.

dwnusbaum: Ok, looking into those test cases, the main difference in the graph of `FlowNodeWrapper`s when the build is running vs completed is that when the build is running, the `FlowNodeWrapper.type` of the node for `inner` is `STAGE`, but when the build completes the `FlowNodeWrapper.type` for `inner` becomes `PARALLEL`. I'll keep looking to understand why and either fix it or at add an ignored test case, although I'll probably open a separate PR for that.

sophistifunk: I couldn't reproduce all of the situations that cause node drift reliably, it's definitely timing-sensitive. I think it occurs only when parallel stages are just starting or just ending, such that the nodes exist but not all the events for those nodes do yet. I don't really have more detail to give, other than my conclusion that it'd be quicker building a new graph visitor that uses the newer pipeline API via feature flag rather than trying to debug the exact event sequence that triggers it. Fresh eyes may be able to find it though.

sophistifunk: Another possibility might be adopting a different id scheme, that is tied to the structure of the graph (akin to an xpath or something) rather than just reusing the ids given to the pipeline nodes by Jenkins.

vivek:
@dwnusbaum Lets get @olamy review and then merge and release it. Great job 🎉

graingert:
@olamy do you know when this will be released?

olamy:
@graingert 1.19.0 has been released with this fix.

feliwir:
I'm still getting this after upgrading to 1.19.0

dwnusbaum:
@feliwir Please post comments like that to the corresponding [Jira issue](https://issues.jenkins-ci.org/browse/JENKINS-58085) or the [jenkinsci-users](https://jenkins.io/mailing-lists/) mailing list instead of GitHub. Make sure to include steps to reproduce the problem, such as the Jenkinsfile you are using, otherwise we are unlikely to investigate because we do not have any leads on what the problem might be.

PR #2020, opened 2019-08-28
validate 'run.causes' is set before checking length

kevinushey:
# Description

See [JENKINS-59131](https://issues.jenkins-ci.org/browse/JENKINS-59131).

# Submitter checklist

This is mostly a bandaid (not sure if it should ever be possible for the `run` object to exist without a `causes` member) but similar checks are done in other places; e.g.

https://github.com/kevinushey/blueocean-plugin/blob/9819a3048d01a0f0979cb9d8346b4ad0145dc3c4/blueocean-dashboard/src/main/js/components/RunMessageCell.jsx#L31

so I believe it makes sense to validate that `run.causes` exists in these places as well.

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/RunDetailsHeader.jsx
+++ b/blueocean-dashboard/src/main/js/components/RunDetailsHeader.jsx
@@ -184,7 +184,7 @@ class RunDetailsHeader extends Component {
         );
 
         const cause = run => {
-            const lastCause = (run && run.causes.length > 0 && run.causes[run.causes.length - 1]) || null;
+            const lastCause = (run && run.causes && run.causes.length > 0 && run.causes[run.causes.length - 1]) || null;
             if (lastCause && lastCause.upstreamProject) {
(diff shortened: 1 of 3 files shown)


halkeye:
Awesome. I think null cause is valid (though a little weird to me), the api is returning it.

Would you be up for adding a test to
https://github.com/jenkinsci/blueocean-plugin/blob/77960a8dc21840ae5f6df784a26f91e09d7da621/blueocean-dashboard/src/test/js/run-message-cell-spec.js

kevinushey:
I gave it a shot! Let me know if that looks sufficient.

halkeye:
I think the ath failing is unrelated, but your new test fails

> AssertionError: expected '–' to equal '-' at Object.

halkeye:
@vivek @dwnusbaum @olamy any objections to me just merging this. It can mask and underlying problem of why the api is returning null, but I rather it just work for people.

kevinushey:
@halkeye thank you for taking the time to review and help guide me through the PR; I appreciate it!

PR #2029, opened 2019-10-08
[JENKINS-59706] PipelineRunImpl.getCommitUrl must not loop over historical builds

jglick:
[JENKINS-59706](https://issues.jenkins-ci.org/browse/JENKINS-59706): #1212 introduced what seems to be a very dangerous loop (without any test or comment explaining why such a loop might be necessary), and #1947 does not seem to have sufficed to ameliorate its effects.

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
@@ -4,8 +4,6 @@ import hudson.Extension;
 import hudson.model.Queue;
 import hudson.model.Run;
 import hudson.model.queue.CauseOfBlockage;
-import hudson.scm.ChangeLogSet;
-import hudson.scm.ChangeLogSet.Entry;
 import io.jenkins.blueocean.commons.ServiceException;
 import io.jenkins.blueocean.rest.Navigable;
 import io.jenkins.blueocean.rest.Reachable;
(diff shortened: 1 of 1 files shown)


halkeye:
This looks good to me, but my only real concern is that I *think* the code was trying to find replays/reruns and show the changelog information based on the original build

Is https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ChangeSetContainerImpl.java#L54-L64 a better solution? or does it suffer from the same problem?

jglick:
https://github.com/jenkinsci/blueocean-plugin/blob/f70deb54fb406a6c6ee21e48e8f086063fab0efb/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/ChangeSetContainerImpl.java#L54-L64 that is. (Press the **Y** key to get a permlink, then paste raw to get automatic preview.)

jglick:
That loop is reasonable—while it does not have a numeric bound, it is limited to actual uses of **Replay**, which are typically at most a handful (and in most cases zero).

jglick:
At any rate, showing less than complete changelog information for replayed builds is surely better than crashing the server…

michaelneale:
Seems sound. Better to not crash for sure. Well spotted.

kshultzCB:
I've done some testing of this versus 1.19. Things seem pretty much the same, with one notable exception. This "Restart Build" button near the bottom right of the screenshot:

image

occasionally, but not always, takes me to a 404 when I'm running master:

image

But it always works fine on 1.19.

Could be a coincidence. I've seen the 404 page three times, all on master and never on 1.19. Once the 404 stops showing up, it never comes back.

jglick:
I _doubt_ that is related, but someone with more familiarity with the codebase should check it. What this PR _may_ have caused is for a replayed build to not display the commit link from the original.

kshultzCB:
The commit does show up on replayed builds. It's not clickable anymore, though.

PR #2030, opened 2019-10-10
Bitbucket server problems with @ in username

shysteph:
# Description

The server user data api expects @ to be turned into _ in a username.  If you try to connect to a bitbucket server repository through blueocean with a username containing @ like an email address the connect button doesn't actually change the UI at all but it does log an exception that it got the wrong response from the api call for BitbucketServerApi.getUser(). The Cloud project has a function that encodes paths in more places but it uses url encoding which server doesn't like.  There are likely more places where the path needs to be encoded but this is what I needed.
I'm fine with using a locally built patch but I figured I would share.  Sorry for the low quality PR but work isn't paying me to write code for jenkins.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerApi.java
+++ b/blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerApi.java
@@ -101,7 +101,7 @@ public class BitbucketServerApi extends BitbucketApi {
     @Override
     public @Nonnull BbServerUser getUser(@Nonnull String userName){
         try {
-            InputStream inputStream = request.get(String.format("%s/%s",baseUrl+"users", userName)).getContent();
+            InputStream inputStream = request.get(String.format("%s/%s",baseUrl+"users", userName.replace('@','_')).getContent();
             return om.readValue(inputStream, BbServerUser.class);
         } catch (IOException e) {
             throw new ServiceException.UnexpectedErrorException(e.getMessage(), e);
--- a/blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerApi.java
+++ b/blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerApi.java
@@ -101,7 +101,7 @@ public class BitbucketServerApi extends BitbucketApi {
     @Override
     public @Nonnull BbServerUser getUser(@Nonnull String userName){
         try {
-            InputStream inputStream = request.get(String.format("%s/%s",baseUrl+"users", userName.replace('@','_')).getContent();
+            InputStream inputStream = request.get(String.format("%s/%s",baseUrl+"users", userName.replace('@','_'))).getContent();
             return om.readValue(inputStream, BbServerUser.class);
         } catch (IOException e) {
             throw new ServiceException.UnexpectedErrorException(e.getMessage(), e);


halkeye:
[2019-10-10T18:26:57.304Z] /home/jenkins/workspace/Plugins_blueocean-plugin_PR-2030/blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/server/BitbucketServerApi.java:[104,129] ')' expected

@shysteph so your missing a )

Do you know how to build this locally to try it?

vivek:
@shysteph, thanks for submitting the PR. However, I am not sure I follow. So you create a username in BB server with `@` in it and use the same username to login in to it and it works. However BB server rest API fails to work with that username with `@` in it? Which version of BB server you are using? 

I think to properly fix this, we should understand whats going on first and then apply a proper fix that handles properly encoded username when calling their API.

shysteph:
I am using BB Server v6.7.0.  Yes, so for example we use email addresses as username so jenkins is jenkins@mywork.com .  I can login to BB just fine and use the web interface as much as I want.  If I try to set up a jenkins project using BB and use the same credentials it fails. When I started inspecting the network traffic I saw that anywhere it would have used the username talking to BB server the @ was included.  If you look at the matching cloud implementation BitbucketCloudApi it always encodes variables being added to the path with encodePath(String).  BB Server uses a different encoding scheme but I suspect the same sort of implementation is the correct way to handle building paths is what is needed.
If I browse around BB using the web interface I can find places where it encodes the username on the path.  For example clicking the profile link in the commit list for a developer with a username of adev@mywork.com you get
"https://scm.mywork.com/users/adev_mywork.com" . A test user I created without an @ in their username has the profile url of "https://scm.mywork.com/users/testuser" which is why this works for most people.
The same is true for rest endpoints such as the one that I "fixed" in the PR: https://scm.mywork.com/rest/api/1.0/users/<username>
With a _ or no @ in the name it works fine (assuming you are authenticated)
https://scm.mywork.com/rest/api/1.0/users/testuser
https://scm.mywork.com/rest/api/1.0/users/jenkins_mywork.com
but the url generated in the BB server api code is
https://scm.mywork.com/rest/api/1.0/users/jenkins@mywork.com
This doesn't work and returns a does not exist type error
`
curl -u 'jenkins@mywork.com' https://scm.mywork.com/rest/api/1.0/users/jenkins@mywork.com
{"errors":[{"context":null,"message":"User jenkins@mywork.com does not exist.","exceptionName":"com.atlassian.bitbucket.user.NoSuchUserException"}]}
`
vs
`
curl -u 'jenkins@mywork.com' https://scm.mywork.com/rest/api/1.0/users/jenkins_mywork.com
{"name":"jenkins@mywork.com","emailAddress":"jenkins@mywork.com","id":1234,"displayName":"Jenkins CI","active":true,"slug":"jenkins_mywork.com","type":"NORMAL","links":{"self":[{"href":"https://scm.mywork.com/users/jenkins_mywork.com"}]}}
`

shysteph:
@halkeye I must have somehow made a typo between building it and actually committing it.  I have it built and running locally.  I'll push an update shortly.

shysteph:
@halkeye @vivek So the real problem is that the api is treating the username and slug as equivalent. 

https://developer.atlassian.com/server/bitbucket/tutorials-and-examples/decorating-the-user-profile/#locate-the-applicationuser 
They note that the slug is some different value that should be used in preference to the username. Reading the source for BB there is no real guarantee for a pattern to generate slugs.  This makes sense since a user may be deleted and a new user later generated with the same name.  This lets them treat the two differently.

https://docs.atlassian.com/bitbucket-server/rest/6.7.0/bitbucket-rest.html#idp392
The best way I can see to get the slug is to make an authenticated GET to /rest/api/1.0/users?filter=<username> and find the result that has a username exactly matching the one you are looking for and use the slug field from there for anything that requires the user slug.  Also update the BBApi classes to better make a distinction between a username and a slug.  Almost everywhere that username is used slug is actually needed and often the slug is referred to as the username.  See for example the method definition for BitbucketApi.getUser vs BitbucketServerApi.getUser.

I'll try to dig deeper in my free time.  It turns out my "fix" just fixed it for the users I was testing anyway and would not be a generic solution.

vivek:
@shysteph Thanks for providing these details, this definitely sounds like a reasonable approach. I tried with BB server `http://localhost:7990/bitbucket/rest/api/1.0/users/\?filter\=vivek@a.b` and it gives only those users who have value matching `vivek@a.b` - better than iterating over all users.

Other option is to perform HEAD call to http://localhost:7990/bitbucket/rest/api/1.0/user and use X-AUSERNAME header. This is standard atlassian response header that represents URL safe username thats used as slug. e.g. `X-AUSERNAME: vivek%40a.b`. It can be used and url decode to start using in API call.

If we go for X-AUSERNAME approach, its simpler change, maybe add apiUsername to [BitbucketAPI](https://github.com/jenkinsci/blueocean-plugin/blob/ccfc733caf6aa505fd77a0e4c08c95f9917e1e68/blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/BitbucketApi.java#L32) and define abstract method getApiUsername with default implementation where it returns `credentials.getUsername()` and its overridden in BitbucketServerAPI to make HEAD call and pick slug from X-AUSERNAME.

PR #2041, opened 2019-11-19
Remove dependencies to joda-time

veita:
With Java 8 or higher Joda Time can be replaced with java.time.

# Description

See [JENKINS-60218](https://issues.jenkins-ci.org/browse/JENKINS-60218).

# Submitter checklist

Unclear how to test this. The fix replaces existing Joda Time code with the equivalent `java.time.*` code. Hereby the system's default time zone is used for formatting, as it was the case with Joda Time.

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-jwt/pom.xml
+++ b/blueocean-jwt/pom.xml
@@ -28,10 +28,5 @@
           <groupId>org.jenkins-ci.plugins</groupId>
           <artifactId>mailer</artifactId>
       </dependency>
-
-      <dependency>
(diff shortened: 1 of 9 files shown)


dwnusbaum:
@veita FWIW, as the failing CI builds show, other plugins (notably `git` and `pipeline-model-definition`) pick up this dependency as well, so it's still a transitive dependency of Blue Ocean, and we ended up reverting part of the change in #2046.

PR #2040, opened 2019-11-14
Move aggregator ('blueocean') docs into git repository

MarkEWaite:
# Move aggregator docs into git repository

See [documentation as code blog post](https://jenkins.io/blog/2019/10/21/plugin-docs-on-github/).  No unit tests because this is purely a documentation change.

Manual test: Review the content in the new adoc file and compare it with the [original wiki page](https://wiki.jenkins.io/display/JENKINS/Blue+Ocean+Plugin).

# Submitter checklist
- [x] Change is code complete and matches issue description
- [x] Appropriate unit or acceptance tests or explanation to why this change has no tests
- [x] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below.

# Reviewer checklist
- [ ] Run the changes and verified the change matches the issue description
- [ ] Reviewed the code
- [ ] Verified that the appropriate tests have been written or valid explanation given

PR #2044, opened 2020-01-27
Replace wiki URL with github URL in pom files

kshultzCB:
# Update project.url in pom files for documentation migration project

Changes the `document.url` field in each submodules' pom.xml to point to instead of `https://wiki.jenkins-ci.org/display/JENKINS/Blue+Ocean+Plugin`. This is the same URL as put into the aggregator by @MarkEWaite in [PR2040](https://github.com/jenkinsci/blueocean-plugin/pull/2040). 

# Submitter checklist
- [x] Link to JIRA ticket in description, if appropriate: N/A
- [x] Appropriate unit or acceptance tests or explanation to why this change has no tests: N/A. Strictly a doc link change in pom files.
- [x] Reviewer's manual test instructions provided in PR description: there are none. Strictly a doc link change in pom files.

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-bitbucket-pipeline/pom.xml
+++ b/blueocean-bitbucket-pipeline/pom.xml
@@ -11,7 +11,7 @@
   <name>Bitbucket Pipeline for Blue Ocean</name>
   <artifactId>blueocean-bitbucket-pipeline</artifactId>
   <packaging>hpi</packaging>
-  <url>https://wiki.jenkins-ci.org/display/JENKINS/Blue+Ocean+Plugin</url>
+  <url>https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean/doc/BlueOcean.adoc</url>
 
   <properties>
     <jacoco.haltOnFailure>true</jacoco.haltOnFailure>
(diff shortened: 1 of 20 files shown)


dwnusbaum:
@kshultzCB For something like this, where your PR requires another PR (the file https://github.com/jenkinsci/blueocean-plugin/blob/master/blueocean/doc/BlueOcean.adoc doesn't exist currently), you can start your PR using the other PR as the base, and then add commits on top of it and submit a PR that has the original commit in addition to your commits.

MarkEWaite:
I'm a little surprised this was merged.  Blue Ocean is composed of multiple plugins.  I assumed in  https://github.com/jenkinsci/blueocean-plugin/pull/2040#discussion_r357928811 that each plugin would have its own adoc page which described the role of that specific plugin within the system that is "Blue Ocean".

Is this a temporary state where those documentation URLs for the component plugins will eventually be replaced by separate adoc pages or is this intended to be the final state?

olamy:
As far I can see the current change is doing the same as today (i.e same page for all modules/plugins) so at this stage it's at least better :) as this is now a page we can update. 
Sure that would be better let's see if we can find some people to contribute this documentation.

MarkEWaite:
Thanks for the clarification @olamy !  I had mistakenly assumed that there were already wiki pages for each of the blue ocean plugins.

PR #2058, opened 2020-03-30
fix code that make the broswer stuck

ghost:
# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


udigot:
async/await may not work for all browsers.
let me know if need to change it

bitwiseman:
It is unclear what this change does.

cliffmeyers:
I think the problem this PR is attempting to solve is a possible browser slowdown issue when a very large data structure is passed to the method, since it is walking the entire tree recursively to inject "capabilities" metadata. Splitting large computation tasks up via `setTimeout` call isn't uncommon in JS (aka "green threading") although in modern browsers you can leverage web workers to run this off the UI thread entirely to avoid jank.

Looking at this impl, it seems a little too conservative to me: it introduces multiple 50 and 100 ms pauses while iterating over the keys of a single object, which I imagine has to slow execution time for even small graphs into the "minutes" range. That's probably why unit test timeout for the relevant spec is being raised to over 15m (!)

A better implementation would probably be to
- Add a counter variable that is incremented each time an object is fully walked,
- Check if the counter exceeds a static threshold and if so
  - Reset the counter variable
  - `await` on a `Promise.resolve` with a `setTimeout` value of 0 or 1, which will push execution into the next iteration of the event loop and give the UI thread time to render,

V8 is pretty fast for this kind of logic, so I have to imagine you could do hundreds, probably many thousands of iterations before breaking the logic up via `setTimeout`

ugatenio:
I created this PR since in our job the extension stuck the browser since **sync** while loop is naive approach in single thread. the sleep method will let the event loop do other stuff between each iteration. BTW who is ghost and why the PR changed like he did it? :)

ugatenio:
@cliffmeyers not sure if I understand you suggestion. what are you think about reduce the sleep time to 1 and 10 ms?

bitwiseman:
According to this site most modern browsers support `await/async`.  The exception being IE11.  
https://caniuse.com/#feat=async-functions

A decision will need to be made about 
whether this is acceptable. 

@cliffmeyers @ugatenio  
What is the bug that this fixes?  Please describe what happens from a user perspective.

ugatenio:
> According to this site most modern browsers support `await/async`. The exception being IE11.
> https://caniuse.com/#feat=async-functions
> 
> A decision will need to be made about
> whether this is acceptable.
> 
> @cliffmeyers @ugatenio
> What is the bug that this fixes? Please describe what happens from a user perspective. If there isn't a bug filed in the Jenkins JIRA, we'll need that done as well.

as the title suggests the web freeze for seconds. I ran chrome performance tool and those functions turn red light. 
At my work we really waiting for this fix since it is very hard to use the extension like that

cliffmeyers:
If this change breaks IE11 support that's probably a concern; @bitwiseman I'm not sure if you all have metrics on browser share for users of Jenkins to assess impact?

@ugatenio how does the extension function in IE11 after these changes? Do the screens still load?

Other info that would be helpful to drill into this:

- Capture a video / giphy of the perf issue and post here
- Capture  the JSON responses from your environment that are triggering the perf issues and include them in a PR along with a unit test that demonstrates the high execution time. Then it would be easier to make performance enhancements and verify they are working as expected

ugatenio:
> If this change breaks IE11 support that's probably a concern; @bitwiseman I'm not sure if you all have metrics on browser share for users of Jenkins to assess impact?
> 
> @ugatenio how does the extension function in IE11 after these changes? Do the screens still load?
> 
> Other info that would be helpful to drill into this:
> 
> * Capture a video / giphy of the perf issue and post here
> * Capture  the JSON responses from your environment that are triggering the perf issues and include them in this PR along with a unit test that demonstrates the high execution time. Then it would be easier to make performance enhancements and verify they are working as expected

sorry I cant expose that.
but you do not need that - to iterate on big graph in sync way is naive approach in single thread.

bitwiseman:
@cliffmeyers 
Good point.  IE11 support is a larger and separate discussion.

@ugatenio  
Cliff's request is the absolute minimum bar for this PR to be considered.
Better would be automated tests that show the behavior. 

I look forward to someone having time to provide the needed information to evaluate this change.

eplodn:
Opened JENKINS-62449 that is related to this problem (me and @ugatenio work together).

BlueOcean is completely unusable in our environment on anything beyond the most basic pipelines.

ugatenio:
can be closed.
I open a new PR under my name: https://github.com/jenkinsci/blueocean-plugin/pull/2104

PR #2064, opened 2020-04-15
[JENKINS-61822] Fix github-api 1.107+ incompatibilities

bitwiseman:
We took dependencies on github-api internals.  This resulted in breakages.

This changes fixes them.

Still in progress. Will continue to break until github-api update is released.

# Description

See [JENKINS-61822](https://issues.jenkins-ci.org/browse/JENKINS-61822).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GHRepoEx.java
```diff
@@ -13,11 +13,4 @@
  */
 @SuppressFBWarnings(value = "EQ_DOESNT_OVERRIDE_EQUALS", justification = "no extra state added to affect superclass equality")
 public class GHRepoEx extends GHRepository {
-    @JsonProperty("private")
```
bitwiseman, line 16: This is no longer needed, but removing the class is more involved, since it is a public class. :sob:

pom.xml
```diff
@@ -645,7 +645,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>github-api</artifactId>
-            <version>1.90</version>
+            <version>1.111-SNAPSHOT</version>
```
bitwiseman, line 648: When it is released...
```suggestion
            <version>1.111</version>
```

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubServerContainer.java
```diff
@@ -97,8 +100,12 @@
                             inputStream = HttpRequest.getErrorStream(connection);
                         }
 
+                        if (inputStream == null) {
+                            throw new IOException();
+                        }
```
dwnusbaum, line 105: Any particular reason for this? It looks kind of unrelated to the other changes. If it occurs in practice it would be nice to an error message.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubRepositories.java
```diff
@@ -64,7 +65,8 @@ public GithubRepositories(StandardUsernamePasswordCredentials credentials, Strin
                     parent.getRepoType(),
                     pageSize, pageNumber), accessToken);
 
-            this.repositories = GithubScm.om.readValue(HttpRequest.getInputStream(connection), GH_REPO_EX_LIST_TYPE);
+            this.repositories = GithubScm.getMappingObjectReader().forType(GH_REPO_EX_LIST_TYPE)
+                .readValue(HttpRequest.getInputStream(connection));
```
dwnusbaum, line 69: Just curious, what would the right approach look like? IIUC, Blue Ocean implemented its own HTTP layer, but is trying to use the data binding layer from github-api, and instead it should be using a higher-level API that does both at once?

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GHRepoEx.java
dwnusbaum: Does removing this field affect backwards compatibility? I am not really sure how the class is used.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubServerContainer.java
bitwiseman: The tests expect the `readValue()` to throw an `IOException` when `inputStream` is `null`.  In the current version of Jackson, it now throws `IllegalArgumentException` producing a different result. 

https://github.com/jenkinsci/blueocean-plugin/pull/2064/files#diff-83cc03e1fffef92465efa766f12eb5f6R117

The right thing is probably to catch `IllegalArgumentException` instead.  I'll fix this up.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubRepositories.java
bitwiseman: Correct, BO shouldn't have reimplemented the only HTTP layer. It should be using the github-api or entirely implementing its own data mapping.   

The objects BO binds are only valid as data containers and even some getters will not work correctly due to the objects having no reference to the expected `GitHub` connection.  

Also, BO doesn't use OkHttp so it gets none of the benefits of any caching. 

I'm sure there were good reasons for this being implemented this way, but it is really unfortunate. 

Whatever behavior was missing could certainly have been added to github-api and in fact most of it has (and even been improved on).  But going through and converting all the places HTTPRequest is used will take more time and have more risk of further breakage.  

This method is a prefect example - it reimplements the paged reading that is already in github-api (including [all the page url parsing](https://github.com/jenkinsci/blueocean-plugin/pull/2064/files#diff-cd796a86ce13b64cde1471f6a4165551R77-R97) ).

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GHRepoEx.java
bitwiseman: It is actually present in the parent class.  I'm not sure what `GHRepository defines _private but doesn't get deserialized reliably with Jackson" means.  It is identical code.

bitwiseman: I'm not kidding.  This was fixed 6 years ago.  

https://github.com/github-api/github-api/blame/e426237c35910500d4484e644862326a79df36b5/src/main/java/org/kohsuke/github/GHRepository.java#L84

I guess if you're using an old enough version of github-api ... but wow.

olamy: well I'm not really it's used except here..... not sure someone will complain if we remove the class... Remove makes sense as this class it's just a hack

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubRepositories.java
olamy: That would be great but so much big change... :(

bitwiseman:
Deployed snapshots: the problem is I have to publish github-api library, wait for it to become visible to CI, then I can publish the snapshot of github-api-plugin.  It limits the usefulness considering how thin a wrapper the plugin is.  I tested it locally though.

bitwiseman:
Build succeeded.  I'm running the tests locally since the CI test server is down.

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubServerContainer.java
```diff
@@ -98,10 +101,13 @@
                         }
 
                         TypeReference<HashMap<String, Object>> typeRef = new TypeReference<HashMap<String, Object>>(){};
-                        Map<String, String> responseBody = GithubScm.om.reader().forType(typeRef).readValue(inputStream);
+                        Map<String, String> responseBody = GithubScm.getMappingObjectReader().forType(typeRef).readValue(inputStream);
 
                         isGithubCloud = code == 200 && responseBody.containsKey("current_user_url");
                         isGithubEnterprise = code == 401 && responseBody.containsKey("message");
+                    } catch (IllegalArgumentException iae) {
```
olamy, line 108: same log so maybe } catch (IllegalArgumentException | IOException ioe)

bitwiseman:
@dwnusbaum @olamy @kshultzCB @car-roll 
As you can see that last five commits contain a bunch of random updates attempting the fix the breaking tests.  And I eventually succeeded.  

I could try reducing the number of changes but I feel that all of these are actually reasonable to do anyway.

dwnusbaum:
Last CI failure was because the Jenkinsfile has a syntax error, [this line](https://github.com/jenkinsci/blueocean-plugin/blob/7b5c10e50a750edb03eb5db1ba49f45fa02bd3d7/Jenkinsfile#L103) needs to be `echo(err.getMessage())` or `echo(err.toString())`, but really it would be better to drop the echo and have the `catch` block `throw` the error at the end so the result from the `FlowInterruptedException` is used if present and so we see the stack trace from whatever actually went wrong.

bitwiseman:
The previous failure was infra. We've achieved fully passing all tests with this change once. Still, I've gone ahead and updated this with feedback.  I'll wait for the results of this CI.

bitwiseman:
Re: catch-throw - I did both. :smile:

PR #2070, opened 2020-04-20
Improve Jenkinsfile error handling by removing unnecessary catch block

dwnusbaum:
# Description

The previous usage of the echo step was incorrect because the argument was not a string. It is generally better to throw errors instead of logging them and using `currentBuild.result` to set the result so that the stack trace of the error is displayed in the build log. The relevant code failed in [this PR build](https://ci.blueocean.io/job/blueocean/job/PR-2064/20/), here is an excerpt from the log:

```
java.lang.ClassCastException: org.jenkinsci.plugins.workflow.steps.EchoStep.message expects class java.lang.String but received class org.jenkinsci.plugins.workflow.steps.FlowInterruptedException
```

We use `catchError` in the `finally` block so that exceptions thrown by that stage do not mask errors throw from the `try` block, which  are more likely to be the root cause of any problems.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -99,22 +99,17 @@ node() {
             }
           }
         }
-      } catch(err) {
-        echo(err)
-        currentBuild.result = "FAILURE"
-
-        if (err.toString().contains('exit code 143')) {
-          currentBuild.result = "ABORTED"
-        }
       } finally {
         stage('Cleanup') {
-          if (params.USE_SAUCELABS) {
-            sh "${env.WORKSPACE}/acceptance-tests/runner/scripts/stop-sc.sh"
-          } else {
-            sh "${env.WORKSPACE}/acceptance-tests/runner/scripts/stop-selenium.sh"
+          catchError(message: 'Suppressing error in Stage: Cleanup') {
+            if (params.USE_SAUCELABS) {
+              sh "${env.WORKSPACE}/acceptance-tests/runner/scripts/stop-sc.sh"
+            } else {
+              sh "${env.WORKSPACE}/acceptance-tests/runner/scripts/stop-selenium.sh"
+            }
+            sh "${env.WORKSPACE}/acceptance-tests/runner/scripts/stop-bitbucket-server.sh"
+            deleteDir()
           }
-          sh "${env.WORKSPACE}/acceptance-tests/runner/scripts/stop-bitbucket-server.sh"
-          deleteDir()
         }
       }
     }


Jenkinsfile
```diff
@@ -99,22 +99,17 @@ node() {
             }
           }
         }
-      } catch(err) {
-        echo(err)
-        currentBuild.result = "FAILURE"
-
-        if (err.toString().contains('exit code 143')) {
-          currentBuild.result = "ABORTED"
-        }
```
dwnusbaum, line 108: If the error is thrown to the top level, which happens automatically with the `catch` block removed, the build result will be set appropriately. IDK if there was an important reason for the `err.toString().contains('exit code 143')` check that set the build result to `ABORTED`. I deleted it because I don't think it really matters what the build result is if it's not a success, just that it wasn't a success. If there is an important reason for it being there, I can add it back.

Jenkinsfile
```diff
@@ -99,22 +99,17 @@ node() {
...
       } finally {
         stage('Cleanup') {
-          if (params.USE_SAUCELABS) {
-            sh "${env.WORKSPACE}/acceptance-tests/runner/scripts/stop-sc.sh"
-          } else {
-            sh "${env.WORKSPACE}/acceptance-tests/runner/scripts/stop-selenium.sh"
+          catchError(message: 'Suppressing error in Stage: Cleanup') {
```
dwnusbaum, line 104: We don't want errors thrown by this stage to mask the errors from the `try` block, so we use `catchError`, which is a step that has similar semantics to `try/catch` but automatically sets the build result and logs the exception if necessary.

Jenkinsfile
kshultzCB: I don't know either. Looks like that was added in [1952](https://github.com/jenkinsci/blueocean-plugin/pull/1952), when @halkeye added support for building on Sauce Labs.

bitwiseman: exit code 143 tends to be when a maven is stopped due to another branch of a parallel pipeline execution.  But I don't feel like this adds much value.

halkeye:
The echo was there because errors were being silently captured and nobody knew what was wrong. Made debugging things super difficult. I don't remember the rest of the reasoning cause it was a year ago

halkeye:
Also it probably predates me but I moved things around

bitwiseman:
@halkeye So, you're okay with giving this change a try?

dwnusbaum:
> The echo was there because errors were being silently captured and nobody knew what was wrong

Yeah makes sense looking at the diff in https://github.com/jenkinsci/blueocean-plugin/pull/1952. This PR changes the `catch` block so that it just rethrows the exception so the message is printed when the build fails in addition to the stack trace, so it should be even more helpful.

halkeye:
Oh I didn't even review the code, just explained when I got tagged

PR #2071, opened 2020-04-20
Move jenkins-war downloading to use maven instead of raw curl

bitwiseman:
# Description
Fixes extremely slow jenkins-war download.

### DIFF
--- a/acceptance-tests/runner/scripts/download.sh
+++ b/acceptance-tests/runner/scripts/download.sh
@@ -63,7 +63,7 @@ download-jenkins() {
     echo "Downloading Jenkins ${JENKINS_VERSION} WAR to ${DOWNTO} ..."
     echo ""
 
-    mvn -B dependency:copy -Dartifact="org.jenkins-ci.main:jenkins-war:${JENKINS_VERSION}:war" ${MAVEN_SETTINGS} && cp "target/dependency/jenkins-war-${JENKINS_VERSION}.war" "${DOWNTO}"
+    mvn -B org.apache.maven.plugins:maven-dependency-plugin:3.1.2:copy -Dartifact="org.jenkins-ci.main:jenkins-war:${JENKINS_VERSION}:war" -DoutputDirectory=target/dependency ${MAVEN_SETTINGS} && cp "target/dependency/jenkins-war-${JENKINS_VERSION}.war" "${DOWNTO}"
 
     if [ $? != 0 ]
       then
--- a/acceptance-tests/runner/scripts/download.sh
+++ b/acceptance-tests/runner/scripts/download.sh
@@ -13,7 +13,7 @@ download() {
         echo ""
         popd
         return
-    fi    
+    fi
 
     echo ""
     echo "Downloading ${URL} to ${DOWNTO} ..."
@@ -31,7 +31,7 @@ download() {
         echo "curl or wget must be installed."
         exit 1
     fi
-    
+
     if [ $? != 0 ]
       then
         popd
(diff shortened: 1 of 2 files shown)


bitwiseman:
@dwnusbaum

 There must some cache setup otherwise these builds would end up downloading all the plugins at the start of each build.

bitwiseman:
@dwnusbaum  
Would you mind if we made the update to the next LTS in a future change?

dwnusbaum:
> There must some cache setup otherwise these builds would end up downloading all the plugins at the start of each build.

Yeah, I agree that it wouldn't make sense, but it's not like Blue Ocean infra is being actively maintained (maybe it is downloading all the plugins every build), so it seems worth checking before we overload Artifactory with war downloads (probably wouldn't matter anyway though since Blue Ocean CI builds are infrequent these days).

Looking at the Jenkinsfile, the ATH builds run on `ci.blueocean.io`, but the normal plugin tests run on `ci.jenkins.io`. The stuff on `ci.jenkins.io` is a [regular `buildPlugin`](https://github.com/jenkinsci/blueocean-plugin/blob/fd3ce2ff2c700a71280a721b865a67ee8c8f7026/Jenkinsfile#L4) call, so I'm not worried about that. For the ATH stuff, there is a [Maven settings file being used](https://github.com/jenkinsci/blueocean-plugin/blob/fd3ce2ff2c700a71280a721b865a67ee8c8f7026/Jenkinsfile#L51), but I don't have access to view it, so IDK how it is set up.

bitwiseman:
The `ci.jenkins.io` build doesn't run tests. All tests are run on ci.blueocean.io. 
I don't see any evidence of plugins and such being re downloaded.  

I just fixed and issue with this change.  
I'll look at the output when this run to see if it is re-downloading or not.

olamy:
@bitwiseman look at this line https://github.com/jenkinsci/blueocean-plugin/blob/master/Jenkinsfile#L51 
`blueocean-maven-settings` 
the content is 
```<?xml version="1.0" encoding="UTF-8"?>
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0" 
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
          xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
<!--  <mirrors>
    <mirror>
      <id>cache.blueocean.io</id>
      <mirrorOf>repo.jenkins-ci.org</mirrorOf>
      <url>http://172.18.128.166/mirrors/jenkins</url>
    </mirror>
  </mirrors> -->
</settings>```

But the build is using `-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn`  this means nothing is displayed and we don't even know if it's downloaded or not. (same as using `-ntp`)
using level info will help to know (but I guess it's downloading the world every build and it's even worst  because of using npm as well :) )

**Comment by bitwiseman** [comment]
https://ci.blueocean.io/blue/organizations/jenkins/blueocean/detail/PR-2071/5/pipeline/57#step-62-log-394  shows things downloading. Testing will continue until results improve.

**Comment by olamy** [comment]
I mean the main build do not show anything :)

File: acceptance-tests/runner/scripts/download.sh [hunk 1]
```diff
@@ -31,15 +31,47 @@ download() {
...
+        echo ""
+        popd
+        return
+    fi
+
+    echo ""
+    echo "Downloading Jenkins ${JENKINS_VERSION} WAR to ${DOWNTO} ..."
+    echo ""
+
+    mvn -B dependency:copy -Dartifact="org.jenkins-ci.main:jenkins-war:${JENKINS_VERSION}:war" && cp "target/dependency/jenkins-war-${JENKINS_VERSION}.war" "${DOWNTO}"
```
olamy, line 66: this will not work :) because per default Maven knows only Centra repo and jenkins artifact are not in central.
you can use a fake pom containing repo-ci as repository or run this download in the `ath` which have a pom with repo-ci configured.

bitwiseman:
Doh, I need to link the `mvn` call to a `pom.xml`.   
Also, yes, it is downloading everything, every time.

bitwiseman:
@olamy 
Thanks!

PR #2072, opened 2020-04-24
Fix tests on Java 11

bitwiseman:
# Description

Due to this project building on separate infrastructure that doesn't run test on Java 11, some tests failed when run as part of other projects which do run tests on Java 11.  The PR switches to running tests on ci.jenkins.io, including Java 8 and 11.  

It also makes a number of other test fixes that are required for that to succeed. 

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


bitwiseman:
The PR-head check is expected to not run.  If this works, we'll see if it is still needed.

bitwiseman:
I keep getting `Could not connect to EC2 (aws) - Ubuntu 18.04 LTS (i-0295191ddd1509923) to send interrupt signal to process`...

pom.xml
```diff
@@ -28,7 +28,7 @@
     <revision>1.23.0</revision>
     <changelist>-SNAPSHOT</changelist>
     <java.level>8</java.level>
-    <jenkins.version>2.150.3</jenkins.version> <!-- Should be kept in sync with the version resulting from Dockerfile FROM statement-->
+    <jenkins.version>2.176.3</jenkins.version> <!-- Should be kept in sync with the version resulting from Dockerfile FROM statement-->
```
dwnusbaum, line 31: Does this need to be updated anywhere else? Maybe here: https://github.com/jenkinsci/blueocean-plugin/blob/12183b7ae4befff6a4eb9158f69835b6d9712d14/Jenkinsfile#L31?

bitwiseman:
Here's the job histories for this change: 

https://ci.jenkins.io/job/Plugins/job/blueocean-plugin/job/PR-2072/
https://ci.blueocean.io/job/blueocean/job/PR-2072/

`sequentialParallelStagesLongRun` is the only remaining consistent failure.

pom.xml
```diff
@@ -1064,6 +1059,22 @@
   </build>
 
     <profiles>
+        <profile>
+            <id>enable-jacoco</id>
```
dwnusbaum, line 1063: I think this is already defined in the parent POM, see https://github.com/jenkinsci/plugin-pom/blob/8c269dbd6f2fd7810efd7ec8df9f6bdba5b37c82/pom.xml#L1059, so I don't think you need it here unless you have special overrides. If you do override it here, you probably want to copy https://github.com/jenkinsci/plugin-pom/blob/8c269dbd6f2fd7810efd7ec8df9f6bdba5b37c82/pom.xml#L1080 to exclude generated classes.

dwnusbaum:
I have definitely seen `sequentialParallelStagesLongRun` flake before when, so I would not worry about it.

Jenkinsfile
```diff
@@ -81,7 +87,7 @@ node() {
 
             junit '**/target/surefire-reports/TEST-*.xml'
             junit '**/target/jest-reports/*.xml'
-            jacoco execPattern: '**/target/jacoco.exec', classPattern : '**/target/classes', sourcePattern: '**/src/main/java', exclusionPattern: 'src/test*'
+            //jacoco execPattern: '**/target/jacoco.exec', classPattern : '**/target/classes', sourcePattern: '**/src/main/java', exclusionPattern: 'src/test*'
```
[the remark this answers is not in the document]
bitwiseman: It does not work after this change. I'm turning off jacoco in CI for this project.  I will remove the line.

pom.xml
bitwiseman: This was a mistake.

pom.xml
bitwiseman: good call.

bitwiseman:
Okay, we have some failure in ATH tests, but these have been flaky for some time. The ci.jenkins.io tests all passed (now with tests) on Java 8 and 11.

PR #2074, opened 2020-04-26
[JENKINS-62058] fix description newline display

pscherz:
# Description

See [JENKINS-62058](https://issues.jenkins-ci.org/browse/JENKINS-62058).

I don't think there are automated tests necessary, and wouldn't know how to. This fix just adds a white-space css member in the build description message box.

Just as described in the Jira ticket, newlines should be displayed as such.

### Build description with '\n' before this change:
BEFORE_single_line_description

### Same build after this change:
AFTER_muiltiple_line_description

## Test instruction
To get a build description with newlines either:
* In Jenkinsfile: set ```currentBuild.description = 'line 1\nline 2\nline 3'```
* In classic UI: Edit any existing build description (I tested with plain text)
 
# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


bitwiseman:
@pscherz At minimum we need screenshots before and after the change.

pscherz:
The images in the linked Jira ticket are actually before/after screenshots.
Is a full page screenshot more acceptable? I can do that.

bitwiseman:
@pscherz 
Ah, thanks.  Having the images here is helpful.  

https://developer.mozilla.org/en-US/docs/Web/CSS/white-space  describes the behavior of `pre-wrap` as: 

> Sequences of white space are preserved. Lines are broken at newline characters, at <br>, and as necessary to fill line boxes.

This seems reasonable to me.

steve-todorov:
This won't work for the old UI since you need to use `<br>` in the description for new lines. Maybe add support for `br` for descriptions in BlueOcean? :)

pscherz:
@steve-todorov There is a way to configure html for build descriptions and it seems we have this configured like that at work, but it looks to me, the default is a plain text formatter, which escapes everything and converts newlines to html breaks.
This is configured with "Global Security"/"Markup Formatter". I don't have a second option in there, just "Plain text".

The classic UI description and BlueOcean description look the same on my instance with this change. So I don't know, BlueOcean doesn't seem very configurable, and someone has to install a plugin before there have to be html breaks in the classic UI description.

 That said, I have no idea of the inner workings of Jenkins. ;)

PR #2084, opened 2020-08-04
Use ACL.as

jglick:
Internal cleanup noticed while investigating uses of relatively exotic Acegi Security APIs for https://github.com/jenkinsci/jenkins/pull/4848.

### DIFF
--- a/blueocean-jwt/src/main/java/io/jenkins/blueocean/auth/jwt/impl/JwtAuthenticationFilter.java
+++ b/blueocean-jwt/src/main/java/io/jenkins/blueocean/auth/jwt/impl/JwtAuthenticationFilter.java
@@ -2,12 +2,11 @@ package io.jenkins.blueocean.auth.jwt.impl;
 
 import hudson.Extension;
 import hudson.init.Initializer;
+import hudson.security.ACL;
+import hudson.security.ACLContext;
 import hudson.util.PluginServletFilter;
 import io.jenkins.blueocean.auth.jwt.JwtTokenVerifier;
 import org.acegisecurity.Authentication;
-import org.acegisecurity.context.SecurityContext;
-import org.acegisecurity.context.SecurityContextHolder;
-import org.acegisecurity.context.SecurityContextImpl;
 import org.apache.commons.lang.StringUtils;
 import org.kohsuke.stapler.Stapler;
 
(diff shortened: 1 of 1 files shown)


blueocean-git-pipeline/src/test/java/io/jenkins/blueocean/blueocean_git_pipeline/GitPipelineCreateRequestTest.java
```diff
@@ -41,27 +43,29 @@ public void createSomeStuff() throws Exception {
...
-                "$class", "io.jenkins.blueocean.blueocean_git_pipeline.GitPipelineCreateRequest",
-                "scmConfig", ImmutableMap.of("id", GitScm.ID, "uri", sampleRepo.toString())))
-            .build(Map.class);
-        assertNotNull(r);
-        assertEquals("pipeline1", r.get("name"));
+        User user = user("vivek", "Vivek Pandey", "vivek.pandey@gmail.com");
+        try (ACLContext ctx = ACL.as(user)) {
+            Map r = new PipelineBaseTest.RequestBuilder(baseUrl)
+                    .status(201)
+                    .jwtToken(getJwtToken(j.jenkins, user.getId(), user.getId()))
```
jglick, line 50: Most of these changes are whitespace, converting to a block so as to reset the authentication at the end. (Setting authentication and never resetting it is very poor form, even in a test.) Turn on the WS-ignoring diff mode to review better.

Note that I have no idea how many of these tests actually relied on `Jenkins.authentication` being switched to the test user; the utility methods did this across the board, but I suspect many just relied on such a user _existing_. Or they may not have relied on the user at all; many of these tests (unlike this one) have no apparent relationship to authentication or access control.

jglick:
> [@olamy](/olamy) [olamy](/olamy) added the [bug](/jenkinsci/blueocean-plugin/labels/bug) label [3 days ago](#event-3681430530)

More like cleanup.

olamy:
I can't see any `cleanup` label here :)  https://github.com/jenkinsci/.github/blob/master/.github/release-drafter.yml

jglick:
`chore` or `internal`: https://github.com/jenkinsci/.github/blob/7baf6188c5dbe8f872bd7f4e13e1aaeee0b2e279/.github/release-drafter.yml#L42-L45

PR #2098, opened 2020-09-04
Ignoring some mock tests which cause problems in PCT

jglick:


olamy:
should be fixed by https://github.com/jenkinsci/blueocean-plugin/pull/2102

PR #2102, opened 2020-09-14
Changed PowerMock Ignore due to a comflict with SSLSocketFactory

joseblas:
# Description

This is a PR for fixing a PCT problem with SSLSocketFactory and PowerMockIgnore


dwnusbaum (approved):
Looks fine, but which tests is this fixing? Is it these ones?

```
io.jenkins.blueocean.blueocean_bitbucket_pipeline.server.BitbucketServerEndpointTest.createThenDelete
io.jenkins.blueocean.blueocean_bitbucket_pipeline.server.BitbucketServerEndpointTest.should404OnDeleteNonexistent
```

joseblas:
exactly. There is another PR from @jglick ignoring them, so I do not know which approach is better

jglick:
compare #2098

jglick (commented):
Great if it works. I suspected this was the right fix, but I did not manage to run this test in PCT, or did not manage to reproduce the failure (cannot recall which at the moment).

PR #2104, opened 2020-09-16
limit fetch call

ugatenio:
# Description

See [JENKINS-63698](https://issues.jenkins-ci.org/browse/JENKINS-63698).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


ugatenio:
https://issues.jenkins-ci.org/browse/JENKINS-63698

blueocean-core-js/src/js/capability/CapabilityAugmenter.js
```diff
@@ -101,9 +102,11 @@ export class CapabilityAugmenter {
                 if (canWalk(value) && nodesAlreadyWalked.indexOf(value) === -1 && ignoredProps.indexOf(key) === -1) {
                     nodesToWalk.push(value);
                 }
+                await sleep(1);
```
ugatenio, line 105: to consider threshold instead

ugatenio:
you can see here a lot of GET calls instead of one (in 30 seconds)
Screen Shot 2020-09-22 at 15 03 16

eplodn:
Our issue has been pinned down to `blueocean-executors-info` plugin.

As of now, that plugin has been uninstalled, as it is completely unclear why it exists or what it brings to the table. The amount of pain that it caused over the years to 150+ developers on our site is hard to describe.

@cliffmeyers @bitwiseman @kzantow Please consider deprecating it altogether, as I suspect there are other installations of it wrecking similar amounts of havoc elsewhere.

bitwiseman:
@eplodn 
We are no longer making major changes to Blue Ocean except for security and regression fixes.  Removing a component would be a major change.  

@ugatenio 
I don't have much context in which to evaluate this.  What you describe as the new behavior looks good, but I'm not sure how to validate it to make sure it doesn't cause a regression.  

@olamy @cliffmeyers  Any thoughts?  The looks like a pretty safe change, but I'd like some other folks to look at it too.

cliffmeyers:
@bitwiseman so I did a little digging on this issue. The last change to the JS file is linked to ticket https://issues.jenkins-ci.org/browse/JENKINS-56773 which describes more or less the same problem: too many API calls to this executor info endpoint. 
Some relevant discussion in the ticket:

> So had a bit of a chat with keith (original executor plugin author) and it was never really designed for more than a proof of concept on how to write a BlueOcean plugin.
> We are pretty sure its caused by https://github.com/jenkinsci/blueocean-plugin/blob/e2dd04dcb3015063a2e31c8d37c90ee5c1ef4d98/blueocean-executor-info/src/main/js/ExecutorInfoService.js#L14 (which will grab executor info every time any pipeline does work)
> So Plans:
> I think the plan is going to be to drop making the executor plugin as a requirement, it'll have to be manually installed again (Sorry Craig)
> Put a throttle on the lookup so it doesn't call nearly as often.
> Since there is a work around in place until the next release, and I personally have a few pressing issues to deal with first, i'm going to leave it unassigned, but we'll get it fixed before we do the next release.

This was the commit that attempted to address the issue:
https://github.com/jenkinsci/blueocean-plugin/commit/7cfca1868b27e227c545a01facce56a18cabf9c3#diff-97663dd2530a551206b11d406543156dR17

I think the original fix is actually be incorrect. Just wrapping the function in lodash `throttle` and immediately invoking it won't have the desired effect; that's just going to create a throttled function, immediately invoke it, and then discard the throttle itself. You need to invoke `throttle` on the function of interest, *hold a reference to the function it returns* and then ensure you are calling that shared reference.

Looking at the changes in this PR, it looks like a correct implementation, since it holds a reference to the return value from `throttle` and invokes that where needed.

eplodn:
@bitwiseman 
In this case we're looking at a plugin that `was never really designed for more than a proof of concept on how to write a BlueOcean plugin` and does not seem to bring any functionality to Jenkins at all.

We have never intended to develop a new blueocean plugin, so I guess it was originally installed at our place by just selecting anything with `blueocean` in its name in Plugin Manager. It took us a couple of years of effort (including trying to engage CloudBees, hiring an outsourcing consultant, but finally just having the right person — @ugatenio — with the skills and willingness to perform a deep dive) to finally figure it out (and solve the issue by uninstalling it). I assume there are other installations with this plugin where people are similarly not aware of it being installed and so these other people might face a similar performance issue, with not as much as a search string to bring them here for a solution.

bitwiseman:
Thanks @cliffmeyers @eplodn @jglick . 
I'll merge this change and add a task to tombstone this plugin.

bitwiseman (approved):

PR #2106, opened 2020-09-21
Trying to make mock tests pass against Spring Security

jglick:
Fixes one failure when run against jenkinsci/jenkins#4848. The other two fail in

```
java.lang.NullPointerException
	at hudson.security.ACL.hasPermission(ACL.java:134)
	at io.jenkins.blueocean.service.embedded.rest.UserImplPermissionTest$3.answer(UserImplPermissionTest.java:130)
	at io.jenkins.blueocean.service.embedded.rest.UserImplPermissionTest$3.answer(UserImplPermissionTest.java:126)
```

which I do not think can be fixed without actually updating the `jenkins.version`, because you cannot mock

```java
public static @NonNull Authentication getAuthentication2()
```

without that dep, and `ACL.hasPermission(Permission)` is `final`.

### DIFF
--- a/blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/rest/UserImplPermissionTest.java
+++ b/blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/rest/UserImplPermissionTest.java
@@ -42,6 +42,7 @@ import org.acegisecurity.AccessDeniedException;
 import org.acegisecurity.Authentication;
 import org.acegisecurity.GrantedAuthority;
 import org.apache.commons.lang.StringUtils;
+import org.junit.AssumptionViolatedException;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


blueocean-rest-impl/src/test/java/io/jenkins/blueocean/service/embedded/rest/UserImplPermissionTest.java
```diff
@@ -143,15 +149,20 @@ public Boolean answer(InvocationOnMock invocation) {
      */
     @Test
     public void useTestAgainstOrgBaseOnFolder() {
+        try { // https://github.com/powermock/powermock/issues/428
```
jglick, line 152: https://github.com/powermock/powermock/issues/428

jglick:
@olamy this is a follow-up to #2084.

olamy (approved):

PR #2116, opened 2020-10-29
Remove executor plugin from required deps

timja:
# Description

People are complaining because they can't uninstall this plugin but it's marked as deprecated, not a good user experience. Can we get a release ASAP for this?

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean/pom.xml
+++ b/blueocean/pom.xml
@@ -108,12 +108,6 @@
             <artifactId>blueocean-jira</artifactId>
         </dependency>
 
-        <!-- TODO remove with next release -->
-        <dependency>
-            <groupId>${project.groupId}</groupId>
-            <artifactId>blueocean-executor-info</artifactId>
-        </dependency>
-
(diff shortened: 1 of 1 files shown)


timja:
cc @olamy

jglick:
Well, there are one of two things you should do ASAP, and then cut a release of the reactor:

* Make this dependency `optional`—but keep it—so that people upgrading `blueocean` are guaranteed to also pick up the tombstone as a dependency bump. What #2114 should have done to begin with.
* Remove the dep, as here, but also delete the whole module from sources (incl. entry in root `pom.xml`). The trouble is that someone running an older release of `blueocean` might upgrade straight from that to this version, skipping the one containing #2114, and not specifically also select the update to `blueocean-executor-info`.

timja:
https://github.com/jenkinsci/blueocean-plugin/pull/2117

PR #2129, opened 2021-02-12
use slf4j 1.7.30

olamy:
# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -38,6 +38,7 @@
     <hamcrest.version>2.2</hamcrest.version>
     <mockito.version>3.3.3</mockito.version>
     <guava.version>18.0</guava.version>
+    <slf4j.version>1.7.30</slf4j.version>
     <jacoco.version>0.8.5</jacoco.version>
     <jacoco.haltOnFailure>false</jacoco.haltOnFailure>
     <jacoco.line.coverage>0.70</jacoco.line.coverage>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -630,6 +631,26 @@
...
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>log4j-over-slf4j</artifactId>
+            <version>${slf4j.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-jdk14</artifactId>
+            <version>${slf4j.version}</version>
+        </dependency>
```
jtnord, line 653: @olamy Jenkins ships with slf4j and thus these should be provided by Jenkins.  if BlueOcean is bundling them, then this is likely not the correct fix, rather the parent should be bumped (to pickup the jenkins bom and the hpi plugin fixed to avoid packaging libraries incorrectly) or the hpi plugin should be updated.  (or the reason for the inclusion should be otherwise tracked down and fixed?
https://github.com/jenkinsci/jenkins/blob/master/core/pom.xml#L451-L463
https://github.com/jenkinsci/jenkins/blob/master/bom/pom.xml#L108-L134

olamy: @jtnord good catch. seems weird to mark them `provided` but save some bytes :)

PR #2132, opened 2021-03-02
[JENKINS-58732] Remove docker-workflow dep from blueocean-pipeline-editor

jglick:
# Description

See [JENKINS-58732](https://issues.jenkins-ci.org/browse/JENKINS-58732). Amends #2056 by @Vlatombe to pick up https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/393 via https://github.com/jenkinsci/bom/pull/415, which in turn requires switching to a somewhat more recent LTS baseline (2.176.x → 2.222.x). While I am here, using the BOM more consistently.

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
     <changelist>-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/blueocean-plugin</gitHubRepo>
     <java.level>8</java.level>
-    <jenkins.version>2.176.4</jenkins.version> <!-- Should be kept in sync with the version resulting from Dockerfile FROM statement-->
+    <jenkins.version>2.222.4</jenkins.version>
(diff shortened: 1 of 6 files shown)


jglick:
@olamy @dwnusbaum et al.

pom.xml
```diff
@@ -29,7 +29,7 @@
     <changelist>-SNAPSHOT</changelist>
     <gitHubRepo>jenkinsci/blueocean-plugin</gitHubRepo>
     <java.level>8</java.level>
-    <jenkins.version>2.176.4</jenkins.version> <!-- Should be kept in sync with the version resulting from Dockerfile FROM statement-->
+    <jenkins.version>2.222.4</jenkins.version>
```
olamy, line 32: just figured out this upgrade is not so trivial... :( because the Jenkinsfile need to be upgraded as well to have BO ATH running with this version of Jenkins. And this not working due to some issue with https://github.com/jenkinsci/java-client-api. I will try to figure out this next week.

PR #2135, opened 2021-03-05
use slf4j, commons-io, commons-lang, commons-codec version coming from bom and mark them as provided

olamy:
# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-jira/pom.xml
+++ b/blueocean-jira/pom.xml
@@ -29,6 +29,12 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>jira</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-api</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/blueocean-jira/pom.xml
+++ b/blueocean-jira/pom.xml
@@ -29,12 +29,6 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>jira</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>org.slf4j</groupId>
-                    <artifactId>slf4j-api</artifactId>
-                </exclusion>
-            </exclusions>
         </dependency>
         <dependency>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -477,26 +476,25 @@
...
             <artifactId>log4j-over-slf4j</artifactId>
-            <version>${slf4j.version}</version>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-jdk14</artifactId>
-            <version>${slf4j.version}</version>
+            <scope>provided</scope>
         </dependency>
```
jtnord, line 498: these should not be here.  they are defined in the Jenkins-bom which is automatically consumed here.
Also the version seems to be higher than provided by the version of jenkins being used - so likely something may not work correctly (but hopefully it will "be fine" :)

blueocean-jira/pom.xml
```diff
@@ -29,6 +29,12 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>jira</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-api</artifactId>
+                </exclusion>
+            </exclusions>
```
jtnord, line 37: just building this project on the `master` branch I did not even see `slf4j` being bundled in the hpi - so it this at all needed?

jtnord (commented):
see #2138

blueocean-jira/pom.xml
olamy: fixed

pom.xml
olamy: this has been moved to `<dependencies>` sections and marked as `provided`

olamy:
I fixed this those and move the provided dependencies by core in `<dependencies>` section and marked them as `provided`

jtnord:
@olamy this did not to the fix you thought it did.
blueocean-bitbucket-pipeline.hpi 1.24.5 still has these embedded in the hpi

guessing as https://github.com/jenkinsci/blueocean-plugin/blob/blueocean-parent-1.24.5/blueocean-bitbucket-pipeline/pom.xml it declares useage of the artifact and the scope is in the dependency section not the dependencyManagement section so it trumps the scope as it is now compile (again)

PR #2252, opened 2022-02-25
Remove parboiled usage

raul-arabaolaza:
It was a transitive dependency from `token-macro` and it has been removed in last releases. Thus making BO to fail if the last version of token-macro is installed.

Plus, there is really no need to add a dependency just to check if a string is empty

# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
+++ b/blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
@@ -10,7 +10,6 @@ import io.jenkins.blueocean.rest.hal.Link;
 import io.jenkins.blueocean.rest.impl.pipeline.scm.Scm;
 import io.jenkins.blueocean.rest.impl.pipeline.scm.ScmFactory;
 import io.jenkins.blueocean.rest.impl.pipeline.scm.ScmServerEndpointContainer;
-import org.parboiled.common.StringUtils;
 
 import javax.annotation.Nonnull;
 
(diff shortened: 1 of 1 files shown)


Vlatombe (approved):

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
```diff
@@ -36,7 +35,7 @@ public GithubEnterpriseScm(Reachable parent) {
         String apiUri = getCustomApiUri();
 
         // NOTE: GithubEnterpriseScm requires that the apiUri be specified
-        if (StringUtils.isEmpty(apiUri)) {
+        if (apiUri == null || apiUri.isEmpty()) {
```
jtnord, line 38: is this the same?   does the upstream not treat a blank (all whitespace) string as empty?
(the java method is the string has length zero)

jtnord: yes it is :-)

https://www.javadoc.io/doc/org.parboiled/parboiled-core/0.11.0/org/parboiled/common/StringUtils.html#isEmpty(java.lang.String)

jtnord (approved):

blueocean-github-pipeline/src/main/java/io/jenkins/blueocean/blueocean_github_pipeline/GithubEnterpriseScm.java
raul-arabaolaza: Yep, I checked, I should have added a comment on the pr

raul-arabaolaza:
Failures can not be caused by my change...

PR #2385, opened 2023-01-18
Ceasing to test against non-block-scoped `stage`

jglick:
See https://github.com/jenkinsci/pipeline-stage-step-plugin/pull/94.

PR #2386, opened 2023-01-18
`RunSearch.collectRunsParallel` returned nondeterministic order

jglick:
I noticed while working on #2385 that I would sometimes get a failure

```
org.junit.ComparisonFailure: expected:<[2]> but was:<[1]>
	at org.junit.Assert.assertEquals(Assert.java:117)
	at org.junit.Assert.assertEquals(Assert.java:146)
	at io.jenkins.blueocean.rest.impl.pipeline.PipelineBaseTest.validateRun(PipelineBaseTest.java:335)
	at io.jenkins.blueocean.rest.impl.pipeline.PipelineBaseTest.validateRun(PipelineBaseTest.java:331)
	at io.jenkins.blueocean.rest.impl.pipeline.MultiBranchTest.getMultiBranchPipelineRuns(MultiBranchTest.java:408)
```

where the test expects to get a REST response with `p/master#2` followed by `p/master#1` but in fact gets the reverse. I was able to trace this to a regression introduced by @olamy in #1744: the `RunList` / `RunMap` serves builds in reverse order but `collectRunsParallel` farms them out to threads which then collect the `BlueRun` objects in a nondeterministic order. Reproduced via

```diff
diff --git blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunSearch.java blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunSearch.java
index 9b93a791d..710c55ec2 100644
--- blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunSearch.java
+++ blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunSearch.java
@@ -162,7 +162,14 @@ public class RunSearch extends OmniSearch<BlueRun> {
                 skipCount--;
             } else {
                 Run r = runIterator.next();
-                callables.add(() -> BlueRunFactory.getRun(r, () ->  parent));
+                callables.add(() -> {
+                    try {
+                        Thread.sleep(r.number * 1000);
+                    } catch (InterruptedException x) {
+                        x.printStackTrace();
+                    }
+                    return BlueRunFactory.getRun(r, () ->  parent);
+                });
             }
             if (callables.size() >= limit) {
                 break;
```

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunSearch.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunSearch.java
@@ -1,5 +1,6 @@
 package io.jenkins.blueocean.service.embedded.rest;
 
+import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Extension;
 import hudson.model.Job;
 import hudson.model.Run;
@@ -24,6 +25,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
 import java.util.concurrent.ArrayBlockingQueue;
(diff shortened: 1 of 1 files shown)


blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunSearch.java
```diff
@@ -102,29 +104,16 @@ public static Iterable<BlueRun> findRuns(Job job, final Link parent){
     }
 
     @SuppressWarnings("unchecked")
-    public static Iterable<BlueRun> findRuns(Job job, final Link parent, int start, int limit){
-        final List<BlueRun> runs = new ArrayList<>();
-        Iterable<Job> pipelines;
-        if(job != null){
-            pipelines = Collections.singletonList(job);
-        }else{
-            pipelines = Jenkins.get().getItems(Job.class);
```
jglick, line 111: AFAICT there were no callers actually passing `null` for `job`, so this code was overly complicated; and the reverse sort would not necessarily be correct if you were actually collecting builds from every job in the system (for some reason).

blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/rest/RunSearch.java
```diff
@@ -209,6 +198,7 @@ private static List<BlueRun> collectRunsParallel(Iterator<? extends Run> runIter
             executorService.shutdownNow();
         }
         LOGGER.debug("runs found:{}", runs.size());
+        runs.sort(Comparator.comparing((BlueRun r) -> Integer.valueOf(r.getId())).reversed());
```
jglick, line 201: Curious that the original `Run.number` is not retained. The `id` is just a stringified version.

PR #2406, opened 2023-02-10
the method is called twice for the same object so at least we can store the result in a field which reduce call by 2 :) and add some caching for completed run

olamy:
1.25.x

### DIFF
--- a/blueocean-github-pipeline/pom.xml
+++ b/blueocean-github-pipeline/pom.xml
@@ -39,8 +39,8 @@
     </dependency>
 
     <dependency>
-        <groupId>org.jenkins-ci.plugins</groupId>
-        <artifactId>pubsub-light</artifactId>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>pubsub-light</artifactId>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
@@ -74,5 +74,21 @@
     </dependency>
 
   </dependencies>
+
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+                <!-- because test GithubOrgFolderPermissionsTest is using some bad static library
+                     it would be better to fix that but maybe next time -->
+              <reuseForks>false</reuseForks>
+            </configuration>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+  </build>
 
 </project>
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeContainerImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeContainerImpl.java
@@ -17,14 +17,15 @@ import java.util.Map;
  * @author Vivek Pandey
  */
 public class PipelineNodeContainerImpl extends BluePipelineNodeContainer {
-    private final WorkflowRun run;
     private final Map<String, BluePipelineNode> nodeMap = new HashMap<>();
 
     private final List<BluePipelineNode> nodes;
     private final Link self;
 
+    private final String jobFullName;
+
     public PipelineNodeContainerImpl(WorkflowRun run, Link parentLink) {
-        this.run = run;
+        this.jobFullName = run.getParent().getFullName();
         this.self = parentLink.rel("nodes");
 
         WorkflowJob job = run.getParent();
@@ -47,11 +48,12 @@ public class PipelineNodeContainerImpl extends BluePipelineNodeContainer {
 
     @Override
     public BluePipelineNode get(String name) {
+
         if(nodeMap.get(name) != null){
             return nodeMap.get(name);
         }
         throw new ServiceException.NotFoundException(String.format("Stage %s not found in pipeline %s.",
-            name, run.getParent().getName()));
+            name, jobFullName));
     }
 
     @Override
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
@@ -1,5 +1,7 @@
 package io.jenkins.blueocean.rest.impl.pipeline;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
 import hudson.Extension;
 import hudson.model.Queue;
 import hudson.model.Run;
@@ -33,6 +35,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.Comparator;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 
 import static io.jenkins.blueocean.rest.model.KnownCapabilities.JENKINS_WORKFLOW_RUN;
@@ -45,6 +48,9 @@ import static io.jenkins.blueocean.rest.model.KnownCapabilities.JENKINS_WORKFLOW
 @Capability(JENKINS_WORKFLOW_RUN)
 public class PipelineRunImpl extends AbstractRunImpl<WorkflowRun> {
     private static final Logger logger = LoggerFactory.getLogger(PipelineRunImpl.class);
+
+    private BluePipelineNodeContainer bluePipelineNodeContainer;
+
     public PipelineRunImpl(WorkflowRun run, Reachable parent, BlueOrganization organization) {
         super(run, parent, organization);
     }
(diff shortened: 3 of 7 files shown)


car-roll (approved):
Maybe some additional caching tests?

olamy:
> ~Maybe some additional caching tests?~ I realized I may have approved too early :)

well the cache looks to work well as I have to change the tests to correctly clear before each run 🤣 (https://github.com/jenkinsci/blueocean-plugin/pull/2406/commits/4f3f4a63d6154cdc525c1119e5f7c196077e38c2 )

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
```diff
@@ -1,5 +1,8 @@
 package io.jenkins.blueocean.rest.impl.pipeline;
 
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.CacheLoader;
+import com.github.benmanes.caffeine.cache.Caffeine;
```
jglick, line 5: Being loaded from the plugin wrapper?

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
```diff
@@ -135,11 +143,33 @@ private boolean isReplayable(ReplayAction replayAction) {
         return replayAction != null && replayAction.isRebuildEnabled();
     }
 
+    static final long PIPELINE_NODE_CONTAINER_CACHE_MAX_SIZE = Long.getLong("PIPELINE_NODE_CONTAINER_CACHE_MAX_SIZE", 10000);
+
+    private static Cache<String, PipelineNodeContainerImpl> PIPELINE_NODE_CONTAINER_LOADING_CACHE = Caffeine.newBuilder()
```
jglick, line 148: How large is the retained size of a `PipelineNodeContainerImpl`? Does it refer to any objects from other plugins esp. `workflow-*`?

jglick, line 148: For future readers, nice to document what the key is. Some `.link.href` which would be, what, `job/mbp/job/main/123/`? Or `/blue/something`?

(Normally a `WorkflowRun` would be keyed by its `externalizableId`.)

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
```diff
@@ -135,11 +143,33 @@ private boolean isReplayable(ReplayAction replayAction) {
...
+        PIPELINE_NODE_CONTAINER_LOADING_CACHE.invalidateAll();
+    }
+
     @Override
     @Navigable
     public BluePipelineNodeContainer getNodes() {
         if (run != null) {
-            return new PipelineNodeContainerImpl(run, getLink());
+            // not using cache if run not completed
+            if (!run.isBuilding() ) {
```
jglick, line 165: You might want
```suggestion
            // not using cache if run not completed
            if (!run.isLogUpdated() ) {
```
which is subtly different.

jglick (approved):
Looks reasonable. Any idea if it helps in practice?

car-roll (approved):

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
olamy: yup. sadly by dependency transitivity and not directly but I don't really want touch too much poms in a maintenance branch.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
olamy: the reference to WorkflowRun has been removed here ae3da2701919e5d6c4dfa1fecd054754f247e2d8 (was not really useful) but some objects a bit deeper in the tree keep reference on it...

olamy: done https://github.com/jenkinsci/blueocean-plugin/commit/ae3da2701919e5d6c4dfa1fecd054754f247e2d8

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
olamy: done https://github.com/jenkinsci/blueocean-plugin/commit/ae3da2701919e5d6c4dfa1fecd054754f247e2d8

olamy:
please note https://github.com/jenkinsci/blueocean-plugin/pull/2407 contains those 4 commits but for a branch based on 12.5.5 (as the current PR is based on 12.5.x which have more commits since 1.25.5)

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
jglick: Probably wise to avoid all `WorkflowRun` references from cache values, since they can be big and are supposed to be possible to unload.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeContainerImpl.java
```diff
@@ -17,14 +17,15 @@
  * @author Vivek Pandey
  */
 public class PipelineNodeContainerImpl extends BluePipelineNodeContainer {
-    private final WorkflowRun run;
     private final Map<String, BluePipelineNode> nodeMap = new HashMap<>();
 
     private final List<BluePipelineNode> nodes;
     private final Link self;
 
+    private final String runParentName;
```
jglick, line 25: `jobFullName` might be a more intuitive name BTW.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
olamy: Yes I definitely agree but it's lot of changes for a bug branch.. anyway we have to do it but I'm not exactly sure how to load? Jenkins.get().getItem? But what is the key/id here?

jglick: For a `Job`, use `.fullName`; for a `Run`, use `.externalizableId`.

olamy: uhm trying to load from `getExternalizableId ` doesn't work.
A quick test such `Jenkins.get().getItem(run.getExternalizableId())` returns null.

olamy: I have some other ways to do it but my concern is to not go to some list search etc... which could potentially load some data we don't need.

jglick: > returns null

Because you are mixing up `Job` and `Run` ids. The first two methods for jobs. The second two for builds.

blueocean-github-pipeline/pom.xml
```diff
@@ -75,4 +75,20 @@
...
+  <build>
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <configuration>
+                <!-- because test GithubOrgFolderPermissionsTest is using some bad static library
+                     it would be better to fix that but maybe next time -->
+              <reuseForks>false</reuseForks>
```
jglick, line 87: Why? This is long since the default in the parent POM: https://github.com/jenkinsci/plugin-pom/pull/453

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
```diff
@@ -157,6 +159,16 @@ static void clearCache() {
         PIPELINE_NODE_CONTAINER_LOADING_CACHE.invalidateAll();
     }
 
+
+    public static WorkflowRun findRun(String externalizableId) {
+        String[] tree = externalizableId.split("#");
+        Item item = Jenkins.get().getItem(tree[0], Jenkins.get());
+        if (item instanceof WorkflowJob) {
+            return ((WorkflowJob)item).getBuild(tree[1]);
```
jglick, line 167: Just use `Run.fromExternalizableId`.

blueocean-github-pipeline/pom.xml
olamy: maybe look at the blueocean itself and not only macro changes from a PR and you will see this https://github.com/jenkinsci/blueocean-plugin/blob/feb97f3796f4f616fec59698f2fe536fd506bc0b/pom.xml#L763

which is perfectly right and definitely faster but not for this module and not for this library called `WireMockRule` (probably because it's a tool old version but here it's a patched branch so touching this is not an option)

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
olamy: argh yes I couldn;t find this static method

blueocean-github-pipeline/pom.xml
jglick: Yeah `reuseForks=false` is set as the default under the assumption that most tests are going to be `JenkinsRule`-based. It does add some overhead if you have tons of short test cases.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
```diff
@@ -157,6 +159,16 @@ static void clearCache() {
...
+        }
+        return null;
```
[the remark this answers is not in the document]
olamy: please look at the code and you will see the code using this method check if run is null or not.
https://github.com/jenkinsci/blueocean-plugin/blob/65121bccdd759b7fdad6f79c898138820c1a07d4/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java#L186
I have added some comments as well..
even the diff show this
Screenshot 2023-02-15 at 7 17 34 am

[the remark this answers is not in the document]
jglick: I was referring for example to https://github.com/jenkinsci/blueocean-plugin/blob/65121bccdd759b7fdad6f79c898138820c1a07d4/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java#L180-L187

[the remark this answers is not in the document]
olamy: look at the constructor of `PipelineNodeImpl` and btw it's only used by `PipelineNodeGraphVisitor` at this stage if the `WorkflowRun` instance doesn't exist it would have been a NPE long way before.
they cannot be called if the run/build doesn't exist because it's a "sort" of graph tree walk on existing objects. such job -> list of run -> detail of the run -> details of the pipeline steps.
so yes this could fail but it's already the case today if between the time you see the list of runs and then you want to go deeper see detail of a run someone or any event delete a run.

olamy:
re open to trigger build

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineRunImpl.java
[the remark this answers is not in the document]
olamy: Spotbugs is enabled already.
anyway I changed to throw NPE so we can move forward with that.

PR #2414, opened 2023-03-16
add this possible heavy computation in the constructor so it is stored and cached then future usage will be really quick...

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

# Description

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
+++ b/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
@@ -57,6 +57,8 @@ public class PipelineNodeImpl extends BluePipelineNode {
     private final Link self;
     private final String runExternalizableId;
     private final Reachable parent;
+
+    private final boolean restartable;
     public static final int waitJobInqueueTimeout = Integer.getInteger("blueocean.wait.job.inqueue", 1000);
 
     public PipelineNodeImpl(FlowNodeWrapper node, Reachable parent, WorkflowRun run) {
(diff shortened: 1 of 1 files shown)


Dohbedoh:
We might need to store the value only when the stage / run is finished. A stage may be non restartable when looking at a running pipeline but will be when completed.

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java
```diff
@@ -67,6 +69,18 @@ public PipelineNodeImpl(FlowNodeWrapper node, Reachable parent, WorkflowRun run)
         this.durationInMillis = node.getTiming().getTotalDurationMillis();
         this.self = parent.getLink().rel(node.getId());
         this.parent = parent;
+        // TODO move to use of https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/596
```
jglick, line 72: https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/596

There is an incremental deployment you could pick up (if this were switched to draft).

olamy: can't really as target branch is `1.25.5.x`

olamy:
> We might need to store the value only when the stage / run is finished. A stage may be non restartable when looking at a running pipeline but will be when completed.

FTR see comment https://github.com/jenkinsci/blueocean-plugin/pull/2415#issuecomment-1471458718

olamy:
mixed with other changes in https://github.com/jenkinsci/blueocean-plugin/pull/2415

PR #2435, opened 2023-05-08
Bump BOM and explicitly depend on `caffeine-api` to fix `LoadingCache.refresh`

jglick:
Otherwise you get constant errors like

```
java.lang.NoSuchMethodError: 'void com.github.benmanes.caffeine.cache.LoadingCache.refresh(java.lang.Object)'
	at io.jenkins.blueocean.rest.impl.pipeline.Caches$ListenerImpl.onUpdated(Caches.java:70)
```

Caused by https://github.com/ben-manes/caffeine/issues/143 as noted in https://github.com/ben-manes/caffeine/issues/543#issuecomment-826987183 and picked up from https://github.com/jenkinsci/caffeine-api-plugin/pull/79. Source-compatible but binary-incompatible, so suffices to rebuild against 3.x.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -259,7 +259,7 @@
         <dependency>
             <groupId>io.jenkins.tools.bom</groupId>
             <artifactId>bom-2.361.x</artifactId>
-            <version>1798.vc671fe94856f</version>
+            <version>2059.v69eec68eb_b_b_e</version>
             <scope>import</scope>
             <type>pom</type>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -496,28 +496,6 @@
             <version>1.24</version>
         </dependency>
         <dependency>
-            <groupId>org.jenkins-ci.plugins</groupId>
-            <artifactId>github-api</artifactId>
-            <version>1.114.2</version>
-            <exclusions>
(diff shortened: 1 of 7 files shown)


jglick:
POM errors, looking…

rsandell (approved):

jglick:
@olamy please consider merging & releasing soon, as the `caffeine-api` update has been out there for a couple weeks and throws errors.

pom.xml
```diff
@@ -792,8 +764,8 @@
                         <rules>
                             <bannedDependencies>
                                 <excludes>
+                                    <!-- TODO very likely wrong -->
```
olamy, line 767: comment here https://github.com/jenkinsci/blueocean-plugin/pull/2433#discussion_r1187977213

jglick: Edit the comment if you like, but can we ship this, or perhaps some more minimal version limited to `caffeine-api` without a general BOM cleanup?

olamy: except the comment. everything looks good.

olamy: https://github.com/jenkinsci/code-coverage-api-plugin/issues/647

PR #2442, opened 2023-06-08
Allow the user to know if their click will be useful

mjaggard:
# Description

When clicking on "Open Blue Ocean" sometimes the click is useful, for example taking me to a page where I can see the progress of my current build. Other times it is useless, for example when I click on "Open Blue Ocean" on a job page. This change allows the user to know before they click whether they'll be taken to a helpful related page or to a generic Blue Ocean home page.
See [JENKINS-71418](https://issues.jenkins-ci.org/browse/JENKINS-71418).

No new tests because I'm not sure how to set up or mock the required build in order to test this.
Manual testing:
* Vew the link on a generic Jenkins page such as http://jenkinsserver/ and check it says "Open Blue Ocean Homepage"
* View the link on a specific numbered Jenkins build page such as http://jenkinsserver/job/foldername/job/jobname/1/ and check it says "Open Blue Ocean"

I did also consider "Open In Blue Ocean" for the helpful link and leaving "Open Blue Ocean" for the generic one.

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/main/resources/io/jenkins/blueocean/service/embedded/Messages.properties
+++ b/blueocean-rest-impl/src/main/resources/io/jenkins/blueocean/service/embedded/Messages.properties
@@ -1,1 +1,2 @@
 BlueOceanUrlAction.DisplayName=Open Blue Ocean
+BlueOceanUrlAction.GenericDisplayName=Open Blue Ocean Homepage
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/BlueOceanUrlObjectImpl.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/BlueOceanUrlObjectImpl.java
@@ -23,7 +23,11 @@ public class BlueOceanUrlObjectImpl extends BlueOceanUrlObject {
 
     @Override
     public @NonNull String getDisplayName() {
-        return Messages.BlueOceanUrlAction_DisplayName();
+        if (this.mappedUrl.equals(BlueOceanUrlMapperImpl.getLandingPagePath())) {
+            return Messages.BlueOceanUrlAction_DisplayName();
+        } else {
+            return Messages.BlueOceanUrlAction_GenericDisplayName();
+        }
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


mjaggard:
I'm unable to see what the CI failure is because I don't have permission.

blueocean-rest-impl/src/main/resources/io/jenkins/blueocean/service/embedded/Messages.properties
```diff
@@ -1 +1,2 @@
-BlueOceanUrlAction.DisplayName=Open Blue Ocean
\ No newline at end of file
+BlueOceanUrlAction.DisplayName=Open Blue Ocean
+BlueOceanUrlAction.GenericDisplayName=Open Blue Ocean Homepage
```
timja, line 2: really not useful 

build url:
https://build.platform.hmcts.net/view/<view name>/job/org-folder/job/repo-n/job/name/master/541/console

image

It opens a run page just fine.

This is really too long for a side panel link

olamy: Agree. This need to be fixed. I will revert this change

olamy: https://github.com/jenkinsci/blueocean-plugin/pull/2474

mjaggard: This was added because every day I get frustrated that the "Blue Ocean" link takes me to a useless homepage that just shows the list of PRs when I click the link from my PR page - it only goes to a useful page when I select it from an actual build.

mjaggard: @timja @olamy

timja: it would be better to fix that instead if possible then?

Blue ocean doesn't have an equivalent page for that view though which is why.

It could go to either:
https://ci.jenkins.io/blue/organizations/jenkins/Core%2Fjenkins/pr
or
Just the last build on the PR

----

see this link here:
https://ci.jenkins.io/job/Core/job/jenkins/view/change-requests/job/PR-5597/134/

the change made here added the 'homepage' link to that view which is wrong because that opens a correct page.

----

Blue Ocean Dashboard would probably be okay as a term for just the case where there's no matching blue ocean page I think, I guess this doesn't affect many people as they will normally open their build from a link on their SCM provider which will have a direct link to the right place so I would say it's a bit unusual to open it from there.

PR #2470, opened 2023-07-31
Tolerate testing in PCT context

basil:
When attempting to add Blue Ocean to `jenkinsci/bom` I discovered this test started failing with the list being of size 3 rather than the expected size 2. Peeking at the contents I see

```
io.jenkins.blueocean.rest.impl.pipeline.StageDurationTrend$FactoryImpl@7e9c962a
io.jenkins.blueocean.service.embedded.rest.junit.BlueJUnitTrend$FactoryImpl@36b5b88a
io.jenkins.blueocean.service.embedded.rest.BlueTrendTest$1@5584dbb
```

whereas when the test is executed outside of PCT only the last two entries are present. This is to be expected, since PCT places all the plugins in the managed set in the class path. To deal with this issue I have made the test tolerate this situation by checking for a size greater than or equal to 2.


basil:
@olamy https://ci.jenkins.io/job/Plugins/job/blueocean-plugin/job/PR-2470/1/ is green

PR #2479, opened 2023-08-16
Merge 1.27.5.x with SECURITY-3116 fix into default branch

daniel-beck:
https://www.jenkins.io/security/advisory/2023-08-16/#SECURITY-3116

Notes on how I see this PR in https://issues.jenkins.io/browse/SECURITY-3116?focusedId=440734&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-440734 (available to maintainers only)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
   <groupId>io.jenkins.blueocean</groupId>
   <artifactId>blueocean-parent</artifactId>
-  <version>${revision}${changelist}</version>
+  <version>1.27.5.1</version>
   <packaging>pom</packaging>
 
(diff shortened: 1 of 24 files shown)


pom.xml
```diff
@@ -25,7 +25,7 @@
   </licenses>
 
   <properties>
-    <revision>1.27.6</revision>
+    <revision>1.27.5.2</revision>
```
jglick, line 28: Is this really supposed to be here @daniel-beck @olamy?

olamy: Good catch nope
I will fix that

olamy: fixed with 15a25be31

PR #2485, opened 2023-08-24
[JENKINS-71803] Blue Ocean tests fail on Java 21 with Mockito errors

basil:
See [JENKINS-71803](https://issues.jenkins.io/browse/JENKINS-71803). Fixes the issue by using the Mockito version from the plugin parent POM (version 5, which supports Java 21) rather than defining an outdated version in this repository.

### DIFF
--- a/blueocean-rest-impl/pom.xml
+++ b/blueocean-rest-impl/pom.xml
@@ -97,6 +97,7 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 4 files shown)


blueocean-rest-impl/pom.xml
```diff
@@ -97,6 +97,7 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
```
basil, line 100: Removes the following unnecessary files from the JPI:

- `byte-buddy-1.12.13.jar`
- `byte-buddy-agent-1.12.13.jar`
- `mockito-core-4.7.0.jar`
- `objenesis-3.2.jar`

basil:
Build is green: https://ci.jenkins.io/job/Plugins/job/blueocean-plugin/job/PR-2485/1/

basil:
A merge/release would assist in the resolution of https://github.com/jenkinsci/bom/issues/2381

PR #2488, opened 2023-08-28
Use JaCoCo version from plugin parent POM

basil:
No need for this plugin to carry its own custom (and out of date) version.


basil:
@olamy Build is green: https://ci.jenkins.io/job/Plugins/job/blueocean-plugin/job/PR-2488/3/

PR #2468, opened 2023-07-28
Flake in `GitPipelineCreateRequestTest`

jglick:
Observed in a PCT run:

```
java.io.IOException: …/work/output-blueocean-git-pipeline/work/blueocean-plugin/blueocean-git-pipeline/target/tmp/j h7427903771576211726/caches/git-2b164368645dee5ba81079f6822e8bec/.git/config
	at org.jvnet.hudson.test.TemporaryDirectoryAllocator.delete(TemporaryDirectoryAllocator.java:144)
	at org.jvnet.hudson.test.TemporaryDirectoryAllocator.delete(TemporaryDirectoryAllocator.java:131)
	at org.jvnet.hudson.test.TemporaryDirectoryAllocator.delete(TemporaryDirectoryAllocator.java:131)
	at org.jvnet.hudson.test.TemporaryDirectoryAllocator.delete(TemporaryDirectoryAllocator.java:131)
	at org.jvnet.hudson.test.TemporaryDirectoryAllocator.dispose(TemporaryDirectoryAllocator.java:99)
	at org.jvnet.hudson.test.TestEnvironment.dispose(TestEnvironment.java:84)
	at org.jvnet.hudson.test.JenkinsRule.after(JenkinsRule.java:524)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:625)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:299)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:293)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.nio.file.DirectoryNotEmptyException: …/work/output-blueocean-git-pipeline/work/blueocean-plugin/blueocean-git-pipeline/target/tmp/j h7427903771576211726/caches/git-2b164368645dee5ba81079f6822e8bec/.git
	at java.base/sun.nio.fs.UnixFileSystemProvider.implDelete(UnixFileSystemProvider.java:247)
	at java.base/sun.nio.fs.AbstractFileSystemProvider.deleteIfExists(AbstractFileSystemProvider.java:110)
	at java.base/java.nio.file.Files.deleteIfExists(Files.java:1181)
	at org.jvnet.hudson.test.TemporaryDirectoryAllocator.delete(TemporaryDirectoryAllocator.java:141)
[...]

PR #2505, opened 2023-09-21
Prevent multiBranchPipelineIndex failure due to DirectoryNotEmptyException

MarkEWaite:
# Fix multiBranchPipelineIndex test failure

- Skip flaky multiBranchPipelineIndex test
- Do not exit test until no activity

https://github.com/jenkinsci/bom/runs/16996428210 shows the failure happened in 2 of the 4 configurations that are tested by the plugin bill of materials.

This pull request offers two different alternatives to resolve the issue.

1. Stop running the test (most likely to resolve the issue, but will hide failures that the test might have detected)
2. Wait for Jenkins to be quiet before exiting the test (may not prevent all test failures)

I'm open to either change and am happy to modify this pull request to include which ever is preferred by @olamy and the other maintainers of the plugin.

We like having the blueocean plugin in the Jenkins plugin bill of materials so that we can detect issues more quickly.

# Submitter checklist

# Reviewer checklist
- [x] Reviewed the code

### DIFF
--- a/blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/MultiBranchTest.java
+++ b/blueocean-pipeline-api-impl/src/test/java/io/jenkins/blueocean/rest/impl/pipeline/MultiBranchTest.java
@@ -318,6 +318,8 @@ public class MultiBranchTest extends PipelineBaseTest {
 
     @Test
     public void multiBranchPipelineIndex() throws Exception {
+        // Skip test to avoid intermittent test failures with DirectoryNotEmptyException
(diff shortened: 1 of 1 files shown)


jglick (approved):
Similar to #2468. TBH it is not clear to me what tests like this are even doing to begin with. That you get a 200 response from some HTTP endpoint? But it does not seem to assert the contents of the response at all. Apparently this is a multibranch project, and perhaps after branch indexing completes there are some builds? Or just branch projects with no builds?

jglick:
Both failures in https://github.com/jenkinsci/blueocean-plugin/pull/2505/checks?check_run_id=17030384497 look like unrelated flakes.

PR #2532, opened 2023-12-04
Upgrade Jira plugin 3.12

basil:
This isn't managed in BOM, and we're currently depending on an out-of-date version. The update to `maven-artifact` is to satisfy a require upper bounds error.


jglick:
@olamy trunk builds of this plugin are currently broken (https://github.com/jenkinsci/blueocean-plugin/pull/2542#issuecomment-1943618055). Perhaps this fixes them? https://github.com/jenkinsci/jira-plugin/pull/471 may be related.

PR #2540, opened 2024-02-01
Replace `DownstreamJobListener` with a `TransientActionFactory` based on `DownstreamBuildAction`

dwnusbaum:
`DownstreamJobListener` here has the same bug as https://github.com/jenkinsci/pipeline-build-step-plugin/pull/127 for completed builds. Prior to https://github.com/jenkinsci/workflow-cps-plugin/pull/807, this only worked as long as you were using the `MAX_SURVIVABILITY` durability level, otherwise it did nothing when the build was already complete. https://github.com/jenkinsci/workflow-cps-plugin/pull/826 then made it so a warning was logged instead of the save being silently ignored in the problematic case, which can lead to log spam.

This PR tries to replace `DownstreamJobListener` with a `TransientActionFactory` that uses `DownstreamBuildAction` from https://github.com/jenkinsci/pipeline-build-step-plugin/pull/128 to avoid the problem. I am not sure about the performance of this approach though.

Untested, because right now Blue Ocean doesn't build for me (I think the main problem is `Cannot download "https://github.com/sass/node-sass/releases/download/v7.0.0/darwin-x64-64_binding.node": ` when trying to build the frontend code).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code


blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/NodeDownstreamBuildAction.java
```diff
@@ -57,4 +70,41 @@ public boolean equals(Object o) {
...
+        @Override
+        public Collection<? extends Action> createFor(FlowNode node) {
+            try {
+                Queue.Executable executable = node.getExecution().getOwner().getExecutable();
+                if (executable instanceof Actionable) {
+                    DownstreamBuildAction action = ((Actionable) executable).getAction(DownstreamBuildAction.class);
+                    if (action != null) {
+                        for (DownstreamBuildAction.DownstreamBuild downstreamBuild : action.getDownstreamBuilds()) {
+                            if (downstreamBuild.getFlowNodeId().equals(node.getId())) {
+                                Run<?, ?> downstream = downstreamBuild.getBuild();
```
dwnusbaum, line 91: Everything before this is probably ok for a `TransientActionFactory`, but in the worst case, this line will have to load the downstream build, which may be slow, so IDK if we are ok with this approach.

Another option could be to keep `DownstreamJobListener` and just change the log level of the warning message to avoid log spam. `NodeDownstreamBuildAction` would not work in cases where the upstream build completes before the downstream build starts, but maybe that's ok. 

A third option could be to delete `NodeDownstreamBuildAction` and switch to using `DownstreamBuildAction` directly, but from a brief look, that seems like it would be a more involved change, and I think it might have similar performance concerns as the `TransientActionFactory` (not 100% sure though).

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/listeners/NodeDownstreamBuildAction.java
```diff
@@ -1,21 +1,34 @@
...
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
+import org.jenkinsci.plugins.workflow.support.steps.build.DownstreamBuildAction;
 import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
 
-import java.util.Objects;
-
 /**
  * Annotates a FlowNode to point to a downstream build triggered by said node. Applied by
  * io.jenkins.blueocean.listeners.DownstreamJobListener in blueocean-pipeline-api-impl
```
jglick, line 27: now incorrect

jglick (commented):
Can we not just delete (or deprecate) `NodeDownstreamBuildAction` and change whatever code is calling is to look up `org.jenkinsci.plugins.workflow.support.steps.build.DownstreamBuildAction` on demand?

jglick:
https://github.com/jenkinsci/blueocean-plugin/blob/a910c5348296bac732149eb7ca4e07d46eec4874/blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeImpl.java#L191-L192 + https://github.com/jenkinsci/blueocean-plugin/blob/a910c5348296bac732149eb7ca4e07d46eec4874/blueocean-dashboard/src/main/js/components/karaoke/components/Pipeline.jsx#L312-L314 IIUC

dwnusbaum:
> Can we not just delete (or deprecate) NodeDownstreamBuildAction and change whatever code is calling is to look up org.jenkinsci.plugins.workflow.support.steps.build.DownstreamBuildAction on demand?

Yes I considered this kind of approach, but I think it would only move the performance issue to graph loading time (a bit better at least since it wouldn't affect non-Blue Ocean use cases). I can switch to it though.

dwnusbaum:
Oh, and I'm not sure if you can just change the code that looks for the action in the special Blue Ocean `FlowNodeWrapper` class in `PipelineNodeImpl`. I think you need to change the code that sets the actions when the graph is being processed so that the action propagates up to the node that is visible in Blue Ocean correctly.

dwnusbaum:
See https://github.com/jenkinsci/blueocean-plugin/pull/2540/commits/b7ef6079aa57050d441d2fd9097a7b5e4d0f1970 for an approach that moves the logic to graph processing time.

Delaying build loading until a user actually clicks on one of the relevant links does not seem possible without UX changes, since we expect the description of the run to be available in the UI.

jglick:
> it wouldn't affect non-Blue Ocean use cases

Which is pretty important I think, since B.O. might be installed yet rarely used.

jglick (approved):
OK AFAICT

blueocean-pipeline-api-impl/src/main/java/io/jenkins/blueocean/rest/impl/pipeline/PipelineNodeGraphVisitor.java
```diff
@@ -97,6 +101,8 @@ public class PipelineNodeGraphVisitor extends StandardChunkVisitor implements No
 
     // Collects instances of Action as we walk up the graph, to be drained when appropriate
     private Set<Action> pipelineActions;
+    // The IDs of nodes which contributed to pipelineActions.
+    private Set<String> pipelineActionsNodes;
```
dwnusbaum, line 105: Needs to be initialized in the constructor.

dwnusbaum: Fixed in e3677b62532f0540c6662f357c8c54e47ae2a3f3.

jglick:
> Blue Ocean doesn't build for me

If you do not need to test the GUI interactively, you can just use https://github.com/eirslett/frontend-maven-plugin?tab=readme-ov-file#skipping-execution (which will be a lot faster too).

blueocean-git-pipeline/pom.xml
```diff
@@ -37,7 +37,14 @@
...
       <!-- TODO this is hard to manage; if needed, should be in BOM -->
-      <version>6.6.1.202309021850-r</version>
+      <version>6.8.0.202311291450-r</version>
+      <exclusions>
+        <exclusion>
+          <!-- Provided by Jenkins core -->
+          <groupId>commons-codec</groupId>
+          <artifactId>commons-codec</artifactId>
+        </exclusion>
+      </exclusions>
```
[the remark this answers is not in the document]
dwnusbaum: The code in `GitUtils` uses Jsch-related classes directly, and it is not immediately obvious to me how to adapt the code to use higher level APIs or to switch to the Apache Mina code, so I will leave this as-is.

wkoot:
> Can we not just delete (or deprecate) `NodeDownstreamBuildAction` and change whatever code is calling is to look up `org.jenkinsci.plugins.workflow.support.steps.build.DownstreamBuildAction` on demand?

Would this by any chance also address https://issues.jenkins.io/browse/JENKINS-38339 ?

dwnusbaum:
> Would this by any chance also address https://issues.jenkins.io/browse/JENKINS-38339 ?

I am not sure, but given that [JENKINS-38339](https://issues.jenkins.io/browse/JENKINS-38339) and [JENKINS-60995](https://issues.jenkins.io/browse/JENKINS-60995) are both many years old, whereas the issue I am trying to fix was caused by https://github.com/jenkinsci/workflow-cps-plugin/pull/826 earlier this year, I suspect that this PR will not fix those issues. [This comment](https://issues.jenkins.io/browse/JENKINS-38339?focusedId=386016&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-386016) seems like a decent hypothesis as to the underlying problem.

If you have a way to reproduce the bugs described in those tickets consistently, then by all means please try out this PR and let us know if it helps.

wkoot:
> If you have a way to reproduce the bugs described in those tickets consistently, then by all means please try out this PR and let us know if it helps, and post your reproduction steps in Jira.

Sadly, it's an on-and-off issue which affects about a quarter of builds. We'll just have to drop BlueOcean and move to Pipeline Graph View, thanks for your quick reply and clarity

jglick:
@olamy is the stalebot (66b3ade60a362d76b22569e3d107c5a965b0fcfa) really constructive here? If you do not want to merge this, close it yourself.

olamy:
> @olamy is the stalebot ([66b3ade](https://github.com/jenkinsci/blueocean-plugin/commit/66b3ade60a362d76b22569e3d107c5a965b0fcfa)) really constructive here? If you do not want to merge this, close it yourself.

This is meant to be constructive. Adding the `stale` label (without closing the PR) is just a way to ping previous participants when a PR hasn’t had attention for a while.
If this PR is still relevant and participants agree, I don’t see any reason we do not merge it.

PR #2560, opened 2024-05-21
Prevent debounceFetchNodes from jumping node view

werkt:
Any of the events which trigger debounceFetchNodes should call fetchNodes[Only] with the current props node specified. This prevents a pipeline view from jumping to the first still-running parallel stage when any non-first stage is focused and a handled non-pipeline_step event is received.

Removed unnecessary fetchNodesOnly parameter.

# Description

Test:
For a running job with parallel stages with multiple steps that emits non-pipeline_step events currently running:
Click to focus on a running step *below* the first still-running step at the top of the pipeline.
Wait for a pipeline_{end,start,block_end,stage} or job_run_ended event to be received.
Without this change, the log/steps view will switch to a view which includes all of the steps of the topmost still-running step (different from focusing on only that step, there is no click-focus that will produce this view).
With this change, the view will remain on the previously focused node after the event has been handled and the debounced fetchNodes triggered.

Reviewer: _If there is test coverage for the React behaviors, I can add them._

See [JENKINS-53296](https://issues.jenkins-ci.org/browse/JENKINS-53296).

# Submitter checklist

# Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-dashboard/src/main/js/components/karaoke/components/Pipeline.jsx
+++ b/blueocean-dashboard/src/main/js/components/karaoke/components/Pipeline.jsx
@@ -50,9 +50,10 @@ export default class Pipeline extends Component {
         this.debounceFetchNodes = debounce((karaokeOut) => {
             logger.debug('sse fetch it', this.karaoke);
             if (karaokeOut) {
-                this.pager.fetchNodesOnly({});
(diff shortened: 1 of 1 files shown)


werkt:
@olamy is this something you could review?

PR #2568, opened 2024-07-30
Only fetch job queue in getCauseOfBlockage

schiasileon:
# Description

This fixes a performance issue on large Jenkins instances with huge queues.

Before this fix, blueocean would fetch all Jenkins queue items when cause of blockage is needed just to find out which item the run belongs to. While doing this, Jenkins has to execute a permission check on all queued Items, possibly leading to performance issues.

We documented and tested this issue on one of our Jenkins instances with 1000+ queue items. The permission check lead to the azure-ad plugin taking up a lot of CPU. Also see previous fix attempts [here](https://github.com/jenkinsci/azure-ad-plugin/pull/599).

Unit test not needed as this doesn't change any behavior.

# Submitter checklist
- [ x ] Link to JIRA ticket in description, if appropriate.
- [ x ] Change is code complete and matches issue description
- [ x ] Appropriate unit or acceptance tests or explanation to why this change has no tests
- [ x ] Reviewer's manual test instructions provided in PR description. See Reviewer's first task below.

# Reviewer checklist
- [ ] Reviewed the code


olamy:
Hi,
Thanks for your PR, this is going to the right direction.
I can reproduce the test failures locally as well. Can you look at those? Maybe it's possible the assert are not correct?

schiasileon:
> Hi, Thanks for your PR, this is going to the right direction. I can reproduce the test failures locally as well. Can you look at those? Maybe it's possible the assert are not correct?

Thx for looking into it. I replaced the approach by getting the items from a different list which does not check permissions and all tests are passing now.
Its now failing on npm install step. No idea why.

schiasileon:
@olamy 
Tests are now all passing

schiasileon:
any feedback pls?
@olamy

olamy:
Hi, 
I'm not really clear if it's normal we are not checking permissions.
I need to check that

olamy:
@schiasileon checking permissions is on purpose (see https://www.jenkins.io/security/advisory/2015-11-11/#queue-api-did-show-items-not-visible-to-the-current-user) 
can you check permissions for those items returned but `getBuildableItems` or ensure we are in context where the current user have enough permissions?

schiasileon:
Hi, I cannot see where permissions would matter in this context. You are just looking for a Item the user already has access to to get cause of blockage.
Please correct me if I'm wrong @olamy

PR #2629, opened 2025-06-02
Replace the use of internal implementation of bitbucket-source-plugin with dedicated API to handle bitbucket endpoint.

nfalco79:
This PR is used to verify the impact of new API in the bitbucket-source-branch-plugin.
The first step is verify the new version of bitbucket-source-branch plugin is compatible with the PR https://github.com/jenkinsci/bitbucket-branch-source-plugin/pull/1056
The second step after release bitbucket-source-branch plugin is replace the use of implementation classes for Bitbucket endpoint and with the new APIs


nfalco79:
step 1 completed and verified

nfalco79:
steps 2 and 3 completed
@jenkinsci/blueocean-plugin-developers this is ready for merge

blueocean-bitbucket-pipeline/src/main/java/io/jenkins/blueocean/blueocean_bitbucket_pipeline/AbstractBitbucketScm.java
```diff
@@ -361,7 +361,35 @@ protected StaplerRequest2 getStaplerRequest2(){
 
     @Restricted(NoExternalUse.class)
     public static @NonNull String normalizeApiUrl(@NonNull String apiUrl){
-        String serverUrl = BitbucketEndpointConfiguration.normalizeServerUrl(apiUrl);
+        try {
```
car-roll, line 364: This looks like it's the same as [URLUtils.normalizeURL](https://github.com/jenkinsci/bitbucket-branch-source-plugin/blob/9888e219b456f7bd4ee3128247d7513490e27edd/src/main/java/com/cloudbees/jenkins/plugins/bitbucket/impl/util/URLUtils.java#L61). Could we just use that instead?

nfalco79: URLUtil is internal

car-roll: whoops, thanks for the clarification!

car-roll (approved):

car-roll:
@olamy would you be able to release this? I haven't been able to release since I changed to my m1 MacBook 😛

olamy:
> @olamy would you be able to release this? I haven't been able to release since I changed to my m1 MacBo

done. Enjoy while it's hot!!! :)

PR #2632, opened 2025-06-16
[JENKINS-75768] Blue Ocean Fails to Load with GitHub Auth: 403 Forbidden on PUT Request

basil:
### Description

See [JENKINS-75768](https://issues.jenkins.io/browse/JENKINS-75768). https://github.com/jenkinsci/github-oauth-plugin/blob/f5ddfba7349dac0eec130482ca98b6ab236f0c49/src/main/java/org/jenkinsci/plugins/GithubRequireOrganizationMembershipACL.java#L84 means that `github-oauth` is limited to working in two configurations:

- Caller (Blue Ocean) is using Acegi, and callee (`github-oauth`) is also using Acegi
- Caller (Blue Ocean) is using Spring Security, and callee (`github-oauth`) is also using Spring Security

Prior to https://github.com/jenkinsci/github-oauth-plugin/pull/285, the former was the case. After https://github.com/jenkinsci/github-oauth-plugin/pull/285, the caller (Blue Ocean) is using Acegi, but the callee (`github-oauth`) is using Spring Security; hence, the `instanceof` check fails and the symptoms reported in the ticket are visible. This PR solves the issue by making the caller (Blue Ocean) also using Spring Security, thus restoring the working status quo.

While I was here, cleaned up after #1571 and #2106.

### Testing done

Reproduced the problem as described in the ticket; could no longer reproduce after this PR.

### Submitter checklist

### Reviewer checklist
- [ ] Reviewed the code

### DIFF
--- a/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/BlueOceanRootAction.java
+++ b/blueocean-rest-impl/src/main/java/io/jenkins/blueocean/service/embedded/BlueOceanRootAction.java
@@ -14,7 +14,7 @@ import io.jenkins.blueocean.commons.ServiceException;
 import io.jenkins.blueocean.rest.factory.organization.OrganizationFactory;
 import io.jenkins.blueocean.rest.model.BlueOrganization;
 import jenkins.model.Jenkins;
-import org.acegisecurity.Authentication;
+import org.springframework.security.core.Authentication;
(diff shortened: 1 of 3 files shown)


basil:
[CI build is green](https://ci.jenkins.io/job/Plugins/job/blueocean-plugin/job/PR-2632/2/)

